PR #70, opened 2026-07-22
Open design-run history in the request builder

### LINKED ISSUE #65: Design-run history should open the request builder, not a second response viewer
## Proposal

When a design-mode run is opened from history, open the **request builder** for
that run's request and inject the stored response into its response pane -
instead of rendering `DesignRunDetail`, which is a second viewer for the same
stored row.

## The architecture already half-believes this

The builder already renders stored design runs. On cold start,
`RequestBuilderProvider` fetches the last design run and rebuilds the response
pane from its trace:

- `useLastDesignRunQuery` (`queries/runs.ts:76`) finds the run by
  `type === "design" && requestId === ...`
- `responseFromRunResult` (`request-builder/utils/restore-response.ts`) maps the
  trace into a `ResponseState`
- `useResponseStore` (keyed by requestId) carries it into the pane

This proposal extends "restore the *last* run" to "restore *any* run". The whole
click path is wiring that exists:

```
RunItem click (design run)
  -> responseFromRunResult(report.results[0])
  -> useResponseStore.setResponse(requestId, restored)
  -> openTab({ type: "request", entityId: requestId })
```

The provider reads the store on tab focus, so the injected response appears. The
next Send overwrites it - correct builder semantics, and the same model Postman
uses (history loads into the builder; there is no third viewer).

## What gets deleted

- **`DesignRunDetail.tsx` entirely** (~180 lines). It is a duplication hotspot:
  a `trace as any` cast, one of the three hand-rolled status chips from #60, and
  one of the four timing renderers.
- **The design branch of the `run` tab** in `HistoryDetail.tsx` - the tab type
  becomes load-test-only.
- **`UnifiedResponseViewer` full mode.** After `DesignRunDetail` goes, the only
  consumer is `SampleRequestCard` (compact mode). Compact mode has no
  conditional tabs, so the second instance of the #59 stranding bug goes with it.

**Where it stops:** load-test samples. `SamplesTab` shows N sampled exchanges
per run and cannot mount a builder per sample, so the compact viewer stays.

## Three gaps the proposal exposes

Two of them are latent bugs on the cold-start restore path **today**; the
proposal just makes them unmissable.

### 1. Response age (fix needed regardless)

The response pane shows no "when". `ResponseState.timestamp` is written by
`restore-response.ts:94` and read by nothing - the codebase's most-repeated
defect class ("written but never read"). A restored response is already
indistinguishable from a fresh one.

Fix: an age chip in `ResponseStatusBar` ("from run - 14:32"), shown when the
response was restored rather than just executed. This also absorbs the Run
Details card `DesignRunDetail` currently shows (runId can live in the chip's
tooltip).

### 2. Sent vs current (fix needed regardless)

The builder's left pane shows the request *as it is now*, possibly edited since
the run. The trace records what was *sent*. Sent headers already survive
(`requestHeaders` -> Headers tab), but `restore-response.ts` drops the sent
body and reduces `rawRequest` to `` `${method} ${url}` `` - flagged in #60.

Fix: rebuild a real raw request from `trace.request` (method, url, headers,
body). Then "what did I actually send" lives fully in the Raw and Headers tabs.
This absorbs the "restored `rawRequest` one-liner" checklist item from #60.

### 3. Orphaned runs (decided: response pane only)

`request_id` is optional engine-side (`run.request_id.value_or("none")` - a run
can be recorded without one), and `delete_request` does not cascade to runs. A
design run with no living request has no builder to open.

Decision: open a tab that renders the builder's own `ResponseViewer` fed from
the trace via a thin provider stub (~10 lines - it needs only `response` and
`isExecuting`). Full fidelity (Timing, Cookies, Headers tabs), no request
editor. **Constraint: this must stay a stub around the existing component - if
it grows its own markup it has quietly become `DesignRunDetail` again.**

## Relationship to open issues

- #59 - the `UnifiedResponseViewer` stranding instance disappears when full
  mode goes; the request-builder instance still needs its fix.
- #60 - absorbs the restored-`rawRequest` item; the sample-view items
  (`SampleRequestCard`, `RequestResponseView`) are untouched and still needed.
- #62 - one fewer consumer of the hand-restated trace shape.

## Suggested order

1. Age chip + real `rawRequest` on the restore path (fixes today's latent bugs,
   no behaviour change to history).
2. RunItem click path: design runs open the builder with injected response.
3. Orphan stub tab.
4. Delete `DesignRunDetail`, shrink `UnifiedResponseViewer` to compact-only.

athrvk:
Closes #65.

A design run is not a document. It is a past state of a request the user still has, so opening it in a read-only viewer of its own put the response and the request that produced it in two different tabs - and meant every fix to the response pane had to be made twice.

Clicking a design run now opens the **request builder** for its request, with that run's response injected into the response pane. The whole path already existed: the builder rebuilds the *last* design run on every cold start (`useLastDesignRunQuery` → `responseFromRunResult` → `useResponseStore`). This extends "restore the last run" to "restore any run", which is the model Postman uses.

### Three things the move exposed

Two are latent bugs on the cold-start restore path today; the move just makes them unmissable.

**Response age.** `ResponseState.timestamp` was written by the restore path and read by nothing - this codebase's most-repeated defect. So a restored response looked exactly like a fresh one, while the request editor beside it showed the request *as it is now*, possibly edited since. It is now `restoredFrom`, read by an age chip in the status bar ("from run - 2h ago"), with the exact time and the run id in its tooltip - which is where the old Run Details card went. The next Send clears it.

**Sent vs current.** The restore path reduced the whole trace to `` `${method} ${url}` ``, so the Raw tab of a reopened run was a single line, and the sent body was reachable nowhere in the app - the old viewer built `requestData.body` from the trace and then rendered a mode that reads only headers. A new shared `buildRawRequest()` mirrors the engine's own assembly (`client.cpp`), so Raw and Headers together now answer "what did I actually send".

**Failed runs.** A request that never reached a server stores no `response` node - `store_result` writes `error_type`/`error_message` instead - and `responseFromRunResult` returned `null` for those. Survivable while a second viewer showed the error in its own callout; not once the builder is the only place a design run appears. Now mapped to the same status-0 shape a live failure produces, so the builder's `ClientErrorView` renders it with icon, hint and code.

### Orphans

`request_id` is optional engine-side, and `delete_request` does not cascade to runs, so a design run can outlive its request. Both cases keep the run tab, which renders the builder's own `ResponseViewer` through a stub context (`DesignRunResponse`) - full tab set, no request editor, and deliberately no markup of its own.

### Deleted

- `DesignRunDetail.tsx` - a `trace as any`, one of three hand-rolled status chips, one of four timing renderers.
- `UnifiedResponseViewer`'s full mode, which had no callers left. It reduces to the embedded sample view, which stays because a samples list cannot mount a builder per sample. `compact`, `showActions`, `trace` and `hiddenTabs` go with it - the last was passed by its one caller as `["request"]`, naming a tab the embedded view never had.

### Verification

`pnpm type-check` and `pnpm test` green (1027 tests, +35). New behavioural tests are mutation-checked - reverting each fix was confirmed to turn them red.

Also driven end-to-end against a live engine and the running app:

- clicking a design run opens its builder with the response and the age chip;
- the Raw tab shows the rebuilt wire request including the sent body;
- an orphaned failed run renders the error view with its `CONNECTION_FAILED` hint;
- clicking a second run for a request whose tab is **already active** swaps the response in place (the store subscription this adds);
- a live Send clears the chip.

The live engine also confirmed the trace field names and that the reconstructed raw request matches the engine's own byte for byte.


athrvk:
Superseded by #75

Issue #126, opened 2026-07-25
engine: constant_rps under-delivers above 1000 RPS, overshoots duration, ramp-down floods; duration units misparsed


athrvk:
Part of #121 (load-generation core hardening). Base: `b4e2e9f`. All anchors verified in `engine/src/core/load_strategy.cpp`.

This is a load-testing tool, so pacing that silently under-delivers or overshoots is a credibility bug: users trust the requested rate/duration.

## Problems

1. **`constant_rps` floors the actual rate to a 1000-multiple above 1000 RPS.** For high RPS the strategy batches per 1ms tick:
   ```cpp
   // load_strategy.cpp:248-253
   if (base_interval_us < 1000) {
       batch_interval_us = 1000;                                  // 1ms tick
       batch_size = std::max(static_cast<size_t>(target_rps / 1000.0), size_t(1));
   }
   ```
   `batch_size = size_t(target_rps/1000.0)` truncates: 1500 RPS -> `size_t(1.5)` = 1 req/ms = **1000 RPS**; 2500 -> 2000; 999999 -> 999000. Any target above 1000 is floored to the nearest 1000. A user asking for 1500 RPS gets 1000 and no warning.

2. **The loop is quota-bound, not time-bound, so an overloaded run overshoots its duration with zero recorded drops.** (`load_strategy.cpp:278` region) When `in_flight()` hits `maxInFlight` the batch loop stalls and `next_batch_time` falls behind wall-clock; the run keeps submitting the backlog past `duration_ms` to hit the request quota rather than ending at the wall-clock deadline, and nothing records the shortfall as dropped/skipped. A 30s run against a saturated target can run well past 30s and still report "on rate".

3. **Ramp-*down* underflows into an unbounded flood.** The ramp interpolation:
   ```cpp
   // load_strategy.cpp:499-503
   return static_cast<size_t>(static_cast<double>(start_concurrency) +
       static_cast<double>(target_concurrency - start_concurrency) * progress);
   ```
   `target_concurrency - start_concurrency` is `size_t`; when `start > target` (ramp down, a legitimate profile) it underflows to ~1.8e19, casts to a huge double, and `maintain_concurrency` tries to hold an astronomical concurrency for the whole ramp - a never-ending max-rate flood instead of a descending ramp.

4. **Duration units other than `s` are misparsed.** The parser strips only the last character and always multiplies by 1000:
   ```cpp
   // load_strategy.cpp:210-213
[...]

Issue #124, opened 2026-07-25
engine: stopping or deleting a run does not actually stop it (drains backlog, no in-flight cancel, DELETE orphans)


athrvk:
Part of #121 (load-generation core hardening). Base: `b4e2e9f`.

## Problem

Three related defects mean a user's stop/cancel of a load run is not honored. Anchors verified.

1. **Stop drains the queued backlog and can hang on a slow upstream.** When the run winds down (including a user stop), `execute_load_test` calls `context->event_loop->stop(true)` (`run_manager.cpp:454`, comment "Wait for pending"). `stop(true)` waits for every already-submitted transfer to finish, including any in-flight transfer against a slow or hung upstream - so a stop against a non-responsive target blocks until curl's own timeout (or forever if `timeout` <= 0, see #122). The queued-but-not-started backlog is drained rather than discarded, so the target keeps receiving traffic after the user hit stop.
2. **No in-flight transfer cancellation.** `event_loop_worker.cpp:257` and the `stop` path have no `curl_multi_remove_handle` / abort of in-progress easy handles; `should_stop` gates the *submission* loop but nothing cancels transfers already handed to curl. Combined with (1), a stop cannot interrupt in-flight work.
3. **`DELETE /runs/:id` on an active run leaves it running and orphans rows.** The handler just calls `db.delete_run(run_id)` with no check that the run is active and no stop of its (detached) worker thread:
   ```cpp
   // routes/runs.cpp:88-101
   auto run = ctx.db.get_run(run_id);
   if (!run) { send_error(res, 404, "Run not found"); return; }
   ctx.db.delete_run(run_id);   // <- no active-run guard, no stop
   ```
   The run row and its metrics/results are deleted while the detached worker (`run_manager.cpp:362`) keeps executing and keeps writing new metrics/result rows referencing the now-deleted run - permanent orphan rows, and the run effectively "un-deletes" partially until it finishes.

## Fix / solution pathway

1. **Stop should discard, not drain.** On `should_stop`, call `event_loop->stop(false)` (discard pending) rather than `stop(true)`. The normal end-of-duration path can keep `stop(true)` to let genuine in-flight requests settle, but bound it (see below). Branch on `context->should_stop` at `run_manager.cpp:454`.
[...]

Issue #123, opened 2026-07-25
engine: extract_port std::stoi aborts the whole daemon on a port &gt; INT_MAX


athrvk:
Part of #121 (load-generation core hardening). Base: `b4e2e9f`.

## Problem

`extract_port()` calls `std::stoi` on an unbounded `\d+` port capture from the user-supplied URL, with no try/catch anywhere on the worker thread. A URL with a port literal above `INT_MAX` throws `std::out_of_range`, which propagates out of the worker thread's initial function and calls `std::terminate` - killing the entire engine daemon and every other active run and SSE stream.

Verified in the current tree:

```cpp
// engine/src/http/event_loop/curl_utils.cpp:37-44
int extract_port (const std::string& url) {
    static const std::regex port_regex(R"(^(?:https?://)?[^:/\s]+:(\d+))", std::regex::optimize);
    std::smatch match;
    if (std::regex_search(url, match, port_regex) && match.size() > 1) {
        return std::stoi(match[1].str());   // <- unguarded, unbounded \d+ capture
    }
    ...
}
```

- Only caller: `setup_easy_handle` (`curl_utils.cpp:99`), invoked per transfer by `run_loop` (`event_loop_worker.cpp:299`, `&dns_cache_` is a static member so the guard always passes) - so `extract_port` runs on every transfer.
- Verified no compensation: `rg "catch" engine/src/http/event_loop/` returns **zero** catch blocks; the thread entry is a bare `std::thread(&EventLoopWorker::run_loop, this)` (`event_loop_worker.cpp:207`) with no `std::set_terminate`; `deserialize_request` (`json.cpp:313-316`) checks only that `url` is present, and `routes/runs.cpp` never validates the URL. `rg "SetInterruptHandler|stoi|catch" engine/src/http/event_loop/*.cpp` -> the only `stoi` is this line.
- The single-request `/execute` path uses `http::Client` (`client.cpp`), which does not call `extract_port`, so this is load-path specific.

## Reproduction

`POST /runs` with request url `http://example.com:99999999999/` (a typo, a pasted timestamp, or an agent-built URL from MCP `start_load_run`). The first worker to pop the request throws `std::out_of_range` from `stoi` inside `run_loop`; the uncaught exception in a thread function -> `std::terminate` -> the daemon dies, taking all concurrent runs and SSE streams with it.

[...]

Issue #121, opened 2026-07-25
Load-generation core hardening (review sweep 2026-07-24)


athrvk:
Tracking issue for hardening the engine's load-generation core: the event loop / curl transfer layer, the load pacing strategies, the run lifecycle, and the metrics collector. Every defect was verified against the tree at commit `b4e2e9f` with file:line anchors in the linked sub-issues.

## Outcome - all nine sub-issues resolved (closed 2026-07-29)

**High:**

- **#122** (PR #132) - `POST /runs` config validated; out-of-range values no longer crash or wedge the daemon.
- **#123** (fix merged) - `extract_port` no longer aborts the daemon on a port > INT_MAX.
- **#124** (PR #138) - stopping or deleting a run actually stops it.
- **#125** (PR #178) - run workers are joined on shutdown instead of detached; no more use-after-free against `Database`/`RunManager`.
- **#126** (PR #136) - `constant_rps` delivers the requested rate, runs end at the wall-clock deadline, duration units parse correctly.

**Medium:**

- **#128** (PR #140) - bounded response buffering, wire method preserved, failed-transfer timing clamped, DNS cache expiry.
- **#129** (PR #169) - bounded metric stores, atomic histograms, real failed counts.

**Low:**

- **#127** (PR #147) - hygiene tail closed: dead config, unguarded throws, orphaned-run reconciliation, `ThreadPool` ODR.

Follow-on work that grew out of this series lives in its own issues: #172 (sampling-path test coverage), #174 (store sampled response headers/bodies), #177 (delete the legacy EAV metrics read path once pre-#146 runs age out), #179 (a racy test assertion from the #129 work), and #173 (error-body shape, cross-cutting).

athrvk:
Adjacent defect found while implementing #122, **not** fixed there (out of its scope) - filing it here so it is not lost.

**`success_sample_rate` is sent as a percentage but read as a sampling period.** The load dialog's control is labelled "Success sample rate - keeping N% of successful responses", with a `0%`-`100%` slider, and `request-builder/index.tsx` maps `data_sample_rate` straight onto the engine's `success_sample_rate`. The engine treats that field as a period: `counter % success_sample_rate == 0`, i.e. "keep 1 in N".

[...]

PR #132, opened 2026-07-25
fix(engine,app): validate POST /runs config; out-of-range values no longer crash the daemon

### LINKED ISSUE #122: engine: validate POST /runs config; out-of-range values crash or wedge the daemon
Part of #121 (load-generation core hardening). Base: `b4e2e9f`.

## Problem

`POST /runs` (and its deprecated alias `POST /run`) deserializes a JSON run config and starts a run with **no range validation** on the numeric fields. Several ordinary-looking values crash the daemon or wedge the run. All anchors verified at the stated base commit.

1. **`success_sample_rate` / `response_sample_rate` = 0 -> divide-by-zero SIGFPE.** `run_manager.cpp:181-190` reads both via `config.value(...)` and casts to `size_t` with no floor. In the hot record path, `metrics_collector.cpp:121` does `counter % config_.success_sample_rate`, and `record_response_sample` does the same with `response_sample_rate`. A `0` is `% 0` -> SIGFPE, the whole daemon dies mid-run.
   ```cpp
   // run_manager.cpp:182
   mc_config.success_sample_rate = static_cast<size_t>(config.value("success_sample_rate", 100));
   // metrics_collector.cpp:121
   if (counter % config_.success_sample_rate == 0) { ... }
   ```
2. **Negative or huge `concurrency` -> `size_t` blow-up -> OOM.** `run_manager.cpp:387-388` does `static_cast<size_t>(config.value("concurrency", default))`; `-1` becomes `1.8e19`. `run_manager.cpp:398` then `std::max(concurrency, size_t(100))` with **no upper clamp**, and that becomes `EventLoopConfig::max_concurrent`, which each worker's `CurlHandlePool` uses as an **eager** pre-allocation count (`event_loop_worker.cpp:122-131`, a `for (i < initial_size) curl_easy_init()` loop that skips nulls rather than breaking). It allocates until malloc fails, then spins the remaining ~1.8e19 iterations. `{"concurrency": -1}` (a natural "unlimited" guess) bricks the run before any traffic.
3. **`timeout` <= 0 -> transfers never expire, run stuck `running`, stop impossible.** `run_manager.cpp:393` `int timeout_ms = config.value("timeout", 30000)`; a `0`/negative flows into the per-transfer curl timeout. A slow/hung upstream then never completes, the run never reaches a terminal status, and (compounded by the stop bug in #124) cannot be stopped.
4. **Numeric (non-string) `duration` -> `type_error` after the run row exists -> row stuck `pending` forever.** `config.value("duration", "60s")` at `run_manager.cpp:162` sits outside the try block (it opens at `:164` and wraps only the `stoll`), so a JSON-number duration - a form `execution.cpp:515-519` explicitly recognizes for logging - throws `nlohmann::type_error.302` out of `RunContext`'s constructor. The route creates the run row before starting the run (`create_run` at `execution.cpp:550` precedes `start_run` at `:559`), so the caller gets an opaque `500` and the already-created row stays `pending` forever.

Note on `liveTickIntervalMs`: the metrics thread does `sleep_for(milliseconds(tick_interval_ms))` (`run_manager.cpp:754`), which would busy-loop at `0` - but the config seed enforces `min_value = "10"` (`database.cpp:~1245-1253`) and `POST /config` validates against the row's min/max, so a `0` is not reachable through any supported path. No fix needed there; listed only so the validation gate below doesn't accidentally regress that floor.

## Fix / solution pathway

Add one validation step in the `POST /runs` handler (or at the top of the run-context construction in `run_manager.cpp`) that checks the config **before** a run row is created, and returns `400` with a specific message on violation - mirror the existing `POST /config` validator that already enforces per-row min/max (`routes/config.cpp`). Concretely:

- `success_sample_rate`, `response_sample_rate`: clamp to `>= 1` (a sample rate of 0 is meaningless; either reject or treat as "sample every request"). Pick one and document it.
- `concurrency`: reject `< 1`; clamp to a sane max (there is already `event_loop::MAX_CONCURRENT` in `constants.hpp` - cap at a multiple of it, or expose an explicit ceiling).
- `timeout`: reject `<= 0` (or define `0` = engine default explicitly, matching how `scriptTimeout` treats `0` since PR #115).
- `duration`: accept a string with a unit and reject other JSON types with a clear `400` **before** the run row is created; see #126 for the unit-parser rewrite (this issue only needs the type/`<= 0` guard so a bad value is a `400`, not a stuck `pending` row).

Defensive belt-and-suspenders even with the gate: guard the `% sample_rate` sites in `metrics_collector.cpp` (treat `0` as 1) so a future caller cannot reintroduce the SIGFPE.

Verify the renderer/MCP never send these values out of range, and that a `400` surfaces. The load-run start path is the request-builder load submit in `app/src/**` and `app/electron/mcp/tools.ts` `start_load_run`. Grep both for `concurrency`, `success_sample_rate`, `response_sample_rate`, `timeout`, `duration`. If any UI control (e.g. a sample-rate slider or concurrency field) can emit `0` or a negative, fix the control's min in the same PR. State "none, verified" if the clients are already in range.

## Tests

- Engine gtest (`engine/tests/run_route_test.cpp` or `run_manager_test.cpp`): `POST /runs` with each out-of-range value returns `400` and creates **no** run row; a mutation-check that removing the guard re-crashes / wedges (for the SIGFPE, a death-test on the `% 0` path).
- Engine gtest for the `metrics_collector` defensive guard: `success_sample_rate = 0` records without SIGFPE.

## Docs (same commit)

- `docs/engine/api-reference.md` - document the accepted ranges and the `400` responses for `POST /runs`.
- `docs/engine/db-schema.md` if any config-row min changes.

## Sequencing

Part of #121. **Parallel-safe** with the `extract_port` fix (#123, curl_utils) and the pacing/duration issue (#126, load_strategy) - three disjoint files. Shares `run_manager.cpp` regions with #124/#125; PRs #115 and #106 have merged since the base commit, so base on current master (the anchors above are exact at `b4e2e9f`; `run_manager.cpp` shifted by roughly a dozen lines). Do this one early - it is the widest crash surface.

## Acceptance criteria

- Each of the four out-of-range inputs returns a `400` (or a documented clamp) and never crashes or wedges the daemon; no run row exists after a rejected request.
- The `% sample_rate` sites cannot SIGFPE even if a `0` reaches them.
- New engine tests cover each case, mutation-checked; client min-values verified or fixed.

athrvk:
## Description

Wave 1 of #121, alongside #123 (PR #131) and #126 - three disjoint files.

**Verified still present at `dcbabe6`**, all four, with the anchors shifted by roughly a dozen lines from the issue's `b4e2e9f` base: the sample-rate reads are now `run_manager.cpp:194/202`, the `% rate` sites `metrics_collector.cpp:129/144`, the `concurrency` cast `:400`, the `timeout` read `:406`, and the `duration` read `:172` - still outside the `try` that opens below it. `start_run` wraps the `RunContext` construction in no `try` either, and neither does its call site at `execution.cpp`, so the `duration` throw does escape to a 500.

**The plan.** The root cause is one shape repeated five times: a field read with `config.value (...)`, cast to `size_t` or handed to a `%`, on a detached worker thread with no `catch` above it. Out of range there is not a bad run, it is a dead daemon. So the fix is a single total gate at the boundary - `validate_run_config`, an extracted testable core in the route, running **before `create_run`** so a rejected config leaves no row behind - plus a clamp inside `MetricsCollector` so the modulo is safe for any caller and not merely for the one route that now guards it.

The alternative I rejected was validating at the top of `RunContext`'s constructor, which the issue also offered. It is closer to the fields and would need no new function, but it is on the wrong side of `create_run`: by then the row exists, so the best it can produce is the stuck-`pending` row that defect 4 is about. Rejecting at the route is what makes "no run row after a rejected request" true, which is half the acceptance criteria.

The rules are a table (`NumericRunField`) rather than five hand-written branches, so a bound and its explanation cannot drift apart, and the message names the field and *why* the bound exists.

**Blast radius traced.** `validate_run_config` sits in `start_load_test`, which is registered at both `POST /runs` and the deprecated `POST /run` alias, so both are covered by construction. The `MetricsCollector` clamp is read by `record_success`, `record_response_sample`, and the constructor's `expected / success_sample_rate` reserve - a third division-by-zero site the issue did not list, which the clamp also closes. No stored shape, no other endpoint, and no success-path payload changes; the only new wire behaviour is a `400` where the daemon used to die.

**Consumers checked, both clients:**
- **Renderer** sends `success_sample_rate` (from the `data_sample_rate` slider), `concurrency`, and a `${n}s` string `duration`; it never sends `timeout` or `response_sample_rate`. The slider was a **live path to the SIGFPE** - see App changes.
- **MCP** `start_load_run` sends `concurrency` (unbounded `z.number()`) and a `z.string()` `duration`; no sample rates, no `timeout`. `checkLoadCaps` only bounded values from above.
- **Import / CLI** do not start runs.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

Not marked breaking: every rejected value either crashed the daemon or wedged the run, so nothing that previously worked stops working.

## Testing

```bash
python build.py -t
cd app && pnpm test && pnpm type-check
```

- **Engine: 374/374 pass**, up from 353 on `master`. The 21 new tests are `engine/tests/run_config_validation_test.cpp`.
- **App: 1150/1150 pass (159 files)**, up from 1143 - **+7** (3 in `LoadTestConfigDialog.test.tsx`, 4 in `electron/mcp/tools.test.ts`; baseline measured by stashing and re-running the same two files: 50 -> 57). `pnpm type-check` clean.
- **Mutation-checked** per CLAUDE.md, both sides, reverted and restored:
  - **Engine** - make `validate_run_config` return `std::nullopt` immediately and delete the collector clamp: **16 fail**, and the decisive one is `MetricsCollectorSampleRateGuard.ZeroSampleRatesDoNotDivideByZero`, which ctest reports as **`NUMERICAL`** - i.e. the test process took the actual SIGFPE, not a failed assertion. Fix restored, 374/374 re-run green.
  - **App** - restore `SAMPLE_RATE_PCT.MIN: 0`, drop the restore-clamp, drop `.int().positive()`: **5 fail** (3 concurrency-schema cases, the slider floor, the restored-0 case). Restored, 57/57 green.
- Cases covered engine-side: each of the four defects at `0`, negative, and (for `concurrency`) far above the ceiling; both `concurrency` bounds asserted inclusive against the constant itself; wrong JSON types for every numeric field and for `duration`; unparseable and zero durations; `"60s"`/`"500ms"`/`"5m"`/`"2h"`/bare `"60"`/`"1.5s"`/whitespace/uppercase all still accepted; absent and explicit-`null` accepted as "use the default".
- `git clang-format --diff --staged --extensions cpp,hpp` -> "clang-format did not modify any files". ESLint clean on all touched TS/TSX. Prettier run on **only** the three files it flagged, each verified prettier-clean on `origin/master` first, so no unrelated reflow.

## Checklist

## Docs

- **`docs/engine/api-reference.md`** - new **Accepted ranges** subsection under `POST /runs`: the range table with the reason each bound exists, the `400 invalid_run_config` shape, the absent/`null` rule, and a note that the ceilings are crash guards rather than policy (every client caps itself lower).
- **`docs/app/api-integration.md`** - why `LOAD_TEST_LIMITS` floors must sit at or inside the engine's, and that "keep no success traces" is the Save timing breakdown toggle rather than a rate of zero.
- **`docs/engine/mcp.md`** - the caps bullet now says `concurrency` is additionally schema-constrained to a positive integer, and points at the range table.
- **`docs/engine/db-schema.md` deliberately untouched**: no config-row min changed. The issue's note about `liveTickIntervalMs` was checked and is correct - the seed's `min_value = "10"` still holds, and nothing here touches it.

## App changes

**Not "none, verified" - the renderer had a live path to the crash.** `LOAD_TEST_LIMITS.SAMPLE_RATE_PCT.MIN` was `0`, and the control is a `type="range"` slider labelled *"0% - errors only"*. Dragging it to its left stop sent `success_sample_rate: 0` and killed the daemon, taking every other active run and SSE stream with it.

- **`src/constants/load-test.ts`** - `SAMPLE_RATE_PCT.MIN: 0` -> `1`, with a comment on why the field is a period rather than a share. Nothing is lost: **Save timing breakdown** is how "keep no success traces" is expressed, and it gates storage entirely.
- **`LoadTestConfigDialog/index.tsx`** - the slider's left-stop label becomes "1% - errors and a trickle", and the restored value is `Math.max`'d against the floor. That clamp is load-bearing: the state is `saved.sampleRate ?? DEFAULT`, and `??` keeps a `0` because `0` is present - so a config persisted under the old floor would otherwise still send it.
- **`electron/mcp/tools.ts`** - `concurrency` becomes `z.number().int().positive()`. The engine's `400` already protects the daemon; this names the field at the tool boundary instead, since `-1`/`0` for "unlimited" is exactly the guess an agent makes.
- **`src/types/api.ts`** - the `// 0-100` comment on `success_sample_rate` was wrong in the direction that mattered; now `1-100` with the reason.

The MCP schema test asserts on the schema object rather than through `dispatchTool`, because `dispatchTool` calls the handler directly and never runs zod - the SDK is what validates args in production, so a `dispatchTool` test would have passed with the guard deleted.

- **`timeout <= 0` is rejected, not treated as "engine default".** The issue allowed either. Rejecting is the loud option, and no client sends `timeout` on `/runs` at all, so nothing has to be migrated to it; silently substituting a default would also mean quietly overriding an explicit "no timeout" the caller asked for.
- **`max_response_samples` is validated too**, though the issue lists four fields. It is the same `static_cast<size_t> (config.value (...))` on the adjacent line of the same constructor, and a negative there reaches `response_samples_.reserve (...)` as ~1.8e19 - an uncaught `length_error` from the same constructor as defect 4, for one row in the table.
- **`duration` accepts a bare `"60"`.** Requiring a unit would have been a tightening that rejects what an agent plausibly sends; this guard only separates "parses to something positive" from "wedges the run". The unit-aware *interpretation* (`5m` currently meaning 5s) stays #126's, and the tests say so.
- **`rampUpDuration` is left alone** - same parser family, but the issue scopes this to `duration`, and #126 owns that file. Noted rather than pulled in.
- **The route-level URL validation and the stop/cancel behaviour are untouched** - #123 (PR #131) and #124 own those.

## Related Issues

Closes #122

Part of #121. File-disjoint from #131 (`curl_utils.cpp`) and from #126 (`load_strategy.cpp`), so all three can land in any order. Shares `execution.cpp` / `run_manager.cpp` regions with #124/#125, which are sequenced after this one.

---
_Generated by [Claude Code](https://claude.ai/code/session_014SvdTXJ8jGzVEmfTQR24qc)_


athrvk:
**Merged `master` in - the app half of this PR is now largely upstream.**

The branch had gone un-mergeable: #170 ("send the sample rate in the engine's unit, and make the load-test ceilings a setting") landed on all four app files this PR touches, and #171 landed in the same `MetricsCollector` constructor. Resolved and pushed as `b07d452`.

**What the resolution kept, and why.** #170 fixed the same defect this PR fixed, from the other end, and more completely - the slider was not only floored at 0, it was sending a percentage into a field the engine reads as a period, so the two ends of the control meant the opposite of what they said. So master wins the app files:

- `LoadTestConfigDialog/index.tsx` and its test: master's wholesale. Its `restore()` helper already clamps the restored sample rate against `LOAD_TEST_LIMITS`, which subsumes this branch's `Math.max`, and its `never offers, or sends, the divide-by-zero value` case covers both floor tests this branch added. The left-stop label is master's `1% - a trickle`, which is the correct wording once the conversion exists.
- `constants/load-test.ts`, `types/api.ts`: master's explanation of the percentage/period distinction, with this PR's fact - that the engine now answers a `0` with a `400` rather than dying - folded into it.
- `docs/app/api-integration.md`: master's two new subsections kept, preceded by this PR's paragraph on route-level range checking and `400 invalid_run_config`.

The engine side merged clean. #171's reserve caps and this PR's sample-rate clamp are independent guards in the same constructor and both remain.

**So the "App changes" section of the description above is now stale in one direction only:** the renderer no longer needs a floor fix from this PR, because it has one. What this PR still delivers app-side is `electron/mcp/tools.ts` - `concurrency` as `z.number().int().positive()` - plus its four tests, and the two comment corrections. The engine validation, the collector clamp and the docs are unchanged.

**Verified after the merge:** engine **411/411** (`ctest --preset linux-dev`), app **1762/1762 across 216 files**, `pnpm type-check` clean. `pnpm lint` reports 114 errors, all pre-existing on `master` and none in a file this merge touches.

---

Issue #125, opened 2026-07-25
engine: shutdown mid-run - detached run worker uses freed Database/RunManager and races curl_global_cleanup


athrvk:
Part of #121 (load-generation core hardening). Base: `b4e2e9f`.

## Problem

The load-run worker thread is **detached** and nothing joins it on shutdown, so shutting the engine down while a run is in flight is a use-after-free plus a curl teardown race.

Verified:

- The run&#39;s execution thread is spawned and immediately detached:
  ```cpp
  // run_manager.cpp:359-362
  context-&gt;worker_thread = std::thread([context, &amp;db, verbose, this]() {
      execute_load_test(context, &amp;db, verbose, *this);
  });
  context-&gt;worker_thread.detach();
  ```
  The lambda captures `&amp;db` (the `Database&amp;`) and `this` (the `RunManager`) by reference.
- `~RunManager()` only calls `stop_sweeper()` (`run_manager.cpp:323-325`) - it does **not** signal `should_stop` on active runs, nor join/drain any detached worker. `get_all_active_runs()` exists (`run_manager.cpp:327`) but is not used in any shutdown drain.
- In `daemon.cpp:170-177`, `db`, `run_manager`, and `server` are stack objects in `main`. On the shutdown path they destruct at scope exit while a detached worker may still be executing `execute_load_test`, dereferencing the now-destroyed `Database`/`RunManager` and writing to freed memory.
- `curl_global_cleanup` at process teardown races any in-flight curl transfers still running on those detached threads.

## Reproduction

Start a load run (any non-trivial duration), then trigger shutdown - OS signal (`SIGINT`/`SIGTERM`, handled in `daemon.cpp`) or `POST /shutdown`. The detached worker keeps running against a destroyed `Database`/`RunManager`; depending on timing this is a crash, a corrupt final write, or a silent UAF. A live SSE `/runs/:id/live` stream at the same time widens the window (`server.cpp:73` teardown vs. the worker still emitting ticks).

## Fix / solution pathway

Give the engine an ordered, blocking shutdown that stops and joins run workers **before** `Database`/`RunManager`/curl teardown:

1. Stop detaching the run worker, or track it so it can be joined. Keep the `std::thread` handle in the `RunContext` (it already exists as `worker_thread`) and join it during shutdown instead of detaching. If detach is kept for the normal path, `RunManager` must hold joinable handles (or `std::jthread`s / a tracked list) for active runs.
[...]

PR #136, opened 2026-07-25
fix(engine,app): deliver the requested RPS, end at the wall-clock deadline, read duration units

### LINKED ISSUE #126: engine: constant_rps under-delivers above 1000 RPS, overshoots duration, ramp-down floods; duration units misparsed

athrvk:
## Description

Wave 1 of #121, alongside #122 (PR #132) and #123 (PR #131) - `load_strategy.cpp` is disjoint from both, so all three can land in any order.

**Verified still present at `2354e2f`**, all five, at the issue's anchors shifted by a few lines: the `batch_size` truncation at `load_strategy.cpp:251`, the quota/backlog end-of-window drain at `:307`, the `size_t` ramp delta at `:501`, and the one-character duration parser at `:214` (constant) and `:457`/`:467` (ramp). One thing the issue describes has already changed: `record_drop_batch` **does** exist and is read all the way through (`run_manager` per-tick rows to `droppedRequests` in the run summary), so the fix wires the shortfall into an existing metric rather than inventing one.

**The plan.** Four of the five defects are one root cause wearing different clothes: a *quantity that only exists as an integer* - a batch of `size_t(rps/1000)`, a delta in `size_t`, a duration whose unit was thrown away, a seed taken before the predicate that bounds it. So the fix pulls each quantity out as a pure, total function (`engine/include/vayu/core/load_pacing.hpp`, the `refill_deficit.hpp` pattern) and makes the strategy loop consume them:

- `take_due_requests` accrues `target_rps × elapsed` and returns the whole requests owed, **carrying the fraction**. Rate fidelity stops depending on the tick length, so 1500 RPS is 1500 rather than 1000, and timer jitter is corrected on the next tick instead of lost.
- `ramp_target_concurrency` interpolates in `double` (signed), clamped at 0, so `start > target` descends instead of underflowing to ~1.8e19.
- `parse_duration_ms` matches `ms`/`s`/`m`/`h` as a **whole suffix** (bare number = seconds), and returns `nullopt` rather than guessing.

The alternative I rejected for the rate was the issue's other option, shrinking the tick below 1ms at high RPS. It fixes 1500 RPS by spinning more, and it fixes nothing at 1001 or 99999 - any rate that is not a multiple of the tick frequency still truncates. The accumulator makes the tick an implementation detail, which is why the 1ms tick could stay.

For the deadline (defect 2), the loop now clamps its accrual to `duration_end`, so the final tick pays out only the slice of the window it covers and no backlog can be pushed past the deadline. Demand refused by `maxInFlight` is **dropped at the instant it came due**, not deferred - which is what makes `sent + dropped == targetRps × duration` true and the shortfall visible. That is Vayu committing to the **open** model for `constant_rps`; the closed-loop modes are unchanged and the docs now say which is which.

**Blast radius traced.** `load_pacing.hpp` is new and consumed only by `load_strategy.cpp`. Inside it, `duration_field_ms` covers both duration fields in both strategies (`duration`, `rampUpDuration`), and `maintain_concurrency`'s new predicate check is shared by `constant_concurrency`, `ramp_up` and `iterations`. The drop counter already had readers: `MetricsCollector::dropped_requests()` -> `run_manager.cpp:537/671` -> `droppedRequests` in the run summary and `dropped_requests` in the per-tick metrics -> the history Performance tab. No endpoint, stored shape or success-path payload changes.

**Consumers checked:**
- **Renderer** - sends `duration` / `rampUpDuration` as `` `${n}s` `` from numeric second inputs, so no unit can reach the engine from the UI; `validateStartConcurrency` already blocks a ramp start above the target. Its `RPS` input allows up to 50000, which is exactly where the flooring bit. No change needed - see App changes.
- **MCP** - `duration` / `rampUpDuration` are free strings an agent writes, so this is where units actually vary. `parseDurationSeconds` was **already** unit-correct (`ms`/`s`/`m`/`h`), so the issue's "the MCP layer caps duration in ms" no longer holds; the real mismatch was the *engine* reading `"5m"` as 5s after the cap had read it as 300s. That mismatch is now gone.
- **Import / CLI** - do not start runs.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

Not marked breaking: every changed behaviour was previously wrong (a floored rate, a flood, a duration in the wrong unit, a seed on a zero-length run). A run that used to silently last 60s after a bad duration now fails - that is the intended loudness, and it is documented.

## Testing

```bash
python build.py -t
cd app && pnpm test && pnpm type-check
```

- **Engine: 378/378 pass**, up from 353 on `master`. **+25**: 18 in the new `engine/tests/load_pacing_test.cpp`, 7 in `load_strategy_test.cpp`.
- **App: 1151/1151 pass (161 files)**, up from **1148 measured on the stashed tree** - **+3** in `electron/mcp/safety.test.ts`. `pnpm type-check` clean.
- The 7 new engine integration tests were run **3 consecutive times** to check for timing flakiness; all green each time (slowest, the 1500 RPS run, 2.1-2.7s).
- **Mutation-checked** per CLAUDE.md, every fix reverted, confirmed red, restored:

  | Reverted | Result |
  |---|---|
  | `parse_duration_ms` -> the old `stoll(substr(0, len-1)) * 1000` | **9 fail** (all of `ParseDurationMs.*`), and the two duration-dependent integration tests stop finishing at all - `"500ms"` becomes a 500s run and `"1500ms"` a 25-minute one, past ctest's 60s per-test TIMEOUT |
  | ramp delta back to `size_t (target - start)` | **3 fail**: `DescendsWhenStartIsAboveTarget`, `RampDownIsMonotonicAndBounded`, and `RampDownDescendsInsteadOfFlooding` hangs - the flood itself, the controller submitting toward a ~1.8e19 target |
  | `take_due_requests` -> truncate per tick, no carried fraction | **4 fail**, printing `delivered 1000 for a target of 1500 RPS`, `delivered 2000 for a target of 2500 RPS`, `delivered 99000 for a target of 99999 RPS`, and the live run `delivered 1826 of 3000` |
  | drop recording removed from the rate loop | **1 fail**: `sent(10) + dropped(0) does not account for the 500 requests the rate owed` |
  | seed-before-predicate check removed | **1 fail**: `a 0s run submitted a seed before checking its own duration` |
  | app: the unreadable-duration guard removed | **2 fail** in `safety.test.ts` |

  Every mutation restored; engine re-run 378/378, app 1151/1151.
- `git clang-format --diff --staged --extensions cpp,hpp` -> clean. ESLint clean on all touched TS. Prettier run on **only** `electron/mcp/tools.ts`, verified prettier-clean on `origin/master` first, so no unrelated reflow.

## Checklist

## Docs

- **`docs/engine/api-reference.md`** (`POST /runs`) - a **Durations** paragraph (the four units, whole-suffix matching, bare number = seconds, and that an unreadable value fails the run); **`constant_rps` is time-bound, and its shortfall is recorded** (fractional accrual, drop-at-due, `sent + dropped == targetRps × duration`); **Ramp semantics** (descending ramps are valid, duration shorter than the ramp stops partway). The request example's `duration` / `rampUpDuration` comments name the units.
- **`docs/engine/benchmarks.md`** - **How `constant_rps` paces** (open, time-bound, tick-independent), and the `maxInFlight` tuning bullet now says that bounding it does not hide demand.
- **`docs/engine/mcp.md`** - the hard-caps bullet now covers rejecting a non-duration.
- `docs/app/api-integration.md` deliberately untouched: the renderer's request shape is unchanged.

## App changes

**Not "none" - but the exposure is MCP, not the UI.**

- **`electron/mcp/safety.ts`** - `checkLoadCaps` rejects a `duration` / `rampUpDuration` that is present but unparseable, naming the units. The engine now fails such a run, so the tool says so before starting one instead of after. `rampUpDuration` was declared on `LoadRunParams` and read by nothing; it is read now. `parseDurationSeconds` gains a comment pointing at the engine parser it must agree with.
- **`electron/mcp/tools.ts`** - `duration` / `rampUpDuration` descriptions name the accepted units, and `startConcurrency` says that a value above `concurrency` ramps down. These are the agent's only hint about a field whose semantics this PR changes. `rampUpDuration` is now passed into the guard's params.
- **Renderer: no change, verified.** `LoadTestConfigDialog` collects whole seconds and the send path formats `` `${n}s` ``, so no unit string is user-authored; `loadTestValidation.ts` already refuses `startConcurrency > concurrency`, so the UI never offered the descending ramp that used to flood. The RPS input's 50000 ceiling needed no change - the rates it allowed simply now arrive.

The MCP guard is tested through `checkLoadCaps` rather than `dispatchTool`, matching the existing file: `dispatchTool` calls handlers directly, so a schema/guard test routed through it can pass with the guard deleted.

- **`rampUpDuration` is fixed too**, though the issue scopes the parser to `duration` (#132 explicitly left it here). It is the same parser on the same two lines of the same file, and leaving one half broken would mean `"5m"` meant two different things in one config.
- **An unreadable duration throws from the strategy** rather than returning a `400`. The strategy runs on the run thread and cannot answer the request; `execute_load_test` already catches and marks the run `Failed` with the message. #122 (PR #132) owns the route-level `400`, and `parse_duration_ms` is a shared header precisely so that validator can adopt it instead of growing a second grammar - worth doing when whichever of the two lands second is rebased.
- **`"0s"` parses to 0 rather than being rejected**; a zero-length run now submits nothing, which is defect 5's stated expectation. Rejecting `duration <= 0` is #122's `400`.
- **A JSON *number* duration is read as seconds** rather than rejected, matching `parseDurationSeconds`. It costs one branch and removes a crash-shaped surprise (`nlohmann`'s `value()` type error) for a shape an agent plausibly sends.
- **The `sleep_time / 2` half-sleep is gone** - it existed to under-shoot a timer the old code could not correct for. Oversleeping is now self-correcting, so the loop sleeps the full remainder; the Windows sub-2ms spin is kept, since 15.6ms rounding would still make sub-tick sleeps bursty.
- **`requests_expected` is left as `rps × duration`** - unchanged, and it now genuinely matches what the generator owes.
- **Untouched:** the stop/cancel and shutdown lifecycle (#124, #125), route-level config validation (#122), `curl_utils` (#123).

## Related Issues

Closes #126

Part of #121.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MHRxAgNkrakNJ81b9TFMep)_


athrvk:
**CI gate went red on `e9ee216` for an infrastructure reason, not a code one - re-run queued.**

Every job that actually tests something passed on the merged head: `Test on ubuntu-latest`, `Test on macos-latest` and `Test on windows-latest` are all green (engine build + ctest, `pnpm type-check`, frontend tests), as are the three `C++ Tests (<os>)` result publishers, `Detect changes` and `Conventional title`.

The only failure is the `CI gate` aggregation job, and it never ran:

- no steps recorded, `runner_id: 0`, empty runner name, failed 51s after being created;
- its log is a 404, because there is no log to download;
- its entire body reads `needs.changes.result` (= `success`) and `needs.test.result` (= `success`) and echoes "CI passed", so had it executed it could not have exited non-zero.

The previous merge commit on this branch (`077608c`, same diff, one master merge earlier) passed the gate on the same code.

I re-ran the failed job. Attempt 2 has been `queued` for ~15 minutes without picking up a runner, and it is not alone - #132 and #138 are queued too, from the same 12:00 batch of master merges. So this is runner-pool congestion on the account rather than anything in this branch, and it should clear on its own; if the gate is still not scheduled in an hour, re-running it again (or a fresh push) is the whole fix.

No code change made, because there is nothing here to change. Nothing in this diff can influence a job that never started a step.

---

Issue #97, opened 2026-07-23
engine(api): reject client-supplied ids on POST create - engine fully owns ID generation


athrvk:
Phase 3 of #94 (parent tracking issue). Builds on #95 (verb split) and #96 (bulk import). Self-contained; see #94 for the overall rationale.

## Problem

After #95, `POST /` still accepts a client-supplied `id` on create - kept only because the import pipeline pre-assigned IDs. #96 removes that need (the orchestrator uses `POST /import/apply` with temp IDs). What remains is a loophole with no legitimate user: any client can still mint arbitrary ID strings, so ID format and uniqueness are policed by convention rather than owned by the engine.

## Fix (engine)

- `POST /collections`, `POST /requests`, `POST /environments`: a body containing `id` -&gt; **`400`** `{&#34;error&#34;: &#34;id is assigned by the engine; omit it (bulk import: POST /import/apply)&#34;}`. The create path always calls `generate_id` (#93).
- `PUT //:id` is untouched - the path parameter is the identity; a mismatched body `id` (if present) should also 400 rather than be silently ignored.
- From here, **every ID in the database came from `generate_id`** - one format (`_`), one authority, and create-vs-update ambiguity is structurally impossible rather than policed. Update the ID wording in `docs/engine/db-schema.md` to state engine ownership outright.

## App changes (required, same PR)

The app must be verified and refactored to never send `id` on create anywhere:

- **Renderer**: sweep `app/src` (`rg &#34;\bid\b&#34; app/src/services/api.ts app/src/queries` plus the create-mutation call sites) - interactive creates already omit `id` after #95, but any payload builder that passes a spread object through (transformers, duplicate-request flows, tab-restore re-saves) must be checked; strip `id` from create payloads explicitly rather than relying on the caller.
- **MCP**: `app/electron/mcp/tools.ts` save tools' id-absent branch must send no `id` key at all (not `id: undefined` serialized oddly); `resolve.test.ts` / tool tests updated.
- **Import**: confirm the orchestrator is fully on `/import/apply` (#96) and `assign-ids.ts` temp IDs never reach a `POST /` body.
- Tests that deliberately asserted the old client-id create path change to expect 400; everything else must pass unchanged.

## Tests

- Engine (per resource): POST with `id` -&gt; 400 with the message above; POST without `id` -&gt; create with a `generate_id`-shaped id; PUT with mismatched body `id` -&gt; 400; PUT without body `id` -&gt; works.
[...]

Issue #94, opened 2026-07-23
engine(api): split create/update verbs, one null-vs-absent rule, engine-owned IDs via a bulk import endpoint


athrvk:
Tracking issue for reworking the resource-write API: split create/update verbs, one null-vs-absent rule, and engine-owned IDs via an atomic bulk import endpoint.

## Outcome - all three sub-issues resolved (closed 2026-07-30)

- **#95** (PR #131) - `POST` is create-only (`409` on an existing id), `PUT` updates (`404` on a missing id); one null-vs-absent rule across all three resources, centralized in the per-resource field appliers (`apply_*_field` in `routes.hpp`) and `apiService.updateX` app-side.
- **#96** (PR #145) - `POST /import/apply`: the whole parsed import persists atomically in one transaction, real ids are generated engine-side, and the temp-id -> real-id map is returned to the orchestrator. Client-side rollback deleted.
- **#97** (PR #205) - `POST` create now rejects a client-supplied `id` outright; every record id comes from `generate_id` (#93, UUIDv4 via PR #99). The upsert-driven silent-merge class is closed end to end.

The original motivation chain held up: millisecond ids could collide (#93), upsert turned a collision into a silent record merge, and a client-supplied id was the remaining loophole - each link is now fixed and test-pinned.

athrvk:
Phase 1 (#95) is implemented in #119. One finding from it bears on this tracker's scope:

**#133 - `apply_json_field` accepts any non-null value.** Phase 1 gives all three resources one null-vs-absent rule, but the shared helper only guards `null`; a wrong-shape value (`variables: 42`, `auth: "bearer"`, `body: [1,2]`) is still dumped verbatim into storage and then silently discarded by every reader. It is the same defect class the phase was created to kill, one level up, so it is a scope question for this plan rather than a loose bug: fold it into a later phase, or leave it standalone.

Two other issues came off the same branch and are **not** in this tracker's scope, linked only so they are findable: #135 (every design run strips `createdAt` from stored variables) and #134 (a script `TypeError` does not name the callee - parked).

Phase 1 also extended the **null rule** to `/globals`, which #95 did not cover. Its verb-split carve-out is unchanged and still correct: a singleton with a hardcoded id has no wrong target to hit, so no `PUT /globals` was added.

Issue #79, opened 2026-07-23
engine(api): collection parentId accepts cycles; cascade delete can infinite-loop and is not transactional


athrvk:
## Problem

`POST /collections` (`engine/src/http/routes/collections.cpp`, ~lines 87-93) accepts any `parentId` without validation: a collection can be made its own parent, or reparented under one of its own descendants, creating a cycle in the `collections.parent_id` tree.

`Database::delete_collection` (`engine/src/db/database.cpp`, ~lines 542-563) then does a BFS with **no visited set**: `to_delete` starts with the root id and appends every child of every entry. With a cycle (A -> B -> A, or `parent_id == id`), `to_delete` grows forever - the engine hangs inside the global DB mutex, which blocks **every** endpoint including `/health`. One bad API call (the engine API is also used by the MCP server and external scripts, not just the app) bricks the daemon until restart.

Separately, the cascade (delete requests + collections, deepest-first) is **not wrapped in a transaction**, so a crash mid-delete leaves a half-deleted subtree.

### 1. Validate on write - `POST /collections` route

After resolving `parentId` in the handler:
- Reject `parentId == id` with `400` `{"error": "A collection cannot be its own parent"}`.
- When updating an existing collection with a non-null `parentId`, walk the ancestor chain of the proposed parent (`get_collection(parent).parent_id` upward). If the chain reaches the collection being updated, reject with `400` `{"error": "Cannot move a collection into its own descendant"}`. Bound the walk with a visited set so pre-existing corrupt data cannot hang the validator itself.
- Do **not** add a parent-existence check in this PR: the import orchestrator (`app/src/services/importers/orchestrator.ts`) creates collections in bulk and an existence requirement would couple this fix to import ordering. Cycle/self checks only.

### 2. Harden the BFS - `Database::delete_collection`

Add a visited set (`std::unordered_set<std::string>`); skip already-seen ids when appending children. This makes deletion terminate even on corrupt data written before the validation existed.

### 3. Make the cascade atomic

Wrap the delete loop in `impl_->storage.transaction([&]{ ...; return true; })` (same pattern as `add_metrics_batch`, `database.cpp` ~line 813). Note `delete_collection` runs under the recursive mutex already, so the transaction lambda calling other locked methods is safe.

## Tests (engine, Google Test in `engine/tests/`)

- Route-level (follow the existing route test pattern if present, else DB-level): self-parent rejected with 400; reparent-into-descendant rejected with 400; a legal reparent still succeeds.
[...]

PR #145, opened 2026-07-25
feat(engine,app): atomic bulk import via POST /import/apply, with an engine-owned id map

### LINKED ISSUE #96: engine(api): POST /import/apply - atomic bulk import returning an id-map; orchestrator switches to one call
Phase 2 of #94 (parent tracking issue). Builds on #95 (verb split). Self-contained; see #94 for the overall rationale.

## Problem

The import pipeline is the only legitimate reason clients supply their own IDs: `app/src/services/importers/assign-ids.ts` pre-assigns UUIDs so the orchestrator (`orchestrator.ts`) can wire `parentId`/`collectionId` across the whole tree before anything is persisted, then posts items individually - a 500-request import is ~500 sequential localhost POSTs. As long as this path exists, `POST /` must accept client IDs, which blocks full engine ID ownership (#97).

## Fix (engine): `POST /import/apply`

One call that persists an entire import atomically:

**Request** (temp IDs are opaque client strings, never stored):
```json
{
  &#34;collections&#34;: [
    { &#34;tempId&#34;: &#34;c1&#34;, &#34;parentTempId&#34;: null, &#34;name&#34;: &#34;...&#34;, &#34;variables&#34;: {}, &#34;auth&#34;: {...},
      &#34;preRequestScript&#34;: &#34;&#34;, &#34;postRequestScript&#34;: &#34;&#34;, &#34;order&#34;: 0 }
  ],
  &#34;requests&#34;: [
    { &#34;tempId&#34;: &#34;r1&#34;, &#34;collectionTempId&#34;: &#34;c1&#34;, &#34;name&#34;: &#34;...&#34;, &#34;method&#34;: &#34;GET&#34;, &#34;url&#34;: &#34;...&#34;,
      &#34;params&#34;: [], &#34;headers&#34;: [], &#34;body&#34;: {...}, &#34;auth&#34;: {...}, &#34;order&#34;: 0 }
  ],
  &#34;environments&#34;: [ { &#34;tempId&#34;: &#34;e1&#34;, &#34;name&#34;: &#34;...&#34;, &#34;variables&#34;: {} } ]
}
```

**Response `200`:** `{ &#34;idMap&#34;: { &#34;c1&#34;: &#34;col_&#34;, &#34;r1&#34;: &#34;req_&#34;, &#34;e1&#34;: &#34;env_&#34; } }`

- Real IDs come from `generate_id` (#93). References are resolved via temp IDs (`parentTempId`, `collectionTempId`); unknown or cyclic references -&gt; `400` with a per-item error (`{&#34;error&#34;: &#34;...&#34;, &#34;item&#34;: &#34;r1&#34;}`), **nothing persisted**.
- **Atomic**: one `storage.transaction` under the DB mutex (pattern of `add_metrics_batch`). Validate everything first (reuse the field validation from `POST /requests`, and #79&#39;s cycle rules for `parentTempId`), then write.
- Extract a testable core `import_apply_response(db, json) -&gt; {status, json}` per the #86 pattern.
- Size guard: cap items per call (e.g. 10,000) with a clear 400, so a malformed payload cannot OOM the engine.

The import pipeline gets a real refactor, not just a URL swap:

- **`assign-ids.ts`** becomes temp-id assignment only (`c1`/`r1` counters or UUIDs - opaque either way); its tests change accordingly.
- **`orchestrator.ts`** builds the single `/import/apply` payload from the parsed draft tree, calls it once, and consumes the returned `idMap` for everything that needs real IDs afterwards - opening imported items in tabs, seeding the TanStack Query caches, success toasts naming counts. Its per-item POST loop and per-item error handling are deleted; a failed import is now all-or-nothing, so the partial-failure UI paths simplify (update or remove the partial-progress reporting accordingly).
- **`api-endpoints.ts` / `apiService`** gain the `IMPORT_APPLY` endpoint and typed request/response shapes in `app/src/types/api.ts`.
- **Tests**: per-format import fixtures (`app/src/services/importers/__fixtures__/`) re-verified end-to-end against the new flow - the imported tree must be identical to what the old per-item path produced; orchestrator tests assert exactly one `/import/apply` call and correct id-map consumption.
- Keep a thin engine test proving per-item `POST` create still works for third-party scripts - the bulk endpoint replaces the app&#39;s usage, not the public create API (that removal is #97).

## Tests

- Engine: id-map correctness (every tempId mapped, prefixes right); parent/collection wiring persisted correctly (nested tree round-trips); atomicity - one invalid item -&gt; 400 + zero rows written; cycle in `parentTempId` -&gt; 400; item cap enforced.
- App: as listed under App changes; `pnpm test`, `pnpm type-check`, `pnpm lint`.

```bash
python build.py -t &amp;&amp; cd engine &amp;&amp; ctest --preset linux-dev --output-on-failure
cd app &amp;&amp; pnpm test &amp;&amp; pnpm type-check &amp;&amp; pnpm lint
```

## Docs (same commit)

`docs/engine/api-reference.md` (new `POST /import/apply` section beside `POST /import/fetch`); `docs/app/import-collections/` (orchestrator flow rewrite - this doc describes the per-item pipeline today); `CLAUDE.md` endpoint table if it stays readable.

## Sequencing

Second of the three #94 sub-issues: after **#95** (verb split); **#93 is done** (PR #99 merged 2026-07-24). Unblocks **#97** (engine-owned IDs). Independent of the main review waves apart from trivial `api-reference.md` adjacency.

## Acceptance criteria

- A full fixture import is one atomic call returning a complete id-map; partial failure persists nothing.
- App import flows produce identical trees to the old per-item path (fixture comparison), in one round trip, with the id-map driving post-import UI.
- Engine + app suites green; docs in the same commit; clang-tidy clean; no em-dashes (use ` - `).

athrvk:
## Description

Phase 2 of #94: the whole parsed import becomes one atomic `POST /import/apply` call, with the engine generating every id and returning a temp-id -> real-id map. The app's per-item POST loop and its best-effort rollback are deleted.

**Plan (root cause, approach, rejected alternative).** The import pipeline was the only legitimate reason `POST /<resource>` accepts a client-supplied `id`: `assign-ids.ts` minted real `col_`/`req_`/`env_` UUIDs so the orchestrator could wire `parentId` / `collectionId` across a draft tree before anything was persisted, then created items one at a time (~500 sequential localhost round trips for a 500-request import) with a rollback that deleted already-created roots when one failed halfway. The fix moves ID ownership to the engine: items reference each other by **opaque temp ids** that are never stored, the engine assigns real ids via `generate_id`, and it returns `idMap`. Validation covers the entire payload before the first write and the write is a single transaction, so a `400` means nothing was persisted and the rollback path has nothing left to do. The alternative I rejected was keeping per-item creates and wrapping them in an engine-side "transaction id" / two-phase handshake - it keeps the round trips, keeps client-owned ids (so #97 stays blocked), and puts partial-state cleanup back on the client, which is exactly the design being removed.

I verified the problem still exists as described before changing anything: `assign-ids.ts` still minted real UUIDs and `orchestrator.ts` still carried `createTree` + `rollback` at the anchors the issue names.

### Engine

- **`POST /import/apply`** (`engine/src/http/routes/import.cpp`), with an extracted testable core `import_apply_response(db, json) -> {status, json}` per the #86 pattern. Passes: claim every `tempId` (unique across all three sections, non-empty, no `id` field) -> resolve `parentTempId` / `collectionTempId` (references may point **forward**, so a child may precede its parent) -> reject cycles in the payload's own parent graph -> build rows -> one write.
- **Cycle rules from #79 applied to the payload graph.** `apply_collection_fields`' stored-tree ancestor walk cannot see a cycle here, because none of the rows exist yet; the payload walk is memoised so it stays linear on a deeply nested import.
- **Shared field appliers, not a second copy.** `apply_collection_fields` / `apply_request_fields` / `apply_environment_fields` are now declared in `routes.hpp` and used by both the single-item POST handlers and bulk import, so the null-vs-absent rule and per-field validation live in exactly one place per resource.
- **`Database::import_apply`** writes collections -> requests -> environments in one `storage.transaction` under `retry_on_busy` (the `add_metrics_batch` pattern, respecting the DB mutex rules).
- **Errors are loud and specific.** Per-item `400`s carry `{"error", "item"}`. A wrong-typed field (`"name": 42`) is caught and returned as a per-item `400` rather than escaping as a `500` - with one transaction, an escaped exception would be a lost import with no explanation. Item cap: 10,000 per call. Bad JSON is a `400`; only a write failure is a `500`.
- **One deliberate deviation from per-item behaviour, documented:** a collection's absent `order`. The per-item default appends after the *stored* siblings, and mid-transaction none of the payload's own siblings are stored, so all of them would land on the same number. Siblings get consecutive slots from that starting point, in payload order. Requests keep the `order: 0` default `POST /requests` has.

### App (required by the issue, included here)

- `assign-ids.ts` -> `assignTempIds`, stamping opaque `c1`/`r1`/`e1` counters; `CollectionDraft`/`RequestDraft`/`EnvironmentDraft` carry `tempId`, not `id`, so a draft can no longer be mistaken for a record.
- `orchestrator.ts` flattens the tree into one `applyImport` payload (depth-first: collection -> its requests -> children, carrying `order` and `bodyType = body.mode`) and **checks the returned id-map covers every item it sent**, throwing `Import incomplete: ...` otherwise - a silently skipped item would otherwise read as a clean import. That check is the id-map's reader; nothing else needs the real ids, because `useImportMutation` invalidates the collection / request / environment queries. Per-item loop, `rollback()`, and the `deleteCollection`/`deleteEnvironment` members of `ImportApi` are gone.
- `IMPORT_APPLY` endpoint, `apiService.applyImport`, and typed `ImportApply*` request/response shapes in `app/src/types/api.ts`.
- The modal's failure path is unchanged and still correct: the error it shows is now the engine's per-item `400`, and there is no partial progress to report.

### Docs (same commit)

`docs/engine/api-reference.md` (new `POST /import/apply` section beside `POST /import/fetch`, plus the "Resource writes" paragraph that claimed import supplies its own ids), `docs/app/import-collections/README.md` (pipeline rewrite), `docs/app/api-integration.md` (new Import block; same stale id claim), `CLAUDE.md` (endpoint table + the POST/PUT bullet). `mkdocs build --strict` passes.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

`POST /<resource>` is untouched and still accepts a client `id` for third-party clients; #97 removes that. Nothing in the app sends one any more.

## Testing

```
python build.py -e -t                                    # clean
cd engine && ctest --preset linux-dev --output-on-failure # 409/409 passed
cd app && pnpm test                                      # 169 files, 1223 tests passed
cd app && pnpm type-check                                # clean
mkdocs build --strict -f .github/mkdocs.yml               # clean
```

- **Engine: 22 new tests** (`engine/tests/import_apply_route_test.cpp`) - id-map completeness and `col_`/`req_`/`env_` prefixes, temp ids never stored, nested tree wired through forward references, field-default parity with per-item create, sibling ordering with and without an explicit `order`, appending after pre-existing collections, and every failure mode (invalid item, null `name`, wrong-typed field, unknown/cyclic/self parent, unknown or non-collection owner, duplicate/missing `tempId`, client `id`, non-array section, non-object body, item cap) each asserting **zero rows written** rather than only a status code. Plus one test that per-item `POST /collections` still creates, for third-party clients.
- **App: per-fixture parity test** (`orchestrator.fixture-parity.test.ts`) reconstructs what the deleted per-item walk would have sent, for all five format fixtures, and compares it field-for-field with the new payload; the fixture list is derived from the directory so a new format cannot skip the check. Mutation-checked: dropping `bodyType` from the flattening fails 5 of 6 cases.
- Orchestrator unit tests cover one-call-only, temp-id wiring, `importEnvironments: false`, engine failure propagation, the id-map completeness guard, and the pre-flight throw when a draft has no temp id.

`pnpm lint` reports 100 pre-existing errors (all `no-explicit-any` in files this PR does not touch, plus one in `api.ts` predating it); `eslint` on the files this PR touches is clean. `clang-tidy` on the changed engine sources is clean apart from one pre-existing `modernize-use-starts-ends-with` in the untouched `import_fetch`. Note the repo's `.clang-tidy` uses `ExcludeHeaderFilterRegex`, which the clang-tidy 18 in this environment cannot parse - I ran it with that key stripped.

## Checklist

## Related Issues
Closes #96

Unblocks #97 (engine rejects a supplied `id` on create).

---
_Generated by [Claude Code](https://claude.ai/code/session_01JFuRHKphmzGv1gvvMnCtMw)_


athrvk:
Merged `master` in - #155 (Postman environment + globals import) landed in the same four files this branch rewrote, so the conflict needed resolving by hand rather than textually. The branch is mergeable again.

**How the two were reconciled.** Both changes are kept whole. Collections, requests and environments still go in the single atomic `POST /import/apply` call this PR introduces; imported **globals** stay the separate read-merge-write #155 added, because they are an engine singleton behind `POST /globals` with no temp id - not a tree item the bulk payload can carry. `applyGlobals` runs after the apply *and* after its id-map check, which preserves the ordering #155 documented as load-bearing: globals is the one write that can destroy data the import did not create, so nothing may fail behind it.

**One behaviour change worth your eye.** #155's `rolls back created collections when the globals write fails` test is *replaced*, not dropped - there is no rollback left to undo an atomic apply that already succeeded. Two tests pin what replaces it: a failed apply never reaches the globals write, and a failed globals write surfaces with the tree already committed. That last one is the only partial outcome the import still has, and it is now stated in `docs/app/import-collections/README.md` and `postman-environment.md` rather than implied. Say the word if you would rather globals became part of the atomic payload instead - that is an engine change and I left it out of scope here.

The fixture-parity guard this PR added earned its keep: it flagged the two fixtures #155 shipped. They are now covered. Since an environment- or globals-only export legitimately parses to no tree, the non-vacuity guard counts every parsed item instead of requiring collections - it still refuses a fixture that parsed to nothing.

Verified on the merge result:

```
python build.py -e -t
cd engine && ctest --preset linux-dev --output-on-failure   # 412/412 passed
cd app && pnpm test                                         # 219 files, 1809 tests passed
cd app && pnpm lint (touched files)                         # clean
mkdocs build --strict -f .github/mkdocs.yml                 # clean
```

Mutation-checked the ordering: moving `applyGlobals` ahead of the apply fails all three ordering tests. Per CLAUDE.md I formatted only `orchestrator.test.ts`, which was prettier-clean beforehand.

---

athrvk:
Merged `master` (2103af5) into this branch - it had gone 6 commits ahead. The merge was textually clean, but one of those commits interacts with this PR by design, so it is worth naming rather than leaving to CI.

**#148 changed `apply_json_field` underneath this branch.** It now rejects a non-object for the six object-shaped fields (collections `variables`/`auth`, requests `body`/`auth`, environments `variables`, globals `variables`) with a 400 naming the field, and is `[[nodiscard]]`. Because this PR deliberately routes bulk import through the *shared* `apply_collection_fields` / `apply_request_fields` / `apply_environment_fields` rather than copying them, `POST /import/apply` picked the guard up for free - a `{"auth": "bearer"}` in an imported request is now a per-item 400 with zero rows written, same as on `POST /requests`. That is the payoff of the shared-applier decision, and it needed no code change here.

Nothing asserted it, though, so a future copy of the applier logic into `import.cpp` would pass all 22 existing tests in the file while silently reopening the hole on the import path only. Added `ImportApplyRouteTest.RejectsAnObjectFieldGivenANonObject` (bd55349): all six fields, three bad shapes each (`42`, a string, an array), asserting the 400 names the offending field and `stored_rows() == 0`. Mutation-checked - with the `is_object` check in `apply_json_field` disabled, the new test fails; restored, it passes.

**Verification after the merge** (the counts in the description predate it):

- `python build.py -e -t` - clean, no new warnings.
- `ctest --preset linux-dev --output-on-failure` - **528/528 passed** (527 on the merge commit, +1 new).
- `cd app && pnpm test` - **1840 passed across 222 files**; `pnpm type-check` clean. No `app/` file changed by this update.

No pre-existing failures observed. `import_apply_route_test.cpp` was not clang-format-clean at the merge commit, so per CLAUDE.md only the added lines were formatted (`git clang-format`).

---

PR #147, opened 2026-07-25
fix(engine): close the hygiene tail - dead config, unguarded throws, orphaned runs

### LINKED ISSUE #127: Engine hygiene tail: dead config/code paths, unguarded exceptions, missing startup reconciliation (low severity)
Part of #121. Base: `b4e2e9f`.

Ten small, independent low-severity engine-side defects in the load-generation core (dead fields, unguarded exceptions, a missing recovery pass, dead code, and a latent ODR risk), each cheap to fix on its own.

## Problem

- [ ] **1. `poll_timeout_ms` config is written, never read.** `EventLoopConfig::poll_timeout_ms` (`engine/include/vayu/http/event_loop.hpp:74`, backed by `POLL_TIMEOUT_MS=10` in `engine/include/vayu/core/constants.hpp:73`) is copied into every worker but the only `curl_multi_poll` call (`engine/src/http/event_loop/event_loop_worker.cpp:377`) hardcodes a literal `1`. Fix: wire `config.poll_timeout_ms` into that call, or delete the unused field/constant.

- [ ] **2. EventLoop rate limiter is a disconnected safety net.** Every one of the N workers is constructed with the full, unmodified `target_rps` (`engine/src/http/event_loop/event_loop_impl.cpp:29-31`, consumed at `event_loop_worker.cpp:173`), so with round-robin sharding the aggregate cap is really `N x target_rps`. Each per-worker token bucket also starts full at `burst_size = 2 x rps` (`engine/src/http/rate_limiter.cpp:20-25`), allowing an instant multi-second burst per worker. The documented blocking `acquire()`/non-blocking `try_acquire()` API (`rate_limiter.hpp:44-63`) has zero callers - only `try_acquire_unlocked()` is used, at `event_loop_worker.cpp:280` - and `IterationsLoadStrategy` parses `rps` (`load_strategy.cpp:408-410`) but never uses it. No user-visible effect today only because the one rps-driven mode (`constant_rps`) paces itself independently of this limiter. Fix: divide `target_rps` by `num_workers` before constructing each worker&#39;s limiter (and size its burst off the divided rate), or remove the per-worker limiter/config plumbing since it enforces nothing currently.

- [ ] **3. No startup reconciliation for crashed/killed runs.** `daemon.cpp` runs `db.init()` (~154) -&gt; starts the sweeper -&gt; starts the server, with no pass over rows left `Running`/`Pending` by a prior crash or the 5s forced-shutdown break (`daemon.cpp:224-228` abandons in-flight runs with no DB status write). `rg RunStatus::Running` shows the only writer is the normal start-of-run transition (`run_manager.cpp:377`) - nothing marks orphans on startup, so `GET /runs` shows them &#34;running&#34; forever until a user manually stops them. Fix: after `db.init()`, scan for runs still `Running`/`Pending` and mark them `Failed`/`Stopped` before the sweeper/server start.

- [ ] **4. Sweeper TTL provider call sits outside its defensive try.** `run_manager.cpp:292` calls `sweeper_ttl_provider_()` before the try block opens at `:298` (whose comment says &#34;never let an exception escape the sweeper thread&#34;). The daemon wires that provider to `db.get_config_int` (`daemon.cpp:171-172`), whose `get_config_entry` call is itself outside its own inner try (`database.cpp:916-929` only wraps the `stoi`). A DB throw here escapes the sweeper thread lambda and calls `std::terminate`, killing the whole daemon over a housekeeping tick. Fix: move the provider call inside the existing try/catch.

- [ ] **5. Outer-catch path never joins `metrics_thread`.** The outer catch in `execute_load_test` (`run_manager.cpp:607-628`) substitutes a 200ms sleep for a real join, unlike the two other failure paths (`:431-432`, `:448`) which do join before retaining. The join is deferred to `~RunContext`, which can then run under `RunManager::mutex_` during a later sweep eviction, and in a contrived tick/TTL configuration can even self-join from the metrics thread itself. Fix: join `context-&gt;metrics_thread` in the outer catch before `retain_run`, matching the other two paths.

- [ ] **6. `handle_result` drops genuine Error results.** `load_strategy.cpp:43-48` treats `result.is_error()` as unreachable (&#34;client.send() now always returns Response&#34;) and just logs + returns without calling `record_error` - but curl-handle-creation failure (`event_loop_worker.cpp:314-323`) and the `stop(false)` cancellation drain (`:228-244`) still produce real Error results. Each dropped result permanently inflates `in_flight()` (since `requests_sent` was already counted at submit) and shrinks effective concurrency for the remainder of the run. Fix: route Error results through `record_error` like every other failure path.

- [ ] **7. Closed-loop modes seed requests before checking duration.** `maintain_concurrency` submits `target(0)` requests unconditionally (`load_strategy.cpp:171-175`) before `should_continue` is evaluated for the first time (`:187`, after a 50ms wait) - so `duration: &#34;0s&#34;` (or negative) still fires the full default concurrency (100) of real requests, unlike `constant_rps` which correctly submits nothing for a 0s window. This is the closed-loop facet of the pacing/duration work already tracked in #126 - cross-linking only, not duplicating the fix here.

- [ ] **8. Numeric duration throws out of `RunContext`&#39;s constructor.** `config.value(&#34;duration&#34;, &#34;60s&#34;)` at `run_manager.cpp:162` sits outside the try block (opens at `:164`, wraps only the `stoll`), so a JSON-number duration - a form `execution.cpp:515-519` explicitly recognizes for logging - throws `nlohmann::type_error.302`, surfacing as an opaque 500 after the run row was already created (`create_run` at `execution.cpp:550` precedes `start_run` at `:559`), leaving that row stuck `Pending` forever. Owned by #122 (config validation); cross-linking only, listed here for completeness.

- [ ] **9. `EventSource`/`SseParser` is dead/test-only code.** `rg &#34;EventSource|SseParser&#34; engine/src engine/tests` shows construction only in `engine/tests/sse_test.cpp`; `engine/src/http/sse.cpp` defines the classes but nothing in production code constructs them - live-metrics streaming uses cpp-httplib&#39;s chunked `sink.write` directly. Fix: delete the unused client (`sse.cpp`/`sse.hpp` + `sse_test.cpp`), or keep it with a comment noting it&#39;s test-only / reserved for a future client use case.

- [ ] **10. Two conflicting `vayu::http::ThreadPool` definitions - confirmed real, not benign.** `engine/include/vayu/http/thread_pool.hpp:26` and `engine/include/vayu/http/event_loop.hpp:230` both declare a class `vayu::http::ThreadPool` in the same namespace with the same public surface but different private impl types (`Impl` vs `detail::ThreadPoolImpl`). Only `engine/src/http/thread_pool.cpp` (built into the `vayu_core` static library, `CMakeLists.txt:237-248`) defines the member functions; `event_loop.hpp`&#39;s declaration has no corresponding `.cpp` (`event_loop.cpp` implements `EventLoop` only). `engine/tests/event_loop_test.cpp` includes `event_loop.hpp` and instantiates `vayu::http::ThreadPool` against that declaration (lines 298+), then links against `vayu_core`&#39;s single implementation - this only works today because both classes&#39; sole data member (`impl_`, a `unique_ptr`) is a single pointer, making the layouts coincidentally ABI-compatible. It is undefined behavior under [basic.def.odr] and fragile to any future divergence. Fix: delete the unused `ThreadPool`/`ThreadPoolImpl` declarations from `event_loop.hpp` and have callers use `thread_pool.hpp`&#39;s definition.

## App changes

None, verified. All ten anchors are engine-only (`engine/src`, `engine/include`, `engine/tests`); nothing in `app/` reads `poll_timeout_ms`, drives the rate limiter, depends on run-status reconciliation, or touches the `ThreadPool`/`EventSource` classes.

## Tests

Two of the ten are genuinely testable as regressions: (3) startup reconciliation - seed a `Running` row directly in the DB before daemon/`RunManager` startup and assert it flips to `Failed`/`Stopped`; (2) rate-limiter aggregate - construct an `EventLoop` with `num_workers &gt; 1` and a `target_rps`, drive it past the initial burst, and assert combined throughput approaches `target_rps` rather than `N x target_rps` (`rate_limit_test.cpp` currently pins `num_workers=1`, which is exactly why this passes unnoticed today). The remaining items are narrow exception-safety/cleanup/dead-code fixes better covered by an assertion added to the existing sweeper/run_manager/event_loop test files than a new suite.

## Docs

None. No doc under `docs/engine/` currently describes `poll_timeout_ms`, the per-worker rate limiter as an enforcement layer, startup recovery semantics, or the SSE client/ThreadPool duplication, so there is nothing to correct; `docs/engine/api-reference.md`&#39;s existing note scoping `targetRps` to `constant_rps` mode already matches item 2&#39;s status quo and needs no change.

## Sequencing

Part of #121; independent of the other sub-issues and can land last, in any order relative to them. Items 7 and 8 are owned by #126 and #122 respectively - this issue only tracks and cross-links them for completeness. PRs #106 and #115 have merged since the base commit, so `run_manager.cpp` line anchors are exact at `b4e2e9f` and shifted by roughly a dozen lines on current master.

Status 2026-07-28 (later sitting): PRs #140 (#128) and #136 (#126) have merged, and PR #147 now conflicts with master in `engine/include/vayu/core/constants.hpp`, `engine/include/vayu/http/event_loop.hpp`, `engine/src/core/load_strategy.cpp`, and `engine/tests/load_strategy_test.cpp` - rebase before merge. Items 1 (poll timeout), 2 (rate limiter), 6 (dropped error completions) may be partially or fully addressed by the merged #140/#136 work - re-verify each against current master during the rebase and drop any item that is already fixed.

## Acceptance criteria

- Each of the 10 items above is either fixed or explicitly marked wontfixed with a one-line rationale in this issue.
- Items 7 and 8 are resolved by #126 and #122 respectively (or explicitly handed back here if those issues decline to take them).
- No behavior change to `app/`; engine test suite (`ctest --preset linux-dev`) stays green.

athrvk:
## Description

Eight of the ten items in #127's hygiene tail. Items 7 and 8 are cross-links the
issue explicitly assigns elsewhere, and both are fixed by the PRs already open
for their owners (verified against the branches, not assumed - see the table).

**Plan.** The root cause each item shares is that nothing forced a reader to
exist: a config field with no consumer, a token bucket whose cap nothing
aggregated, a status transition no startup path performed, a `catch` that did
not cover the call that throws, a `Result` branch dismissed as unreachable that
two live code paths still produce. So the approach is per-item and local - wire
the value or delete it, split the budget where the workers are built, reconcile
where the other startup housekeeping already runs (`Database::init()`), widen
the guard to cover the throwing call, record the completion the collector was
missing - rather than one shared abstraction over ten unrelated defects. The
alternative I rejected for item 2 was deleting the per-worker limiter outright
(the issue offers it): it is the only backstop below a strategy's own pacing,
and a limiter that enforces the aggregate the user asked for is worth more than
one less class. For item 3 I rejected putting the reconciliation in `daemon.cpp`
- `init()` already owns startup housekeeping, runs before the sweeper and server
start by construction, and is reachable from tests without a daemon.

| # | Item | Status |
|---|------|--------|
| 1 | `poll_timeout_ms` written, never read | Fixed - field deleted, constant now carries the 1ms the loop actually polls with |
| 2 | Rate limiter is a disconnected safety net | Fixed - aggregate budget split per worker; dead `rps` parse in `IterationsLoadStrategy` deleted |
| 3 | No startup reconciliation for crashed runs | Fixed - `Database::reconcile_orphaned_runs()`, called from `init()` |
| 4 | Sweeper TTL provider outside its defensive try | Fixed |
| 5 | Outer catch never joins `metrics_thread` | Fixed |
| 6 | `handle_result` drops genuine `Error` results | Fixed |
| 7 | Closed-loop modes seed before checking duration | Owned by #126 - fixed in #136 (`if (context->should_stop \|\| !should_continue (0)) return;` before the seed loop) |
| 8 | Numeric duration throws out of `RunContext`'s ctor | Owned by #122 - fixed in #132 (`validate_run_config` rejects it with a 400 before `create_run`) |
| 9 | `EventSource`/`SseParser` is dead/test-only code | Fixed - deleted (`sse.cpp`, `sse.hpp`, `sse_test.cpp`, and `SseEvent`/`EventSourceState`, which nothing else used) |
| 10 | Two conflicting `vayu::http::ThreadPool` definitions | Fixed - the undefined declaration in `event_loop.hpp` removed; `event_loop_test.cpp` includes `thread_pool.hpp` |

Notes on two judgement calls:

- **Item 1** could have been fixed the other way (wire the 10ms config into the
  poll). That would have been a silent 10x change to the hot loop's idle wait
  for a knob no caller sets, so the value the code has always used wins and the
  constant is corrected to match it.
- **Item 2's burst floor** is not defensive padding. `refill_tokens()` clamps
  the bucket to `burst_size`, so a share below one token could never accumulate
  one and that worker would never start a transfer - reachable by any run with
  `target_rps < num_workers`. `FloorsTheBurstAtOneTokenSoAWorkerCanNeverStall`
  pins it.
- **Item 3** leaves `end_time` as recorded rather than stamping the restart
  time: when the process died is unknowable at restart, and stamping it would
  invent a duration spanning however long the daemon was down.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`python build.py -e -t` then `ctest --preset linux-dev --output-on-failure`:
**370/370 passed** (8 new; the deleted `sse_test.cpp` took 20 tests for the
now-deleted client with it). App suite unchanged and green - **1215 tests /
168 files** - plus `pnpm type-check` clean; no `app/` file is touched, which is
the issue's "App changes: none, verified".

New tests, each mutation-checked (fix reverted, failure confirmed, restored):

- `PerWorkerConfigTest.*` (4) - the split, the derived default burst, the
  one-token floor, and the untouched unlimited/single-worker cases.
- `RateLimiterTest.EnforcesTargetRPSAcrossMultipleWorkers` - the end-to-end
  counterpart the existing `num_workers=1` test cannot see. Without the split
  this lands near 400 RPS against a 100 RPS budget.
- `DatabaseTest.ReconcileMarksRunsAbandonedByAPreviousProcessFailed` and
  `ReconcileKeepsTheEndTimeTheRunAlreadyRecorded`.
- `LoadStrategyTest.TransportErrorResultIsRecordedAndClearsInFlight` - asserts
  the error is recorded *and* that `in_flight()` returns to 0, which is the
  consequence that mattered.

Items 4, 5 and 10 have no added test: 4 and 5 are exception-safety/ordering
fixes whose failure mode is `std::terminate` or a join under a mutex during a
sweep (not reachable from a gtest without contriving the DB to throw on a
sweeper tick), and 10 is a link-time ODR fix - the existing `ThreadPoolTest`
suite now exercises the single surviving declaration. Called out here rather
than left to be noticed.

Clang-tidy clean on every changed source. Formatting: `rate_limit_test.cpp` was
clang-format-clean before, so it is fully formatted; the other files were not,
so only the lines this PR adds were normalised (per CLAUDE.md).

## Checklist

`docs/engine/architecture.md` in the same commit: the "Poll timeout: 10ms" line
was already wrong and is corrected to 1ms; the rate-limiting bullet now says
`targetRps` is an aggregate budget split N ways; and the Database section gains
the startup housekeeping order (reconcile, then prune, both best-effort). The
issue's "Docs: none" was written when item 3 added no behaviour - it does now.

## Related Issues
Closes #127

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kvet5RYR5egU4fioRhpKYf)_


athrvk:
Merged `master` (2103af5) into this branch - it had gone 25 commits ahead and the PR was in a conflicted state. Four files conflicted; three resolutions were mechanical, one needed a decision worth flagging.

**Item 6 collided with a fix that landed on master independently.** #140 fixed the same dropped-`Error` defect in `handle_result`, so the two versions are combined rather than one replacing the other:

- Master's control flow wins - record the error, notify the closed-loop controller, return early. This branch's version fell through the `else if` chain and had no `notify_refill` on that path.
- This branch's error type wins - `error_type_name (error.code)` rather than master's hardcoded `"cancelled"`. Both producers of a real `Error` (the `stop(false)` cancellation drain at `event_loop_worker.cpp:105` and curl-handle creation failure at `:410`) raise `ErrorCode::InternalError`, so nothing at that point can tell them apart, and labelling a handle-creation failure "cancelled" names a cause that did not happen. `TransportErrorResultIsRecordedAndClearsInFlight` asserts `internal_error` in the trace and would fail against the hardcoded string.

The other three:

- `constants.hpp` / `event_loop.hpp` - kept `POLL_TIMEOUT_MS = 1` and the deletion of the unread `EventLoopConfig::poll_timeout_ms`, alongside master's new DNS-cache and response-cap wording. Re-verified against 2103af5 that the field still has no reader anywhere under `engine/`, so item 1's premise holds.
- `load_strategy_test.cpp` - union of this branch's transport-error test and master's open-loop pacing suite.
- `load_strategy.cpp` - master's `duration_field_ms` moved into the anonymous namespace this branch reopens after `handle_result` (which it lifted to namespace scope so the test can call it).

**Verification after the merge** (numbers in the description above predate it):

- `ctest --preset linux-dev --output-on-failure` - **488/488 passed**, including all 8 tests this PR adds.
- `cd app && pnpm test` - **1828 passed across 221 files**; `pnpm type-check` clean. Still no `app/` file touched by this PR.

No pre-existing failures observed.

---

athrvk:
Merged `master` (0797fd0) into this branch - it had gone 4 commits ahead (#146 wide metric ticks, #145 bulk import, #150 variable `createdAt`, #178 joined run workers) and the PR was conflicted again.

**One conflict, mechanical.** #146 inserted the `metric_tick` declarations into exactly the block where this branch declares `reconcile_orphaned_runs` in `database.hpp`. Both are kept; they are unrelated additions to the same neighbourhood.

**#178 narrows item 3's rationale, not its need.** The comment on `reconcile_orphaned_runs` and the startup-housekeeping section of `docs/engine/architecture.md` both cited "the daemon's 5s forced-shutdown break" as an abandonment path. Master fixed that: `RunManager::shutdown` now signals `should_stop` on every active run and joins every worker, so a graceful shutdown writes the terminal status. Reconciliation still covers a crash or a kill, and both places now say so rather than naming a path that no longer exists.

**Verification after the merge** (numbers in earlier comments predate it):

- `ctest --preset linux-dev --output-on-failure` - **538/538 passed**, including all 8 tests this PR adds.
- `cd app && pnpm test` - **1846 passed across 223 files**; `pnpm type-check` clean. `git diff master -- app/` is still empty on this branch.

No pre-existing failures observed.

---

Issue #112, opened 2026-07-24
docs(engine): scripting.md documents non-existent script APIs; prune dead script config (contextPoolSize, TestsValidating)


athrvk:
## Problem

`docs/engine/scripting.md` has drifted into advertising a script API the runtime does not provide, and two script-related config knobs have zero readers. None of these individually is severe; together they make the scripting surface lie about itself. This issue is cheap hygiene, batched so an implementer clears it in one pass. All verified in the current tree (c502e6b).

**Doc drift (scripting.md documents things that do not exist):**

1. `pm.response.body` - `scripting.md:72` documents it as &#34;Raw body string&#34;. `setup_pm_response` (`engine/src/runtime/script_engine.cpp:831-880`) sets `json()`/`text()`/`headers` but never a `body` property, so it is `undefined`. Either point users at `pm.response.text()` or remove the line.
2. `pm.variables.get/set` - `scripting.md:108-118` documents it in full, with a resolution order. `setup_pm_object` never binds `pm.variables`, so calling it throws. `docs/app/pm-api-compatibility.md:52-53` correctly lists it as unsupported. **The two docs contradict each other.** Fix scripting.md to match reality (unsupported); implementing `pm.variables` is a separate enhancement, out of scope here.
3. Language-feature &#34;Limitations&#34; are wrong - `scripting.md:212` claims &#34;No optional chaining (`?.`), nullish coalescing (`??`)&#34; and calls them &#34;ES2021+&#34;. Both are ES2020 and are supported by the vendored quickjs-ng. The stated limitation is misleading; remove or correct it. (PLAUSIBLE - confirm by running a one-line script through the engine test harness before editing.)
4. Header casing - `scripting.md:71` and `pm-api-compatibility.md:20` say `pm.response.headers` has &#34;lowercase keys&#34;. `setup_pm_response` (`script_engine.cpp:870-873`) preserves the casing as received, and the engine test asserts `headers[&#34;Content-Type&#34;]` (mixed case) works (`engine/tests/script_engine_test.cpp:285`). The &#34;lowercase&#34; claim is false in both docs; correct to &#34;keyed as received&#34;.
5. `version()` hardcodes &#34;QuickJS 2024-01-13&#34; (`script_engine.cpp:1276`) regardless of the vendored quickjs-ng version. Minor: derive it from the vendored version macro, or drop the date.

**Dead config / metric (written, never read):**

[...]

PR #149, opened 2026-07-25
docs(engine): pm.request is a read-only snapshot, not a mutable request

### LINKED ISSUE #109: engine(runtime): pre-request scripts cannot mutate the outgoing request, but scripting.md says they can
## Problem

`docs/engine/scripting.md` teaches pre-request scripts as a way to sign/mutate the outgoing request, but the runtime discards every such write. A user who follows the documented pattern ships a request with no signature and gets no error.

Verified in the current tree (c502e6b):

- `pm.request` is a read-only snapshot. `ScriptContext::request` is `const Request*` (`engine/include/vayu/runtime/script_engine.hpp:42`), and `setup_pm_request` (`engine/src/runtime/script_engine.cpp:883-918`) builds plain JS values (url/method/headers/body) with **no** write-back into C++.
- The execute path sends the request built **before** the script ran. `engine/src/http/routes/execution.cpp:386` moves `built.request`; the pre-request script runs against a snapshot; then the **original** `request` is sent (`execution.cpp:417`). Nothing reads the JS `pm.request` object back. This is the codebase&#39;s signature &#34;written but never read&#34; pattern, at the script boundary.

Yet the docs present mutation as supported:

- `docs/engine/scripting.md:166-178` shows `pm.request.headers[&#39;X-Timestamp&#39;] = ...` under &#34;Modify request before sending&#34;.
- `docs/engine/scripting.md:224` states plainly &#34;Can modify `pm.request` (headers, body)&#34;.
- The header comment says &#34;Request to potentially modify&#34; (`script_engine.hpp:97`) and `execution.cpp:384-385` comments &#34;It may be further modified by the pre-request script&#34;.

The **other** doc gets it right: `docs/app/pm-api-compatibility.md:64-93` documents `pm.request` as read-only with mutations discarded. So the two docs contradict each other, and the aspirational one is the one users reach for first.

## Fix / solution pathway - owner decision required

Two honest resolutions; pick one.

**Option A - document the truth (small, safe, recommended for now).** Make `scripting.md` match `pm-api-compatibility.md`: state that `pm.request` is a read-only snapshot and mutations are ignored; remove the &#34;modify request before sending&#34; example (`:166-178`) or relabel it as unsupported; fix the two misleading code comments (`script_engine.hpp:97`, `execution.cpp:384-385`). No engine behavior change. This closes the correctness gap (users stop being misled) with near-zero risk.

**Option B - implement request mutation (larger feature).** After the pre-request script runs, read `pm.request.{headers,body,url,method}` back out of the JS context and apply the diffs to the C++ `Request` before it is sent. Requires: a non-const path for the request, read-back plumbing in `execute()` (the context opaque already carries `request`), and defining precedence vs. engine-side auth/variable resolution (auth is applied engine-side per `docs/engine/architecture.md`, so a script-set `Authorization` header could be clobbered - decide the order). This is a genuine feature and should be its own scoped issue if chosen, not folded silently.

Recommendation: ship **Option A** now (it is a doc-and-comment PR), and open a separate enhancement for Option B if request-signing is a wanted capability. Do not leave the docs claiming a behavior that does not exist.

Option A: none (docs + engine comments only). Option B: the renderer sends `preRequestScripts` already; no client change needed for read-back since the mutated request is used server-side - but verify no client assumes the sent request equals the composed one. (&#34;Option A: none, verified.&#34;)

## Tests

- Option A: no code test; this is a doc/comment correction. Optionally add a gtest asserting the documented-as-ignored behavior is in fact ignored (`pm.request.headers[&#39;X&#39;]=&#39;y&#39;` does not change the sent request), to pin the contract so a future Option B is a deliberate change.
- Option B: gtest that a pre-request script setting a header/body changes the actual sent `Request`; plus a precedence test against engine-applied auth.

## Docs (same commit)

- `docs/engine/scripting.md:166-178,224` - correct or remove the mutation claims.
- Align the two docs so `pm-api-compatibility.md` and `scripting.md` agree on `pm.request` read-only semantics.

## Sequencing

Part of the QuickJS runtime review series. Option A is docs + `setup_pm_request` region comments (~883-918), disjoint from the timeout and matcher fixes - parallel-safe. If Option B is chosen it overlaps `execute()` and should be ordered after the timeout fix - in flight as PR #115 - merges (both touch `execute`). Independent of the engine schema/API series and the #94 chain. Lower severity than the timeout and matcher issues (silent, but only affects users who attempt request mutation). Needs the A-vs-B decision before an implementer starts.

Status 2026-07-28: Option A (docs-only) was chosen - fix in flight as PR #149, which merges cleanly onto current master.

## Acceptance criteria

- `scripting.md` no longer claims pre-request scripts can mutate the outgoing request (Option A), OR they actually can and a gtest proves it with a defined auth-precedence rule (Option B).
- The two script docs agree.
- The misleading comments at `script_engine.hpp:97` and `execution.cpp:384-385` are corrected.

## Series index

**QuickJS runtime review series** (a PR per issue; all independent of the engine schema/API series #78/#89 -&gt; #90 and the #94 chain):

- #107 - script timeout never enforced (bug, high)
- #108 - pm.expect matchers: silent no-ops + throwing matchers (bug, high)
- #109 - pre-request request mutation is discarded but documented (bug/docs) &lt;- this issue
- #110 - pm.*.set() strips secret flag / type (bug)
- #111 - script errors + per-test failures never displayed (bug, app-only)
- #112 - scripting.md drift + dead config cleanup (docs, land last)

athrvk:
## Description

`docs/engine/scripting.md` taught pre-request scripts as a way to sign or stamp the outgoing request, but the runtime discards every such write. A user who followed the documented pattern shipped an unsigned request and got no error. `docs/app/pm-api-compatibility.md` already documented the truth, so the two docs contradicted each other and the aspirational one is the one users reach for first.

**Plan (root cause, approach, rejected alternative).** The root cause is that `setup_pm_request` (`script_engine.cpp`) copies `url`/`method`/`headers`/`body` out of the C++ `Request` into a plain JS object and nothing ever reads that object back; `execution.cpp` sends the request built *before* the script ran. This is the codebase's signature "written but never read" pattern at the script boundary. The approach is Option A from the issue - make the docs and the code comments tell the truth, and pin the contract with a test so a future write-back implementation is a deliberate change. The rejected alternative is Option B, implementing request mutation: it needs a non-const request path, read-back plumbing in `execute()`, and a defined precedence rule against engine-applied auth (a script-set `Authorization` would otherwise be clobbered by the engine's own auth resolution). That is a genuine feature, and the owner chose Option A in [a comment on the issue](https://github.com/athrvk/vayu/issues/109#issuecomment-5073671779).

Verified the problem still exists at `6dd7901` exactly as the issue describes: `ScriptContext::request` is still `const Request*`, `setup_pm_request` still has no write-back, `execution.cpp` still sends `request` unchanged after `execute_script(..., "Pre-request")`, and both misleading comments are still present.

Changes:

- **`docs/engine/scripting.md`** - state the read-only-snapshot contract under **Request Object (`pm.request`)**; replace the "Modify request before sending" example (which showed `X-Timestamp` / `X-Signature` header writes) with an honest inspect-and-stage-a-variable example plus a **Not supported** callout; correct the "Can modify `pm.request` (headers, body)" bullet in **Script Execution Context**. Both new mentions link to `pm-api-compatibility.md#request-mutation--url-variables`, so the two docs now agree and point at each other.
- **`engine/include/vayu/runtime/script_engine.hpp`** - `execute_prerequest`'s `@param request Request to potentially modify` corrected; the doc block now says the request is exposed as a read-only snapshot and is never written back.
- **`engine/src/http/routes/execution.cpp`** - "It may be further modified by the pre-request script" corrected to state that this is what gets sent and writes to `pm.request` are discarded.
- **`engine/tests/script_engine_test.cpp`** - new `PreRequestWritesToPmRequestAreDiscarded`.

**Blast radius traced.** `rg "pm\.request"` over `app/src` and `app/electron` is empty - the app never references it (no editor autocomplete entry, no script template), so **App changes: none, verified**. The only other doc mentions (`api-reference.md:732`, `architecture.md:200`) say auth is resolved *before* the script so `pm.request` is accurate - both already correct, left alone. `execute_prerequest` has no production callers (only tests); noted below rather than touched.

**Deliberate scope limits.** `scripting.md` also documents `pm.variables` (unsupported) on an adjacent line and has other drift - that is #112's explicit remit, which is sequenced to land last and sweep the residue, so it is left untouched here to avoid taking its diff. Only the mutation claims this issue names are corrected.

## Type of Change
- [x] Bug fix (a documented behavior that does not exist; no runtime behavior change)
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- `cd engine && ctest --preset linux-dev --output-on-failure` - **388/388 passed** (387 before, +1 new).
- `cd app && pnpm test` - **1215 passed across 168 files**; `pnpm type-check` clean. No app files changed; run to confirm nothing regressed.
- `mkdocs build -f .github/mkdocs.yml --strict` - clean, which validates the new relative cross-link and its heading anchor.
- No pre-existing failures observed on the base commit.

The new test is mutation-checked: flipping `EXPECT_EQ(req.url, "https://api.example.com/users")` to the value the script assigns makes it fail, then restoring makes it pass - so the assertions really do read the C++ `Request`. It also asserts the script *did* write successfully on the JS side (`pm.environment.set("jsSideUrl", pm.request.url)` comes back as the assigned URL), so the test cannot pass by the script silently no-opping. Implementing write-back now breaks this test, which is the point: it forces the docs to be updated with it.

## Checklist
- [x] My code follows the style guidelines (no em-dashes; `prettier` deliberately **not** run on `docs/engine/scripting.md`, which was not prettier-clean before this change, per CLAUDE.md)
- [x] I have updated documentation (this change *is* the documentation fix; done in the same commit)

## Related Issues
Closes #109

Adjacent, not fixed here (flagging for #112, the dead-code sweep): `ScriptEngine::execute_prerequest` has no callers outside `script_engine.cpp` and the test suite, and its `Request&` (non-const) parameter is itself a leftover claim that pre-request scripts mutate. Its doc comment is corrected here; whether to delete the method or make it `const Request&` belongs in that cleanup, not this diff.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jnm3Bp7TjhJjitSibzeS7n)_

### DIFF
--- a/docs/engine/scripting.md
+++ b/docs/engine/scripting.md
@@ -93,6 +93,14 @@ pm.request.headers           // Request headers (object)
 pm.request.body              // Request body (string, if any)
 ```
 
+`pm.request` is a **read-only snapshot** of the request the engine is about to send
+(auth already resolved into headers/URL). It is a plain JavaScript object built for
+the script and never read back afterwards, so assigning to it - `pm.request.url = …`,
+`pm.request.headers['X-Sig'] = …` - changes only that throwaway object and is silently
+discarded. There are no `headers.add/upsert/remove` mutators either. See
+[Request mutation & URL variables](../app/pm-api-compatibility.md#request-mutation--url-variables)
+for the full compatibility notes and what it would take to support this.
+
 ## Environment Variables (`pm.environment`)
 
 Access and modify environment variables:
@@ -165,17 +173,23 @@ pm.environment.set('token', json.token);
 
 ### Pre-request Script
 
-Modify request before sending:
+Inspect the outgoing request and stage variables for later runs:
 
 ```javascript
-// Add timestamp header
-pm.request.headers['X-Timestamp'] = Date.now().toString();
+// Inspect what is about to be sent (read-only)
+console.log(pm.request.method, pm.request.url);
 
-// Add signature header
-const secret = pm.environment.get('secret');
-const data = pm.request.body + secret;
-pm.request.headers['X-Signature'] = computeHash(data);
+// Stage a value for subsequent runs - it does NOT affect this request
+pm.environment.set('lastRunAt', Date.now().toString());
 ```
+
+> **Not supported:** a pre-request script cannot change the outgoing request.
+> Writing to `pm.request` (headers, body, URL, method) is discarded, and setting a
+> variable that a `{{placeholder}}` in the URL references does not re-resolve it -
+> placeholders are substituted app-side before the payload reaches the engine, so the
+> new value only applies to later runs. Signing or stamping a request from a script is
+> therefore not possible today; see
+> [Request mutation & URL variables](../app/pm-api-compatibility.md#request-mutation--url-variables).
 
 ### Response Time Assertion
 
@@ -223,7 +237,8 @@ QuickJS supports ES2020 features with some limitations:
 ### Pre-request Scripts
 
 - Execute before sending the HTTP request
-- Can modify `pm.request` (headers, body)
+- Can **read** `pm.request` (method, url, headers, body) - it is a snapshot; writes to it
+  are discarded and the request is sent unchanged
 - Can access `pm.environment` and `pm.variables`
 - Cannot access `pm.response` (request hasn't been sent yet)
 
--- a/engine/include/vayu/runtime/script_engine.hpp
+++ b/engine/include/vayu/runtime/script_engine.hpp
@@ -93,8 +93,14 @@ class ScriptEngine {
 
     /**
      * @brief Execute a pre-request script
+     *
+     * The request is exposed to the script as `pm.request`, a read-only
+     * snapshot. Nothing is read back out of the JS object, so the script
+     * cannot change what is sent - see docs/engine/scripting.md and
+     * docs/app/pm-api-compatibility.md.
+     *
      * @param script JavaScript code
-     * @param request Request to potentially modify
+     * @param request Read by the script via `pm.request`; never written back
      * @param env Environment variables
      * @return Script result
      */
--- a/engine/src/http/routes/execution.cpp
+++ b/engine/src/http/routes/execution.cpp
@@ -417,8 +417,9 @@ void register_execution_routes (RouteContext& ctx) {
         }
 
         // Take the request built above (auth already resolved into headers/url,
-        // so pm.request reflects the real outgoing set). It may be further
-        // modified by the pre-request script.
+        // so pm.request reflects the real outgoing set). This is what gets
+        // sent: the pre-request script only reads it, and writes to
+        // `pm.request` are discarded (see docs/app/pm-api-compatibility.md).
         auto request = std::move (built.request);
 
         // Auth failure: record a failed result against the run and return the
--- a/engine/tests/script_engine_test.cpp
+++ b/engine/tests/script_engine_test.cpp
@@ -538,6 +538,40 @@ TEST_F (ScriptEngineTest, RequestHeaders) {
     EXPECT_TRUE (result.success);
 }
 
+// `pm.request` is a read-only snapshot: built as a plain JS object for the
+// script and never read back, so a pre-request script cannot change what is
+// sent. Both docs say so (docs/engine/scripting.md,
+// docs/app/pm-api-compatibility.md); this pins the contract so implementing
+// write-back becomes a deliberate change that also updates them.
+TEST_F (ScriptEngineTest, PreRequestWritesToPmRequestAreDiscarded) {
+    Request req;
+    req.method       = HttpMethod::GET;
+    req.url          = "https://api.example.com/users";
+    req.headers      = { { "Content-Type", "application/json" } };
+    req.body.mode    = BodyMode::Text;
+    req.body.content = "original";
+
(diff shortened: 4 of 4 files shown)


athrvk:
Closing this per your instruction on #109: "Close the open PR, and create a new PR which enables this capability rather than just updating the docs."

Option B is now open as #176 - a pre-request script's `pm.request` edits (method, url, headers, body) are read back and applied to the request before it is sent, with a script-set `Authorization` overriding the engine-applied one. This PR's docs correction is superseded rather than dropped: `scripting.md` and `pm-api-compatibility.md` are still made to agree there, in the other direction, and the two misleading comments this PR fixed (`script_engine.hpp:97`, `execution.cpp:384-385`) are corrected in #176 as well - they now state the rule instead of a possibility. The gtest here that pinned "a `pm.request` mutation does not change the sent request" is replaced by 17 tests pinning the opposite for pre-request scripts, one of which keeps that assertion for *test* scripts, where it is still true.

Nothing from this branch is left unaddressed.

---

PR #150, opened 2026-07-26
fix(engine,app): keep variables' createdAt through a design run

### LINKED ISSUE #135: app,engine: every run strips variables' createdAt, then a newly added variable sorts above the old ones
Add a variable to a collection that has existing variables, press Tab, and the new row jumps **above** the older ones instead of appearing at the bottom.

The sort is not at fault. Rows are ordered by a per-variable `createdAt` (`app/src/modules/variables/main/VariableTableEditor.tsx:216`), oldest first, and the stored data really does say the new row is older:

```
new test  createdAt 1784967810149
test      createdAt 1784967811254   <- 1.1s NEWER than "new test"
sdf       createdAt 1784967872689
```

The row was stamped wrong. Two defects combine to do it.

## 1. The engine wipes `createdAt` on every design run

`vayu::Variable` (`engine/include/vayu/types.hpp:463`) has four fields - `value`, `secret`, `enabled`, `type`. There is no `createdAt`, so `serialize_variables_json` (`engine/src/http/routes/execution.cpp:114`) writes those four and drops everything else.

`persist_script_variables` calls it **unconditionally after every `POST /execute`** - for the active environment, for globals, and for the executed request's collection. It runs whether or not a script touched a variable. Every Send therefore strips the ordering key from all three scopes.

Observed in a live database, which makes a clean natural experiment:

| Scope | `createdAt` present? |
|---|---|
| an environment that is **not** active | kept |
| the **active** environment | stripped |
| globals (persisted on every run) | stripped |
| the collection owning the executed request | stripped, then re-stamped |

## 2. The app re-stamps a stripped row at save time, not load time

- Load passes the value through as-is, `undefined` for a stripped row (`VariableTableEditor.tsx:394`).
- A row the user types gets `Date.now()` on the **first keystroke** (`VariableTableEditor.tsx:418`).
- A stripped row only gets a timestamp when the **save** fires: `createdAt: v.createdAt ?? Date.now()` (`VariableTableEditor.tsx:240`).

The save happens after the keystroke, so the pre-existing row receives the newer timestamp and leapfrogs the row just created. That is the 1.1s gap in the data above.

## Not surface-specific, and not recent

`VariablesTab` renders the same `VariableTableEditor` as the Variables screen (`app/src/modules/collections/CollectionDetail/VariablesTab.tsx:32`) - one component, one code path. It shows up on whichever scope has been stripped, which is why the collection whose requests were run is where it was noticed. Globals are stripped too and should reproduce it.

Both halves arrived in the same commit, `2f769a7` ("script vars persistence"), which added the `createdAt` ordering and the persistence that destroys it. Latent since then.

## Fix sketch (to be explored properly, not prescribed here)

- **Engine:** give `Variable` a `created_at`, read it in `parse_variables_json` and write it in `serialize_variables_json`, so the round trip stops discarding it. Worth deciding at the same time whether `persist_script_variables` should write at all when no script touched a variable - today every run rewrites three scopes unconditionally, which is also what makes this fire so often.
- **App:** stop backfilling to `Date.now()` at save. An unknown timestamp means "older than everything", so a stable sentinel keeps legacy rows at the top rather than firing them to the bottom one at a time as they happen to be saved.
- **Migration question:** rows already stripped have no true creation time. Decide whether to leave them tied (stable alphabetical among themselves, above everything timestamped) or to backfill once at load in current display order.

The two halves are independent: the engine fix stops new damage, the app fix stops already-stripped data from shuffling. Both are needed.

## Done when

- Running a request no longer changes the stored shape of any variable it did not modify.
- Adding a variable to a scope with stripped rows puts the new row at the bottom and leaves the others where they are.
- A test covers the full round trip (`POST /execute` on a request whose collection has variables, then assert the stored blob still carries `createdAt`), not just the app-side sort.
- The engine's variable round-trip is pinned field by field, so the next field added to `VariableValue` cannot be silently dropped the same way.

athrvk:
## Description

Adding a variable to a scope that already had some put the new row **above** the older ones. Verified against current `master` (6dd7901) before changing anything: both halves of the diagnosis in #135 still hold, at the same anchors.

**The plan.** Root cause is a round trip that is not one. `vayu::Variable` had four fields and no `created_at`, so `serialize_variables_json` - which `POST /execute` runs over the active environment, globals, *and* the executed request's collection after every Send, script or no script - wrote four keys and dropped the app's ordering key. The app then compounded it: a row with no `createdAt` was backfilled with `Date.now()` at **save** time, which is after the keystroke that stamped the new row, so the pre-existing row got the newer timestamp and leapfrogged.

The approach is to make the round trip whole and to stop rewriting what nobody changed: `Variable` gains an optional `created_at`; the blob parse/serialize pair moves out of `execution.cpp`'s anonymous namespace into `vayu::json::parse_variables` / `serialize_variables` so the two halves sit together and are unit-testable; and `persist_script_variables` compares the parsed on-disk blob with the in-memory one per scope and writes only on a difference. On the app side, unknown stays unknown - the sort already reads a missing value as older than everything, which is stable, and no row is stamped except at creation.

**Alternative rejected:** preserving the raw JSON per variable and merging modelled fields on serialize, which would make *any* unknown future field survive. It needs `Variable` to carry a JSON blob everywhere it is used (scripts, auth, the resolver) to fix one field, and the skip-when-unchanged rule already protects the untouched case, which is the common one. The enumerable-mistake guard instead lives in a test (`RoundTripKeepsEveryField`) and in a comment on the struct, per the issue's own fix sketch.

**Deliberate additions beyond the two halves**, both invited by the issue's "worth deciding at the same time":

- `POST /execute` now skips a scope no script changed. A run that writes no variable performs no DB write at all, so sending a request no longer moves a collection's or environment's `updatedAt` (only reader is the "Updated" stat in `InfoTab`, checked). This is also what protects an untouched legacy row from *any* future serializer gap.
- `pm.*.set()` stamps a key it creates. It is the one place the engine legitimately knows a creation time; stamping an existing key would sort it below rows the user added after it, which is the very bug. Same reasoning applied to the request builder's inline variable add, the app's other creation site.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

```
python build.py -e -t
cd engine && ctest --preset linux-dev --output-on-failure   # 394/394 passed (was 387, +7 new)
cd app && pnpm test                                        # 169 files, 1218 tests passed (+3 new)
cd app && pnpm type-check                                  # clean
```

New engine tests - `engine/tests/script_variables_test.cpp` (7):

- `RoundTripKeepsEveryField` - every field survives parse → serialize, plus a whole-struct equality check that catches the *next* field rather than this one.
- `AbsentCreatedAtIsNotInvented`, `NonNumericCreatedAtIsTreatedAsAbsent`, `MalformedBlobParsesToEmptyRatherThanThrowing` - the boundary cases: a legacy blob, a hand-edited one, a corrupt one.
- `ARunThatChangesNothingLeavesEveryScopeUntouched` - the reported bug, asserted on the **stored blob** and on `updated_at`, since a map-level check passes on the broken code.
- `AScriptWriteKeepsTheOrderingKeyOfEveryOtherVariable` - when a script does write, the other variables keep `createdAt`/`secret`/`type`, and the scopes it did not touch stay unwritten.
- `ARunWithoutAnEnvironmentOrRequestWritesOnlyGlobals` - the lookups that return nothing must not fail or write.

New app test - `app/src/modules/variables/main/created-at-ordering.test.tsx` (3): asserts the **saved payload**, with the clock pinned at two positions so the defect is deterministic - no invented timestamp on the untimestamped row, a stamp on the row just typed, and payload key order `stripped, kept, fresh`.

Mutation-checked, each reverted and confirmed failing, then restored:

| Reverted | Fails |
|---|---|
| `createdAt` dropped from `serialize_variables` | `RoundTripKeepsEveryField`, `AScriptWriteKeepsTheOrderingKeyOfEveryOtherVariable` |
| skip guard replaced with `if (true)` | `ARunThatChangesNothingLeavesEveryScopeUntouched`, `AScriptWriteKeepsTheOrderingKeyOfEveryOtherVariable` |
| app backfill restored to `v.createdAt ?? Date.now()` | 2 of the 3 app tests |

**Pre-existing, not touched:** `ThreadPoolTest.CreateWithSpecificThreads` timed out once on this machine mid-run and passed on a re-run and in two other full runs; it shares no code with this change. `pnpm lint` fails on `master` (114 errors) and fails identically here - the only delta is line-number shifts of pre-existing findings in a file I edited. `docs/` builds clean under `mkdocs build --strict`. clang-tidy reports nothing on the four C++ files touched. Per CLAUDE.md I formatted only `RequestBuilderProvider.tsx`, the one file that was prettier-clean before.

## Checklist

Docs in the same commit: `docs/engine/db-schema.md` (VariableValue gains `createdAt`, who may stamp it, and the "add a field to both halves" rule), `docs/engine/api-reference.md` (`POST /execute` persists only the scopes a script changed), `docs/engine/scripting.md` (`set()` preserves an existing variable's fields, stamps a new one). The scripting.md edit is two sentences in the `pm.environment` section, kept away from the lines #112 and #149 are rewriting.

## Related Issues
Closes #135

---
_Generated by [Claude Code](https://claude.ai/code/session_01DTqgdBH2eygZs7ZE3d1wgM)_


athrvk:
Merged `master` (2103af5) into this branch - it had gone 23 commits behind, so the green CI on this PR was measured against a base from two days ago. The merge was textually clean; the reason for doing it now rather than at merge time is that two of those commits land in the code this PR rewrites, and one of them left a stale reference behind.

**#148 touches the same write path.** It made `apply_json_field` reject a non-object for the six object-shaped fields (including `collections.variables` / `environments.variables`) with a 400. That is upstream of this PR's concern - it guards the blob's *shape* on write, where this PR guards the *fields inside it* through the design-run round trip - so the two compose without a code change. Re-verified after the merge that `parse_variables` / `serialize_variables` are still the only place in `engine/` that reads or writes a variables blob field by field, i.e. #148 did not add a second copy that would drop `createdAt` again.

**One fix in this update (1514f8a).** #148's comment on `apply_json_field` names `parse_variables_json` as the reader that degrades quietly on a bad blob. This branch renamed that helper to `vayu::json::parse_variables` when it moved out of `execution.cpp` into `utils/json`, so after the merge the comment pointed at a symbol that no longer exists. Renamed in the comment; no behaviour change.

**Verification on the merge result** (the counts in the description predate it):

```
python build.py -e -t                                       # clean, no new warnings
cd engine && ctest --preset linux-dev --output-on-failure    # 512/512 passed
cd app && pnpm test                                          # 222 files, 1834 tests passed
cd app && pnpm type-check                                    # clean
cd app && prettier --check (the four files this PR touches)   # clean
```

512 is up from 488 pre-merge purely because master added tests; all 8 this PR adds still pass, and the app suite includes this PR's 3 + 3 + 3.

**One pre-existing failure worth naming, since it is not mine to fix here.** A repo-wide `pnpm lint` reports 114 errors across 20-odd files (`electron/`, `components/ui/`, and others this PR never opens) - mostly the newer `react-hooks/set-state-in-effect` and `react-hooks/purity` rules. Of the two source files this PR edits, the flagged lines are untouched by it and the same code is present verbatim on `master` (e.g. `VariableTableEditor.tsx`'s `setVariables(rows)` inside an effect, master line 404). No workflow runs `pnpm lint` repo-wide, so this does not gate CI. Per CLAUDE.md I did not run a repo-wide fix; the four files this PR touches are prettier-clean.

---

Issue #76, opened 2026-07-23
Consolidate the HTTP-exchange UI: one component per concern across builder, dashboard, and history


athrvk:
## Summary (re-verified 2026-07-28 against master `f73b0c9`)

This issue originally described five duplication gaps across the request builder, the live load-test dashboard, and the history views. Since it was filed, most of them were fixed - primarily by PR #114 (#60, shared response-viewer primitives in the sample paths) and the app-UI rework series (PRs #142-#164). This body has been rewritten against the current tree: the resolved items are recorded below so nobody re-litigates them, and the two gaps that remain are re-anchored. **What remains is a consistency/maintainability refactor, not a defect** - smaller than originally scoped, and still gated on the Open questions at the end.

## Resolved since filing (do not redo)

- **Status chips (was gap 2): done.** Both `dashboard/components/RequestResponseView.tsx` (`:373`) and `history/main/components/SampleRequestCard.tsx` (`:67`) now render `StatusCodeBadge` from `components/shared/response-viewer`. No hand-rolled status `Badge` remains in either.
- **Raw `<pre>` header blocks (was gap 3): done.** `RequestResponseView` renders headers through `CompactHeadersViewer` (`:554`) and bodies through `ResponseBody` (`:569`); `SampleRequestCard` imports `HeadersViewer` / `UnifiedResponseViewer`. The one `<pre>` left in `RequestResponseView` (`:466`) is an error-message block, which is legitimate. `RequestResponseView.primitives.test.tsx` guards the primitive adoption.
- **The three-way shared header component family (was part of gap 3): mostly collapsed.** `HeadersViewer` and `CompactHeadersViewer` are now one file with two exported variants (`shared/response-viewer/HeadersViewer.tsx`); `ResponseHeadersPanel` survives as a 54-line wrapper used only by the builder's `ResponseViewer`. Folding that wrapper is optional hygiene, not a duplication problem.
- **`MetricCard` twice (was gap 5): done.** Zero `MetricCard` files or references remain anywhere in `app/src`.
- **Design-run side (PR #75) and the history load-test view reusing the dashboard** (`LoadTestDetail` -> `reportToDerived` -> `HeroRow`/`ModeStatsRow`/`charts/uplot`): still true, still the model to follow.
- The response-viewer rework (PR #164: tabs, toolbar, Raw, Console, Tests) further reinforces that the request builder's `ResponseViewer` is the mature reference primitive - it is not the thing that changes here.

[...]

Issue #60, opened 2026-07-22
Load-test and history sample views hand-roll the shared response-viewer primitives


athrvk:
## The pattern

`components/shared/response-viewer/` exists because the same response UI was
written twice and the copies drifted. The components there carry the scar tissue
in their own doc comments - `StatusCodeBadge` says a copy "had lost the
`status === 0` branch, so a connection failure rendered as a literal `0` chip
instead of `ERR`", and `HeadersViewer` documents a border that measured **1.002**
against its own background in dark, i.e. invisible.

The extraction stopped at the request builder and the history design-run viewer.
The load-test sample paths - `dashboard/RequestResponseView.tsx` and
`history/.../SampleRequestCard.tsx` - still hand-roll all of it, so none of those
fixes reach them. This is the repo's own rule from CLAUDE.md:

> A hand-rolled copy of a primitive does not receive the primitive's fixes.

## Headers: four implementations

| Where | How |
|---|---|
| `shared/response-viewer/HeadersViewer.tsx` | collapsible table, `border-rule`, `--status-*-text` colours |
| `shared/response-viewer/HeadersViewer.tsx` `CompactHeadersViewer` | `key: value` list on `surface-sunken` |
| `history/main/components/SampleRequestCard.tsx:114-127` | `<pre>{JSON.stringify(headers, null, 2)}</pre>` |
| `dashboard/components/RequestResponseView.tsx:553` | hand-rolled `Object.entries().map()` divs |

`SampleRequestCard` dumps request headers as raw JSON **immediately above** a
`UnifiedResponseViewer` that would render them properly (line 130). The two
treatments sit adjacent in the same card.

## Status chip: one shared component, one caller, three copies

`StatusCodeBadge` is imported by exactly one file (`ResponseStatusBar`). The
`statusCode === 0 ? "ERR" : statusCode` logic it was extracted to own is
re-implemented in two places (a third copy in `DesignRunDetail.tsx` was deleted
along with the file by PR #75), each with its own variant mapping:

- `history/main/components/SampleRequestCard.tsx:64-69`
- `dashboard/components/RequestResponseView.tsx:342-351` (plus a local
  `getStatusBadgeVariant` helper at line 58 duplicating `httpStatusClass`)

## Timing breakdown: four renderers, and a formatter that is imported but skipped

- `request-builder/.../ResponseTimingTab.tsx` (213 lines, timeline track)
- `history/main/components/TimingBreakdown.tsx` (108 lines)
[...]

PR #163, opened 2026-07-27
fix(app): give the timing waterfall five real hues, and the Raw tab a language

athrvk:
> **Stacks on [#162](https://github.com/athrvk/vayu/pull/162)**, which stacks on
> [#160](https://github.com/athrvk/vayu/pull/160). Merge in order and this diff
> collapses to its own work.

## Timing — two phases were outside the categorical set

TTFB took `--primary`; Download took `--success`.

| Phase | Was | Now |
|---|---|---|
| DNS | `--chart-2` teal | unchanged |
| Connect | `--chart-4` amber | unchanged |
| TLS | `--chart-5` rose | unchanged |
| TTFB | **`--primary`** | `--chart-3` violet |
| Download | **`--success`** | `--chart-6` moss |

The design system forbids a chart series on `--primary` because it tracks the
user's accent, and this is exactly what that costs:

- **Green scheme:** `--primary` is hue **142**, `--success` is hue **142** — three
  points of lightness apart. Two of five phases render as the same swatch.
- **Default orange:** TTFB sits **14°** from Connect's amber.

`--success` was independently wrong — a status token spent on a series with no
status, in the pane where the status bar sits directly above it.

The root cause is that this chart needs five series and the set offered four
fixed hues, so **`--chart-6` (moss)** is added. It sits in the widest gap in the
ring (38 → 172), 67° from its nearest neighbour.

## Raw — the language never existed

Monaco ships ~90 basic languages and `http` is not one of them: it has `html`,
`hcl`, `handlebars`, nothing for a protocol exchange. **Unknown ids fall back to
plain text**, so the one tab whose entire job is reading an HTTP exchange has
rendered as undifferentiated grey since it was written. The prop looked correct
and did nothing.

`lib/http-language.ts` is a small Monarch grammar: request line, status line,
header names, and the body after the blank line.

## Raw — an empty state that could never render

`if (!rawRequest && !response)` where `response` is a required object — always
truthy, so always false.

**Rewriting the condition was the first attempt and was also dead.**
`buildRawResponse` always emits a status line, and the tab is mounted behind
`hasRaw = !!response.rawRequest`. Nothing this component renders can be empty;
the check is the tab's *existence*, one level up. A second dead guard would have
been the same defect in better clothes, so the state is deleted.

## Two defects only running it could find

Both surfaced by tokenizing a sample exchange in the browser, neither reachable
by any test I wrote:

1. **The separator and the grammar drifted within one commit.** I changed the
   separator from 60 box characters to a commented line and left the grammar
   matching the old shape — so the tokenizer never left its `body` state and the
   whole *response* half was unhighlighted. One constant now, defined beside the
   grammar that has to recognise it.
2. **Even then the response head stayed plain**, because the blank line *after*
   the separator re-entered body-reading before the status line arrived. A blank
   line is what ends a head in HTTP and the grammar uses it as exactly that, so
   the join keeps its blank before the separator and drops the one after.

Verified against the real join:

| Line | Tokens |
|---|---|
| `GET /orders?limit=5 HTTP/1.1` | `keyword\|string\|comment` |
| `Host: api.example.com` | `type\|delimiter\|string` |
| `# ─── response ───` | `comment` |
| `HTTP/1.1 200 OK` | `keyword\|number\|string` |
| `Content-Type: application/json` | `type\|delimiter\|string` |
| `{"a":1}` | plain — correctly *not* read as a header |

## Raw — now curl notation, after checking what other tools do

Asked whether Postman / Bruno / JMeter show raw this way. **None of them do**, in
two respects:

- **The word means something else.** In Postman and Bruno, "Raw" in a response
  pane is the *unformatted response body* — one of Pretty / Raw / Preview, which
  is exactly what Vayu's own body toolbar segment means. The whole exchange lives
  elsewhere: Bruno calls it the **Timeline**, Postman puts it in the **Console**.
- **None concatenates the halves into one blob** with a rule through the middle.
  JMeter splits them across Sampler result / Request / Response data.

So direction is a prefix now — `>` sent, `<` received, as `curl -v` and HTTPie
print. Every line says which half it belongs to, an excerpt stays unambiguous
without its boundary, and there is no marker string for the component and the
grammar to disagree about — which they had already managed once.

Line numbers are off, with the gutter. Every line already opens with a marker, so
a number column is a second left-hand rail competing with the one that carries
meaning.

**The tab is still called "Raw"**, which collides with the body toolbar's Raw.
Left as-is pending a naming decision; `Timeline` and `Wire` were both considered
and both collide inside Vayu (with the Timing tab, and with the Timing summary's
own "Wire" label).

## Raw bodies — a false alarm with two real hazards behind it

Reported that the Body tab's Raw looked formatted. It was not: the engine already
sends untouched bytes (`json["bodyRaw"] = response.body`) and stores the trace
body as an opaque string. Three stored runs read back out of the engine:

| Server | Stored body |
|---|---|
| httpbin `/cookies` | `{
  "cookies": {}
}` — indented |
| jsonplaceholder `/todos/5` | `{
  "userId": 1,...` — indented |
| graphqlzero `/api` | `{"errors":[{"message":...` — **compact** |

The first two pretty-print their own JSON; the third does not, and Vayu shows it
compact. Nothing was being reformatted.

Two latent hazards would have broken precisely that, and are closed:

- `restore-response.ts` **never set `bodyRaw`**. Raw was correct only by accident,
  because `ResponseBody` falls back to `body` and it happened to be the same
  string. The moment a trace holds a parsed object — the branch above it exists
  because that can happen — `body` becomes pretty-printed and Raw would silently
  show formatting no server sent.
- `execute-mapping.ts` built its `bodyRaw` fallback with
  `JSON.stringify(body, null, 2)` — indenting the one field whose whole job is to
  be unformatted.

## A Monarch bug only running it could find

`next` rides on a *group's action*, not as a third element of the rule. Monaco
accepts `[regex, action, next]` only when the action is a single token and
silently ignores it beside a per-group array — so the tokenizer never left its
body state and every response header after the status line fell through as a bare
comment. Verified against the real join:

| Line | Tokens |
|---|---|
| `> POST /orders HTTP/1.1` | `comment\|keyword\|string\|comment` |
| `> Host: api.example.com` | `comment\|type\|delimiter\|string` |
| `> {"qty":2}` | `comment` — body, not mis-read as a header |
| `< HTTP/1.1 201 Created` | `comment\|keyword\|number\|string` |
| `< Location: /orders/9` | `comment\|type\|delimiter\|string` |

## Headers — already shared, nothing to do

Asked whether the response viewer hand-rolls headers. It does not, and hasn't
since **#60** closed. All four implementations that issue enumerated are gone:
`SampleRequestCard`'s `<pre>{JSON.stringify(...)}</pre>` now uses `HeadersViewer`,
and `RequestResponseView`'s hand-rolled `Object.entries().map()` now uses
`CompactHeadersViewer`. The only remaining `Object.entries(headers)` calls are
variable *resolution*, not display. `PHASE_TIPS` and `formatPhaseDuration` from
the same issue are shared too.

The two survivors are both in the shared file and differ by design — a table and a
compact list. #162 incidentally finished the job by aligning the one thing that
had drifted between them: the table used `--status-*-text`, the list used
`--primary`; both now use muted key / foreground value.

**#76 remains open** for the parts that are still duplicated — the status chip,
the four timing renderers, and the 599-line `RequestResponseView`. That one says
it needs UI/UX discussion before code, so it is not folded in here.

## Mutation check

| Mutation | Fails |
|---|---|
| TTFB back on `--primary` | 2 |
| Download back on `--success` | 2 |
| Two phases share a hue | 1 |
| The fixed-width box rule returns | 1 |
| Language never registered | 2 |
| Line numbers turned back on | 1 |

## Verification

`pnpm type-check`, `pnpm build`, full suite — **205 files, 1613 tests**.
`design-system-doc.test.ts` passes against the new token. Lint clean on touched
files (1 pre-existing error elsewhere). `docs/design-system.md` gains `--chart-6`,
why it exists, and the rule that a series never takes `--primary` or `--chart-1`.

**Grammar verified by live tokenization** (table above). The rendered colours of
those token types in each theme are not verified.


athrvk:
Superseded by the branch cut fresh from master after #160 and #162 merged - this one could not be rebased cleanly, and it was also missing the toolbar-background fix, which #162 merged just before.

PR #167, opened 2026-07-28
refactor(app): consolidate the HTTP-exchange UI - one timing descriptor, one sampled-exchange shell

### LINKED ISSUE #76: Consolidate the HTTP-exchange UI: one component per concern across builder, dashboard, and history

athrvk:
## Description

Closes the last two gaps in #76: the five network phases were declared five times, and the sampled-exchange shell was implemented twice.

**The issue scoped gap 2 at three timing renderers. There are five**, and the two it did not name are where the damage was:

- `dashboard/components/charts/TimingWaterfall.tsx` still painted **TTFB with `--primary` and Download with `--success`** - the exact bug `ResponseTimingTab`'s header comment describes fixing ("they are `--chart-3` and `--chart-6` now"). The fix reached the list it was written against and could not reach the other. `--primary` follows the user's accent, so under the green scheme it lands on `--success`'s hue and two of the five bars render as one colour. The guard that enforces this (`charts/uplot/status-code-series.test.ts`) reads chart specs in `uplot/`; the waterfall sits one directory above, outside its scan.
- The same file carried its own longer copies of the five tooltip sentences, written before `phase-tips.ts` and never migrated - so the module whose docstring says these live in one place was describing four renderers out of five.
- `RequestResponseView`'s run-level averages card is a fifth hand-rolled copy.

### Commit 1 - one timing-phase descriptor

`shared/response-viewer/timing-phases.ts` holds `TIMING_PHASES` (label, `longLabel`, hue, tip, and the trace/average field each phase reads) plus `phasesFromTrace` / `phasesFromAverages`. All five renderers read through it - add a phase there and every one picks it up.

The two copies of the tile grid become `TimingPhaseTiles`. `InfoChip` moves to `components/ui`: the request builder had hand-rolled its own `InfoTip`, and the copy is the one that got the `border-rule` fix, so the dashboard's original stayed outline-less in dark.

Two fixes fall out of the consolidation:

- the history card gated its "Timing Breakdown" heading on `sample.trace` while the breakdown inside returned null unless DNS/connect/TLS was present, so a TTFB-and-download-only trace printed a heading over nothing;
- `TimingWaterfall` formatted with `.toFixed(0)`, rendering any sub-millisecond average as a flat `0`.

### Commit 2 - `SampledExchange`

One shell for the dashboard's live sample list and the history detail's stored one. It owns the summary row, the expansion, the error block and the timing tiles; it is presentational over already-shaped data and expansion stays the parent's state (the dashboard keeps a `Set` of open indices, the history detail one). Site-specific sections are slots - `details` before the tiles, `children` after - not booleans, following the reasoning that kept the two response viewers separate.

The row is a real `<button>` with `aria-expanded` now; neither hand-rolled version exposed the state.

### One deviation from the issue's "visually identical" constraint

**The history card's timing tiles lose their pastel tints.** They were `bg-blue-50 dark:bg-blue-950/30` etc.; the shared tile is neutral (`surface-card` + `border-rule`), byte-identical to what the dashboard already rendered. Two components that look different cannot both survive a merge unchanged, so phase hue moved to `--chart-*` for the two renderers where colour is actually the encoding (the builder's timeline segments, the waterfall's bars). That also empties the raw-palette exception list in `docs/design-system.md`, which that section itself warns decays. Reverting is a one-file change - the descriptor already carries a hue per phase.

Two smaller deliberate changes: the history row adopts the dashboard's icon set (`AlertCircle`/`Clock`/`CheckCircle2`) and chevron in place of a hand-drawn CSS triangle, and gains the slow-request state it never had.

**Not done, deliberately:** folding `ResponseHeadersPanel` into the `HeadersViewer` family - the issue marks it optional hygiene and it is a 54-line wrapper with one caller.

## Type of Change
- [x] Bug fix - the waterfall's accent-tracking series, the sub-millisecond `0`, the heading over nothing
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- 211 test files / 1700 tests green; `tsc --noEmit` clean; no new lint errors or prettier drift in touched files.
- Two new guards, both **mutation-checked**:
  - `timing-phases.test.tsx` mocks in a **sixth** phase and asks each of the five renderers to show it. Asserting that today's five render would pass a renderer that hardcoded the same five, so that is not the test. Reverting the waterfall to a local array fails that case and only that case.
  - `sampled-exchange-adoption.test.tsx` replaces the shell with a sentinel, so a view that hand-rolls a row again fails - asserting on markup would not, since a hand-rolled row renders the same chip and latency. Reverting `SampleRequestCard` fails only its case.
  - `SampledExchange.test.tsx` covers the shell's own behaviour (parent-held expansion, section order, first-clause error preview, zero-status handling, heading gated on content).
- **Verified in a browser, not only in tests** - Chromium, light and dark, both sample views and both timing breakdowns against fixture data. The waterfall now shows five distinct hues where TTFB and Download previously tracked the accent, and the tiles' outlines survive dark mode, which is what `surface-card` + `border-rule` was for.
- **Still worth a manual pass:** the live-updating dashboard under a real SSE stream. Driving the packaged Electron app needs the engine daemon and actual run data, which this environment could not provide.

## Checklist

Docs updated in the same commits: `docs/app/COMPONENTS.md` (both new shared components plus `ui/info-chip`) and `docs/design-system.md` (the timing-phase token table, and the now-empty decorative-palette exception list). The stale `TimingBreakdown` exclusion note in `history-status-tokens.test.ts` is corrected too.

## Related Issues
Closes #76

https://claude.ai/code/session_01S4AAeWm6BZiTGH9g7H1ki8

---
_Generated by [Claude Code](https://claude.ai/code/session_01S4AAeWm6BZiTGH9g7H1ki8)_


athrvk:
Pushed two commits beyond the #76 scope described above, so the diff isn't a surprise in review. Both are the same defect shape #76 is about, found while verifying its scope - say the word and I'll split them into their own PR.

**`711ddd0` - fold `ResponseHeadersPanel` into the headers family.** #76's third suggested-direction item, marked optional there. `HeadersViewer` and `CompactHeadersViewer` already shared a file; the Headers *tab* was the last variant of the family living alone. Its docstring had gone stale - it was extracted because the builder's tab and an inline block in `UnifiedResponseViewer` were the same panel, but the stored viewer's Headers tab went back to a pair of `CompactHeadersViewer`s when it became compact-only (#75), so the two callers it was extracted for are one. Corrected the comment rather than deleting it, and did **not** unify the two treatments: a table needs width the compact pane does not have. Added the four tests it never had - the empty-state fallback is the only behaviour that is this panel's own, and it is the behaviour a copy of this panel has already lost once.

**`45836a2` - one `Eyebrow`, and stop re-typing its class string.**

- `Eyebrow` was two components: `components/ui/eyebrow.tsx` takes a `className`, `modules/dashboard/components/shared.tsx` declared a second one that dropped it. Both imported the same `EYEBROW_CLASS`, so nothing ever looked wrong - the copy just made the prop unreachable for dashboard callers. That file re-exports the primitive now, as it already does for `EYEBROW_CLASS` and `InfoChip`.
- Twelve files still typed the full class string out. `eyebrow.tsx` already recorded that the remaining hand-typed copies were "a separate sweep"; this is that sweep. Eleven `<p>` sites in the settings panels and welcome screens become `<Eyebrow>` - a pure substitution, same element and byte-identical class string.

Two are not pure, both deliberate: `GraphQLBody`'s `PaneTitle` keeps its `<span>` and takes `EYEBROW_CLASS` (it is an inline label in a flex bar, where the primitive's `<p>` is the wrong element), and `InheritanceChain`'s label had drifted to `tracking-[0.07em]` - 0.11px per character at 11px, a copy nobody could diff rather than a variant.

**Deliberately left alone:** `CollectionDetail/shared.tsx` (`Field`, `SectionLabel`, `Stat`) and `ChainCard` run a denser 10px tier, and `ResponseBody`'s content-type chip is 11px *without* the semibold because it is inline in a 32px toolbar band. Those are different things that happen to be uppercase; folding them in would resize four surfaces and wants a design call, not a sweep.

`eyebrow.test.ts` guards the narrow rule - the exact string may appear once - and asserts it scanned a non-empty tree, since a source scan that reads nothing passes forever. Mutation-checked: restoring any single copy names that file in the failure.

214 test files / 1721 tests green, `tsc` clean, ESLint clean on every touched file.

---

PR #175, opened 2026-07-28
feat: support HTTP/2, selectable per request

athrvk:
Adds a protocol selector - `auto` / `http1.1` / `http2` - to the request builder's **Settings** tab, and makes it real by linking nghttp2 into libcurl.

## The bug underneath this

`engine/src/http/event_loop/curl_utils.cpp` has set `CURLOPT_HTTP_VERSION` to `CURL_HTTP_VERSION_2TLS` for as long as the load path has existed, and `docs/architecture.md` claimed libcurl gave us "HTTP/1.1, HTTP/2, HTTP/3". None of it was true. vcpkg's curl port compiles HTTP/2 out unless the `http2` feature is requested, and `engine/vcpkg.json` asked for a plain `"curl"`:

```
libcurl/8.20.0 Schannel zlib/1.3.2
CURL_VERSION_HTTP2: False
```

Every request went out as HTTP/1.1 regardless of what the code asked for, and the two comments asserting multiplexing described something that could not happen. `engine/tests/http_version_support_test.cpp` now asserts the capability bit so a rebuild cannot silently drop back to a 1.1-only libcurl.

## Design decisions worth keeping

**One control, governing both modes.** The protocol lives in the request builder's Settings tab and applies to Send and load test alike. An earlier revision of this branch gave the load test dialog its own per-run override; it was removed. `followRedirects` and `maxRedirects` have no per-run override, so protocol having one made one of that tab's three fields mean something different from its neighbours. The load dialog decides load *shape* - rps, duration, concurrency - not request semantics. `POST /runs` still carries `httpVersion`, because that is how a run states its protocol at all and how MCP's ad-hoc runs (which have no saved request to read from) specify one.

**The global default is a seed, not a fallback.** `config_entries.defaultHttpVersion` supplies the value when a request is *created* without one. It is never consulted at execution time, so changing it does not retroactively alter saved requests. That is deliberate: for a load-testing tool, a global toggle silently changing every saved request's protocol would invalidate comparisons against historical runs. The rejected alternative - a nullable column where NULL means "resolve later" - would have given NULL a second meaning, and this codebase already has a documented null-vs-absent rule where NULL means "reset to a concrete default" at write time.

**Requested and negotiated are different value spaces, deliberately never unified.** Three things are called `httpVersion`. On a request it is `auto` / `http1.1` / `http2` - what was asked for. On a response or a stored trace it is a display string - `HTTP/1.1`, `HTTP/2`, or `""` when nothing was negotiated. `auto` is a request; `HTTP/2` is an outcome. Rendering one where the other belongs would look authoritative and be wrong, so the two are documented at every declaration and checked at every crossing.

**`auto` maps to curl's own default**, established by probing five hosts rather than reading docs - three HTTPS and two cleartext. The cleartext rows are the load-bearing ones: over TLS an unset option and an explicit `2TLS` cannot diverge, so HTTPS alone would not have tested the question. `CURL_HTTP_VERSION_2TLS` means "h2 for HTTPS, 1.1 for HTTP", and it did fall back on cleartext, matching the unset default exactly.

**The protocol is always sent, never elided** - on execute, save and load, even when it equals the default. Same rule the redirect fields follow, same reason: an omitted field lets an engine-side default win silently, which is not the client's call to make.

**`http2` against a plain `http://` URL negotiates 1.1 silently, and the CRUD layer does not warn.** A URL is frequently an unresolved `{{variable}}` at write time, so a warning would either block legitimate saves or fire on most templated URLs. The honest place for that feedback is execution, where the negotiated protocol is knowable - which is what the Raw tab now shows.

## A subtlety in reporting historical runs

A run stored before this branch has no `httpVersion` in its snapshot, and the summary normalises that to `"auto"`. Read that as **"no protocol was recorded"**, not "auto was chosen" - a load run from before this branch hardcoded `2TLS`, and everything before it went out as HTTP/1.1 anyway because nghttp2 was not linked. The displayed value is right; the reasoning behind it is narrower than it looks.

## Pre-existing bugs found and fixed in passing

- **A stored `http2` request loaded into the builder as `auto`**, and the next save silently downgraded it - `initialRequest` copied `followRedirects` and `maxRedirects` but not the new field.
- **`HttpResponse` never declared `httpVersion`**, so the live Raw tab would have kept printing a hardcoded `HTTP/1.1` forever and that whole change would have shipped as a no-op with its tests green.
- **`LoadTestDetail`'s config row was gated on `mode || comment`**, so the protocol chip would have been invisible on an iterations-only load run.
- **`docs/engine/building.md` told readers to run `vcpkg install curl ...`** in its linker-errors section. Following it now installs curl without http2 and fails the new gate test with no clue why.
- **`GET /config` would 500 on one malformed row.** The config serializer parsed an entry's `options` unguarded; a single bad row took down the whole settings screen rather than one entry. The POST validator already guarded the same parse.
- **`followRedirects` / `maxRedirects` were stored on every run and reported nowhere.** They now ride into both `GET /runs`'s summary and the run report. No screen renders them yet - this makes them available to anything reading the API.

## Testing

Engine 439, app 1659, `tsc --noEmit` clean, `mkdocs build --strict` passes.

**Verified manually, because no test covers it:** neither suite performs a real HTTP/2 transfer. The Raw tab was confirmed against a live ALPN-negotiating host showing `GET / HTTP/2` and `HTTP/2 200 OK`.

Behavioural tests are mutation-checked - the fix reverted, the failure observed, the fix restored - for the reporting allowlists (each independently, proving they are not both satisfied by one edit), the badge predicate, the settings label rendering, the load payload's protocol source, and the raw-view builders.

`SettingsMain.enum.test.tsx` carries a C++/TypeScript parity guard: it reads `engine/include/vayu/types.hpp` for the ordered enum and both switch maps, and `engine/src/db/database.cpp` to pin that the seeded options are still *derived* from `all_http_versions()` rather than a literal someone swapped in. It asserts both files read non-empty, because a source scan of `""` satisfies every assertion below it.

## Known gaps and follow-ups

- **Linux and macOS are unproven.** Everything was validated on Windows/MSVC/Schannel. Those platforms link curl statically - a materially different path. CI covers all three; watch the first run.
- **The first CI run will be slow.** Changing `vcpkg.json` rotates the binary-cache key, so curl and nghttp2 rebuild from source on every platform.
- **The load path's negotiated protocol is captured but unread.** `curl_utils.cpp` fills `Response::http_version` for symmetry with the single-request driver; nothing downstream consumes it, since a load run has many exchanges and no single negotiated value to report.
- **`POST /execute` is the one write path that coerces an invalid protocol rather than rejecting it.** `POST /runs` and the requests CRUD both return 400. Neither shipped client can send an invalid value - the renderer sends a typed union, MCP validates with `z.enum` - so this is a consistency gap, not a live hole.
- **The load path has not been manually protocol-tested**, only the single-request path.

Separately ticketed, all pre-existing: duplicating a request drops its entire Settings tab (`followRedirects`, `maxRedirects` and now `httpVersion` are all missing from the duplicate payload); the requests/collections/environments routes emit a flat error body the app's HTTP client cannot read, so this branch's carefully-worded 400s surface as a bare "HTTP 400"; and the engine test suite leaves untracked `*.db.bak` files at the repo root.


athrvk:
Pushed `35ac59a` - a bug the second catch-up merge (`332c299`) left behind in the MCP layer.

**What was wrong.** `start_load_run` advertises an `httpVersion` argument, but on the saved-request branch nothing read it. The catch-up merge put `httpVersion` in `buildExecutionPayload`'s key loop, and `composeLoadRunRequest` only calls that builder for a URL-only run - the `requestId` branch builds its payload from `composeSavedRequest` plus `readRequestOverrides`. Since `composeSavedRequest` always emits a protocol, `start_load_run { requestId, httpVersion: "http2" }` started the run at the request's **stored** protocol, silently.

Confirmed before fixing, against the branch as it stood: a saved request storing `http1.1` plus an agent asking for `http2` put `"httpVersion": "http1.1"` on the wire.

That is the "written but never read" shape CLAUDE.md calls this codebase's most repeated defect, and it was invisible to the existing tests - the only `httpVersion` forwarding test used the URL-only path, which the old placement satisfied. It is a merge artifact, not a mistake in either parent: neither side had a saved-request load path *and* an `httpVersion` override until they met.

**The fix.** Read `httpVersion` with the other agent-stated overrides, where the saved-request path already picks up url/method/headers/body. The ad-hoc path is unaffected - it spreads those overrides into its payload, so a URL-only run still forwards a stated protocol and still omits an absent one.

Two tests: the override beating a stored `http1.1` (so a pass cannot come from both sides agreeing on the `auto` default), and the stored protocol surviving when nothing is stated. Both mutation-checked - reverted to the previous placement, watched the first fail and only it, restored.

Also fixed in the same commit: the merge had spliced `mcp.md`'s **Protocol** bullet into the middle of the **Scripts** bullet, orphaning its continuation. Moved it after, corrected it to describe the saved-request override, and added `httpVersion` to the list of fields that override a composed saved request.

**Verification.** App suite 1924 passed (230 files), `tsc --noEmit` clean, prettier clean on both touched files, `mkdocs build --strict` passes.

The engine was **not** built here, and this is worth knowing for any agent session on this branch: vcpkg cannot fetch `nghttp2` in this environment - `github.com` and `codeload.github.com` return 403 under the sandbox's egress policy, and everything else this branch needs is already cached, so it is only the new dependency that cannot be resolved. CI is the gate for the engine build and its test suite. My change is app-only, so nothing in it depends on that.

Two pre-existing lint failures I did not touch, both reproducing on this branch before my commit: `src/types/domain.ts` (4 x `prettier/prettier`) and `src/modules/settings/main/SettingsMain.tsx:141` (`react-hooks/set-state-in-effect`). For context, `pnpm lint` also fails on master itself (100 errors), so this is not a regression from this PR.

---

PR #176, opened 2026-07-29
feat(engine): pre-request scripts can mutate the outgoing request

### LINKED ISSUE #109: engine(runtime): pre-request scripts cannot mutate the outgoing request, but scripting.md says they can

athrvk:
## Description

Option B for #109, per your instruction on the issue ("close the open PR, and create a new PR which enables this capability rather than just updating the docs"). #149 - the docs-only Option A - is closed in favour of this.

**Plan.** The root cause is a one-way boundary: `setup_pm_request` copies the C++ `Request` into a JS object and nothing ever reads it back, so `pm.request.headers['X-Signature'] = …` mutated a throwaway and the composed request went out unsigned - this codebase's "written but never read" pattern, at the script boundary, while `scripting.md` taught request signing as a supported pattern. The fix closes the loop: after the pre-request script returns, `pm.request`'s `url`/`method`/`headers`/`body` are read back and applied to the same `Request` before `client.send()`. The JS object is treated as **authoritative, not a diff** - the header set it holds is the header set that is sent. The alternative I rejected was diffing the JS object against the original snapshot and applying only what changed: it reads as the safer option, but it cannot express `delete pm.request.headers['Authorization']` (an absent key is indistinguishable from an untouched one), and it breaks entirely for a script that assigns `pm.request = {…}` wholesale. Authoritative is the only rule that stays true in both cases, and it is the one Postman's live-binding model implies.

**Precedence over engine-applied auth is ordering, not a special case.** `build_request` resolves auth (bearer/basic/apikey/oauth2) into the request *before* the script runs - that is pre-existing and is why `pm.request` was accurate in the first place - so a script-set `Authorization` is simply later, and wins. `PreRequestScriptHeaderOverridesEngineAppliedAuth` drives that through the real `build_request` rather than hand-building the header.

**Failure path.** A value the engine cannot send - a non-string `url`, an unknown verb, an object header value, an empty URL - rejects the **whole** write-back, leaves the request byte-for-byte unchanged, and is reported as `preScriptError` (which `ResponseViewer` already renders in the Console tab). All-or-nothing: a script that sets a good header and a bad method sends neither, rather than half of what it asked for. Header values that are numbers or booleans are converted, because `pm.request.headers['X-Timestamp'] = Date.now()` is the shape people actually write and it has one honest wire form.

**Deliberate call worth your eye:** the write-back runs even when the script threw. The edits made before the throw already happened, the request is sent either way, and silently dropping them is the exact defect being fixed. `PreRequestScriptEditsMadeBeforeAThrowStillApply` pins it; the failure is still reported.

**Blast radius traced.** Only `POST /execute` runs pre-request scripts - load runs execute the `tests` (post-request) script only - so no load path changes. `Client::send` sets `response.request_headers`/`raw_request` from the request it is given, and `store_result` stores the same object, both *after* the script, so the Raw tab, `requestHeaders` and the restored trace all report what was actually sent. **No client assumes the composed request equals the sent one** (the issue's App-changes question): the app's Raw tab consumes the engine's `rawRequest`, not a locally composed string. The only app change is one line making the new capability discoverable in the pre-request panel's quick reference.

`ScriptContext` gains `mutable_request`, settable only through `make_request_mutable(Request&)` so the read snapshot and the write target cannot drift apart; a test script leaves it null, since its request has already gone out and a write-back there could only misreport what was sent.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

```
python build.py -e -t
cd engine && ctest --preset linux-dev --output-on-failure   # 521/521 passed (17 new)
cd app && pnpm test                                         # 221 files, 1828 tests passed
cd app && pnpm lint + prettier --check (touched file only)   # clean
mkdocs build --strict -f .github/mkdocs.yml                  # clean
```

**Mutation-checked**: with the `apply_pm_request_writeback` call disabled in `ScriptEngine::Impl::execute`, **14 of the 17** new tests fail. The 3 that still pass are the ones that assert the *old* behaviour is preserved - a test script's mutations are discarded, and a script that edits nothing leaves the request identical.

The 17 cover: header add / delete / number-and-boolean stringification, URL and (case-insensitive) method change, body added to a bodyless request getting `Text` mode so `apply_method_and_body` does not drop it, an edited JSON body keeping `Json` mode, `delete pm.request.body` clearing it, auth precedence, edits surviving a throw, and five rejection cases each asserting nothing was applied.

No pre-existing failures observed. Two files I touched (`script_engine.cpp`, `execution.cpp`) were already not clang-format-clean at the base commit, so per CLAUDE.md only my changed lines were formatted (`git clang-format`); `script_engine_test.cpp` was clean and is formatted in full - its diff is pure insertions.

## Checklist

Docs updated in the same commit: `docs/engine/scripting.md` (new "Mutating the request" section with the authoritative-object, precedence, refusal and `{{var}}` rules; the two execution-context bullets corrected), `docs/app/pm-api-compatibility.md` (the "cannot change the outgoing request" section rewritten - the two docs now agree, in the other direction), `docs/engine/api-reference.md` (`POST /execute`), `docs/engine/architecture.md` (design-mode flow + script engine capabilities). The misleading comments the issue names at `script_engine.hpp:97` and `execution.cpp:384-385` are corrected - they now state the rule rather than a possibility.

## Related Issues
Closes #109

---
_Generated by [Claude Code](https://claude.ai/code/session_018hfndY2LPPzgfAYMGgsKJU)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -368,12 +368,17 @@ of that duplication: both clients now collect an ordered list of `ScriptPart`s
 (root-to-leaf chain, then the request's own, each naming its origin) and send
 the list as `preRequestScripts` / `postRequestScripts` on `POST /execute` - and
 the **engine** joins them with `"\n\n"` and runs the result. The renderer's load
-path sends the same kind of list as `tests` on `POST /runs`. MCP has no
-chain-composing `/runs` caller: `start_load_run` takes an agent-supplied ad-hoc
-`tests` string (like its ad-hoc `preRequestScript`/`postRequestScript`, see
-`tools.ts::buildExecutionPayload`), not a chain-built list, so this is not "both
-clients send the list on `/runs`". Each client still builds its own script-part
-list itself (the `scriptParts` helper in
+path sends the same kind of list as `tests` on `POST /runs`; MCP's
+`start_load_run` sends it as `postRequestScripts` when given a `requestId`
+(`tools.ts::composeLoadRunRequest`, reusing `composeSavedRequest`), or an
+agent-supplied ad-hoc `tests` string for a URL-only run. **Both names reach the
+same script**: `read_post_request_script` (`engine/src/http/script_parts.cpp`)
+owns every spelling the post-request script answers to - stored as
+`postRequestScript`, `postRequestScript(s)` on `/execute`, `tests` on `/runs` -
+and both routes read through it, so a payload composed for one endpoint can
+start the other kind of run unchanged. Add a spelling to that table, never to a
+route. Each client still builds its own script-part list itself (the
+`scriptParts` helper in
 `app/src/modules/request-builder/utils/script-parts.ts` and in
 `app/electron/mcp/resolve.ts` - the same intentional duplication, since MCP
 cannot import from `app/src/`), so a change to the list-building rule (e.g. what
--- a/app/electron/mcp/engine-client.ts
+++ b/app/electron/mcp/engine-client.ts
@@ -107,6 +107,16 @@ export class EngineClient {
 		);
 	}
 
+	/**
+	 * Fetch a single saved request. Unlike environments, the engine does have a
+	 * `GET /requests/:id` route, so this is one round trip rather than a scan of
+	 * every collection's list. A `404` surfaces as an `EngineRequestError`, which
+	 * is what distinguishes "that request was deleted" from an unreachable engine.
+	 */
+	getRequest(id: string, signal?: AbortSignal): Promise<unknown> {
+		return this.request("GET", `/requests/${encodeURIComponent(id)}`, undefined, signal);
+	}
+
 	listEnvironments(signal?: AbortSignal): Promise<unknown> {
 		return this.request("GET", "/environments", undefined, signal);
 	}
--- a/app/electron/mcp/tools.test.ts
+++ b/app/electron/mcp/tools.test.ts
@@ -373,6 +373,44 @@ describe("dispatchTool", () => {
 		});
 	});
 
+	test("run_request forwards an ad-hoc pre-request script the agent supplied", async () => {
+		// Parsed through the tool's own inputSchema first, because that is the
+		// only thing standing between the agent and the engine: `registerTool`
+		// hands the SDK this shape, and a zod object *strips* keys it does not
+		// declare. `buildExecutionPayload` has always read `preRequestScript`
+		// off `args`, but until it was declared here nothing could put it there
+		// - so a request the agent asked to have signed went out unsigned.
+		// dispatchTool alone does not validate, so asserting on it would pass
+		// with the field removed and prove nothing.
+		const tool = TOOLS.find((t) => t.name === "run_request");
+		const args = z.object(tool!.inputSchema as Record<string, z.ZodTypeAny>).parse({
+			url: "https://api.example.com/users",
+			preRequestScript: "pm.request.headers['X-Signature'] = 'abc';",
+			postRequestScript: "pm.test('ok', function () {});",
+		});
+
+		const client = fakeClient();
+		const res = await dispatchTool(
+			"run_request",
+			args,
+			ctxWith(client, { allowlist: ["api.example.com"] })
+		);
+
+		expect(res.isError).toBeFalsy();
+		const payload = (client.executeRequest as ReturnType<typeof vi.fn>).mock.calls[0][0];
+		expect(payload.preRequestScript).toBe("pm.request.headers['X-Signature'] = 'abc';");
+		expect(payload.postRequestScript).toBe("pm.test('ok', function () {});");
+	});
+
+	test("start_load_run does not offer a pre-request script", () => {
+		// POST /runs never runs one - only the deferred `tests` script - so
+		// offering the field would promise a hook that silently does nothing.
+		const loadRun = TOOLS.find((t) => t.name === "start_load_run");
+		expect(loadRun).toBeDefined();
+		expect(Object.keys(loadRun!.inputSchema)).not.toContain("preRequestScript");
+		expect(Object.keys(loadRun!.inputSchema)).toContain("tests");
+	});
+
 	test("run_request resolves {{variables}} in the URL from the environment", async () => {
 		const client = fakeClient({
 			getEnvironment: vi.fn().mockResolvedValue({
--- a/app/electron/mcp/tools.test.ts
+++ b/app/electron/mcp/tools.test.ts
@@ -405,11 +405,120 @@ describe("dispatchTool", () => {
 	test("start_load_run does not offer a pre-request script", () => {
 		// POST /runs never runs one - only the deferred `tests` script - so
 		// offering the field would promise a hook that silently does nothing.
+		// The renderer's load path sends no pre-request script either, so this
+		// asymmetry is the app's, not MCP's.
 		const loadRun = TOOLS.find((t) => t.name === "start_load_run");
 		expect(loadRun).toBeDefined();
 		expect(Object.keys(loadRun!.inputSchema)).not.toContain("preRequestScript");
 		expect(Object.keys(loadRun!.inputSchema)).toContain("tests");
 	});
+
+	test("both execute-shaped tools name the validation script the same way", () => {
+		// One field in the app - the request builder's Tests tab - drives both
+		// Send and a load run. It reached the engine under two names
+		// (`postRequestScript` on /execute, `tests` on /runs), and MCP exposed
+		// whichever name its endpoint used, so a script an agent wrote for
+		// run_request could not be handed to start_load_run unchanged.
+		for (const name of ["run_request", "start_load_run"]) {
+			const shape = TOOLS.find((t) => t.name === name)!.inputSchema as Record<
+				string,
+				z.ZodTypeAny
+			>;
+			expect(Object.keys(shape)).toContain("postRequestScript");
+			// The engine's own spelling stays accepted on both: a zod object
+			// strips what it does not declare, so dropping it from either tool
+			// would turn a script the agent believes is running into silence.
+			expect(Object.keys(shape)).toContain("tests");
+		}
+	});
+
+	/**
+	 * Parse through the tool's own `inputSchema` before dispatching, as the SDK
+	 * does in production: a zod object strips undeclared keys, so a test that
+	 * hands `dispatchTool` the argument directly passes even with the field
+	 * removed from the schema and proves nothing.
+	 */
+	const parseArgs = (name: string, args: Record<string, unknown>) =>
+		z
+			.object(TOOLS.find((t) => t.name === name)!.inputSchema as Record<string, z.ZodTypeAny>)
+			.parse(args);
+
+	test("start_load_run sends postRequestScript to /runs under the key it reads", async () => {
+		const client = fakeClient();
+		const res = await dispatchTool(
+			"start_load_run",
+			parseArgs("start_load_run", {
+				url: "https://api.example.com",
+				confirmed: true,
+				postRequestScript: "pm.test('ok', function () {});",
+			}),
+			ctxWith(client, { allowlist: ["api.example.com"] })
+		);
+
+		expect(res.isError).toBeFalsy();
+		const payload = (client.startRun as ReturnType<typeof vi.fn>).mock.calls[0][0];
+		expect(payload.tests).toBe("pm.test('ok', function () {});");
+		// Forwarding it verbatim would look right and validate nothing: the
+		// engine's run config reads `tests` and never `postRequestScript`.
+		expect(payload.postRequestScript).toBeUndefined();
+	});
+
+	test("start_load_run still accepts the engine's own `tests` spelling", async () => {
+		const client = fakeClient();
+		const res = await dispatchTool(
+			"start_load_run",
+			parseArgs("start_load_run", {
+				url: "https://api.example.com",
+				confirmed: true,
+				tests: "pm.test('a', () => {});",
+			}),
+			ctxWith(client, { allowlist: ["api.example.com"] })
+		);
+
+		expect(res.isError).toBeFalsy();
+		const payload = (client.startRun as ReturnType<typeof vi.fn>).mock.calls[0][0];
+		expect(payload.tests).toBe("pm.test('a', () => {});");
+	});
+
+	test("run_request accepts `tests` as the same script as postRequestScript", async () => {
+		const client = fakeClient();
+		const res = await dispatchTool(
+			"run_request",
+			parseArgs("run_request", {
+				url: "https://api.example.com/users",
+				tests: "pm.test('b', () => {});",
+			}),
+			ctxWith(client, { allowlist: ["api.example.com"] })
+		);
+
+		expect(res.isError).toBeFalsy();
+		const payload = (client.executeRequest as ReturnType<typeof vi.fn>).mock.calls[0][0];
+		expect(payload.postRequestScript).toBe("pm.test('b', () => {});");
+	});
+
+	test.each(["run_request", "start_load_run"])(
+		"%s rejects both names for one script rather than picking one",
+		async (name) => {
+			// Two different scripts under two names for one slot: whichever is
+			// dropped, the agent is told the run validated something it did not.
+			const client = fakeClient();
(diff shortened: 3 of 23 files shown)


athrvk:
Does MCP understands this ?

athrvk:
Good question - it did not, and the reason was a gap worth fixing. Pushed `183b86d`.

**Two halves to the answer.**

**Where MCP runs a saved request, it already understood it - for free.** `run_collection_smoke` composes the collection-chain + request's own pre-request scripts and sends them as `preRequestScripts` on `POST /execute`. The write-back is entirely engine-side, so those scripts now mutate the request with no MCP change at all. Same for anything else that replays a stored script.

**Where an agent writes its own script, it could not - and not because of this PR.** `run_request`'s `inputSchema` declared method / url / headers / body / auth / requestId / environmentId / collectionId and nothing else. `buildExecutionPayload` has copied `preRequestScript` / `postRequestScript` off the arguments since it was written, but `registerTool` hands the SDK that Zod shape, and a Zod object *strips* keys it does not declare - so nothing could ever put those keys there. A read with no reachable writer, which is the same "written but never read" defect in mirror image. It cost nothing while a pre-request script could not change anything; now that its `pm.request` edits reach the wire, it is the difference between an agent being able to sign a request and not.

`run_request` now declares both, and its description says what a pre-request script can do. **`start_load_run` deliberately does not get one:** the engine runs a pre-request script on `POST /execute` alone - `POST /runs` runs only the deferred `tests` script - so offering the field there would promise a hook that silently never fires. A test pins each side.

The forwarding test parses its arguments through the tool's own `inputSchema` before dispatching, because `dispatchTool` does not validate - asserting on it directly would have passed with the field removed and proved nothing. Mutation-checked: remove `preRequestScript` from the schema and it fails; restore it and 34/34 pass.

`docs/engine/mcp.md` gains a **Request mutation** bullet in Request composition covering all three tools, including why `start_load_run` is the exception.

Verified on the new head: app **1830 passed across 221 files** (2 new), `pnpm type-check` clean, eslint + prettier clean on the two touched files, `mkdocs build --strict` clean. Engine untouched by this commit, and its 521/521 from `0d82873` was green on all three OS runners.

---

athrvk:
Engine handles postTestScript and tests as same, should this be exposed via MCP too? 
this is an issue because UI and MCP should have consistency!

athrvk:
Checked this in the engine, and the premise needs one correction before the rest follows - the engine does **not** treat the two keys as the same:

| where | key it reads |
|---|---|
| storage (`requests`, `collections`) | `postRequestScript` |
| `POST /execute` | `postRequestScripts` / `postRequestScript` (`execution.cpp:465`) |
| `POST /runs` | **`tests` only** (`run_manager.cpp:206-211`, `read_script(config, "tests", "tests")`) |

Send `postRequestScript` to `POST /runs` and it is silently ignored; send `tests` to `POST /execute` and the same. One concept, three names, no interchangeability - the equivalence lives in the *clients*, which map the one stored script onto whichever key the target endpoint reads.

**So on consistency, MCP already matches the UI exactly.** The app sends the stored script as `postRequestScripts` on a Send (`request-builder/index.tsx`) and as `tests` on a load run (`index.tsx:460`, `scriptParts(...)` → `tests`). MCP does the same: `run_request.postRequestScript` → `/execute`, `start_load_run.tests` → `/runs`. Neither client is inventing anything the other does not do.

What was genuinely bad is that an **agent** reading two tool schemas saw `postRequestScript` on one and `tests` on the other with nothing connecting them - they read as unrelated features. Pushed `0dae2a6`: each description now names the other and says they are the same script under two endpoint names, and `docs/engine/mcp.md` states the three-name mapping. No key renamed, no alias, no behaviour change - aliasing would put two names for one field inside a single schema and raise a "what if both are set" question, which trades one confusion for another.

**Two things I did *not* do, both your call:**

1. **Unify the engine's names** (accept `tests` on `/execute`, or `postRequestScript` on `/runs`, or converge on one). That is the only change that removes the inconsistency at the root rather than papering over it, and it is a public API change - it belongs in its own issue, not in this PR.

2. **A real UI/MCP gap in this same area, which is bigger than the naming:** `start_load_run` never sends a *saved* request's stored test script, or its collection chain's. The app's load run does - that is what `scriptParts(collectionAncestors, …)` at `index.tsx:460` is for. So "load test this saved request" through MCP silently runs no assertions, while the same action in the UI runs all of them. `CLAUDE.md` and `docs/engine/mcp.md` both record the current split as deliberate ("MCP has no chain-composing `/runs` caller"), so I have not touched it - but it is the concrete version of the consistency problem you are pointing at, and I think it deserves an issue. Say the word and I will file both.

Verified on the new head: app **1830 passed across 221 files**, `pnpm type-check` clean, eslint + prettier clean on the touched file, `mkdocs build --strict` clean. Engine untouched by the last two commits.

---

athrvk:
- accept postRequestScript on /runs unifying things, basically in all the app tests and postRequestScript are aliases of the same thing
 - Fix this as well - A real UI/MCP gap in this same area, which is bigger than the naming: start_load_run never sends a saved request's stored test script, or its collection chain's.
 
 Make sure it's not a patch work, but a maintanable solution!

athrvk:
You are right, and documenting the split was not enough - pushed `2e036f5`, which makes the two tools name it the same way.

**What the inconsistency actually was.** In the app it is *one* field: the request builder's **Tests** tab. Clicking Send sends it as `postRequestScripts` on `POST /execute`; starting a load test sends the same script as `tests` on `POST /runs`. MCP was exposing whichever name its endpoint happened to use - `postRequestScript` on `run_request`, `tests` on `start_load_run` - so an agent that wrote assertions for one could not hand them to the other unchanged, and read as two tool schemas they looked like unrelated features. The engine's two key names are a wire detail; there is no reason for the agent-facing surface to inherit it when the app's does not.

**The fix.** Both tools now declare `postRequestScript`, and `start_load_run` maps it to `tests` when it builds the run payload. Three decisions worth your eye:

- **`tests` stays accepted on both**, not removed. A Zod object *strips* keys it does not declare, so dropping it would turn a script an agent believes is running into silence - the same failure mode as the `run_request` gap in the last commit, in reverse. Declaring it on `run_request` too closes the mirror hole (an agent that learned `tests` from the load tool would otherwise have its assertions silently dropped there).
- **Both names at once is an error, not a precedence rule.** They are one slot; whichever we dropped, the agent would be told a run was validated by assertions that never ran. `ToolArgError` → `errorResult`, before any traffic is generated.
- **`buildExecutionPayload` no longer forwards the script keys at all.** It was copying `postRequestScript` onto the `/runs` body too, where the engine reads only `tests` - inert only because `start_load_run` did not declare the field, and a silent no-op the moment it did. Each handler now adds the script under the key its own endpoint reads.

**`start_load_run` still offers no pre-request script**, and that one is not an MCP inconsistency: the renderer's load path does not send one either (`request-builder/index.tsx` builds `tests` only), because the engine runs a pre-request script on `POST /execute` alone. UI and MCP already agree there. The test now says so.

**On the concurrent commit.** `0dae2a6` landed on this branch while I was working and answered the same question with docs only. This merges it rather than overwriting: its "one post-request script, three engine names" bullet survives (reworded, since the engine keys still are not interchangeable on the wire even though MCP now hides the split), and its two schema descriptions are superseded by the shared ones.

**Verification on the new head** (which also merges `master` at `0797fd0`, resolving the branch's *behind* state - the merge was clean and both suites were re-run after it):

- `python build.py -e -t` then `ctest --preset linux-dev --output-on-failure` - **571/571 passed**.
- `cd app && pnpm test` - **1836 passed across 221 files** (6 new); `pnpm type-check` clean; eslint + prettier clean on the touched TS files; `mkdocs build --strict` clean. `docs/engine/mcp.md` was already not prettier-clean at the base commit, so per CLAUDE.md it is left unformatted.

Mutation-checked, each independently: drop the `tests` mapping in `start_load_run` and the two forwarding tests fail; remove the both-names rejection and both rejection tests fail; remove `tests` from `run_request`'s schema and three fail. The dispatch tests parse their arguments through the tool's own `inputSchema` first, as the SDK does - handed to `dispatchTool` directly they would pass with the field deleted and prove nothing.

`docs/engine/mcp.md` gains the mapping and the rule in the same commit.

---

athrvk:
Document the most complex possible cases of pre req script modification available via this new capability. also the UI needs to impart this info to the user.
Also the auto completions in the monaco editor needs to be added since we have new functionality available.

athrvk:
All three done in `f887fff`, plus a merge of the concurrent naming commit in `d568d20`. One of them turned up something that changes what the docs can honestly say.

**The sandbox has no crypto, and the docs have been lying about it for a long time.** Before writing signing examples I checked what a script can actually reach: QuickJS plus exactly two globals, `pm` and `console`. No `crypto`, no `btoa`/`atob`, no `TextEncoder`, no `URL`/`URLSearchParams`, nothing async. So a real HMAC is **not possible in-script today** - and `scripting.md`'s pre-request example has called `computeHash(data)` for as long as it has existed, a function that has never been defined anywhere. Same defect as the one this branch set out to fix, one section further down. It is gone; the signing example is now a pure-JS FNV-1a with "not a cryptographic hash - do not use it as one" on it, and a **What a script can compute** section states both lists. Two tests pin the present and absent globals, so if `crypto` ever lands, the note that says it hasn't fails rather than rots.

**1. Worked examples** (`Worked examples: rewriting a request`): rewriting a JSON body and deriving a header from the *result*, adding a query parameter across its three cases (no query / absent / already present, fragment preserved) since there is no `URL` to lean on, switching method and body together, swapping engine-applied auth for a custom scheme, the checksum above, and handing a nonce to the test script. The non-trivial ones **execute against a real request** in `script_engine_test.cpp` rather than being eyeballed.

That paid for itself immediately: I wrote that header names are case-insensitive so a lower-case `delete` works. **It does not.** `pm.request.headers` is a plain JS object, so `delete pm.request.headers['authorization']` leaves a capitalised `Authorization` standing - the test failed on the sentence I had just written. The doc now shows the `Object.keys(...).forEach` loop instead, and I closed the ambiguity underneath it: two keys differing only in case are **rejected** rather than resolved by enumeration order, because one of them is usually the Authorization header and "whichever came last" is not a rule anyone should rely on.

**2. The panel** now carries the rules a snippet cannot show, as a list under the quick reference: the object is authoritative, a script beats the **Auth** tab, names are case-sensitive, a bad value refuses the whole edit with the reason in the Console tab, and load tests run no pre-request script at all. The Tests panel says the mirror image - `pm.request` is readable there but writing does nothing. Its intro also now states the capability outright. Guarded by `script-panels.test.tsx`, which asserts the list is non-empty before checking its contents.

**3. Completions** are engine-served (`scripting.cpp` → `useScriptCompletionProvider`), so that is where they were added. The four fields now read `(writable pre-request)` with the rules in their hover docs, and there are three new snippets: set a header, remove a header, rewrite a JSON body. `script_completions_test.cpp` guards the shape Monaco needs - notably that a snippet carrying `${1:...}` is marked `InsertAsSnippet`, since without it the placeholders insert as literal text and read as a typo. Mutation-checked: drop that rule from one snippet and the test fails.

Verified on the new head: engine **589/589**, app **1864 passed across 223 files**, `pnpm type-check`, eslint, prettier and `mkdocs build --strict` all clean. Only my changed lines were clang-formatted in the files that were already format-dirty at base.

---

PR #178, opened 2026-07-29
fix(engine): join run workers on shutdown instead of detaching them

### LINKED ISSUE #125: engine: shutdown mid-run - detached run worker uses freed Database/RunManager and races curl_global_cleanup

athrvk:
## Description

**Plan.** The root cause is ownership: `start_run` detached the run worker, so nothing could join it, and the daemon's drain used `active_count() == 0` as its "worker is done" signal. That signal is wrong twice over - a worker reaches it at `retain_run`, its *final statement*, with the thread still unwinding, and the loop gave up outright after 5s - after which `main` ran `curl_global_cleanup` and destroyed the `Database` and `RunManager` the worker holds by reference. The fix makes `RunManager` own the worker handles and adds an ordered `shutdown()`: signal `should_stop` on every active run, wait up to `RUN_SHUTDOWN_GRACE_MS` for them to settle, then **join** every worker, before curl teardown and before `db` / `run_manager` leave scope.

The alternative I rejected was keeping the handle in `RunContext::worker_thread` (the field that already existed) and joining it there. It cannot work: the worker captures a `shared_ptr` to its own context, so a retained run swept by the TTL sweeper while its worker is still unwinding drops the last reference *on that worker*, and `~RunContext` would join the calling thread to itself - `std::terminate`. The manager joins from a thread that is never the worker. The now-dead `worker_thread` field is deleted rather than left as a field nothing writes.

I verified the anchors against current master first, as the issue asks: PR #138's stop/DELETE rework moved the surrounding lifecycle code but left the defect exactly as described - `run_manager.cpp` still spawned and detached, and `~RunManager` still only called `stop_sweeper()`.

**Two consequences worth naming, both deliberate:**

- **The grace period bounds the wait, not the join.** A worker that has not settled in 5s is logged and still joined. Letting go of it is precisely the use-after-free being fixed, so the bound exists to make a slow shutdown *visible*, not permissible.
- **`POST /runs` gains a `503`.** The HTTP server stops before the drain, but a request already inside the handler could otherwise spawn a worker past it. `start_run` now returns `false` once the drain has begun and the route answers `503 {"error": "Engine is shutting down"}` rather than a `202` for a run nothing will execute. `workers_mtx_` is held from that check through the handle insert, because anything narrower lets `shutdown()` snapshot and join in between.

A thread also cannot reap itself, so a finished worker's handle outlives it; the next `start_run` collects them, which keeps a long-lived daemon from accumulating one per run.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

New `engine/tests/run_shutdown_test.cpp` (5 tests), driving real runs through `RunManager` against the in-process mock's `/hang` endpoint:

- `ShutdownStopsAndJoinsAWorkerMidRun` - the handle is tracked while the worker runs, `shutdown()` returns with `active_count()` and `tracked_worker_count()` both zero, and the run's status is already terminal.
- `DestructorDrainsAnActiveRun` - the backstop, asserted from outside the manager's scope.
- `StartRunAfterShutdownIsRefused`, `ShutdownWithNoActiveRunsIsPromptAndIdempotent`, `FinishedWorkersAreReapedByTheNextStartRun`.

**Mutation-checked, both directions:**

- Restore `detach()` on the spawned worker → `ShutdownStopsAndJoinsAWorkerMidRun`, `DestructorDrainsAnActiveRun` and `FinishedWorkersAreReapedByTheNextStartRun` fail.
- Remove the `shutdown()` call from `~RunManager` → `DestructorDrainsAnActiveRun` aborts (a joinable `std::thread` destroyed unjoined is a `std::terminate`, which is the loud version of the same defect).

Full verification on this branch:

```
python build.py -e -t                                      # clean, no new warnings
cd engine && ctest --preset linux-dev --output-on-failure   # 510/510 passed (505 before, +5 new)
cd app && pnpm test                                         # 221 files, 1828 tests passed
bash scripts/pre-commit                                     # clang-tidy clean on all 6 touched files
mkdocs build --strict -f .github/mkdocs.yml                 # clean
```

No pre-existing failures observed. `run_manager.cpp`, `run_manager.hpp` and `execution.cpp` were not clang-format-clean at the base commit, so per CLAUDE.md only the added lines were formatted (`git clang-format`).

**App changes: none, verified.** The issue asks for this explicitly. `app/electron/sidecar.ts` is the only `POST /shutdown` caller; the route sets its content and *then* signals the daemon from a separate thread, so the client still gets its `200` before the drain begins - unchanged by this PR. Worth knowing for a future sitting: the app SIGKILLs the engine `ENGINE_GRACEFUL_EXIT_TIMEOUT_MS` (5s) after that response, so a drain that outruns the grace period can still be killed - safely now, since the process dies rather than the worker outliving its database.

## Checklist

Docs in the same commit: `docs/engine/architecture.md` (Run Manager - the drain-then-join ordering and why the manager owns the handles; Thread Model - the full shutdown order) and `docs/engine/api-reference.md` (the new `503` on `POST /runs`, plus the status-code table).

## Related Issues
Closes #125

---
_Generated by [Claude Code](https://claude.ai/code/session_018d4HAMSBYD1CTP9udnNuQe)_

### DIFF
--- a/docs/engine/api-reference.md
+++ b/docs/engine/api-reference.md
@@ -926,6 +926,13 @@ The sample rates are additionally clamped to &ge; 1 inside the metrics
 collector, so the modulo cannot divide by zero even for a caller that bypasses
 this route.
 
+**Shutdown refuses new runs.** Once the daemon has begun draining its run
+workers, `POST /runs` answers `503 {"error": "Engine is shutting down"}` rather
+than accepting a run nothing will ever execute. The window is small - the HTTP
+server stops before the drain begins - but it is not empty, and a request
+already in a handler when the drain starts must not be able to spawn a worker
+past it (see `RunManager::shutdown`).
+
 **Auth pre-flight.** When `auth.mode` is `oauth2`, the run route resolves the
 token **before** creating the run and warms the cache for the workers. An
 unauthorizable config is rejected up front with `409` (interactive sign-in
@@ -1483,6 +1490,7 @@ Get script engine API completions for UI autocomplete.
 | 409 | OAuth 2.0 interactive authorization required (`/run` pre-flight, `/oauth2/token`) |
 | 500 | Internal server error |
 | 502 | Upstream network error (OAuth 2.0 token endpoint, `/import/fetch` proxy) |
+| 503 | Engine is shutting down (`POST /runs` only - see below) |
 
 ## Notes
 
--- a/docs/engine/architecture.md
+++ b/docs/engine/architecture.md
@@ -115,8 +115,17 @@ Manages the lifecycle of load test runs:
   counts still agree.
 - **The move to the retained map is the "worker is finished" signal**: it happens after the
   final metrics flush and status update, which is what `DELETE /runs/:id` waits on before
-  removing rows (see the API reference).
-- **Graceful shutdown**: Stops active runs on daemon shutdown
+  removing rows (see the API reference). It is *not* the "worker thread has exited" signal -
+  the move is the worker's last statement, and the thread is still unwinding after it.
+- **Graceful shutdown drains, then joins**: `RunManager::shutdown()` sets `should_stop` on
+  every active run, waits up to 5s (`RUN_SHUTDOWN_GRACE_MS`) for them to reach a terminal
+  status, and then **joins** every worker thread. The manager owns those thread handles -
+  a worker holds a `shared_ptr` to its own `RunContext`, so a context that owned its thread
+  could end up joining itself. The wait is bounded; the join is not, because abandoning a
+  worker leaves it writing through references to a `Database` that `main` is about to
+  destroy. A run started while the drain is in progress is refused with a `503`.
+- **Finished workers are reaped on the next `start_run`**: a thread cannot join itself, so
+  its handle outlives it until another thread collects it.
 
 ### Metrics Collector
 
@@ -364,6 +373,12 @@ prompt cancellation rather than waiting for in-flight requests to drain.
 - **Worker Threads**: One per active load test (executes load strategy)
 - **Metrics Thread**: One per active load test (aggregates and streams metrics)
 - **Event Loop Threads**: One per CPU core (handles curl_multi I/O)
+
+Shutdown unwinds that in a fixed order, because every one of these threads
+holds references to state `main` owns: HTTP server stopped → run workers
+signalled and joined (each joins its own metrics thread and stops its event
+loop first) → `curl_global_cleanup` → `Database` / `RunManager` destroyed at
+scope exit. Nothing is detached.
 
 ## Performance Characteristics
 
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -400,6 +400,7 @@ if(VAYU_BUILD_TESTS)
         tests/run_manager_test.cpp
         tests/run_route_test.cpp
         tests/run_stop_test.cpp
+        tests/run_shutdown_test.cpp
         tests/run_config_validation_test.cpp
         tests/script_compose_test.cpp
         tests/encoding_test.cpp
--- a/engine/include/vayu/core/constants.hpp
+++ b/engine/include/vayu/core/constants.hpp
@@ -156,6 +156,12 @@ constexpr int DEFAULT_LIVE_REPLAY_WINDOW_MS = 300000;
 constexpr size_t DEFAULT_MAX_LIVE_TICKS = 50000;
 /// Size of the context pool for request handling
 constexpr size_t CONTEXT_POOL_SIZE = 64;
+/// How long `RunManager::shutdown` waits for signalled runs to reach a terminal
+/// status before it logs that they have not. The *wait* is bounded; the join
+/// that follows it is not, because abandoning a still-running worker is exactly
+/// the use-after-free the drain exists to prevent. Matches the 5s the daemon
+/// waited before the drain was ordered.
+constexpr int64_t RUN_SHUTDOWN_GRACE_MS = 5000;
 } // namespace server
 
 /**
--- a/engine/include/vayu/core/run_manager.hpp
+++ b/engine/include/vayu/core/run_manager.hpp
@@ -8,6 +8,7 @@
  */
 
 #include <atomic>
+#include <chrono>
 #include <condition_variable>
 #include <deque>
 #include <functional>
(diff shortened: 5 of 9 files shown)


athrvk:
CI went red on `C++ Tests (windows-latest)` with one failure - `MetricsCollectorTest.SuccessCountNeverExceedsTotalRequests`. **It is not this PR's, and it is not a Windows quirk.** I have re-run the failed jobs; the detail is below because the test is genuinely racy and will bite again.

**Not mine, by construction.** `metrics_collector.cpp`, `metrics_collector.hpp` and `metrics_collector_test.cpp` are byte-identical to master on this branch (`git diff origin/master...HEAD` on those three paths is empty). This PR touches `run_manager`, `daemon`, the `POST /runs` handler, and its own new test file.

**Reproduced locally on Linux** - 1 failure in 200 runs of the existing binary:

```
tests/metrics_collector_test.cpp:449: Failure
Expected: (collector->success_count ()) <= (collector->total_requests ()), actual: 1 vs 0
terminate called without an active exception
```

**Root cause - two defects in the test, both latent since #169:**

1. **The assertion samples its two operands in an order C++ does not fix.** `ASSERT_LE` passes both values as arguments to gtest's comparison helper, so their evaluation order is unspecified, and every compiler I checked evaluates the right one first. `record_error` bumps `total_requests_` and *then* `total_errors_` (`metrics_collector.cpp:173-174`), so a reader between those two increments sees `total=1, errors=0` and `success_count()` returns 1 - correctly, that window is exactly what the guarded subtraction documents. The failing interleaving is the first iteration: `total_requests()` is evaluated first and reads `0`, the writer then performs its first `record_error`, and `success_count()` reads the gap as `1`. Hence `1 vs 0`. Sampling into two locals in a defined order makes the assertion mean what it says.

2. **The `ASSERT_` aborts the test body while `writer` is still joinable**, so `~std::thread` calls `std::terminate` - the `terminate called without an active exception` line above. The abort destroys the evidence: the process dies before gtest prints the summary. `ASSERT_` inside a fixture needs the thread joined first (or the check needs to be an `EXPECT_` with a `break`).

I have deliberately **not** fixed it here - it is unrelated to #125 and belongs in its own change rather than hidden inside this diff. Happy to send that as a separate PR on request; it is a few lines in the test only, no production change.


---

Issue #192, opened 2026-07-29
Importer hardening (review sweep 2026-07-29)


athrvk:
Tracking issue for hardening the import pipeline (`app/src/services/importers/`): the Postman, Insomnia, and OpenAPI parsers, and the post-#145 orchestrator/`POST /import/apply` flow. Every defect is verified against master `4136c66` with file:line anchors in the linked sub-issues; each sub-issue is self-contained and implementer-ready.

## Sub-issues

- **#195 - Postman importer.** Headline: the AWS Signature auth type on the wire is `awsv4`, but `mapPostmanAuth` matches `"aws"`, so every real Postman AWS export silently imports as `mode: "none"` with the config discarded AND the `nonExecutableAuth` warning counter suppressed (`shared.ts:107`, verified against the live v2.1.0 schema). Plus: explicit folder-level `noauth` collapses to `none` which the inheritance walk then skips (requests regain the parent's credentials - fix spans `auth-resolution.ts` and the MCP `resolve.ts` copy together); GraphQL `variables` kept as a JSON-encoded string; `protocolProfileBehavior` redirect settings never read; a `decodeURIComponent` crash on malformed percent-sequences fails the whole import; three smaller gaps.
- **#196 - Insomnia + OAuth2 importer.** Headline: non-listed text bodies (XML etc.) dropped wholesale with no skipped counter (`insomnia-v4.ts:105`). Plus: bearer `prefix` and OAuth2 `tokenPrefix` silently dropped (Authorization scheme changes), folder-level scripts dropped (real Insomnia feature; export key shape to be confirmed against a live 9.x export before implementation), header/param descriptions dropped, per-request redirect settings not importable, unhelpful crash on shape-invalid files, a dead doc path.
- **#193 - OpenAPI importer.** Path-item-level `$ref` never resolved (whole path silently vanishes); urlencoded/multipart body schemas behind `$ref`/`allOf` yield zero form fields; Swagger 2 `formData` ignores `consumes`; server URL template variables/defaults dropped; 3.1 claimed by the detector but 3.1 keywords ignored by the sampler; TRACE dropped without a skipped record; non-array `parameters` crashes parse.
[...]

Issue #189, opened 2026-07-29
app: burn down the ESLint debt (100 errors) and enforce lint in the PR CI matrix


athrvk:
Base: master `4136c66` (2026-07-29).

## Problem

`cd app &amp;&amp; pnpm lint` fails with **100 errors and 8 warnings**, and **no CI workflow runs it** - `rg &#34;lint&#34; .github/workflows/*.yml` matches only the PR-title linter, and the freshly split PR matrix (PR #168: app / engine / installer jobs) runs tests and type-check but not ESLint or `format:check`. So the debt is invisible in review and grows silently; `CLAUDE.md`&#39;s own rule (&#34;Strict TypeScript - no `any`&#34;) is unenforced.

Breakdown by rule (verified at base):

| count | rule |
|---|---|
| 78 | `@typescript-eslint/no-explicit-any` |
| 8 | `react-hooks/set-state-in-effect` |
| 6 | `react-refresh/only-export-components` |
| 6 | `prettier/prettier` |
| 3 | `react-hooks/exhaustive-deps` |
| 3 | `@typescript-eslint/no-unused-vars` |
| 2 | `no-useless-escape` |
| 1 each | `vayu/app`, `react-hooks/purity`, `react-hooks/preserve-manual-memoization` |

6 errors are auto-fixable with `--fix`.

## Fix / solution pathway

Two-step, in one PR or two:

1. **Make `pnpm lint` exit 0.** The 78 `no-explicit-any` errors are the bulk: replace each `any` with a real type, `unknown` + narrowing, or - where a third-party boundary genuinely requires it - a line-level `eslint-disable` **with a justification comment** (the repo convention already demands justification for `@ts-ignore`; hold `any` to the same bar). The `react-hooks/*` errors are behavioral - fix them properly, do not disable (each is a real stale-closure / effect-loop hazard). Run `--fix` for the mechanical 6. Per CLAUDE.md, do NOT run prettier/`eslint --fix` repo-wide - fix only the flagged files, and never format `docs/design-system.md`.
2. **Enforce it.** Add `pnpm lint` (and `pnpm format:check` if wanted) to the app job in `.github/workflows/` alongside the existing test/type-check steps, so a PR reintroducing an error fails CI.

## App changes

This is all app-side; no engine involvement.

## Tests

[...]

PR #208, opened 2026-07-30
fix(app): openapi importer - resolve $ref'd path items and form bodies, honor consumes

### LINKED ISSUE #193: app(import): OpenAPI importer - unresolved $refs drop paths and form bodies, server variables lost, 3.1 gaps
Part of #192. Base: master `4136c66`.

## Problem

Seven related gaps in the OpenAPI 3.0 / Swagger 2.0 importers (`app/src/services/importers/openapi-v3.ts`, `openapi-v2.ts`, `schema-sampler.ts`). All anchors verified against the base commit above by reading the current file. The two silent-loss items - a `$ref`'d path item dropping a whole path, and a `$ref`/`allOf` form-body schema yielding zero fields - are the core of this issue: both produce a clean-looking import with data missing and no signal anywhere in the UI.

### 1. Path-item-level `$ref` never resolved - a whole path silently vanishes

`openapi-v3.ts:68-72`:

```ts
for (const [path, pathItem] of Object.entries(spec.paths ?? {})) {
    const pathParams = (pathItem as any)?.parameters ?? [];
    for (const method of HTTP_METHODS) {
        const op = (pathItem as any)?.[method];
        if (!op) continue;
```

OpenAPI 3.1 allows a path item to be `{"$ref": "#/components/pathItems/X"}`; 3.0 also supports path-item refs. Such an item has no method keys, so `op` is always `undefined` and every operation under that path is skipped - no request built, `requestCount` not incremented, and `meta.skipped` stays `[]` (hardcoded at `openapi-v3.ts:107`; identical structure and identical hardcoded `[]` at `openapi-v2.ts:69-75,119`). `$ref` resolution exists for parameters (`openapi-v3.ts:147`, `openapi-v2.ts:150`) and `requestBody` (`openapi-v3.ts:188`) but never at the path-item level. The doc's "Dropped / not represented" list (`docs/app/import-collections/openapi-v3.md:161-169`) does not mention path-item refs, and `docs/app/import-collections/openapi-v2.md:222-232` likewise omits it.

**Failure scenario:** A bundled single-file spec hoists a shared path item into `components.pathItems` and references it (`"/users/{id}": {"$ref": "#/components/pathItems/UserOps"}`). Import succeeds, the preview shows a plausible collection, `requestCount` simply excludes the missing endpoints, and nothing flags the drop. The user discovers the missing `/users/{id}` operations only by comparing against the source spec.

### 2. Urlencoded/multipart body schema behind `$ref` or `allOf` yields zero form fields

`openapi-v3.ts:199-206`:

```ts
for (const ct of ["application/x-www-form-urlencoded", "multipart/form-data"] as const) {
    if (content[ct]) {
        const props = content[ct].schema?.properties ?? {};
        const fields = Object.keys(props).map((k) => ({ key: k, value: "", enabled: true }));
        return {
            mode: ct === "multipart/form-data" ? "form-data" : "x-www-form-urlencoded",
            fields,
        };
    }
}
```

Unlike the JSON branch (`openapi-v3.ts:192-196`, which runs the schema through `sampleSchema` - full `$ref`/`allOf`/`oneOf`/`anyOf` resolution, see `schema-sampler.ts:30-47`) and unlike `requestBody`-level `$ref` (resolved at `openapi-v3.ts:188`), the form branch reads `content[ct].schema?.properties` directly with no ref/composition resolution. A form schema written as `{"$ref": "#/components/schemas/LoginForm"}` (the norm for generated specs - e.g. FastAPI's `Body_login_...` schemas behind `OAuth2PasswordRequestForm`) has no literal `properties`, so `fields` comes out `[]`. The body mode is still set correctly; every field name defined in the spec is silently lost. `openapi-v3.test.ts` has no form-body test at all, so nothing pins the behavior. The doc (`docs/app/import-collections/openapi-v3.md:108-112`) promises "one entry per `schema.properties` key" but the promise silently fails when `properties` sits behind a `$ref`.

**Failure scenario:** `requestBody: {content: {"application/x-www-form-urlencoded": {schema: {$ref: "#/components/schemas/TokenRequest"}}}}` where `TokenRequest` (in `components`) has `grant_type`/`username`/`password`. Import produces `{mode: "x-www-form-urlencoded", fields: []}`. The user sees the right mode with an empty field list, gets no skipped-item notice, and must reconstruct every field name by hand from the original spec.

### 3. Swagger 2 `formData` always becomes multipart `form-data`, ignoring `consumes`

`openapi-v2.ts:183`:

```ts
if (formFields.length > 0) body = { mode: "form-data", fields: formFields };
```

`consumes` is already computed at `openapi-v2.ts:139-145` and used to pick JSON vs. text for `in: "body"` parameters (`openapi-v2.ts:173-175`), but the `formData` branch (`openapi-v2.ts:178-183`) never reads it. Swagger 2.0 ties `formData` parameter encoding to `consumes`: `application/x-www-form-urlencoded` vs. `multipart/form-data` are distinct wire encodings, and Vayu models them as distinct `BodyMode`s (`app/src/types/domain.ts:12`) which the v3 parser maps correctly (`openapi-v3.ts:199-206`). The v2 parser collapses both to `"form-data"` unconditionally. `openapi-v2.test.ts` has no `formData` test. The doc (`docs/app/import-collections/openapi-v2.md:137`) pins the unconditional mapping, but the pin is wrong against the Swagger 2.0 spec.

**Failure scenario:** A classic Swagger 2 login/token endpoint: `consumes: ["application/x-www-form-urlencoded"]`, parameters `in: formData` for `username`/`password` (the canonical Swagger petstore `updatePetWithForm`-style pattern). Imports as multipart `form-data`; when the user fills values and sends, the request goes out as `multipart/form-data` to a server expecting urlencoded and is rejected (400/415). The fields are present and correct, so the encoding mismatch is non-obvious until execution.

### 4. Detector claims OpenAPI 3.1 support; sampler ignores 3.1 schema keywords

`docs/app/import-collections/openapi-v3.md:24` states the detector matches `"3.0.0, 3.0.3, and 3.1.x"` (code: `openapi-v3.ts:44-47`, `v.startsWith("3.")`), but `schema-sampler.ts:49-72` switches on `schema.type` for string values only:

```ts
switch (schema.type) {
    case "string": ...
    case "integer":
    case "number": ...
    case "boolean": ...
    case "array": ...
    case "object":
    default: { /* walks properties, else {} */ }
}
```

OpenAPI 3.1 adopted plain JSON Schema, where a nullable field is commonly written as a type array (`type: ["string", "null"]`) rather than 3.0's `nullable: true`. A type array matches no `case`, falls to the `object`/default branch, and - having no `properties` - returns `{}` where a scalar (`""`/`0`) belongs. `schema.const` and `schema.examples` (both 3.1/JSON-Schema keywords) are never read either (only the singular `example` at `schema-sampler.ts:42`). `schema-sampler.test.ts` has no type-array or `const` case.

This is real but narrower than it first looks: the `anyOf: [{type:"string"},{type:"null"}]` nullable style (common in FastAPI-era 3.1 generators) is handled correctly via the `allOf`/`oneOf`/`anyOf` first-branch rule (`schema-sampler.ts:44-47`), and the output is a synthetic body stub the user edits anyway - not user data. Pick one of two fixes, do not leave the mismatch standing:

- **(a) Support the keywords** - handle `Array.isArray(schema.type)` (sample using the first non-`"null"` entry) and read `schema.const` (return it verbatim, same precedence tier as `example`) in `schema-sampler.ts`, so the detector's 3.1 claim is actually true for schema sampling.
- **(b) Narrow the claim** - stop `detect()` matching `3.1.x`, or add an explicit doc caveat that 3.1-only schema keywords (type arrays, `const`) are not sampled and degrade to `{}`.

The doc must state which was chosen either way.

### 5. TRACE operations dropped with no skipped record

`openapi-v3.ts:20`: `HTTP_METHODS = ["get", "post", "put", "patch", "delete", "head", "options"]` omits `trace`, which OpenAPI 3's Path Item Object supports. A `trace:` operation is never read, never counted in `requestCount`, and never recorded in `meta.skipped` (hardcoded `[]`, `openapi-v3.ts:107`). This is forced in part: `app/src/types/domain.ts:10` defines `HttpMethod` without `"TRACE"`, so the operation genuinely cannot be represented today. `openapi-v2.ts:20`'s identical list omitting `trace` is spec-correct (Swagger 2.0's Path Item Object has no `trace`), so this item is v3-only.

The doc's method list (`docs/app/import-collections/openapi-v3.md:35`) documents which methods the loop processes but does not say TRACE is intentionally excluded, and the "Dropped / not represented" section (`openapi-v3.md:161-169`), which is exactly where a reader would look for this, lists scripts/environments/additional-servers/response-schemas/cookie-params/multi-tag but omits TRACE entirely - this is the same "written but never read" doc gap pattern as the other findings in this issue, not a documented, deliberate limitation. Confirmed against the tracking issue's own summary (#192), which independently lists this as a real gap in the OpenAPI slice.

**Failure scenario:** A spec with a `trace:` operation imports with that operation missing; `requestCount` in the preview matches only the non-trace operations, so the import looks complete, and no skipped item explains the absence.

### 6. Non-array `parameters` crashes the whole parse

`openapi-v3.ts:146`: `for (const param of [...pathParams, ...(op.parameters ?? [])])` spreads `pathParams` (`openapi-v3.ts:69`: `(pathItem as any)?.parameters ?? []`, no `Array.isArray` check) and `op.parameters`. A malformed spec with `parameters` written as an object instead of an array (a common hand-edited-YAML mistake - a missing `-` under `parameters:`) throws `TypeError: ... is not iterable` at the spread, aborting the entire import with no indication of which path caused it. Identical pattern at `openapi-v2.ts:70,149`. `ImportModal.tsx:79-90` catches the exception and displays the raw message, so this is loud, not silent, but the message names no offending path and blocks importing every other valid endpoint in the file. Neither `openapi-v2.test.ts` nor `openapi-v3.test.ts` covers a non-array `parameters` shape.

**Failure scenario:** Hand-authored YAML with `parameters:` accidentally written as a mapping. Import fails wholesale with a cryptic "is not iterable" error instead of skipping the malformed block or naming the offending path.

## Fix pathway

1. **Path-item `$ref` (item 1).** Before the `HTTP_METHODS` loop in both parsers, resolve a `pathItem.$ref` the same way `resolveRef` already resolves parameter/`requestBody` refs (`openapi-v3.ts:51-59`, `openapi-v2.ts:47-55`): `const resolvedItem = (pathItem as any)?.$ref ? resolveRef((pathItem as any).$ref) : pathItem;` and iterate methods on `resolvedItem`.
2. **Form-body `$ref`/`allOf` (item 2).** Route `content[ct].schema` through the existing `sampleSchema`/`resolveRef` machinery before reading `.properties`, e.g. `const resolved = content[ct].schema ? walkToProperties(content[ct].schema, resolveRef) : {}` - reuse the resolution `sampleSchema` already performs (`schema-sampler.ts:30-47`, the `$ref` and `allOf`/`oneOf`/`anyOf` first-branch handling) rather than re-implementing it; the simplest route is exporting a properties-resolving variant (or calling `sampleSchema` and reading `Object.keys()` off the produced stub object) so form fields see the same `$ref`/`allOf` resolution JSON bodies already get.
3. **Swagger 2 `formData` consumes (item 3).** In `openapi-v2.ts`, mirror the v3 content-type branch: decide the form `BodyMode` (`"form-data"` vs `"x-www-form-urlencoded"`) from the already-computed `isJsonConsume`-style check against `consumes`/`spec.consumes`, using the same `application/x-www-form-urlencoded` vs `multipart/form-data` test `openapi-v3.ts:199` uses, instead of unconditionally setting `"form-data"` at line 183.
4. **3.1 keyword gap (item 4).** Pick (a) or (b) above and implement it: either extend `schema-sampler.ts`'s `switch` to handle `Array.isArray(schema.type)` and read `schema.const`, or narrow `detect()`/the doc to stop claiming full 3.1 support. State the choice explicitly in the PR description and the doc update.
5. **TRACE (item 5).** Record TRACE (and any other operation whose method Vayu cannot represent) in `meta.skipped` instead of silently omitting it from `requestCount` - add a `SkippedItem` kind for "unsupported method" (`app/src/services/importers/types.ts`) and push one entry per dropped operation, so the previously-hardcoded `skipped: []` gains its first real producer in this parser. Do not attempt to add `"TRACE"` to `HttpMethod`/`domain.ts` as part of this fix - that is a larger, unrelated change to the execution model.
6. **Non-array `parameters` (item 6).** Guard both parsers' parameter assembly with a shape check before the spread, e.g. `const rawParams = Array.isArray(pathParams) ? pathParams : []` and `Array.isArray(op.parameters) ? op.parameters : []`, so a malformed `parameters` block is treated as empty (and, ideally, recorded via the same skipped mechanism as item 5) rather than throwing and aborting the whole file.

## App changes

None, verified. All seven items are contained within `app/src/services/importers/openapi-v3.ts`, `openapi-v2.ts`, and `schema-sampler.ts` - pure parsing logic that runs before any engine call. The engine is not involved in OpenAPI/Swagger parsing (`app/src/services/importers/` builds `CollectionDraft`/`RequestDraft` client-side; only the resulting `POST /import/apply` payload reaches the engine, and none of these fixes change that payload's shape). No MCP surface touches these parsers either (`rg -n "openapi-v3|openapi-v2|schema-sampler" app/electron` returns no hits).

## Tests

Vitest, fixture-driven, mutation-checked (apply the fix, write the test, revert the fix locally, confirm it fails, restore the fix) per this repo's convention:

- `openapi-v3.test.ts`: a spec with a `components.pathItems`-backed `$ref` path item asserts the operations under it are imported (not silently absent); a `$ref`'d urlencoded/multipart `requestBody` schema asserts non-empty `fields` matching the referenced schema's `properties`; a `trace:` operation asserts a `SkippedItem` is recorded and `requestCount` is unaffected; a non-array `parameters` on a path item or operation asserts the parser does not throw (either treats it as empty or records a skip).
- `openapi-v2.test.ts`: a `formData` parameter set with `consumes: ["application/x-www-form-urlencoded"]` asserts `body.mode === "x-www-form-urlencoded"`, and with `consumes: ["multipart/form-data"]` (or absent, defaulting per the chosen rule) asserts `body.mode === "form-data"`; same non-array `parameters` guard as v3.
- `schema-sampler.test.ts`: if item 4 takes route (a), a `type: ["string", "null"]` node samples to `""` (not `{}`) and a `const`-only node samples to the `const` value; if route (b), no sampler test is needed but `openapi-v3.test.ts`'s detect-boundary test should assert the narrowed version match.

## Docs

Same commit:

- `docs/app/import-collections/openapi-v3.md`: update the "Dropped / not represented" list (lines 161-169) to include path-item `$ref` (until fixed) and TRACE; update the request-body table (lines 104-112) to state form bodies now resolve `$ref`/`allOf` like JSON bodies once item 2 lands; resolve the 3.1 claim (line 24) per whichever of 4(a)/4(b) is chosen - either document type-array/`const` sampling support, or narrow the stated version match and explain why.
- `docs/app/import-collections/openapi-v2.md`: update the `formData`/`consumes` section (lines 139-158, and the mapping table row at line 88/134/137) to state `formData` now honors `consumes` for the urlencoded/multipart choice once item 3 lands; update the "Dropped / not represented" list (lines 222-232) for path-item `$ref`.

## Sequencing

Part of #192; parallel-safe with the tracking issue's other sub-issues (Postman, Insomnia + OAuth2, orchestrator/import-apply) - none touch `openapi-v2.ts`, `openapi-v3.ts`, or `schema-sampler.ts`. No dependency on in-flight PR #175 (that PR touches request-settings fields `followRedirects`/`maxRedirects`/`httpVersion`, none of which this issue's parsers read or write).

## Acceptance criteria

- A path item expressed as `{"$ref": ...}` (3.0 or 3.1) imports its operations instead of silently vanishing.
- A urlencoded/multipart `requestBody` schema behind `$ref` or `allOf` produces the same field list an equivalent inline schema would.
- Swagger 2 `formData` parameters import as `x-www-form-urlencoded` or `form-data` matching the operation's (or spec's) `consumes`, not unconditionally multipart.
- The detector's 3.1 claim and the sampler's actual keyword support agree - either both cover type-array/`const`, or the claim is narrowed and documented.
- A dropped TRACE operation (or any operation whose method Vayu cannot represent) is recorded in `meta.skipped`, not silently excluded from `requestCount`.
- A non-array `parameters` value on a path item or operation does not abort the whole import.
- No change to any engine file or `POST /import/apply` payload shape.

athrvk:
## Description

Part of #192. All seven findings in #193 were re-verified against current master (`9fc4b9d`) before editing - the line anchors had shifted slightly, but every defect is still present exactly as described.

**Plan (root cause, approach, rejected alternative).** The root cause of the two silent-loss items is the same one twice: the parsers resolve `$ref` at *some* levels and not others, and the unresolved level then reads as "nothing here". A path item is only checked for method keys, so a `{"$ref": ...}` item contributes none and its whole path disappears; the form-body branch reads `content[ct].schema?.properties` directly, so a schema behind `$ref`/`allOf` yields zero fields while the JSON branch two lines up resolves the same shape fine. So the approach in both cases is to route the read through the resolution the parser already owns: `resolvePathItem` before the method loop, and `schemaFieldNames` - which samples the schema and takes the stub's own keys - for the form branch. For the form fields I rejected writing a second properties-walker that merges `allOf`: it would resolve *further* than the JSON body next to it, so the same schema would import differently depending on its media type. Going through `sampleSchema` gives form and JSON bodies provably the same reach (first branch only), which is the parity the issue asks for.

The helpers live in a new **`openapi-shared.ts`**, imported by both parsers, rather than as a copy in each. The two parsers are structural clones - which is precisely how both grew the same unguarded `parameters` spread and the same hardcoded `skipped: []` - and three of these fixes had to land in both. For item 4 I took route **(a), support the keywords**, over (b) narrowing the detector: `detect()` matching `3.1.x` is right (a 3.1 spec imports fine), the gap was only in the sampler, and telling users "3.1 detected but not really sampled" is a worse contract than reading three keywords.

## Findings, and what changed

| # | Finding | Fix |
|---|---|---|
| 1 | Path-item-level `$ref` never resolved - a whole path silently vanishes | `resolvePathItem` (`openapi-shared.ts`) resolves a `{"$ref": ...}` path item before the `HTTP_METHODS` loop, in **both** parsers. One hop only, matching the parameter / `requestBody` refs (a ref-to-a-ref would need a cycle guard and is not a shape generators emit). A path item that is not an object, or whose `$ref` does not resolve to one, is now counted rather than skipped in silence. |
| 2 | Urlencoded/multipart body schema behind `$ref` or `allOf` yields zero form fields | Field names come from the new `schemaFieldNames(schema, resolveRef)` (`schema-sampler.ts`), which samples the schema and returns `Object.keys()` of the stub. `$ref` and `allOf` now resolve for form bodies exactly as far as they already did for JSON bodies - first branch only, deliberately, so the two cannot disagree. A schema that samples to a non-object contributes no names. |
| 3 | Swagger 2 `formData` always multipart, ignoring `consumes` | `buildSwaggerOp` derives `formMode` from the operation's (else the spec's) `consumes`: urlencoded when it names `application/x-www-form-urlencoded` and **not** multipart; `form-data` when it names multipart, names both (only multipart can carry a `type: file` field), or names neither - so today's default is preserved for a spec with no `consumes`. Compared on the media type alone, so `; charset=utf-8` still matches. |
| 4 | Detector claims 3.1; sampler ignores 3.1 keywords | Route **(a)**: `schema-sampler.ts` handles `Array.isArray(schema.type)` (samples the first non-`"null"` member; an only-`"null"` type and the scalar `type: "null"` sample as `null`) and reads `const`. `const` is placed **above** `example` - JSON Schema makes it the only permitted value, where `example` is an annotation. |
| 5 | TRACE dropped with no skipped record | A `trace` operation is counted as a new `unsupported_method` `SkippedItem` and stays out of `requestCount`. `UNSUPPORTED_METHODS` is the list and `trace` is its only member (a path item defines exactly eight methods). `HttpMethod` is **not** touched, per the issue. |
| 6 | Non-array `parameters` crashes the whole parse | Both parsers' parameter lists go through `SkipTally.params`, which returns `[]` for a non-array and counts a `malformed_spec`. Absent stays uncounted - only a present-but-wrong shape is a defect. Every other path in the file still imports. |

**Beyond the issue's literal list** (both noted in the docs): the sampler also reads 3.1's plural **`examples`** (first entry, when there is no singular `example`). The issue's Problem section names it as never read and route (a) is "support the keywords"; leaving it out would have meant a 3.1 spec's request-body example still being ignored, and the doc caveating it. Second, item 6's "ideally, record it" is implemented rather than skipped - silently treating malformed input as empty is the same silent-loss shape as the findings being fixed here.

**`meta.skipped` gains its first producer in these parsers**, so the two new kinds have a reader before they have a writer: `ImportModal.tsx:506-514` already renders `${count} ${kind}` for every `meta.skipped` entry, which is where `1 unsupported_method` / `2 malformed_spec` surface in the preview. No new UI was needed and none was added.

**Blast radius traced.** `SkippedItem.kind` is read in exactly two places - `ImportModal`'s preview line and `insomnia-v4.ts`'s cast - and neither switches exhaustively on the union, so widening it changes no existing behaviour (`tsc` confirms). `sampleSchema` has three call sites (both OpenAPI parsers' body paths); the added `const`/`examples`/type-array handling only fires on schemas that previously fell through to `{}` or ignored a pinned value, so existing fixtures are unaffected - the two parser suites and `orchestrator.fixture-parity.test.ts` pass unchanged. `schemaFieldNames` is v3-only (Swagger 2 form fields come from `formData` parameters, not a schema). No engine file, no `POST /import/apply` payload shape, and no MCP surface is touched - `rg -n "openapi-v3|openapi-v2|schema-sampler" app/electron engine` is still empty, matching the issue's "App changes: none".

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

All run on this branch at `eeee26d`, on Linux.

- `cd app && pnpm test` - **1947 passed, 230 files** (1926 on the base commit; **21 tests added**).
- `mkdocs build --strict -f .github/mkdocs.yml` - clean (the new `#consumes--form-encoding` and cross-doc anchors resolve).
- `npx eslint` on the eight touched app files - **29 problems, down from 31 on the base commit**; every one is a pre-existing `no-explicit-any` in these parsers belonging to #189's lint burn-down (two went away because the `pathParams: any[] = []` defaults did). No new diagnostics.
- `npx prettier --check` on the eight touched app files - clean (all seven pre-existing ones were prettier-clean before, so formatting them was in scope). The three `docs/app/import-collections/*.md` files are **not** prettier-clean and were not clean on the base commit either, so they were deliberately left unformatted per CLAUDE.md.
- **Engine build and `ctest` were not run**: no C++ changed (the issue's own acceptance criteria include "no change to any engine file"), so no engine test could go from green to red.

New coverage: `openapi-v3.test.ts` (a `components.pathItems` `$ref` path item incl. its shared parameters, an unresolvable one, a `$ref`'d urlencoded body, an `allOf` multipart body, TRACE, a non-array `parameters` on both a path item and an operation, and a clean-fixture assertion that `skipped` stays `[]`), `openapi-v2.test.ts` (the six `consumes` → form-encoding cases, spec-level `consumes` fallback, field rows preserved, path-item `$ref`, the `parameters` guard), and `schema-sampler.test.ts` (type arrays in three shapes, only-null, `const` incl. beating `example` and `const: null`, `examples[0]`, plus a `schemaFieldNames` block).

Every fix was mutation-checked by script - reverted in place, its covering tests run, then restored from a copy. All fourteen reverts fail, and only the intended tests fail:

| Reverted | Failing tests |
|---|---|
| `resolvePathItem` in v3 → raw path item | 2 (`$ref`'d item, unresolvable item) |
| `schemaFieldNames` → `schema?.properties` | 2 (`$ref` form body, `allOf` form body) |
| the `UNSUPPORTED_METHODS` tally | 1 (TRACE) |
| `tally.params(op.parameters)` in v3 | 1 |
| `tally.params(pathItem.parameters)` in v3 | 1 |
| `mode: formMode` → `mode: "form-data"` | 3 (per-`consumes`, field rows, spec-level fallback) |
| `resolvePathItem` in v2 | 1 |
| `tally.params` in v2 | 1 |
| the type-array normalisation | 2 |
| the `const` branch | 1 |
| the `examples[0]` branch | 1 |
| `case "null"` | 1 |
| `schemaFieldNames`' non-object guard | 1 |
| `SkipTally.params`' malformed count | 2 (one per parser) |

The two `skipped` assertions on the committed fixtures are what stop the tally from becoming noise: a producer that fires on a healthy spec would fail them, not merely add a warning nobody notices.

## Checklist

## Related Issues
Closes #193

Docs updated in the same commit: `docs/app/import-collections/openapi-v3.md` (path-item `$ref` resolution and the `malformed_spec` count in **Tree structure**, a new `trace` paragraph, the form-body table row plus a paragraph on why field names go through the sampler, the rewritten `const`/`example`/`examples` precedence and type-array bullets, the `null` row, the parameter-guard paragraph, and a `meta.skipped` kind table replacing "this parser never emits `SkippedItem`s"), `docs/app/import-collections/openapi-v2.md` (a new **`consumes` → form encoding** section with the four-case table, the `formData` mapping-table row, path-item `$ref` and parameter-guard paragraphs, the sampler bullets, the corrected "Dropped / not represented" and `meta` lines, and two rows in the v2-vs-v3 contrast table), and `docs/app/import-collections/README.md` (the widened `SkippedItem` union with what each new kind means, and the `sampleSchema` / `schemaFieldNames` bullets).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Tk6Eijq5Hp9ya9Z4gcBKJh)_


athrvk:
Fix the merge conflict

athrvk:
Merged `master` (`e666276`) in - conflict resolved, branch is mergeable again.

**One conflict, in `importers/types.ts`:** #207 and this branch both widened `SkippedItem.kind`. Kept all three new kinds (`malformed_item` from #207, `unsupported_method` + `malformed_spec` from here) and rewrote the doc comment to separate the kinds that mean "Vayu cannot represent this" from the three that do not.

I also updated the `SkippedItem` line in `docs/app/import-collections/README.md`. Git took this branch's version of that block wholesale, and master had left it listing only the five original kinds (#207 documented `malformed_item` in `postman.md` instead), so the auto-merged result would have described an 8-member union with 7 members - it now names all eight.

Everything else auto-merged. Worth noting the two semantic overlaps that a clean auto-merge does *not* prove correct, both checked: #207's `normalizeVars(path, { pathTemplates: true })` call sites in the two OpenAPI parsers survived (the `{{baseUrl}}/users/{{id}}` assertion in the new path-item-`$ref` test would fail otherwise), and #207's new optional `RequestDraft.followRedirects` / `maxRedirects` are untouched here - OpenAPI has no equivalent field, so these parsers correctly leave both absent.

Re-verified on the merged tree (`6ca3523`): `pnpm test` **1985 passed, 231 files**, `pnpm type-check` clean, `mkdocs build --strict` clean, prettier clean on the touched files, and eslint still at the same 29 pre-existing `no-explicit-any` from #189. The test/verification numbers in the PR body above were measured pre-merge at `eeee26d`.

---

PR #209, opened 2026-07-30
fix(app): insomnia importer - text bodies, auth prefixes, folder scripts, five smaller gaps

### LINKED ISSUE #196: app(import): Insomnia importer - text bodies dropped wholesale, auth prefixes lost, and five smaller gaps
Part of #192. Base: master `4136c66`.

## Problem

Eight defects in the Insomnia v4 / OAuth2 import path (`app/src/services/importers/insomnia-v4.ts`, `oauth2-import.ts`), re-verified against current code. All anchors below are exact at `4136c66`.

### 1. Non-listed text bodies (XML, YAML, CSV, "Other") dropped wholesale, no skipped counter - HIGH (`insomnia-v4.ts:76-107`)

```ts
function insomniaBody(body: any): RequestBody {
	const mime = (body?.mimeType ?? "").split(";")[0].trim();
	switch (mime) {
		case "application/json": ...
		case "text/plain": ...
		case "application/graphql": ...
		case "application/x-www-form-urlencoded": ...
		case "multipart/form-data": ...
		default:
			return { mode: "none" };   // :105-106
	}
}
```

Any `mimeType` outside those five - `application/xml` (Insomnia's stock SOAP/XML body type), `application/yaml`, "Other" - falls to `default` and `body.text` is discarded entirely. Nothing increments a counter for it: `SkippedItem.kind` (`types.ts:17`) has no matching kind, and `skippedCounts` is fed only by resource `_type` during the tree walk (`insomnia-v4.ts:178`), not by body shape. The only warning surface, `ImportModal.tsx:506-512`, reads `meta.skipped` / `meta.nonExecutableAuth` - neither fires. A collection of SOAP/XML requests imports clean-looking, every body silently emptied.

This is representable: `RequestBody`'s `text` mode exists (`domain.ts` body modes) and the sibling Postman importer already proves the pattern - `postman.ts:44` routes unrecognised raw bodies through `shared.ts:120-129`'s `rawBody()`, which JSON-sniffs and falls back to `{ mode: "text", content }` rather than dropping. The doc (`docs/app/import-collections/insomnia-v4.md:112`) pins the drop with the rationale "Vayu can't represent it" - false for any text-bearing mime, so the doc's own contract is wrong here, not merely undocumented.

### 2. Bearer auth `prefix` silently dropped - MEDIUM (`insomnia-v4.ts:42-43`)

```ts
case "bearer":
	return { mode: "bearer", token: normalizeVars(asString(auth.token)) };
```

Insomnia's bearer auth carries a `prefix` field (UI "PREFIX", e.g. `Token` for GitLab/Django-style APIs). It is never read, and Vayu's bearer mode has no slot for it (`domain.ts:97`: `{ mode: "bearer"; token: string }`). The imported request silently sends `Authorization: Bearer <token>` instead of the configured scheme. Unlike `digest`/`ntlm`/`iam`, this is not counted in `nonExecutableAuth` - the import looks fully faithful and the loss surfaces only as an unexplained 401 at execution.

### 3. OAuth2 `tokenPrefix` (header prefix) silently dropped - MEDIUM (`oauth2-import.ts:87-123`)

`mapInsomniaOAuth2` reads `grantType`/`usePkce`/`authorizationUrl`/`accessTokenUrl`/`redirectUrl`/`clientId`/`clientSecret`/`scope`/`username`/`password`/`audience`/`resource`/`credentialsInBody` - no prefix-shaped key. `config.headerPrefix` therefore inherits `defaultOAuth2Config()`'s `"Bearer"` (`app/src/services/oauth/defaults.ts:23`). This is a real asymmetry, not a model limitation: `OAuth2Config.headerPrefix` exists precisely so the sibling Postman mapper can set it -

```ts
// oauth2-import.ts:77, mapPostmanOAuth2
headerPrefix: d.headerPrefix ? nv(d.headerPrefix) : "Bearer",
```

Vayu executes OAuth2 imports (unlike digest/ntlm/aws), so a non-Bearer Insomnia `tokenPrefix` silently becomes `Bearer` at execution time - the import reports success, then every request 401s.

### 4. Folder-level scripts dropped where the Postman path imports the equivalent - MEDIUM (`insomnia-v4.ts:161-195`)

```ts
const buildCollection = (node: Resource, isWorkspace: boolean): CollectionDraft => {
	...
	return {
		...
		preRequestScript: "",     // :190
		postRequestScript: "",    // :191
		children,
		requests,
	};
};
```

Every `workspace` and `request_group` gets hardcoded empty scripts, even when `opts.importScripts` is true. `CollectionDraft` carries both fields and the orchestrator sends them (`orchestrator.ts:127-128`), and the **Postman** importer already imports folder-level events at the equivalent point:

```ts
// postman.ts:115, pmFolder
const events: any[] = Array.isArray(node.event) ? node.event : [];
...
preRequestScript: ctx.opts.importScripts ? joinExec(events.find((e) => e.listen === "prerequest")) : "",
postRequestScript: ctx.opts.importScripts ? joinExec(events.find((e) => e.listen === "test")) : "",
```

Insomnia's own docs confirm folder-level scripts are a real, executing feature - not just folder-scoped variable access: request execution runs "scripts defined on the request or its parent folders," and script authors reach a folder's own scope via `insomnia.parentFolders.get(...)`. So an Insomnia 9.x+ export with a folder-level auth-refresh script imports cleanly with that script gone, and requests fail in ways the (correctly imported) request-level scripts don't explain. No `meta.skipped` entry fires - `skippedCounts` is keyed only by the five resource `_type`s at `insomnia-v4.ts:178`.

`docs/app/import-collections/insomnia-v4.md` pins the empty-string behavior three times (lines 68, 155, 164), so code matches doc - but the doc's framing ("workspace/group-level scripts are not imported") reads as a deliberate limitation rather than the data loss it is. No fixture or test in this repo carries a `request_group` with `preRequestScript`/`afterResponseScript`, so **the exact export key names should be confirmed against a real Insomnia 9.x+ folder-script export before implementing** - the field names are assumed to mirror the request-level keys already read at `insomnia-v4.ts:156-157`, which is the natural expectation but is unverified in-repo.

### 5. Header/param `description` dropped - LOW (`insomnia-v4.ts:140-153`)

```ts
params: mapKeyValues(
	((r.parameters as any[]) ?? []).map((p) => ({ key: p.name, value: p.value, disabled: p.disabled }))
),
headers: mapKeyValues(
	((r.headers as any[]) ?? []).map((h) => ({ key: h.name, value: h.value, disabled: h.disabled }))
),
```

Both projections drop `description` before calling `mapKeyValues`, even though `mapKeyValues` forwards it (`shared.ts:64`: `...(r.description ? { description: r.description } : {})`) and the Postman path passes rows through raw so descriptions survive. Metadata-only loss, no behavior change; doc-consistent (`insomnia-v4.md:81-82` documents the `{name,value,disabled}` projection), so this is hygiene, not drift.

### 6. Per-request redirect settings not importable - LOW (pipeline-wide gap)

Vayu stores `followRedirects`/`maxRedirects` per request (`domain.ts:161-163`) and always sends both explicitly because the engine's default is `true` (an omitted `false` would silently follow a redirect the user disabled). Insomnia serializes the equivalent as `settingFollowRedirects` (`"off"` when a user disables following) per request. `insomnia-v4.ts`'s `buildRequest` never reads it, and there is nowhere to put it if it did: `RequestDraft` (`types.ts:37-49`) and `ImportApplyRequestItem` (`api.ts:347-362`) have no redirect fields. Every imported request lands on the engine default (follow = true), silently changing behavior for a request the user configured to stop at a 3xx. This is a pipeline gap shared with the Postman importer's identical `protocolProfileBehavior` miss (tracked in the Postman sub-issue) - the fix is client-only, since the engine's `/import/apply` already accepts `followRedirects`/`maxRedirects` on create.

### 7. Shape-invalid Insomnia files crash parse with raw `TypeError`s - LOW (`insomnia-v4.ts`, several sites)

No shape guards exist anywhere in this parser, unlike its Postman sibling (`postman.ts:95`, `Array.isArray(item.event)`):

- `:122` - `for (const r of resources)` throws if `resources` is non-array.
- `:141`, `:148` - `.map` on `r.parameters` / `r.headers` throws if non-array.
- `:89`, `:97` - `.map` on `body.params` throws if non-array.
- `:77` - `.split(";")` throws if `body.mimeType` is a non-string.
- `:161-195` - `buildCollection` recurses with no visited-set; a self/ancestor-referential `_id` (which Insomnia cannot itself emit) would `RangeError`.

All are caught by `ImportModal.tsx:79-91`'s generic `try/catch`, which sets `phase = "error"` and shows `(e as Error).message` verbatim - so there is no data loss and no partial import, only an unhelpful raw JS error (`"Cannot read properties of undefined (reading 'map')"`) instead of a message naming the format and the problem. Reaching any of these requires a hand-edited or script-mangled export (the Insomnia app itself only emits arrays and string mime types), so this is rare-input-but-loud, not silent data loss - message-quality hygiene, matching the guard style `postman.ts` already uses elsewhere in this codebase.

### 8. Dead doc path cited in `oauth2-import.ts` - LOW (`oauth2-import.ts:13`)

```ts
/**
 * ...
 * Field references: docs/oauth/02-postman-bruno.md §1.5 (Postman v2.1 keys).
 */
```

`docs/` has no `oauth/` directory (`ls docs/` confirms) and the only repo-wide match for `02-postman-bruno` is this comment. The file the OAuth2 field-mapping contract claims to be verified against does not exist, so a contributor adjusting grant-type or prefix mapping (items 2 and 3 above) has no in-repo authority to check against. Pure hygiene, zero runtime impact.

## Fix pathway

1. Map unknown text-bearing `body.mimeType` values to `{ mode: "text", content }` with the mime preserved, mirroring `shared.ts`'s `rawBody()` fallback, instead of dropping to `{ mode: "none" }`. Reserve `{ mode: "none" }` for genuinely binary/file-only bodies, and count those in a new `meta.skipped` kind (or `meta.nonExecutableAuth`-style counter) so a real drop is still visible.
2. Import bearer `prefix` / OAuth2 `tokenPrefix` into the stored auth config, or at minimum count the loss visibly (increment a counter surfaced in the preview) so it isn't silently discarded and misreported as a faithful import.
3. Same fix as (2) for `mapInsomniaOAuth2`: read the Insomnia field and set `config.headerPrefix`, mirroring `mapPostmanOAuth2`'s `d.headerPrefix ? nv(d.headerPrefix) : "Bearer"` at `oauth2-import.ts:77`.
4. Import folder-level scripts the way the Postman path already does - cite `postman.ts:115,132-136` (`pmFolder`'s `node.event` handling) as the pattern to copy. Confirm the real Insomnia 9.x+ export key names for `request_group` scripts first (no in-repo fixture pins them); if the export shape does not carry them the way requests do, downgrade this item to a documented limitation instead.
5. Forward `description` in both `mapKeyValues` call sites in `buildRequest` (params and headers), matching the Postman path.
6. Wire `settingFollowRedirects`/(if it exists) a max-redirects equivalent to the request settings: add the fields to `RequestDraft` and `ImportApplyRequestItem`, plumb them through the orchestrator, and read them in `insomnia-v4.ts`'s `buildRequest`. In-flight PR #175 (HTTP/2 support) has since merged into master (`b3e2e9f`, commit after `4136c66`) and touches this same request-settings surface (`followRedirects`/`maxRedirects`/`httpVersion`); land this item on top of, or rebased over, #175's merged form rather than in parallel with it.
7. Wrap the parse in shape validation (or targeted guards at each unguarded site) producing a named error ("Malformed Insomnia export: <detail>") instead of letting a raw `TypeError`/`RangeError` propagate to `ImportModal`.
8. Fix or delete the dead doc path: either restore a field-reference doc under `docs/app/import-collections/` and repoint the comment at it, or delete the stale reference and point at `docs/app/import-collections/insomnia-v4.md` / `postman.md` instead.

## App changes

Engine: none, verified - every fix above is contained to `app/src/services/importers/insomnia-v4.ts`, `oauth2-import.ts`, `shared.ts`, `types.ts`, `api.ts`, and (item 6) the orchestrator/apply-payload plumbing already exercised by the existing `followRedirects`/`maxRedirects` fields on other draft types. No `engine/` file is touched by anything in this issue.

App: `app/src/services/importers/insomnia-v4.ts`, `app/src/services/importers/oauth2-import.ts`, `app/src/services/importers/shared.ts` (if the text-body fallback is generalized), `app/src/services/importers/types.ts`, `app/src/types/api.ts`, and `app/src/services/importers/orchestrator.ts` (item 6 plumbing only).

## Tests

Vitest, fixture-driven, mutation-checked (apply the fix, add the test, revert the fix locally, confirm the test fails, restore the fix) - in `insomnia-v4.test.ts` and `oauth2-import.test.ts`:

- Item 1: a fixture request with `body.mimeType: "application/xml"` (or `text/csv`) imports with the body **present** (`mode: "text"`, content intact) - the test must assert the imported request HAS the body, not merely that import doesn't throw.
- Item 2: a bearer auth object with `prefix: "Token"` either round-trips the prefix or is counted as lossy - assert whichever the fix chooses, not silence.
- Item 3: an Insomnia `oauth2` block with a header-prefix-equivalent field produces `config.headerPrefix` equal to that value, not the `"Bearer"` default.
- Item 4: a `request_group` fixture carrying the confirmed script field(s) produces a non-empty `preRequestScript`/`postRequestScript` on the corresponding `CollectionDraft`.
- Item 5: a header/param row with `description` set preserves it on the mapped `KeyValueEntry`.
- Item 6: a request fixture with the redirect-disabling field set produces a `RequestDraft` (and downstream `ImportApplyRequestItem`) with `followRedirects: false`.
- Item 7: each previously-unguarded shape (non-array `resources`, non-array `parameters`/`headers`, non-string `mimeType`) produces a named, format-specific error rather than a raw `TypeError`.
- Item 8: no test needed; verify the doc link resolves (or is deleted) as part of the normal `mkdocs build --strict` gate.

## Docs

`docs/app/import-collections/insomnia-v4.md` updated in the same commit as the corresponding code change:

- Body mapping table (line ~112): replace "anything else ... -> none" with the new text-fallback rule and the narrowed "truly unrepresentable" case.
- Auth mapping table (line ~122) and lossy-summary (lines ~158-165): document bearer `prefix` / OAuth2 `tokenPrefix` handling once fixed.
- Field-mapping table (line ~68, ~155, ~164): update the "always \"\" for collections" language once folder scripts are imported (or leave as-is with an explicit note if item 4 is downgraded after the export-key check).
- Add `description` to the params/headers mapping table (line ~81-82).
- Note the new redirect-setting field once item 6 lands.
- `oauth2-import.ts:13`'s doc reference fixed to point at a real path.

## Sequencing

Part of #192; parallel-safe with the other #192 sub-issues (Postman, OpenAPI, orchestrator) - none touch the same files. Item 6 (redirect settings) should land after, or rebased over, PR #175 (`b3e2e9f`, "feat: support HTTP/2, selectable per request"), which has already merged into master ahead of this issue's base commit and touches the same request-settings surface. Items 1-5, 7, 8 are independent of #175 and of each other.

## Acceptance criteria

- An Insomnia export with an XML/YAML/CSV/"Other" text body imports with that body present under `mode: "text"`, not silently emptied; a genuinely unrepresentable (binary/file-only) body is still dropped but counted.
- A bearer `prefix` and an OAuth2 `tokenPrefix` are either preserved through import or counted as a visible loss in the preview - never silently replaced with `Bearer`.
- Folder-level Insomnia scripts import the same way Postman folder-level events do, once the real export key names are confirmed; if they cannot be confirmed, the doc is updated to state the limitation is deliberate rather than implying data survives.
- Header/param `description` values survive Insomnia import.
- A request's redirect-following setting is either imported or the loss is documented and counted; the fix lands after PR #175's merged request-settings surface.
- A shape-invalid Insomnia file fails import with a named, format-specific error, not a raw `TypeError`/`RangeError`.
- The `oauth2-import.ts` doc reference resolves to a real, currently-existing file.
- `docs/app/import-collections/insomnia-v4.md` reflects every mapping-table change made in this issue's commits.

athrvk:
## Description

Part of #192, implementing #196. All eight findings were re-verified against master before editing.

**Rebase note (2026-07-30).** Master has since taken #205, #206, #207, #208 and the 0.12.0 release. This branch is merged up to `5281c96` and the conflicts flagged below are resolved: `RequestDraft.followRedirects` / `ImportApplyRequestItem.followRedirects` and the orchestrator's forwarding line now come from **#207's** version, which is a superset (it carries `maxRedirects` too), so those three files are no longer part of this diff at all. Only the field's doc comment gained a clause naming Insomnia as a second producer. The duplicate orchestrator test this branch had added was dropped for the same reason - #207 landed an equivalent one. Net diff against master is now 7 files: the Insomnia parser and its tests, the OAuth2 mapper and its test, one line of `types.ts`, and two docs.

**Plan (root cause, approach, rejected alternative).** Six of the eight are one-site mapping or robustness bugs in `insomnia-v4.ts` / `oauth2-import.ts`. The two that needed a decision are the bearer prefix and the redirect setting.

*Bearer prefix.* Vayu's bearer mode is `{mode:"bearer", token}` and the engine writes `Authorization: Bearer <token>` from it, so there is nowhere to put Insomnia's PREFIX. I rejected widening `RequestAuth`'s bearer variant with an optional `prefix`: the engine would ignore it, which is exactly the "written but never read" defect CLAUDE.md calls this codebase's most repeated one, and making it real means an engine change the issue rules out. I also rejected the issue's fallback of merely *counting* the loss - the counter surfaces as "N auth not executed", which is wrong here (the auth executes, just with the wrong scheme). Instead the prefix is preserved through the mode that already carries a verbatim header: `{mode:"apikey", key:"Authorization", value:"<prefix> <token>", in:"header"}`. `apply_auth`'s `ApiKeyAuth` branch writes `req.headers[key] = value` unchanged (`engine/src/http/auth_resolver.cpp`), so the wire bytes match what Insomnia sent, with no engine or domain change. A prefix differing only in case keeps the native bearer mode (RFC 7235 §2.1 makes schemes case-insensitive).

*Redirects.* This branch produces **only** `followRedirects`, never `maxRedirects`: Insomnia's redirect limit is an app-wide setting, not a per-request field, so a value here would have no source. Absent is meaningful and is preserved end to end - the engine defaults `follow_redirects` to `true`, so an omitted `false` would silently follow a 3xx the user disabled, and an absent value written as `false` would stop one they never touched. The draft/payload fields and the orchestrator plumbing came from #207; no engine change was needed either way, since `POST /import/apply` passes each item's fields through the shared `apply_request_fields` (`engine/src/http/routes/import.cpp`).

## Findings, and what changed

| # | Finding | Fix |
|---|---|---|
| 1 | Non-listed text bodies (XML, YAML, CSV, "Other") dropped wholesale, no counter | `unlistedBody` returns `{mode:"text", content}` for a non-empty string `body.text`, mirroring `shared.ts`'s `rawBody()` fallback. No JSON sniffing - Insomnia states the mime, so an XML body stays `text`. `{mode:"none"}` is now reserved for a genuinely binary body (`fileName`, no text), which increments the existing `file_body` `SkippedItem` kind. **`multipart/form-data` file parts are counted too** (they were dropped silently, and the doc pinned that as uncounted) - parity with the Postman parser's `skippedFileBody`. |
| 2 | Bearer `prefix` silently dropped | Preserved as an explicit Authorization header via `apikey` (see plan). `Bearer`/`bearer`/empty/absent keep `{mode:"bearer"}`. |
| 3 | OAuth2 `tokenPrefix` silently dropped | `mapInsomniaOAuth2` sets `headerPrefix: auth.tokenPrefix ? nv(auth.tokenPrefix) : "Bearer"`, the same line `mapPostmanOAuth2` already had. |
| 4 | Folder-level scripts dropped where the Postman path imports the equivalent | `buildCollection` reads `preRequestScript` / `afterResponseScript` off the `request_group` (and workspace), gated on `importScripts`, exactly as `buildRequest` does. **See the caveat below.** |
| 5 | Header/param `description` dropped | Both projections go through a shared `kvRow`, which forwards a string `description` (any other type is ignored) into `mapKeyValues`. |
| 6 | Per-request redirect setting not importable | `settingFollowRedirects` (`"off" → false`, `"on" → true`, `"global"`/absent → omitted) → `RequestDraft.followRedirects`, which the orchestrator forwards only when set. |
| 7 | Shape-invalid files crash with a raw `TypeError`/`RangeError` | Guards at every previously-unguarded read throw `Malformed Insomnia export: <detail>`, which `ImportModal` already shows verbatim. A `parentId` cycle (reachable only via a duplicated `_id`) is caught by a visited set instead of blowing the stack. Absent stays absent everywhere - only a present-but-wrong shape is an error. |
| 8 | Dead doc path in `oauth2-import.ts` | Repointed at `docs/app/import-collections/postman.md` and `insomnia-v4.md`. |

**Deliberate deviation - item 4's key-name confirmation.** #196 asks that the `request_group` script key names be confirmed against a live Insomnia 9.x export first, and downgraded to a documented limitation if they cannot be. I could not confirm them from this environment: the sandbox's network policy answers `raw.githubusercontent.com` and GitHub blob/tree pages with 404, so Insomnia's `models/request-group.ts` is unreadable, and no repo fixture or public doc pins the v4 shape (Kong's import/export reference documents only `_type`). What I did establish is that folder-level scripts are a real 9.3+ feature and that the v4 export writes model fields verbatim, and PR Kong/insomnia#7558 confirms the *request* model's field names are exactly `preRequestScript` / `afterResponseScript` - which is what this parser already reads at the request level. So rather than pre-emptively documenting data loss, the read is keyed on those same names, and the failure mode if Insomnia spells them differently is an absent field, i.e. byte-for-byte today's behaviour (empty strings, no crash). Both the code comment and `insomnia-v4.md` state that the names are unverified against a live export. Say the word if you would rather have the documented-limitation version instead.

**Blast radius traced.** `RequestDraft.followRedirects` has a producer (this parser) and a reader (`orchestrator.ts`'s `flatten` → the apply payload → `apply_request_fields`). `mapInsomniaOAuth2` is called only from `insomniaAuth`. `SkippedItem.kind` is unchanged: `file_body` already existed and `ImportModal` already renders `${count} ${kind}` for every entry, so the new counts surface with no UI change. Nothing in `app/electron/mcp/` or `engine/` reads these files (`rg` confirmed), matching the issue's "Engine: none".

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

All re-run after the merge, on the merged head, on Linux.

- `cd app && pnpm test` - **2002 passed, 231 files** (1985 on `5281c96`; **17 tests added**).
- `cd app && pnpm type-check` - clean. `npx tsc -p tsconfig.node.json --noEmit` (the electron project, which `type-check` does not build) - clean.
- `npx prettier --check` on the touched app files - clean. `insomnia-v4.test.ts` was prettier-clean on the base commit, so formatting it was in scope; the two `docs/app/import-collections/*.md` files are **not** prettier-clean and were not clean before either, so they were left unformatted per CLAUDE.md.
- `npx eslint` on the touched files - **11 problems in `insomnia-v4.ts`, unchanged from 11 on the base commit**; every one is a pre-existing `no-explicit-any` in this parser belonging to #189's lint burn-down. No new diagnostics, no new file with diagnostics.
- **Engine build and `ctest` were not run**: no C++ changed (the issue's own acceptance criteria say "no `engine/` file is touched"), so no engine test could go from green to red.

New coverage in `insomnia-v4.test.ts` (16) and `oauth2-import.test.ts` (1): an XML body kept as text, a binary body dropped **and counted**, a multipart file part counted with the text field kept, the clean fixture asserted to emit **no** `file_body` (so the counter cannot become noise), a `Token` prefix, the four prefix spellings that must stay native bearer, folder scripts on and off, descriptions on both rows, the four `settingFollowRedirects` cases, the seven malformed shapes, and `tokenPrefix`.

Every fix was mutation-checked by script - reverted in place, its covering tests run, then restored. All reverts fail, and only the intended tests fail:

| Reverted | Failing tests |
|---|---|
| the text-body fallback | 1 |
| the binary `file_body` count | 1 |
| the multipart file-part count | 1 |
| `insomniaBearer`'s prefix branch | 1 |
| folder scripts → `""` | 1 |
| `kvRow`'s `description` | 1 |
| `followRedirects` on the draft | 1 |
| `rowsOrThrow` → tolerant `[]` | 3 |
| the `resources` array guard | 1 |
| the resource-entry object guard | 1 |
| the `mimeType` string guard | 1 |
| the cycle guard | 1 |
| `headerPrefix` in `mapInsomniaOAuth2` | 1 |

## Checklist

## Related Issues
Closes #196

Docs updated in the same commit: `docs/app/import-collections/insomnia-v4.md` (rewritten body-mapping table with the text fallback and the two counted drops, two bearer rows plus a PREFIX note and a `tokenPrefix` note in Auth mapping, the folder-script row with its caveat, `description` on the params/headers rows, a new `settingFollowRedirects` row, the `file_body` paragraph in the skip-counting nuance, the corrected `importScripts` bullet and lossy summary, the corrected `meta` line, and a new **Malformed exports** section) and `docs/app/import-collections/README.md` (the `RequestDraft` shape, now listing both redirect fields with each format's producer).


athrvk:
fix merge conflicts

PR #212, opened 2026-07-30
fix(engine)!: split expect.equal from .eql, add the missing chai matchers

### LINKED ISSUE #183: engine(scripting): `expect.equal` and `.eql` are the same function, and both disagree with chai
## Problem

`equal` and `eql` are bound to the same C++ function (`engine/src/runtime/script_engine.cpp:889-890`):

```cpp
JS_SetPropertyStr (ctx, obj, "equal", JS_NewCFunction (ctx, expect_equal, "equal", 1));
JS_SetPropertyStr (ctx, obj, "eql",   JS_NewCFunction (ctx, expect_equal, "eql", 1));
```

`expect_equal` (`:289-337`) compares numbers, strings, bools, null and undefined by value, and **falls through to `JSON.stringify` comparison for everything else** (`:319-328`).

In chai - which Postman uses - `equal` is strict `===` and `eql` is deep equality. So we are wrong in both directions:

| Assertion | Postman | Vayu | Consequence |
|---|---|---|---|
| `expect({a:1}).to.equal({a:1})` | **fails** (different references) | **passes** | false pass |
| `expect({a:1,b:2}).to.eql({b:2,a:1})` | **passes** (order-insensitive) | **fails** (JSON strings differ) | false failure |

Verified by executing both against the engine at `4136c66`.

The false pass is the one that matters: a ported test asserting reference identity silently becomes a value comparison.

## Missing matchers

Also absent from `create_expectation` (`:831-910`), all of which throw `TypeError` today - loud, so lower severity than the above, but each is a Postman suite that will not run:

`deep.equal` (the `deep` chainer itself is missing, so `deep` is `undefined`), `oneOf`, `keys`, `members`, `throw`, `instanceOf`, `closeTo`, `satisfy`, `string`, `nested.property`, and `.and` chaining.

Present today: `equal`, `eql`, `above`, `below`, `least`, `most`, `include`, `contain`, `length`, `lengthOf`, `a`, `an`, `match`, `property`, plus the paren-less `true`/`false`/`null`/`undefined`/`ok`/`empty`/`exist` and `not`.

## Fix

1. **Split the two.** `equal` becomes strict equality (`JS_StrictEq`, i.e. reference identity for objects); `eql` keeps deep semantics.
2. **Make `eql` a real deep compare**, not `JSON.stringify`. Stringify is key-order sensitive, drops `undefined` members, and cannot compare anything non-serialisable. A recursive compare over object keys handles all three.
3. **Add the `deep` chainer** so `to.deep.equal` aliases `eql`, which is how most Postman suites are written.
4. **Add the missing matchers above.** Each must be a real assertion, not a no-op - see #TBD_TO_BE for what a stubbed matcher costs.

Breaking-change note worth calling out in the PR: an existing Vayu test using `to.equal` on an object starts failing. That is the point (it agrees with Postman now), but it should be named in the release notes rather than discovered.

## App changes (required)

None in behaviour. The **completion list** (`engine/src/http/routes/scripting.cpp`) should gain the new matchers and correct the `equal` / `eql` descriptions, which is engine-side; guarded by `script_completions_test.cpp`.

## Tests

- Mutation-checked pair, the core of this issue: `expect({a:1}).to.equal({a:1})` **fails**, `expect({a:1}).to.eql({a:1})` **passes**. Revert the split and one of them flips.
- `eql` is key-order insensitive: `{a:1,b:2}` eqls `{b:2,a:1}`.
- `eql` handles nesting, arrays, and mixed types; `NaN`, `null` vs `undefined`, and nested arrays are the edges worth pinning.
- One passing and one failing case per new matcher, since a matcher that cannot fail is the #108 defect.
- `to.deep.equal` behaves as `eql`.

## Docs (same commit)

- `docs/engine/scripting.md` - the `### pm.expect()` list is the full matcher inventory; update it and state the `equal` vs `eql` distinction explicitly.
- `docs/app/pm-api-compatibility.md` - its matcher table is the compatibility record; move the newly supported ones and keep the still-missing ones listed.

## Sequencing

Part of #180. Touches `create_expectation` and `expect_*` (`script_engine.cpp:289-910`), which no other sub-issue in the series touches - parallel-safe with the scopes, request/response and dynamic-variable issues. Best done **after** #TBD_TO_BE, only because that one is a false pass and this one is mostly loud failures.

## Acceptance criteria

- `equal` is strict and `eql` is deep, each proven by a test that fails if they are re-aliased.
- `eql` is key-order insensitive and does not go through `JSON.stringify`.
- `to.deep.equal` works.
- Every added matcher has a passing and a failing case.
- Both script docs list the same matcher set as the runtime.

athrvk:
## Description

`equal` and `eql` were bound to the same C++ function doing a `JSON.stringify` comparison, so Vayu was wrong in both directions: `expect({a:1}).to.equal({a:1})` **passed** (Postman fails it - different references) and `expect({a:1,b:2}).to.eql({b:2,a:1})` **failed** (Postman passes it - key order is not part of deep equality). Re-verified against `e5eaf0b` before editing: the aliasing is at `script_engine.cpp:889-890` and `expect_equal`'s stringify fallthrough at `:319-328`, exactly as #183 describes.

**Plan (root cause, approach, rejected alternative).** The root cause is one comparison serving two different assertions. `equal` is now strict and `eql` is a real recursive compare, split at the registration so neither can drift back into the other. The strict path is hand-written rather than delegated: `JS_IsStrictEqual` exists only in quickjs-ng (Windows), while the Unix build vendors Bellard's QuickJS, so a portable `===` is applied in `js_strict_equal` - no cross-type equality, `NaN !== NaN`, objects by reference. The deep path deliberately does **not** go through `JSON.stringify`, which is key-order sensitive, drops `undefined` members, and cannot see a value it fails to serialise.

I rejected the smaller-looking alternative of keeping one comparison and having `equal` merely *also* accept reference identity: it leaves the false pass in place, since two structurally equal objects would still compare equal. I also rejected treating `Map` / `Set` / typed arrays structurally - their contents are not own properties, so every distinct pair would report equal, which is the same class of false pass this series exists to remove. They report *unequal* instead; `Date` compares by instant and `RegExp` by pattern.

**Blast radius traced.** `create_expectation` backs `pm.expect` only - `pm.response.to` is a separate exotic-class object (`create_response_to_object`) and is untouched, so no response assertion changes. No existing test or fixture asserts `to.equal` on an object (checked across `engine/tests`, `app/src`, `app/electron`, `docs`, `scripts`), so nothing in the repo flips. The app has no matcher list of its own; the editor's is engine-served from `routes/scripting.cpp`, updated here in the same commit.

### What else changed, and why it is in scope

Adding the `deep` chainer made two comparisons that shared the old stringify approach reachable as `deep.include` / `deep.property`, so leaving them would have shipped a new false-pass surface:

- **array membership in `include`** compared `JS_ToCString` forms, under which every object member matched every object needle - both render as `[object Object]`, so `expect([{a:1}]).to.include({b:2})` passed.
- **the optional value argument to `property`** compared JSON strings; it is now strict, or deep under `.deep`, as in chai. `property` had to be touched anyway for `nested.property`, and shipping a new matcher through a known-wrong comparison seemed worse than the small behaviour change.

Deliberate deviations from the issue spec, both narrowing:

- `.include.keys` (chai's subset form) and `.any.keys` are **not** implemented; `have.keys` is all-and-only, as in chai. `all` is accepted as a no-op passthrough (it is chai's default), `any` is not - aliasing it to `all` would assert more than the script asked. Both recorded in `pm-api-compatibility.md` under *Not (yet) supported*.
- `nested.property` does not support chai's `\.` escape for a literal dot in a path segment. A plain `property('a.b')` still treats the dot literally, which is tested.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Breaking change
- [x] Documentation update

**Breaking:** a script asserting `to.equal` on an object or array now fails. That is the intended behaviour (it agrees with Postman), but it is a behaviour change rather than something to discover - worth a line in the release notes when this version ships.

## Testing

**The C++ suite could not be built in this session.** `python build.py -e -t` fails in vcpkg: `https://github.com/google/googletest/archive/v1.17.0.tar.gz` returns **403** from this environment's egress proxy, and every other github-hosted vcpkg distfile (`nlohmann-json`, `sqlite-orm`) is blocked the same way, so no engine target configures. The same failure happened during session setup, before any edit - it is an environment limitation, not something this branch introduced. **CI's `C++ Tests (ubuntu / macos / windows)` jobs are the authoritative run for the 20 new test cases below.**

What was verified locally instead, in place of `ctest`:

- **`script_engine.cpp` compiles clean** with `-std=c++20 -Wall -Wextra -Wpedantic` against the vendored Bellard QuickJS, adding no warning class the file did not already have.
- **Every assertion in the new tests was executed for real.** The vendored QuickJS was built directly and linked with `script_engine.cpp` plus a driver that runs each script through `ScriptEngine::execute_test` and checks PASS/FAIL: **90/90 cases behaved as asserted**, covering the strict/deep pair, key order, nesting, `NaN`, `null` vs `undefined`, dropped-`undefined` members, arrays vs array-likes, `Date`, `RegExp`, `Map`, cycles, `.and`, and one passing plus one failing case for each new matcher including its bad-argument path.
- **Mutation-checked.** Re-aliasing `equal` back to the deep comparison flips exactly three driver cases - `equal` on two objects, `equal` on `NaN`, and `not.equal` on two objects - which are the cases `ExpectEqualIsStrictAndEqlIsDeep` and `ExpectEqualKeepsPrimitiveSemantics` assert. Reverting the fix fails the suite.
- **The completion list builds and the new drift guard's logic was exercised** against the runtime: `get_script_completions()` compiles standalone and yields 91 items / 33 dotted chains, every item satisfies the existing shape and `insertTextRules` rules, and all 33 chains resolve in the runtime (function members callable, getters present as own property descriptors). The guard requires at least 20, so it cannot pass vacuously.
- **`mkdocs build --strict -f .github/mkdocs.yml`**: clean.

New/changed tests: 14 new `TEST_F`s in `script_engine_test.cpp` (about 100 `pm.test` cases) and 1 new `TEST` in `script_completions_test.cpp`.

The app suite was not run - this branch touches no app file. It was run green on the three open importer PRs earlier in this session (231 files / 1964 tests) while checking they still merge cleanly with current `master`.

## Checklist

Docs updated in the same commit: `docs/engine/scripting.md` (the `pm.expect()` inventory is now the full matcher set, with the `equal` vs `eql` distinction stated first) and `docs/app/pm-api-compatibility.md` (matcher table, the still-missing names, and a correction: that section's heading claimed the chains belonged to `pm.response.to` as well, which they never did).

Not run: `clang-format` on the touched files. All four were already not clang-format-clean at `HEAD` under the locally available clang-format 18 - it disagrees with the repo's formatting on pre-existing lines in both directions - so per CLAUDE.md ("format only files you touched that were clean before") they were left alone and the new code follows the surrounding style by hand.

## Related Issues
Closes #183

---
_Generated by [Claude Code](https://claude.ai/code/session_012tCgYMQCWFRFXD1HKfSZdz)_


athrvk:
- Add a link to the scripting doc on the pre request script and tests tab on the request builder so any one can easliy refer it link it via the https://athrvk.github.io/vayu/ page
- Verify all these additions are available in the monaco auto complete

athrvk:
Both done, plus the `master` merge. Pushed as `1a49534` (follow-up) and `a192986` (merge of `8870126`).

### 1. Scripting doc link in the script tabs

`ScriptPanel` is shared by the **Pre-request** and **Tests** tabs, so one link covers both - it sits beside "Quick Reference:" and opens through the keyed `openAppLink("scripting")` channel, the same one the welcome screen's footer uses. That channel is the only way the renderer can reach the system browser; an anchor would spawn an unmanaged Electron window.

The links now point at the published site rather than raw GitHub markdown, per your note:

| constant | before | after |
|---|---|---|
| `SCRIPTING_DOCS_URL` | `github.com/athrvk/vayu/blob/master/docs/engine/scripting.md` | `athrvk.github.io/vayu/engine/scripting/` |
| `DOCS_URL` | `github.com/athrvk/vayu#readme` | `athrvk.github.io/vayu/` |

**`DOCS_URL` was not something you asked for** - I changed it because leaving "Documentation" on the GitHub README while "Scripting Guide" moved to the site makes the Help menu point at two different docs. Say the word and I will put it back. Both constants also feed the Help menu and the welcome footer, so those move to the site too.

I built the site (`mkdocs build --strict`) and confirmed `engine/scripting/index.html` exists, so the URL the button opens is a real page, not a 404 - `use_directory_urls` is on, hence the trailing-slash form rather than `.md`.

Two new tests in `script-panels.test.tsx` (run for both variants): the link is present and calls `openAppLink("scripting")`, and clicking it does not throw when the preload bridge is absent.

### 2. Monaco autocomplete - verified in both directions

The gap was real: **`eqls`, `key`, `throws` and `all` were bound by the runtime but not offered**, so they would never have appeared in the editor. Added `to.eqls`, `to.have.key`, `to.have.all.keys` and `to.throws` (the list already offers aliases like `to.contain` and `to.have.lengthOf`, so this follows it).

Rather than eyeball the list, there is now a test each way, and neither hardcodes the matcher set:

- `EveryOfferedExpectMatcherExistsInTheRuntime` - every offered chain is looked up in the runtime (functions must be callable; getters must exist as own property descriptors, since evaluating one *asserts*). Requires at least 20 so it cannot pass vacuously.
- `EveryExpectMemberTheRuntimeBindsIsOffered` - the new one, and the one that would have caught the four missing aliases: it asks the runtime for `Object.getOwnPropertyNames(pm.expect(1))` and requires every name to appear as a segment of some offered chain. A matcher added later cannot go missing from the editor by omission.

Executed locally against the real runtime (same linked-driver approach as before, since the vcpkg gtest download is still 403 here): **37/37 offered chains resolve** and **41/41 bound members are offered**.

On the renderer side the provider is a pass-through - it maps every completion the engine serves straight into Monaco - so the remaining risk is the replace-range, not the list. The matcher chains are now up to four segments (`to.have.nested.property`, `to.have.all.keys`), which is deeper than anything shipped before, so `script-completion-range.test.ts` gains a case pinning that `pm.expect(x).to.have.nested.`, `pm.expect(x).to.deep.` and `pm.expect(x).to.have.all.k` all replace from the start of the chain. Replacing only the last segment would have inserted the full path *after* the typed `to.have.nested.`.

### Merge of `master`

Two conflicts, both from #213 landing in the same files:

- `script_engine.cpp` - adjacent includes (`<cmath>` vs `<iterator>`); kept both.
- `script_completions_test.cpp` - both sides appended a new test at the same spot. Kept both as separate tests; #213's `TheOfferedVariableScopeMethodsAreExactlyWhatTheRuntimeHas` is byte-identical to its version on `master`.

Re-verified after the merge: the merged `script_engine.cpp` compiles clean, all 90 assertion behaviour cases still pass, both completion directions still hold against the merged tree, `mkdocs build --strict` is clean, and the app suite is **231 files / 1990 tests green** with `tsc --noEmit`, ESLint and Prettier clean on the touched files. #213's own tests use `to.equal` only on strings, so the strict-equality change does not touch them.

---

Issue #184, opened 2026-07-29
engine(scripting): variable scopes have only get/set - add has/unset/clear/toObject and the merged `pm.variables`


athrvk:
## Problem

All three variable scopes expose exactly two methods. `setup_pm_environment` (`engine/src/runtime/script_engine.cpp:1509-1518`):

```cpp
JS_SetPropertyStr (ctx, env, "get", ...);
JS_SetPropertyStr (ctx, env, "set", ...);
JS_SetPropertyStr (ctx, pm, "environment", env);
```

`setup_pm_globals` (`:1557-1566`) and `setup_pm_collectionVariables` (`:1605-1614`) are the same shape.

Missing on every scope: `has`, `unset`, `clear`, `toObject`, `replaceIn`, and `pm.environment.name`. Missing entirely: **`pm.variables`**, the merged accessor that reads across all scopes in precedence order - which `docs/engine/scripting.md` currently teaches anyway (tracked as its own bug in #TBD_PHANTOM; this issue is the implementation that supersedes that doc change).

`unset` is the one whose absence has no workaround: a script cannot remove a variable it created, and setting it to `""` leaves an enabled empty variable behind, which is not the same thing for `{{template}}` resolution.

Bruno covers the same ground with `getEnvVar`/`setEnvVar`/`hasEnvVar`/`deleteEnvVar`/`getAllEnvVars`, plus runtime vars with a five-level precedence chain, so this is table stakes rather than a Postman quirk.

## Fix

Add to each of `pm.environment`, `pm.globals`, `pm.collectionVariables`:

- `has(key)` - true when the key exists **and is enabled**, matching what `get` already honours (`js_pm_environment_get` skips disabled entries, `:1248-1264`).
- `unset(key)` - remove it.
- `clear()` - remove all in that scope.
- `toObject()` - plain key/value snapshot.

Add `pm.variables` with Postman's precedence: **local → data → environment → collection → global**. Vayu has no local or data scope, so it resolves environment → collection → global, which is the order `docs/app/variable-resolution.md` already documents for the app. `pm.variables.set` writes to the **local** scope in Postman; with no local scope here, the honest choices are to write to environment or to reject the write loudly. Pick one and document it - do not silently pick environment and let a user believe it is ephemeral.

[...]

Issue #182, opened 2026-07-29
docs(engine,app): `pm.variables` and `pm.response.headers.get()` are taught but do not exist


athrvk:
## Problem

Two APIs are actively taught to users and are not in the runtime. Both verified absent by probing the engine at `4136c66` (`typeof` each name in a real script context).

**1. `pm.variables` - documented in one doc, denied in the other.**

`docs/engine/scripting.md:149-159` has a whole section:

```markdown
## Variables (`pm.variables`)

const value = pm.variables.get('baseUrl');
pm.variables.set('baseUrl', 'https://api.example.com');
```

…and `:438` / `:446` say pre-request and test scripts "Can access `pm.environment` and `pm.variables`".

`setup_pm_object` installs exactly seven members (`script_engine.cpp:1208-1624`): `response`, `request`, `environment`, `globals`, `collectionVariables`, `test`, `expect`. **`pm.variables` is `undefined`**, so `pm.variables.get('x')` throws `TypeError: cannot read property 'get' of undefined`.

Meanwhile `docs/app/pm-api-compatibility.md:56` correctly lists `pm.variables.*` under "Not (yet) supported". **The two docs contradict each other**, and the aspirational one is the one a script author reads first - the exact shape of #109, which was just fixed one section away in the same file.

**2. `pm.response.headers.get()` - taught by the app itself.**

`app/src/modules/request-builder/components/RequestTabs/panels/script/script-variants.tsx:131` puts this in the Tests panel's quick reference, on screen next to the editor:

```js
pm.response.headers.get("Content-Type")
```

`setup_pm_response` builds `headers` as a plain object of key/value pairs (`script_engine.cpp:1197-1202`). There is no `.get`, so copying the line the app suggests throws `TypeError: not a function`. Postman's `pm.response.headers` is a `HeaderList` with `.get()`/`.has()`; ours is a bag.

## Fix

**Make them honest now; implementing them is a separate issue.** This issue is the cheap half - stop misleading people today.

- `scripting.md`: delete the `## Variables (pm.variables)` section, or relabel it "not yet supported" with a link to the implementing issue. Fix the two "Can access … and `pm.variables`" bullets. Point readers at the scoped accessors that do exist.
- `script-variants.tsx`: replace the `headers.get(...)` line with the form that works - `pm.response.headers["content-type"]` - and note the keys are lower-cased by the engine.
[...]

Issue #180, opened 2026-07-29
Postman `pm.*` compatibility sweep (2026-07-29): silent false-passes, phantom APIs, and the missing surface


athrvk:
Tracking issue for `pm.*` compatibility with Postman. Every finding below was **probed against the running engine** at commit `4136c66` (a throwaway gtest that `typeof`-checks each name and executes each assertion), not read off the docs - two of the findings are cases where the docs and the runtime disagree, so the docs could not be the source of truth.

The severity ordering is deliberate: a **silently passing assertion is worse than a missing feature**, because a missing feature throws and a false pass ships a broken API with a green suite.

## Sub-issues

**High - wrong answers, not gaps:**

- **#181** - `pm.response.to.be.*` is an empty placeholder, so `pm.response.to.be.ok` on a **500 passes**. Verified: a `pm.test` containing only that assertion reports PASS against a 500 response.
- **#182** - two APIs we teach and do not have: `docs/engine/scripting.md:149-159` documents `pm.variables` (absent from the runtime, and `pm-api-compatibility.md:56` correctly lists it as unsupported - the two docs contradict each other), and the app's Tests panel offers `pm.response.headers.get("Content-Type")` (`script-variants.tsx:131`), which throws because `pm.response.headers` is a plain object.

**Medium - incompatible semantics and missing surface:**

- **#183** - `equal` and `eql` are the same function (`script_engine.cpp:889-890`) doing JSON-string comparison. In chai `equal` is strict `===` and `eql` is deep, so we pass where Postman fails and fail where Postman passes (key order). Plus the common matchers that are absent.
- **#184** - variable scopes have only `get`/`set` (`script_engine.cpp:1509-1516`). Postman has `has`/`unset`/`clear`/`toObject`/`replaceIn`, plus the merged `pm.variables` accessor.
- **#185** - request/response surface: `pm.request.headers.add/upsert/remove/has`, URL parts, `pm.response.reason()`, `size()`, `headers.get()/has()`.
- **#186** - dynamic variables (`$guid`, `$timestamp`, `$randomInt`, …). Highest value per line of code here, and both Postman and Bruno have them.

**Large - new capability, own design:**

[...]

PR #214, opened 2026-07-30
feat(engine,app): request/response header methods, `reason()` and `size()`

### LINKED ISSUE #185: engine(scripting): request/response surface - header helpers, `reason()`, `size()`, URL parts
## Problem

`pm.request` and `pm.response` expose plain data where Postman exposes objects with methods, so idiomatic Postman scripts throw rather than run.

**Response** (`setup_pm_response`, `engine/src/runtime/script_engine.cpp:1150-1209`) has `code`, `status`, `responseTime`, `responseTimeWire`, `responseTimeQueueWait`, `errorCode`/`errorMessage`, `json()`, `text()`, `headers`, `to`. Missing:

- `headers.get(name)` / `headers.has(name)` - `headers` is a plain object (`:1197-1202`), so `.get` throws. The app's own Tests panel suggests this call (tracked in #TBD_PHANTOM).
- `reason()` - the status text. The engine **already has it** as `Response::status_text`; it is simply not exposed. Written but never read, at the binding.
- `size()` - Postman returns `{ body, header, total }`.
- `responseSize`, `cookies`.

**Request** (`setup_pm_request`, `:1211-1247`) has `url`, `method`, `headers`, `body`, all now writable via the #109 write-back. Missing:

- `headers.add/upsert/remove/get/has` - Postman's `HeaderList` methods. Our plain object supports assignment and `delete`, which covers the same ground but does not run a Postman script unchanged.
- URL parts: `url.query`, `url.path`, `url.host`. Ours is a bare string, and there is no `URL` global in the sandbox to parse it with, so a script must do string surgery (`docs/engine/scripting.md` now shows exactly that, at length, because there is no alternative).
- `auth`, `name`, `id`.

Bruno covers the same ground with `getHeader`/`setHeader`/`deleteHeader`/`headerList` and `getHost`/`getPath`/`getQueryString`/`getPathParams`.

## Fix

Ordered by value, and worth splitting across PRs if it gets long:

1. **`headers.get()` / `has()` on both** request and response. Smallest change, most-used idiom, and it makes the app's existing quick-reference line true. Case-insensitive lookup, since HTTP header names are - note the response object's keys are lower-cased by the engine and the request's are not.
2. **`reason()`** - one line, the data is already on `Response`.
3. **`size()`** - `{ body, header, total }` from the body length and the serialised header bytes.
4. **`pm.request.headers.add/upsert/remove`** as methods over the same object the #109 write-back reads. They must mutate the same JS object, or a script mixing methods and assignment gets two views of one header set.
5. **URL parts** - `pm.request.url` stays a writable string for compatibility with #109; add `pm.request.url.*` accessors only if they can coexist with that. A string with attached properties is a footgun; a separate `pm.request.getUrlParts()`-style accessor may be the honest shape. **Decide this explicitly in the PR** rather than by accident.

Point 5 is the one with a real design question in it. It is fine to land 1-4 and leave 5 to a follow-up.

## App changes (required)

Once `headers.get` exists, revert the workaround from #TBD_PHANTOM: `script-variants.tsx` can go back to suggesting `pm.response.headers.get("Content-Type")`, and its test updates with it. That is a deliberate reversal and should be named in the PR.

Completions (`engine/src/http/routes/scripting.cpp`) gain every new member - engine-side, guarded by `script_completions_test.cpp`.

## Tests

- `headers.get` is case-insensitive on both request and response; a missing header returns `undefined`, not a throw.
- `reason()` returns the status text, including for a synthetic status 0 (client-side failure) where `status_text` comes from `vayu::http::status_text`.
- `size()` totals body and headers; an empty body reports 0 rather than being absent.
- `pm.request.headers.add`/`upsert`/`remove` are visible to the #109 write-back - i.e. a header added by method call actually goes on the wire. This is the interaction most likely to break: assert the sent `Request`, not the JS object.
- Mixing a method call and a direct assignment in one script yields one coherent header set.

## Docs (same commit)

- `docs/engine/scripting.md` - the `## Request Object` and `## Response Object` blocks are the API inventory.
- `docs/app/pm-api-compatibility.md` - move the newly supported members out of the unsupported list.

## Sequencing

Part of #180. Touches `setup_pm_request` / `setup_pm_response` (`script_engine.cpp:1150-1247`), which is the same region #109's write-back landed in - re-verify anchors on pickup. Parallel-safe with the assertion and variable-scope sub-issues.

Do it **after** #TBD_PHANTOM, so the app's quick-reference line is corrected first and restored here rather than left wrong in between.

## Acceptance criteria

- `pm.response.headers.get()` / `has()` work, case-insensitively.
- `reason()` and `size()` return what Postman returns.
- Request header methods and direct assignment agree, proven against the sent `Request`.
- Whatever is decided for URL parts is documented, including if it is deferred.
- Both script docs match the runtime.

athrvk:
## Description

`pm.request` and `pm.response` exposed plain data where Postman exposes objects with methods, so idiomatic Postman scripts threw rather than ran. Items 1-4 of #185's Fix list are implemented; item 5 (URL parts) is deliberately deferred with the reason recorded in both script docs.

**Plan (root cause, approach, rejected alternative).** The root cause is not that the methods were missing but *where* they would have to live: `apply_pm_request_writeback` reads `pm.request.headers`' own **enumerable** string properties as the header set that goes on the wire, so any method placed on that object naively would be read as a header whose value is a function and would fail the whole write-back. The approach is therefore to install `get`/`has` (and, on the request, `upsert`/`add`/`remove`) as **non-enumerable** properties of that same object, and to `JS_DefinePropertyValueStr` header entries with `JS_PROP_C_W_E` rather than `JS_SetPropertyStr` so a header field literally named `get` overwrites the method's attributes too and still reaches the wire. Living on the same object is what makes a method call and a plain assignment reach one property set rather than two views of it - the interaction #185 flagged as most likely to break, and the one every request-side test here asserts against the sent `Request`, not the JS object. The alternative I rejected was a separate `HeaderList`-style wrapper object mirroring the plain one: it reads more like Postman, but it reintroduces exactly the "two views of one header set" bug, and the write-back would have to learn about a second shape.

**Verified against reality first.** `setup_pm_request` / `setup_pm_response` are still where #185 says (`script_engine.cpp`, now ~1660-1760 after the write-back landed above them), `Response::status_text` is still populated and unread by the binding, and #182's app-side workaround is still in place. One anchor moved: the guard test #182 left behind (`ResponseHeadersIsAPlainObjectReadByLowerCasedKey`) explicitly says "put it back in the Tests panel quick reference" when `.get` exists, so it is rewritten rather than deleted.

### Three deliberate divergences from Postman

- **The methods are case-insensitive; indexing stays case-sensitive.** `upsert('authorization', v)` writes through an existing `Authorization` rather than adding a second spelling - which the write-back refuses outright as a clash, since `Headers` is a case-insensitive map.
- **`add` throws on a name already present** and names `upsert` in the error. Postman's `HeaderList` holds duplicates and `add` appends one; a single-valued `Headers` map cannot represent that, and silently behaving as `upsert` would hide the difference rather than report it. This matches the write-back's existing refusal of a two-casing header set right next door.
- **`reason()` always returns a string.** Postman returns `null` when there is no reason phrase; vayu's synthetic status `0` (client-side failure) has a canonical answer in `vayu::http::status_text`, so it reads `"Error"`.

### URL parts: deferred, with a reason

`pm.request.url.query` / `.path` / `.host` cannot coexist with what `pm.request.url` already is. The write-back requires it to still be a **string** when the script returns (`JS_IsString`, else the request is rejected), and a JS string primitive cannot carry properties - boxing it into a `String` object to hang them off would reject every request. Any accessor therefore has to be a separate member (`pm.request.getUrlParts()`-shaped), and that shape has not been decided, so URL parsing stays string work. Documented in `docs/engine/scripting.md` ("URL parts are not exposed") and in the not-supported list of `docs/app/pm-api-compatibility.md`, and cross-linked from the sandbox-limitations section that already explains there is no `URL` global.

### App changes (required by the issue)

#182's workaround is reverted: the Tests panel quick reference suggests `pm.response.headers.get("Content-Type")` again, plus `reason()` and `size().total`, and the pre-request panel gains `upsert`. Its guard test flips from "suggests only header reads the runtime supports" (asserting the absence of `.get(`) to pinning the restored form on both variants - so dropping the runtime methods without pulling the suggestion again fails there.

Completions (`routes/scripting.cpp`) gain all nine new members, guarded against the runtime by a new `typeof` check in `script_completions_test.cpp`, mirroring the existing `pm.response.to.be.*` guard. That guard is the reason #182 happened: the list and the runtime are different translation units.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

All green on this branch, run exactly as #185 specifies:

- `python build.py -e -t` then `cd engine && ctest --preset linux-dev --output-on-failure` - **655/655 passed** (640 before; 15 new).
- `cd app && pnpm test` - **1931 passed across 230 files**.
- `mkdocs build --strict -f .github/mkdocs.yml` - clean (both new anchors resolve).
- Prettier on the one app source file touched (it was clean before); no repo-wide formatting.

**Mutation-checked.** The load-bearing claim is that the methods must be non-enumerable. Flipping `HEADER_METHOD_FLAGS` to `JS_PROP_C_W_E` and rebuilding fails 5 of the 8 header tests - the three that assert against the sent `Request` (`PreRequestHeaderUpsertReachesTheOutgoingRequest`, `...ReplacesThroughADifferentCasing`, `...RemoveIsCaseInsensitiveAndIdempotent`), `PreRequestHeaderMethodsAndAssignmentShareOneSet`, and `HeaderMethodsAreInvisibleToEnumerationAndJson`. Restored afterwards.

New engine tests, covering the failure path as well as the happy one:

- `get`/`has` case-insensitive on both objects; an absent header is `undefined`, not a throw.
- Methods invisible to `Object.keys` and `JSON.stringify`.
- Bad input refused by name: no argument, a number, an empty name, and a method detached from its object.
- No mutators on `pm.response.headers`.
- `reason()` for a real phrase and for the status-0 fallback.
- `size()` totals, `size().body === text().length`, and `0` for an empty body.
- Request methods reaching the wire: `upsert` object and two-arg forms, replacement through a different casing, `add` refusing an existing name (and leaving the request untouched), `remove` case-insensitive and idempotent, methods and assignment composing into one set, a non-primitive value refused, and a header literally named `get` surviving to the wire.

## Checklist

## Related Issues
Closes #185

Part of #180 (wave 2). Parallel-safe with #183 (PR #212) and #184 (PR #213); the only shared file is `docs/app/pm-api-compatibility.md`, where the edits are in different sections.

---
_Generated by [Claude Code](https://claude.ai/code/session_0142XB9QvBtndZA6koKgrYg5)_


athrvk:
fix merge conflict

athrvk:
Merged `master` in and resolved. Both conflicts were with #213 (the variable-scope work from #184), which landed in the same two files:

**`docs/app/pm-api-compatibility.md`** - the supported-surface table and the not-supported list. Took master's expanded scope rows plus its new "Merged variables" row, kept this branch's "Request headers" row. In the not-supported list two bullets became false in opposite directions and both are gone: this branch's `pm.variables.*` (master implements it now) and master's `pm.response.headers.get/has` (this branch implements it now). Master's other three bullets - `pm.variables.set`, `replaceIn`, `pm.environment.name` - are kept as written.

**`engine/tests/script_completions_test.cpp`** - both sides added a guard immediately after `EveryOfferedResponseStatusClassExistsInTheRuntime`. They check different surfaces, so both are kept rather than merged into one.

`script_engine.cpp` auto-merged and I checked it for semantic overlap: `setup_pm_variable_scope` and the header accessors are separate functions, and the shared `js_type_name` helper this branch moved above the write-back is still defined exactly once.

Worth noting one interaction the merge creates, since it is the kind that passes both suites separately and could have failed together: #213's new completions guard enumerates each variable scope with `for (var key in scope)` and requires every enumerable function to be in the completion list. This branch's header methods are deliberately **non-enumerable**, so they are invisible to that loop - correctly, since they are not variable-scope methods. They are covered by their own `typeof` guard instead.

Re-verified after the merge, all green:

- `python build.py -e -t`; `ctest --preset linux-dev --output-on-failure` - **666/666** (655 on this branch before the merge, 651 on master).
- `pnpm test` - **2002 passed across 231 files**.
- `pnpm type-check` - clean.
- `mkdocs build --strict` - clean; the cross-link master added to `scripting.md#variables-pmvariables` and the two anchors this branch added both resolve.

The test counts in the PR description above are the pre-merge ones; these are the current numbers.

---

athrvk:
Master moved again while the first merge was in CI - #212 landed, and it touches every file this branch does. Merged it in rather than leaving the branch behind, since a textually clean merge is not proof of a semantically clean one when the overlap is this complete.

It was clean both ways. #212's work is in `create_expectation`; this branch's is in `setup_pm_request` / `setup_pm_response`, and its new script-panel test asserts the scripting-guide link rather than the quick-reference lines this branch rewrites.

Branch head is `5cb9203`, zero commits behind master, and the PR diff is still the same **8 files / +919 / -79** it was when opened - neither merge widened the change.

CI green on the new head, all 19 checks:

- Engine `682/682` on ubuntu, macOS and Windows.
- App `2007/2007` across 231 files on all three platforms.
- `pnpm type-check`, `mkdocs build --strict`, CI gate - all clean.

Nothing further from me; it is ready for your review.

---

PR #221, opened 2026-07-30
feat(app): dynamic variables (`{{$guid}}`, `{{$timestamp}}`, `{{$random*}}`)

### LINKED ISSUE #186: engine,app: dynamic variables (`$guid`, `$timestamp`, `$randomInt`, …)
## Problem

Vayu has no dynamic variables. Postman resolves `{{$guid}}`, `{{$timestamp}}`, `{{$randomInt}}` and around a hundred `{{$random*}}` faker values anywhere a `{{variable}}` is accepted - URL, headers, body, scripts - and Bruno has an equivalent set. A collection imported from either brings requests whose URLs and bodies contain these, and in Vayu they resolve to an **empty string** today, because unknown names resolve to empty.

That last part is the sharp edge: an imported request with `{{$guid}}` in its body silently sends an empty field rather than failing, so the import looks like it worked.

The most-used handful, in rough order: `$guid`, `$randomUUID`, `$timestamp`, `$isoTimestamp`, `$randomInt`, `$randomAlphaNumeric`, `$randomEmail`, `$randomFirstName`, `$randomLastName`, `$randomFullName`, `$randomCompanyName`, `$randomUrl`, `$randomIP`, `$randomBoolean`, `$randomPassword`.

## Fix

**Where this resolves is the whole design question, and it is not obvious.**

`{{…}}` interpolation happens **app-side** today, before the payload reaches the engine (`useVariableResolver.ts` in the renderer, `resolve.ts` for MCP - the known duplication CLAUDE.md documents). So the natural home is a resolver in the app, applied wherever a `{{name}}` is looked up and found to start with `$`.

But `pm.variables.get('$guid')` in a script would then need an engine-side copy, which is a third implementation of the same table (renderer, MCP, engine). Options:

1. **App-side only.** `{{$guid}}` works in URL / headers / body; `pm.variables.get('$guid')` does not. Cheapest, and matches where interpolation already lives. Needs the renderer and MCP copies to stay in step - the duplication that already exists and is already guarded by `resolve.test.ts`.
2. **Engine-side only**, resolved after the script runs. Wrong today: it would require moving `{{…}}` resolution into the engine, which is deferred backlog item **A1** in `docs/plans/pending-backlog.md`. Do not start that here.
3. **Both**, with one shared table expressed twice. Honest, and the same trade the script-part builders already make.

**Recommendation: option 1 first**, scoped to interpolation, with the script-side accessor called out as not covered. Revisit when A1 lands.

Each generator must be a **function evaluated per occurrence**, not a value computed once: two `{{$guid}}` in one request must differ, which is what people use them for. A load run must generate per-iteration values, or every request in the run carries the same "unique" id - the case where getting this wrong is least visible and most damaging.

## App changes (required)

This is mostly app work: the generator table plus the resolver hook, and the same table in `app/electron/mcp/resolve.ts` if MCP is in scope. The variable autocomplete (`variable-completion.ts`) should offer them, or they stay undiscoverable.

## Tests

- Each generator produces a value of the right shape (`$guid` matches the UUID pattern, `$timestamp` is a plausible epoch, `$randomInt` is in range).
- **Two occurrences in one string differ** - the per-occurrence rule, and the one a naive implementation gets wrong.
- A load run's iterations differ from each other.
- An unknown `$name` does something defined and loud rather than resolving to empty - this is the current defect, so the fix must not preserve it for typos.
- Renderer and MCP agree, extending `resolve.test.ts`, if MCP is in scope.

## Docs (same commit)

- `docs/app/variable-resolution.md` - the resolution rules; this adds a new class of name.
- `docs/engine/scripting.md` / `docs/app/pm-api-compatibility.md` - state clearly whether the script-side accessor is covered, since the answer under option 1 is "no".
- `docs/app/import-collections/` - imported collections are the main source of these; note that they now resolve.

## Sequencing

Part of #180. Independent of the other sub-issues: touches the app's resolver rather than `script_engine.cpp`. Highest value per line of code in the series, and the only one whose absence corrupts **imported** collections rather than hand-written scripts.

Related to deferred backlog item **A1** (consolidating resolution into the engine) but must not depend on it.

## Acceptance criteria

- The documented generator set resolves in URL, headers and body.
- Two occurrences in one request produce different values, and load-run iterations differ.
- An unknown `$name` is not silently empty.
- Renderer and MCP agree if both are in scope.
- The chosen option, and what it excludes, is written down.

athrvk:
## Description

A `{{name}}` starting with `$` now names a **generator** instead of a lookup. Postman and Bruno both have these, so an imported collection arrives full of them; before this they resolved to an empty string, which is the sharp edge issue #186 names - a request whose body carries `{{$guid}}` sent an empty field and the import looked successful.

**Root cause, approach, and the alternative rejected.** The defect is not a missing table so much as a missing *case*: `resolveString` had one branch for "the map has this name" and one for everything else, and the second branch returns `""`. Anything the user writes that no scope defines - a typo, a Postman faker name - lands in it silently. The fix adds a third case before the fallback, driven by a table of functions, and makes the `$` prefix meaningful: a name declared as a generator that Vayu does not have keeps its braces rather than joining the silent-empty path. Resolution stays **app-side, at interpolation time** (option 1 in the issue), which means the table is duplicated for the MCP client exactly as variable and auth resolution already are. The rejected alternative is option 2/3, resolving `$` names engine-side: it is the only way to get per-iteration values in a load run, and it requires `{{…}}` interpolation to move into the engine - deferred backlog item **A1**, which the issue says explicitly not to start here. That exclusion is not hidden; it is written into the docs and surfaced in the load-test dialog (see below).

**Three rules the implementation turns on**

1. **Scopes win.** A workspace defining a variable literally named `$guid` keeps its value; only a name nothing defines reaches a generator. Adding the table cannot change what an existing request sends.
2. **One value per occurrence.** The table holds functions, called inside the `replace` callback, so two `{{$guid}}` in one body are two different ids. A table of precomputed values would pass every shape test and fail this one.
3. **An unknown `$name` keeps its braces.** `{{$randomInteger}}` is sent as that literal text, not `""`, so it is visible in the request and the token stays marked unresolved in the UI. Ordinary unknown names still resolve to `""`, unchanged.

**The set:** `$guid`, `$randomUUID`, `$timestamp`, `$isoTimestamp`, `$randomInt`, `$randomAlphaNumeric`, `$randomBoolean`, `$randomEmail`, `$randomFirstName`, `$randomLastName`, `$randomFullName`, `$randomCompanyName`, `$randomUrl`, `$randomIP`, `$randomPassword`.

**Discoverability** (the issue calls it out: without it they stay invisible, since no scope can list them). The `{{` list offers them under a **Dynamic** heading in the plain fields (`VariableAutocomplete`) and in the body editors (the Monaco provider), sorted below the workspace's own names and dropped when a real variable shadows one - the generator would not run there.

**Two deviations worth naming**, both because the alternative would have been a UI that contradicts the behaviour:

- `DynamicVariableToken` is a new sibling of `EditableVariable` rather than a flag on it. `EditableVariable` paints a name nothing defines in `text-destructive-text` with a "not defined" tooltip and opens a popover to edit the stored value. For a generator both are wrong: it resolves every time, and there is no stored value - any value the popover showed would be one it had just invented rather than the one the request will carry.
- The load-test dialog gains a `hasDynamicVariables` notice, following the existing `hasPreRequestScript` / `OAuth2LoadTestGuard` precedent. A run is interpolated once, here, and the engine repeats the payload, so every iteration carries the same generated value. That is the least visible way to get a load test wrong - it succeeds and the data is quietly degenerate - so it is stated rather than left to be discovered. Advisory, not gating.

**Excluded, and written down** (`docs/app/variable-resolution.md` → *What this does not cover*): `pm.variables.get("$guid")` in a script (scripts run engine-side; the engine does no `{{…}}` interpolation), and per-iteration load-run values (needs A1).

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

App-only change - no engine file is touched, so no C++ test can move; per CLAUDE.md ("ask what a failure would even look like") the engine suite was not run.

- `pnpm test` - **2072 passed across 237 files**. Measured on `master` (`5ff501f`) for comparison rather than assumed: **2007 across 231**, so **+65 tests in 6 new files**.
- `pnpm type-check` - clean.
- `pnpm lint` - **100 errors / 8 warnings, identical to the base commit** (pre-existing debt, issue #189). Verified by stashing the branch and re-running; this PR adds none.
- `npx prettier --check` on every file touched - clean. No repo-wide formatting.
- `mkdocs build --strict -f .github/mkdocs.yml` - clean; the four new cross-links to `#dynamic-variables` resolve.

New coverage, written mutation-check style:

- `lib/dynamic-variables.test.ts` (25) - shape per generator, uniqueness, and the per-occurrence rule (50 `$guid` draws must all differ; a precomputed table fails here alone).
- `hooks/useVariableResolver.dynamic.test.tsx` (9) - the wiring: shadowing, per-occurrence through `resolveString`/`resolveObject`, unknown `$name` keeps its braces, ordinary unknown still empties, `hasUnresolvedVariables` agrees.
- `electron/mcp/resolve.test.ts` (+7) - the same behaviour through `composeSavedRequest`, plus a **parity guard that imports both tables** and compares names and descriptions. Test-only cross-boundary import; production code in `electron/` still cannot reach `app/src/`.
- `useVariableCompletionProvider.dynamic.test.tsx` (5) - drives the provider through a Monaco stub rather than scanning source, per the class-through-a-binding lesson in CLAUDE.md.
- `variable-autocomplete.dynamic.test.tsx` (5), `dynamic-variable-token.test.tsx` (5) - rendered, asserting on `element.className` for the token colour.
- `utils/dynamic-variable-scan.test.ts` (5) and `LoadTestConfigDialog.test.tsx` (+3) - the notice fires on exactly the fields the load path resolves, ignores disabled rows and unknown `$name`s, and does not gate Start.

## Checklist

Docs updated in the same commit: `docs/app/variable-resolution.md` (the set, the three rules, the two exclusions, the MCP copy), `docs/app/pm-api-compatibility.md` and `docs/engine/scripting.md` (script side explicitly not covered), `docs/app/import-collections/README.md` (imported collections now resolve these), `docs/engine/mcp.md` (MCP behaviour and the once-per-run caveat), `docs/app/COMPONENTS.md` (the new token component and why it is separate).

## Related Issues
Closes #186


athrvk:
Merged `master` in and resolved. Branch head is `bbf7812`, zero commits behind, base is `950ecf5`.

**One conflict, in `docs/app/pm-api-compatibility.md`.** #214 removed the `pm.response.headers.get/has(...)` bullet from *Not (yet) supported* - it implements those now - and this branch had added its dynamic-variables bullet directly after that same bullet, so the two edits landed on adjacent lines of the same list. Took master's removal and kept this branch's bullet; both changes are in the result, neither is a partial.

**`docs/engine/scripting.md` auto-merged, and I checked it rather than trusting the merge.** #214's additions are in the request/response sections; this branch's paragraph sits under `replaceIn()` in the variables section. They make the same claim about the engine doing no `{{var}}` interpolation, from the two directions it matters in, so they read as one argument rather than a repetition.

Nothing else overlapped: #214's work is in `script_engine.cpp` and the script panels, this branch's is in the app's variable resolver and the MCP copy. The PR diff is still **25 files / +1567 / -35** - the merge widened nothing.

Re-verified on the merged tree:

- `pnpm test` - **2072 passed across 237 files** (unchanged; #214 added engine tests, not app ones, so the app baseline is still 2007/231).
- `pnpm type-check` - clean.
- `mkdocs build --strict` - clean, including the cross-link into `#dynamic-variables` from the file that conflicted.
- `prettier --check` on the files touched - clean.

The engine suite still was not run locally: this branch touches no engine file, and everything the merge brought into `engine/` arrived already green on master. CI covers it either way.

---

Issue #188, opened 2026-07-29
engine(scripting): `pm.sendRequest`, `pm.cookies`, `pm.info` and flow control are absent (scoping issue)


athrvk:
> **Split, 2026-08-04.** This issue has done its job as a scoping record. Each of the four capabilities now has its own issue with a Problem section anchored against master `6b5cc15`; this issue is the group record and the detail below is kept for the reasoning, not as the work item. **Implement the sub-issues, not this one.**
>
> | | Capability | Issue | Size |
> |---|---|---|---|
> | 1 | `pm.info` | #300 | small - plumbing plus one client-side half |
> | 2 | `pm.cookies` / `pm.response.cookies` | #301 | small read half, large jar half |
> | 3 | `pm.sendRequest` | #302 | medium, gated on three decisions |
> | 4 | Flow control + `pm.iterationData` | #303 | gated on a collection runner that does not exist |
>
> Three claims in the text below were checked against current code while splitting and turned out to be **wrong or stale**. They are corrected in the sub-issues rather than silently fixed here, so the reasoning that produced them stays readable:
>
> - **Sync vs async is no longer an open question** (item 1). #187 answered it and shipped the answer: the sandbox has `Promise` but nothing drains the job queue, so Web Crypto shipped as a synchronous `pm.crypto` rather than a promise-based API wearing a standard name. Postman's `sendRequest` is callback-based, and a callback can be invoked synchronously - see #302.
> - **The load-metrics worry does not apply as stated** (item 1). Load runs do not execute test scripts on the hot path; `validate_scripts` runs them once per *sampled* response after the run has finished (`engine/src/core/run_manager.cpp:47-160`, called at `:705-709`). Nothing a script sends can inflate the run's numbers. The real cost is different and is recorded in #302.
> - **The parsed-but-unread cookie state is gone** (item 2). `attrs` is rendered at `ResponseCookies.tsx:69-71`; `pm-api-compatibility.md` no longer makes that claim. See #301.
>
> A fourth thing the split turned up, which nothing here anticipated: `pm.sendRequest` would bypass the MCP allowlist entirely, because that allowlist is enforced client-side in the MCP server before it calls `POST /execute`. #302 treats it as a design constraint on the feature rather than a footnote.
>
> The three `#TBD_*` placeholders below resolve to #187 (crypto), #181 (`to.be` placeholder) and #182 (phantom APIs).

## Problem

[...]

PR #230, opened 2026-07-31
feat(scripting): add pm.crypto and base64 globals so scripts can sign requests

### LINKED ISSUE #187: engine(scripting): no crypto or base64 in the sandbox - request signing is impossible in-script
## Problem

The script sandbox is QuickJS plus exactly two globals, `pm` and `console` (`setup_pm_object`, `engine/src/runtime/script_engine.cpp:1616-1629`, and the `console` install at `:229`). There is **no** `crypto`, `btoa`/`atob`, `TextEncoder`, `URL`, `URLSearchParams` or `setTimeout`.

Pinned by `NoCryptoBase64OrUrlParserIsExposed` in `engine/tests/script_engine_test.cpp`, added when #109 landed precisely so this claim cannot rot.

The consequence is specific: **#109 gave pre-request scripts the ability to rewrite the outgoing request, and the headline use case for that - signing it - cannot be done.** HMAC, SHA, and base64 are all unavailable. `docs/engine/scripting.md` currently works around this with a pure-JS FNV-1a checksum labelled "not a cryptographic hash - do not use it as one", which is honest but is not what someone signing an AWS or webhook request needs.

Two smaller consequences: no base64 means no hand-built Basic auth header (the Auth tab covers it, so this is minor), and no `URL` means query-string edits are string surgery, which `scripting.md` spends a whole worked example on.

For scale: Postman exposes **Web Crypto globally** (`Crypto`, `CryptoKey`, `SubtleCrypto`) and has deprecated `crypto-js` in favour of it; it also ships `require()` for ajv, chai, cheerio, lodash, moment, uuid, xml2js, csv-parse and postman-collection, plus Node builtins. Bruno allows `require()` in Developer Mode. Vayu has none of this.

## Fix

**Scope this deliberately - "add Web Crypto" is much larger than it sounds.** Ordered by value-to-cost:

1. **`btoa` / `atob`.** Pure C++, no dependency, a few dozen lines. Unblocks base64 for signatures computed elsewhere and for Basic-style headers.
2. **A hashing surface**: SHA-256 at minimum, plus HMAC-SHA256. This is what "sign the request" actually needs. OpenSSL is **already a transitive dependency** (curl links it, and `engine/vcpkg.json` resolves `openssl` - see the vcpkg output in any build), and `picosha2` is already vendored for PKCE. Either could back this without a new dependency, which is the main reason to prefer a small surface over a full Web Crypto port.
3. **Shape**: a real `crypto.subtle` is Promise-based, and the sandbox has **no event loop** - `Promise` exists but nothing drains the job queue, and `setTimeout` is absent - so an async API would hang. A **synchronous** helper is the honest shape here. Naming it `crypto.subtle` while making it synchronous would be a lie of a familiar kind; prefer an explicit `pm.crypto`-style or global `vayuCrypto`-style surface, or implement the async one only alongside a job-queue pump. **Decide and document.**
4. **`TextEncoder`** if the hashing API needs byte input; otherwise take strings and skip it.
5. `URL` / `URLSearchParams` - separate concern, lower value, worth its own issue if wanted.

Explicitly out of scope: `require()` and a library bundle. That is a much larger surface and a licensing question (the engine is AGPL-3.0); it should be its own issue if it is ever wanted.

## App changes (required)

None for the runtime. The **completion list** (`engine/src/http/routes/scripting.cpp`) must gain whatever is added, or it is undiscoverable - engine-side, guarded by `script_completions_test.cpp`.

## Tests

- Known-answer vectors, not round-trips: SHA-256 and HMAC-SHA256 against published test vectors; base64 against RFC 4648 examples including padding cases and an empty string.
- Non-ASCII input, since encoding is where a hand-rolled base64 or a string-to-bytes step goes wrong.
- **Update `NoCryptoBase64OrUrlParserIsExposed`** - it currently asserts these are *absent* and will fail by design. That failure is the reminder to rewrite the "you cannot sign in a script" note in `scripting.md`; it says so in its own message.
- An end-to-end pre-request test: sign a canonical string, set the header, assert the sent `Request` carries it - the #109 write-back and this feature meeting.

## Docs (same commit)

- `docs/engine/scripting.md` - **`### Sign a request with a checksum you can actually compute`** and **`### What a script can compute`** both exist only because this is missing. Both need rewriting, and the FNV-1a example should become a real signing example.
- `docs/app/pm-api-compatibility.md` - the sandbox-globals list.
- `docs/engine/architecture.md` - the Script Engine section, if the dependency surface changes.

## Sequencing

Part of #180. **Do it last of the code changes.** It is the only sub-issue that may add a dependency or a new global surface, and it should not sit in front of the correctness fixes (#TBD_TO_BE, #TBD_PHANTOM, #TBD_CHAI), which are small and fix active misbehaviour.

Independent of every other sub-issue in file terms - it touches the sandbox setup rather than the `pm` bindings.

## Acceptance criteria

- Base64 and a hashing surface exist, verified against published vectors.
- The sync-vs-async decision is documented with its reason.
- The sandbox-surface tests are updated rather than deleted.
- `scripting.md`'s signing example computes a real signature, and the "no crypto" note is gone or corrected.

athrvk:
## Description

**Plan.** The root cause is that #109 gave pre-request scripts the ability to rewrite the outgoing request while the sandbox had no way to compute a signature over the result: QuickJS plus `pm` and `console`, no hash, no base64. `scripting.md` papered over it with an FNV-1a checksum explicitly labelled "not a cryptographic hash - do not use it as one". The approach is the issue's items 1 and 2 only, built on dependencies the engine already carries: `btoa` / `atob` as globals, and a `pm.crypto` object with `sha256` and `hmacSha256` backed by the vendored picosha2 that PKCE already uses. **The alternative I rejected is a real `crypto.subtle`**: Web Crypto is Promise-based, and this sandbox has no event loop and no `setTimeout`, so nothing drains the job queue - an `await crypto.subtle.digest(...)` would never resume and the script would report a timeout rather than a result. Naming a synchronous API `crypto.subtle` is exactly the lie of a familiar name that `NoCryptoBase64OrUrlParserIsExposed` was written to catch, so the surface takes a name of its own and the sync decision is documented with its reason in three places. I also rejected implementing a job-queue pump, which is a much larger change than "let me sign this request".

I verified the problem still exists as described on current master before starting: the two globals, the `NoCryptoBase64OrUrlParserIsExposed` test, and the FNV-1a doc section were all where the issue said.

**What landed:**

- `pm.crypto.sha256(data, encoding?)` and `pm.crypto.hmacSha256(key, data, encoding?)`. `encoding` is `'hex'` (default), `'base64'`, `'base64url'` or `'bytes'`.
- `btoa` / `atob` globals with standard web semantics - binary strings, one byte per code unit, so `btoa` throws above U+00FF rather than silently UTF-8 encoding.
- `utils/hmac_sha256` (RFC 2104 over the existing `utils/sha256`), `utils/base64_decode` and `utils/hex_encode`. **No new dependency; OpenSSL is still not linked.**

**Two design points worth a reviewer's eye:**

1. **What a string means as bytes.** `pm.crypto` hashes a string's **UTF-8** bytes - the bytes the engine puts on the wire, so a digest here matches `sha256sum` over the same text. `btoa` keeps **Latin-1** code-unit semantics, because that is what the web name promises. The two conventions differ, deliberately, and each is pinned by a test.
2. **`'bytes'` output.** AWS SigV4 derives its signing key in four HMAC rounds, each keyed by the *raw* digest of the previous one; with only text outputs that chain cannot be expressed. `'bytes'` returns a `Uint8Array`, which is also accepted as key or data. Pinned against the published SigV4 example key.

**Blast radius traced:** the sandbox surface has one reader in the runtime and three documented readers - the completion list (engine-side, served to Monaco), `docs/engine/scripting.md`, and the app's pre-request quick reference in `script-variants.tsx`, which asserted "No crypto, base64 or `URL` in the sandbox" and is now two-thirds false. All three are updated. Per CLAUDE.md's "written but never read", the completion entries have a test that *calls* each offered name in the runtime rather than checking a string is present.

**Deliberate deviations from the issue spec**, both toward a smaller surface:

- **`TextEncoder` skipped** (the issue's item 4 makes it conditional). Accepting `Uint8Array` directly covers byte input without a second global.
- **No `URL` / `URLSearchParams`** - the issue scopes them out as their own issue.
- `base64_decode` is slightly **stricter than WHATWG `atob`**: it rejects trailing bits that decoding would drop, e.g. `atob('Zm9vYmF')`. Truncated base64 failing loudly beats returning bytes the author never encoded. Whitespace is still forgiven.

**Cross-platform note:** the implementation reads and builds typed arrays through `JS_GetTypedArrayBuffer` / `JS_NewArrayBufferCopy` rather than `JS_GetUint8Array` / `JS_NewUint8ArrayCopy`. The engine vendors Bellard's QuickJS on Linux/macOS and quickjs-ng on Windows, and only the latter has the Uint8Array shortcuts - the first build failed on exactly this.

**Adjacent things I left alone:** `docs/engine/scripting.md`'s Limitations section, and the `crypto`-adjacent items in #188 (`pm.sendRequest`, `pm.cookies`). #187 is the last code item in the #180 series, so #112's "land after the series drains" gate is now down to #188, which is a scoping record rather than code.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

All run on this branch, all green:

- `python build.py -e -t` - clean, 0 errors.
- `cd engine && ctest --preset linux-dev --output-on-failure` - **706 passed, 0 failed** (+24 net new; 26 added, 2 renamed).
- `cd app && pnpm test` - **2098 passed, 240 files**.
- Prettier + ESLint on the one touched `.tsx`; `git clang-format` on changed lines only, leaving the pre-existing (non-clean) formatting of those files untouched per CLAUDE.md.

New coverage, chosen as known-answer vectors rather than round-trips (a round-trip passes over two mirrored bugs):

- SHA-256 against FIPS 180-4; **HMAC-SHA256 against RFC 4231 cases 1, 2, 3 and 6** - case 6's 131-byte key exercises the "hash the key down first" branch of RFC 2104 that short keys never reach.
- Base64 against RFC 4648 §10 including every padding case and the empty string, both directions, plus a full 0-255 byte round trip.
- Non-ASCII in three shapes: `sha256('héllo')` equals `printf 'h\xc3\xa9llo' | sha256sum`; `btoa('héllo')` gives the **Latin-1** encoding, not the UTF-8 one; `btoa('naïve €')` throws.
- Failure paths asserted, not assumed: malformed base64, an object / number / null / missing argument to `sha256`, an unknown encoding name (and that the message names the valid ones), a missing HMAC argument.
- End to end: a pre-request script signs a canonical string and the sent `Request` carries the header - #109's write-back and this feature meeting.

**Mutation-checked** (applied the mutation, rebuilt, confirmed red, reverted):

| Mutation | Result |
|---|---|
| `key.size() > block_size` → `>=` in `hmac_sha256` | `HmacSha256.KeyLengthsAroundTheBlockBoundary` fails |
| `btoa` passes UTF-8 through instead of Latin-1 | 3 tests fail (`BtoaRejectsCharactersAboveLatin1`, `BtoaEncodesLatin1SupplementCharacters`, `Base64CarriesHighBytesThroughTheJsBoundary`) |

The first is the reason the boundary test pins *known answers* rather than "the two differ" - an off-by-one there still yields two different MACs, so the weaker assertion would have passed while a 64-byte key was quietly being hashed down.

## Checklist

## Related Issues
Closes #187

Part of #180 - this is the last of that series' code changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_018Xq11zNfSUizKCNxF3FxTF)_

### DIFF
--- a/app/src/modules/request-builder/components/RequestTabs/panels/script/script-variants.tsx
+++ b/app/src/modules/request-builder/components/RequestTabs/panels/script/script-variants.tsx
@@ -112,7 +112,10 @@ export const SCRIPT_VARIANTS: Record<ScriptVariant, ScriptVariantConfig> = {
 				tab.
 			</>,
 			<>
-				No crypto, base64 or <code className={CODE_CLASS}>URL</code> in the sandbox, and
+				Sign what you send with <code className={CODE_CLASS}>pm.crypto.hmacSha256</code> or{" "}
+				<code className={CODE_CLASS}>pm.crypto.sha256</code> (synchronous, hex by default);{" "}
+				<code className={CODE_CLASS}>btoa</code> / <code className={CODE_CLASS}>atob</code>{" "}
+				are there too. No <code className={CODE_CLASS}>URL</code> parser in the sandbox, and
 				load tests do not run pre-request scripts at all.
 			</>,
 		],
--- a/docs/app/pm-api-compatibility.md
+++ b/docs/app/pm-api-compatibility.md
@@ -26,6 +26,8 @@ intent is that the most common Postman scripts paste in and run unchanged.
 | Globals             | `pm.globals.get/set/has/unset/clear/toObject`                                    |
 | Collection vars     | `pm.collectionVariables.get/set/has/unset/clear/toObject`                        |
 | Merged variables    | `pm.variables.get(name)`, `.has(name)`, `.toObject()` - read-only, see below     |
+| Crypto              | `pm.crypto.sha256(data, encoding?)`, `.hmacSha256(key, data, encoding?)` - synchronous, see below |
+| Base64              | `btoa(binaryString)`, `atob(base64)` - globals, standard web semantics           |
 | Console             | `console.log/info/warn/error`                                                    |
 
 `pm.response.headers` is a plain object keyed by the **lower-cased** header name, not
@@ -51,6 +53,37 @@ because Postman writes it to a per-request *local* scope that Vayu does not have
 alternatives (persisting to the environment, or dropping the write) would misrepresent
 what happened. The error names the three scoped setters. See
 [scripting.md](../engine/scripting.md#variables-pmvariables).
+
+### Hashing (`pm.crypto`) is Vayu's own name, and it is synchronous
+
+Postman exposes **Web Crypto** globally (`crypto.subtle`), whose every method
+returns a Promise. Vayu's sandbox has no event loop and no `setTimeout`, so
+nothing drains the job queue: an `await crypto.subtle.digest(...)` would never
+resume and the script would report a timeout rather than a result. Rather than
+wear a familiar name with unfamiliar behaviour, the surface is `pm.crypto` and
+it returns its result directly.
+
+```javascript
+pm.crypto.sha256('abc');                        // hex, 64 chars
+pm.crypto.sha256('abc', 'base64');              // 'hex' | 'base64' | 'base64url' | 'bytes'
+pm.crypto.hmacSha256(secret, canonicalString);  // hex by default
+```
+
+Strings are hashed as their **UTF-8** bytes. Pass a `Uint8Array` to hash bytes
+directly, and ask for `'bytes'` to get one back - that is what multi-round key
+derivation (AWS SigV4) needs, since each round is keyed by the raw digest of the
+previous one. Anything that is neither a string nor a byte-sized typed array
+throws: stringifying an object would hash the text `[object Object]` and return
+a digest that looks perfectly valid.
+
+`btoa` / `atob` keep their web semantics deliberately - they operate on **binary
+strings**, one byte per code unit, so `btoa` throws on a code point above U+00FF
+instead of silently UTF-8 encoding it. A signature over quietly-substituted
+bytes verifies nowhere.
+
+Not provided: `crypto`/`crypto.subtle` under those names, `TextEncoder`, MD5,
+SHA-1, and any asymmetric algorithm. The engine-side detail is in
+[scripting.md](../engine/scripting.md#what-a-script-can-compute).
 
 ### Assertion chains (`pm.expect`)
 
--- a/docs/engine/architecture.md
+++ b/docs/engine/architecture.md
@@ -169,6 +169,11 @@ JavaScript execution engine for pre-request and test scripts:
 - **Postman-compatible API**: `pm.test()`, `pm.expect()`, `pm.response`, etc.
 - **Mutable request**: a pre-request script's `pm.request` writes are applied to the
   request before it is sent (see `docs/engine/scripting.md`)
+- **Signing**: `pm.crypto.sha256` / `.hmacSha256` plus the `btoa` / `atob` globals, so
+  a pre-request script can sign the request it just rewrote. Backed by the already
+  vendored picosha2 (`utils/sha256.hpp`, shared with PKCE) and `utils/encoding.hpp` -
+  **no new dependency, and OpenSSL is still not linked**. Synchronous, because the
+  sandbox has no event loop to settle a Promise on
 - **Memory limit**: 64MB per script execution
 - **Timeout**: 5 seconds per script
 - **Sandboxed**: No filesystem or network access
--- a/docs/engine/scripting.md
+++ b/docs/engine/scripting.md
@@ -590,25 +590,12 @@ properties but one HTTP header, so the write-back rejects it rather than
 picking a winner, and the request is sent unchanged with the reason in
 `preScriptError`.
 
-### Sign a request with a checksum you can actually compute
+### Sign a request
 
-**There is no crypto in the sandbox** - no `crypto`, no `TextEncoder`, no
-`btoa`. A real HMAC is therefore not possible in-script today, and any example
-claiming otherwise is wrong. What *is* possible is a pure-JS digest over a
-canonical string, which is enough for a checksum, a cache key, or a test double
-of a signing scheme:
+`pm.crypto` gives a pre-request script a real HMAC, so the request it rewrites
+can be signed for what it actually became:
 
 ```javascript
-// FNV-1a, 32-bit. Not a cryptographic hash - do not use it as one.
-function fnv1a(text) {
-  var hash = 0x811c9dc5;
-  for (var i = 0; i < text.length; i++) {
-    hash ^= text.charCodeAt(i);
-    hash = (hash + ((hash << 1) + (hash << 4) + (hash << 7) + (hash << 8) + (hash << 24))) >>> 0;
-  }
-  return ('00000000' + hash.toString(16)).slice(-8);
-}
-
 var timestamp = Date.now().toString();
 var canonical = [
   pm.request.method,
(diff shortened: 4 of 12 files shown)


athrvk:
Why are we hand rolling things ? Can't we use something globally accepted ? an open source utility ?

athrvk:
Fair challenge, and it turned up one thing I should have caught. Splitting it into what is actually hand-rolled here, because it is less than the diff makes it look:

| Piece | Where it comes from |
|---|---|
| SHA-256 | **picosha2** (MIT, vendored) - not written here, and already the engine's hash since PKCE |
| `base64_encode` | pre-existing `utils/encoding.hpp`, used today for Basic auth and PKCE - untouched by this PR |
| **HMAC-SHA256** | ~25 new lines: the RFC 2104 construction over picosha2 |
| **`base64_decode`** | ~50 new lines, the missing half of the existing encoder |
| **`hex_encode`** | ~10 new lines |

So no new hash function, and no new cipher. The genuinely new code is the HMAC construction and a base64 decoder.

**The thing you're right about, and I've pushed it** (cef24ca): `pkce_test.cpp` had its own local hex table and loop, which my `hex_encode` turned into a duplicate. That is exactly the "hand-rolled copy of a primitive" CLAUDE.md warns about, and it's now one encoder in one place.

**On OpenSSL specifically - the issue's premise doesn't survive contact with the Windows build.** #187 says "OpenSSL is already a transitive dependency (curl links it, and `engine/vcpkg.json` resolves `openssl` - see the vcpkg output in any build)". That is true on Linux and macOS and **false on Windows**. From the vcpkg curl port manifest:

```json
"ssl": {
  "description": "Default SSL / TLS implementation.",
  "dependencies": [
    { "name": "curl", "features": ["openssl"], "platform": "uwp | !windows" },
    { "name": "curl", "features": ["sspi"],    "platform": "windows & !uwp" }
  ]
}
```

Windows gets **Schannel**, so there is no OpenSSL in that build at all. `engine/vcpkg.json` also does not list `openssl` as a direct dependency - what shows up in the Linux vcpkg output is curl's transitive resolution, and building against a transitive dependency's headers without declaring it is fragile anyway.

So "use OpenSSL" is not free here - it means adding a **new direct dependency**, and on Windows it means building and shipping all of OpenSSL for one HMAC. It would also reverse a decision that is written down in `utils/sha256.hpp` today: *"so the engine neither hand-maintains a hash implementation nor links OpenSSL."*

**Why I'd still argue this specific 25 lines is the right side of the "don't roll your own crypto" line.** That rule is about primitives - hash functions, ciphers, RNGs - where a subtle error is invisible and catastrophic. HMAC is not a primitive; it is a fixed construction (pad the key to the block, XOR with ipad/opad, hash twice) with **official test vectors published in RFC 4231**. It is either right or it fails the vectors. This one is checked against cases 1, 2, 3 and 6 - case 6 being the 131-byte key that exercises the "hash the key down first" branch a short key never reaches - plus a boundary test pinned to known answers, which I mutation-checked: flipping `>` to `>=` in the key-length test turns it red. The base64 decoder is checked against every RFC 4648 §10 vector, both directions, plus a full 0-255 byte round trip.

The one property a library would give us that this does not is **constant-time execution**, and I've documented that limit in `sha256.hpp` rather than leaving it implicit: these compute a signature to *send*, never to compare against an attacker-supplied one. If we ever add verification, that needs a constant-time compare regardless of which library computes the digest.

**Your call, and I'll take either.** If you'd rather have OpenSSL, the switch is straightforward - declare it in `vcpkg.json`, `find_package(OpenSSL)`, swap the two functions for `HMAC()` / `EVP_Digest` - with two caveats worth knowing before you say yes: it puts OpenSSL into the Windows build for the first time, and OpenSSL's own base64 (`EVP_DecodeBlock`) is a well-known footgun that doesn't report the true unpadded length, so we'd end up wrapping it and keeping most of the decoder anyway. A middle option is vendoring a small audited HMAC header, though that trades 25 lines of RFC 2104 for another vendored dependency.

Say the word and I'll do it in this PR.

---

athrvk:
Checked this properly against the headers actually installed in `engine/vcpkg_installed/`, rather than from memory. Short version: **nothing the engine already links provides base64 decode or plain hex**, and the one dependency that does provide them is not on Windows and gets both of them wrong for our use.

### What each current dependency actually exposes

| Dependency | base64 encode | base64 decode | hex |
|---|---|---|---|
| cpp-httplib | `detail::base64_encode` | **none** | none |
| curl | none public (`Curl_base64_encode` is internal) | none public | none |
| nlohmann-json, gtest, sqlite-orm | none | none | none |
| sqlite3 | none | none | SQL `hex()` only, no C API |
| OpenSSL | `EVP_EncodeBlock` | `EVP_DecodeBlock` | `OPENSSL_buf2hexstr` |

Two things about the httplib entry, since it looks like the obvious candidate: it is **encode only**, and it lives in `detail::` - and `httplib.h:4588` introduces it with a link to *"base64-decode-snippet-in-c"* on StackOverflow. So it is a hand-rolled base64 as well, just someone else's, with no stability guarantee attached. It is also not reachable from where this code lives: `vayu_core` deliberately does not link httplib (only `vayu-cli`, `vayu-engine` and `vayu_tests` do, verified in `CMakeLists.txt`), which is exactly what the existing comment in `encoding.hpp` already says.

### OpenSSL, measured rather than asserted

I compiled a probe against the installed OpenSSL 3.6.2 and ran it:

```
EVP_DecodeBlock("Zg==") -> returned len=3, bytes: 66 00 00   (true payload is 1 byte: 66)
EVP_DecodeBlock("Zm8=") -> returned len=3               (true payload 2)
EVP_DecodeBlock("Zm9$") -> -1                            (rejects invalid input, good)
OPENSSL_buf2hexstr({ba,78,16}) -> "BA:78:16"
```

- `EVP_DecodeBlock` **zero-pads and reports the padded length**. It will not tell you the payload is 1 byte; the caller has to count trailing `=` itself and subtract. Handing that raw to `atob` would return `"f\0\0"` for `atob('Zg==')`.
- `OPENSSL_buf2hexstr` returns **colon-separated uppercase**. A signature needs `ba7816...`, so every digest would be allocated, then rewritten to strip colons and lowercase it.

So adopting OpenSSL for these two would not remove the wrapper, it would add a dependency *and* keep a wrapper - on top of the Windows problem from my last comment, where curl uses Schannel and OpenSSL is not in the build at all.

### Why I think this is a different question from the HMAC one

Base64 and hex are **encodings, not crypto**. "Don't roll your own" earns its force from primitives whose bugs are silent and catastrophic. A wrong base64 produces a visibly wrong string and dies on the first RFC 4648 vector - which is why these are pinned to all seven §10 vectors in both directions, plus a full 0-255 byte round trip and six malformed-input cases. Different risk class from inventing a hash.

Worth noting too that the scope here is smaller than the question implies: `base64_encode` was **already in the tree** and already shipping for Basic auth and PKCE - this PR did not add it. What's new is the decode half next to it, and hex **encode** only (there is no hex decode in this PR, nothing needs one).

### If you still want a library, here's the coherent version

The thing I'd push back on is a *half* migration: decoding via a vendored library while `base64_encode` stays hand-written leaves two base64 implementations in the tree, which is the duplication problem you're pointing at, made worse. The coherent version is to replace the whole surface - encode and decode together - with a vendored single-header (`cpp-base64` or `modp_b64`, both BSD), which also means touching the OAuth and PKCE paths that use the current encoder. That is a bigger, more interesting change than this PR and belongs in its own issue.

Happy to file that and do it. My recommendation is to keep this PR as-is and take that separately, because it trades ~50 lines of vector-tested in-repo code for a ~200-line third-party file of comparable provenance, and the win is real but modest. But it is your call and I'll do it either way - say which and I'll get on it.

---

PR #247, opened 2026-08-01
feat(engine,app): enforce a single active environment, and stop design runs storing an indeterminate end_time

### LINKED ISSUE #242: engine(db): design-run rows insert an uninitialized end_time (garbage after a crash), and environments.is_active is a dead column
Two survivors from re-verifying an old dead-field suspect list, originally at master `b5db20d` and **re-verified post-#231 at `9afeaa8`** (the rest of that list cleared: `order` quoting is safe via sqlite_orm, OAuth `expires_in`/`expires_at` units are consistent, #135's `createdAt` fix holds, and `body_type` is a documented deliberate write-only mirror). Filed together because both are small `engine/src/db` + route-layer touches; honest severity: rare, low impact, cheap to fix.

## Problem

1. **`Run::end_time` has no default initializer and the design-run insert never sets it.** `engine/include/vayu/types.hpp` declares the field bare; the design-run insert (`engine/src/http/routes/execution.cpp:517` sets `start_time`, nothing sets `end_time`) leaves the value indeterminate, while the load-run insert seeds it (`execution.cpp:715-716`, `run.end_time = run.start_time`). Normal completion is fine - `update_run_status` stamps `end_time` on every status write (`engine/src/db/database.cpp:804-816`) and all in-process terminal writers go through it (verified across `run_manager.cpp`, `execution.cpp`, `runs.cpp`). The gap is the **crash path**: `reconcile_orphaned_runs` (`database.cpp:977-1001`) marks orphaned runs Failed while deliberately leaving `end_time` "as recorded", and its comment (:986-989) claims "create_run already seeds end_time = start_time" - true only for the load path. A design run orphaned by a daemon crash therefore ends Failed with garbage `end_time`; the reader guard (`runs.cpp`, `end_time > 0 ? end_time : now_ms()`) rescues it only if the garbage happens to be non-positive.
2. **`environments.is_active` is a dead column.** Written on create and update (`engine/src/http/routes/environments.cpp:49`), stored (`database.cpp:256`), serialized (`json.cpp:306`) - zero logic readers engine-side (verified: exactly 3 hits, all plumbing). Active-environment tracking is app-side (`session-store.ts`), and the app's only read (`VariableTableEditor.tsx:287`) just echoes the stored value back on save; the actual "is active" check uses the session store. The column is a false affordance: a client could set `is_active` on two environments and nothing anywhere would notice or care.

## Fix / solution pathway

1. Give `Run::end_time` a default (`int64_t end_time = 0;` in `types.hpp`) - the zero then flows through the existing `> 0` guards - and/or seed `run.end_time = run.start_time` in the design insert to match the load path. Fix the `reconcile_orphaned_runs` comment (:986-989) to state the real invariant. Prefer the field default: it closes every future insert site, not just this one.
2. Decide `is_active`'s fate: dropping the column is a schema migration (check `docs/engine/db-schema.md` and sqlite_orm's sync behaviour); the cheaper honest option is documenting it as accepted-but-unused in `db-schema.md` and having `apply_environment_fields` keep accepting it for wire compatibility. Either is fine; silent false affordance is not.

## App changes

None required for (1) - verified, the app reads `endTime` through the already-guarded serializer path. For (2), if the column is dropped: remove the echo write in `VariableTableEditor.tsx:287` and the field from the environment transformer/domain type (grep `isActive` app-side at pickup); if documented instead, no app change.

## Tests

- Engine gtest: insert a design run, simulate the orphan path (`reconcile_orphaned_runs`), assert the reconciled row's `end_time` is sane (equals `start_time` or 0-and-guarded) - mutation-check by reverting the default. `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`.
- App: none (or `cd app && pnpm type-check` if (2) drops the field).

## Docs (same commit)

- `docs/engine/db-schema.md` - the runs-table `end_time` semantics, and `is_active`'s status whichever way (2) goes.

## Sequencing

- **Unblocked** - PR #231 merged; anchors above are refreshed against merged master `9afeaa8`. Independent of the #235 state-layer series (engine-side, disjoint files). Parallel-safe with everything else open.

## Acceptance criteria

- No insert path can produce a run row with indeterminate `end_time`; the reconcile comment states the true invariant.
- `is_active` is either gone (with migration + app cleanup) or documented as accepted-but-unused; it no longer poses as functional.

---

athrvk:
## Description

Both survivors from #242's dead-field re-verification, confirmed still live at `469b90c` before any edit. The `is_active` half was rewritten after review feedback - see the second plan paragraph.

**Plan 1 - `end_time` (root cause, approach, rejected alternative).** The design-mode run insert (`execution.cpp`) stamped `start_time` and nothing else, and `db::Run::end_time` had no default initializer, so the row reached SQLite carrying whatever was on the stack. Normal completion masked it - `update_run_status` stamps `end_time` on every terminal write - but the crash path does not: `reconcile_orphaned_runs` marks an orphaned run Failed and deliberately leaves `end_time` *as recorded*, so the garbage survived into `GET /runs/:runId/report`, where the reader's `end_time > 0 ? end_time : now_ms()` guard rescued it only when the garbage happened to be non-positive. The fix is the issue's preferred pathway plus its alternative, because they cover different holes: the field now defaults to `0` (the "no end recorded" sentinel every reader already guards on), which closes *every* future insert site, and both route inserts seed `end_time = start_time` through a single `seed_run_times` helper, which is what makes a crash-orphaned run report a zero duration instead of one spanning the daemon's downtime. I rejected leaving the design insert unseeded (default `0` only): a design run orphaned by a crash would then hit the `now_ms()` fallback and report a duration stretching to whenever the report is next read - precisely the invented duration `reconcile_orphaned_runs`' comment says it refuses to create. That comment claimed "create_run already seeds end_time = start_time", which was true only of the load path; it now states the real invariant.

**Plan 2 - `is_active` (rewritten after review).** The first version documented the column as accepted-but-unused, which the issue permitted. Per review that was the wrong call, so this now makes the column mean what it says. Two things were actually wrong: nothing stopped two environments carrying the flag, and the choice that drove variable resolution lived only in `localStorage`, so it was per-machine, per-profile, and invisible to the engine storing a column for it. **The invariant is enforced in the DB layer, not the routes**: every write path that can store an active environment calls `deactivate_other_environments_locked` inside the same transaction, so activating one deactivates the previous one atomically - no reader observes two actives, and none observes an intermediate zero. Three paths reach this table (`POST`, `PUT`, and `POST /import/apply`), and a rule living in the handlers would have to be repeated in each - which is exactly how this column previously ended up honoured on create but not on update. Writing `isActive: true` is the whole switch: one PUT, no companion request, and clearing is `isActive: false` on the row that holds it. I rejected adding a dedicated `PUT /environments/:id/active` endpoint - the existing verb already carries the field, and a second spelling of the same write is more surface to keep consistent, not less.

App-side, the session store becomes a *cache* of engine state rather than a second source of truth. `useSetActiveEnvironmentMutation` is the only writer: it PUTs, updates the store optimistically, and rolls back if the engine refuses, because a selection the engine did not accept must not survive in the UI - the next launch would silently disagree with the one before it. It invalidates the environments list rather than patching the response into it, since the *deactivated* row changed server-side without appearing in any response body. `useActiveEnvironmentRestore` reconciles at launch: adopt the engine's active environment, or push a persisted id the engine does not know about (the upgrade path, so nobody loses a selection made before the column was real), once per session. Both directions wait on `isSuccess` - an unreachable engine returns an empty list that is indistinguishable from "no environments exist", and adopting from that would clear a good selection on every cold start. The variables editor also stops echoing `isActive` on save: absent means "keep" on a PUT, and echoing a cached value let a variable edit re-activate an environment from a stale read, deactivating whichever one the engine actually held.

I rejected deriving the active environment purely from the query and deleting the store field. It is the tidiest single-source design, but the resolver and every composed payload read the id synchronously, so a cold start would send requests with no environment until the list loaded - trading a state-sync problem for a correctness one.

**Merged with master (#244, #245, #246, #248, #249, #250).** The only conflict was `App.tsx`, where #246 landed `useActiveEnvironmentGuard` on the same import block and hook list. Both hooks stay and the guard runs first: it answers "does this id still exist", the restore hook answers "which id does the engine hold", so a dangling id is dropped rather than pushed back at the engine as a selection. The rest auto-merged; `docs/app/state-management.md` gained a sentence tying the two paragraphs together now that both describe `activeEnvironmentId`.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Re-run in full after the merge, not just before it.

**Engine** - `python build.py -e -t` then `ctest --preset linux-dev --output-on-failure`: **747/747 passed** (737 on master, 10 added), 117.6s.
- `run_row_seed_test.cpp` (new) - the `end_time` invariant both inserts share, plus the field-default backstop.
- `db_test.cpp` - `ReconcileLeavesNoIndeterminateEndTimeOnAnUnseededRun` drives the real crash path; five environment tests covering activate-switches-the-previous, an inactive save leaving the active one alone, deactivating to none, **surviving a database reopen**, and the bulk-import path (which writes rows directly, so it needs the rule applied there too). The assertions read the *whole table* as a set - a per-row check would have passed on the pre-fix behaviour.
- `resource_write_route_test.cpp` - the same switch over the wire, on both POST and PUT.

**App** - `pnpm test`: **2235/2235 passed** (260 files), `pnpm type-check` clean.
- `active-environment-restore.test.tsx` (new, 7 cases) - the switch reaching the wire as one PUT, clearing via the flag-holder, rollback when the engine refuses, adopt-on-launch, the one-shot upgrade push, no-clear when the engine is unreachable.
- `env-switch-toast.test.tsx` - now async, since the toast reports what the engine accepted; added the refused-switch case (error toast, old selection kept).
- Three editor test files mock `@/queries` by explicit export list and needed the new hook added (the third, `concurrent-edit-clobber.test.tsx`, arrived with #244 in the merge).

**Mutation-checked by reverting each fix and rebuilding**, not mentally: dropping the `end_time` seed turns `SeedRunTimes.SeedsEndTimeAlongsideStartTime` red; dropping the mutation's `onError` rollback turns the rollback case red; dropping the once-per-session ref turns the failure-path case red at **78 calls instead of 1**. That last test is deliberately built around a *deferred* rejection - an immediately-rejected mock is batched into one commit and hides the loop entirely, which is how the first version of the test passed against the broken code.

Docs: `mkdocs build --strict` passes.

## Checklist

Style note: the touched C++ files carry pre-existing clang-format violations in untouched regions; I formatted only the line ranges I edited, so each file's violation count is identical to master's (`database.cpp` 51, `types.hpp` 31, `db_test.cpp` 1). New files are clean. `clang-tidy` could not run in this container - the pinned `.clang-tidy` uses `ExcludeHeaderFilterRegex`, which needs clang-tidy 19+ and only 18.1.3 is installed; CI covers it. The one ESLint error in `VariableTableEditor.tsx` (`Date.now()` purity) is pre-existing and untouched - part of #189.

## Related Issues
Closes #242


athrvk:
Don't patch fix environment.is_active.

It should have a proper fix so that single env can be active at a time, and closing and opening the app sets the previously set value.
Force it on db/engine layer rather than application layer, but it app layer give a easier maintanable solution then go for it

athrvk:
Don't patch fix environment.is_active.

It should have a proper fix so that single env can be active at a time, and closing and opening the app sets the previously set value.
Force it on db/engine layer rather than application layer, but it app layer give a easier maintanable solution then go for it

athrvk:
Agreed - the documentation-only treatment is gone, replaced with a real fix. Pushed in e03e8c9.

**Single active, enforced engine-side.** Every write path that can store an active environment calls `deactivate_other_environments_locked` inside the same transaction, so activating one deactivates the previous one atomically. It lives in the DB layer rather than the handlers because three paths reach that table (`POST`, `PUT`, `POST /import/apply`) - a rule in the handlers would have to be repeated in each, which is how this column previously ended up honoured on create but not on update. Writing `isActive: true` is the entire switch: one PUT, no companion request to clear the old one; clearing is `isActive: false` on the row that holds it. No new endpoint - the existing verb already carries the field.

**Restored on relaunch.** The selection now lives in the database instead of `localStorage`, so it survives a restart and a reinstall, and two clients on the same database agree. The session store stays as a synchronous cache: `useSetActiveEnvironmentMutation` is the only writer (optimistic, rolled back if the engine refuses), and `useActiveEnvironmentRestore` reconciles at launch - adopt the engine's active environment, or push a persisted id the engine does not know about, once per session. That push is the upgrade path, so no existing user loses the environment they had selected before the column was real.

I kept the store rather than deriving the active id purely from the query: the resolver and every composed payload read it synchronously, so a cold start would send requests with no environment until the list loaded. That trades a state-sync problem for a correctness one.

One thing worth your eye: **this overlaps #246**, which owns the app-side active-environment lifecycle right now. I wrote this to be additive - its `useActiveEnvironmentGuard` answers "does this id still exist", the new restore hook answers "which id does the engine hold" - so the two compose rather than duplicate. Whichever merges second rebases; the touch points are `queries/environments.ts` and `docs/app/state-management.md`, in different regions.

Verification: engine 747/747 (10 new, including one that reopens the database to prove the choice persists), app 2144/2144 (7 new), type-check clean, docs strict build clean. Each fix was mutation-checked by reverting it and rebuilding - notably the once-per-session guard, without which the failing-write path makes 78 requests instead of 1.

---

athrvk:
fix merge conflict

Issue #235, opened 2026-07-31
Renderer state-layer hardening (review sweep 2026-07-31)


athrvk:
Tracking issue for the renderer state-layer review sweep of 2026-07-31, run at master `b5db20d` with six parallel readers followed by an adversarial refutation pass on every headline claim.

**Closed 2026-08-01:** all six sub-issues (and the related #227, #228, #242) shipped via PRs #243-#254 and were **re-verified fixed at acceptance-criteria level on master `6939690`** - including the mutation-checked guards each issue demanded (shared save-flusher test, Dock pending/engine-status tests, cycle-guard test, the `state-management-doc.test.ts` doc guard, and the single-source `VARIABLE_PATTERN` test). Two closures went beyond their spec: `environments.is_active` became fully functional (engine-enforced single active environment, restored on launch) rather than documented-as-dead, and the docs rewrite added the identifier-guard test this issue only suggested.

One small follow-up found during closure verification and filed separately: **#256** - `isTabDirty`'s `collection` case misses the `collection-<id>-info/-auth/-script` draft-context keys (guard-consistency only; no reachable data-loss path today).

## Sub-issues (all verified complete)

- [x] #236 - Save pipeline data loss and visibility
- [x] #237 - Per-editor save wiring
- [x] #238 - Run lifecycle state
- [x] #239 - Zombie ids and invalidation gaps
- [x] #240 - Dead state surface and persistence hardening
- [x] #241 - Docs rewrite + doc-guard test

Related, also verified complete: #242 (engine/db: `Run::end_time` defaulted and seeded, `is_active` made real), #227 (resolver hygiene), #228 (introspection now composes via `POST /compose`).

## Found healthy, for the record

Verified clean at sweep time: the quit path flushes end to end; `import.ts`'s complete invalidation; `layout-store`'s versioned persistence; `requestDetailOptions`' 404/transport discipline; the #63 save-model consolidation; `failSave` toast wiring (post-#173); the response-store request-switch race handling. The old dead-fields suspect list mostly cleared: `order` quoting, OAuth expiry units, #135's `createdAt` fix, and `body_type` (documented deliberate mirror) are all fine.

---
Generated by [Claude Code](https://claude.ai/code)

PR #248, opened 2026-08-01
fix(app): surface engineError, evict stale responses, harden store persistence

### LINKED ISSUE #240: app: dead state surface and persistence hardening - engineError shown nowhere, eight dead actions, unbounded response map, unversioned persists
Part of #235 (state-layer sweep). Anchors verified at master `b5db20d`. Honest calibration up front: nothing here is urgent - this is the hygiene tail of the sweep, filed because each item is either the repo's documented written-but-never-read class or a forward hazard that is cheap now and expensive later.

## Problem

1. **`engineError` is written on every health-poll failure and displayed nowhere.** Writer: `queries/health.ts:41,46` (`setEngineError`). Readers of the field: zero outside `engine-store.ts` (verified; the `engineErrorResult` hits in `electron/mcp/tools.ts` are an unrelated helper). The two connection surfaces (`App.tsx:28,65`, `Dock.tsx:124,170-174`) render only `isEngineConnected` - the user sees "Disconnected" with no way to learn ECONNREFUSED vs timeout vs TLS. The defect class CLAUDE.md calls the codebase's most repeated, live in the engine-status pipeline.
2. **Eight dead store actions** (each verified zero callers incl. tests unless noted): engine-store `setPendingRestart` (:52) and `reset` (:66); dashboard-store `getLatestMetrics`/`getMetricsWindow` (:258-273; `dashboard/index.tsx:241-244` re-derives inline instead) and `reset` (:239, whose only trace is a comment saying not to call it); tabs-store `replaceActiveTab` (:155, with a stale doc comment claiming WelcomeScreen uses it - it uses `openTab`, `WelcomeScreen.tsx:81`) and `closeAll` (:178); toast-store `dismissAll` (:160-163). Each is API surface a future caller may trust as exercised.
3. **The response map only grows.** `clearResponse`/`clearAll` (`response-store.ts:79-87`) have zero callers; nothing evicts on tab close, LRU eviction, request delete, or cascade delete, and each entry holds `body` plus `bodyRaw` (roughly double). Memory-only by design (header comment :15-17), so the bound is one session - but a long session over many requests retains every last response, including for deleted requests.
4. **Persistence versioning is inconsistent.** `layout-store` does it right (v3 + real migrate). `tabs-store` and `session-store` are `version: 1` with no `migrate` - a future version bump silently discards users' persisted state. `client-settings-store` has **no version at all** (:164-205) and zustand's shallow top-level merge means a persisted nested object replaces its default wholesale: `loadTestCeilings` and `autoSave.delayMs` are explicitly defended (:199-202, with a comment saying exactly why), but `editor`, `notifications`, and the rest are not - the next key added to `EditorPrefs`/`NotificationPrefs` ships `undefined` to every existing user (e.g. a `notifications` key feeding `slice(-undefined)` disables the toast cap). Not live at this commit; a hazard every settings addition walks past.
5. **Save-store dead fields** `lastSavedAt`/`pendingSaveId` - owned by #236 (listed there because the fix belongs with the pending-status work); noted here only for the inventory.

## Fix / solution pathway

- (1) Surface the reason: Dock tooltip or the disconnected banner shows `engineError` text (one `useEngineStore` read in `Dock.tsx:170-174`); clear it on reconnect (`health.ts` already does).
- (2) Delete the eight actions (and the stale `replaceActiveTab` comment). Any future need re-adds them with a caller and a test.
- (3) Wire eviction: `clearResponse(requestId)` from `useDeleteRequestMutation` and from `closeTabsForEntities`; optionally `clearAll` from Clear-history. Do not persist responses - the by-design comment stands.
- (4) Give `client-settings-store` `version: 1` plus a `merge`/`migrate` that spreads defaults under each nested object (generalize the `loadTestCeilings` defense); add pass-through `migrate` stubs to tabs/session stores so a future bump has somewhere to go instead of silently discarding.

## App changes

App-only. No engine or API change.

## Tests

- `cd app && pnpm test` - `engineError` renders when set (extend the Dock tests; mutation-check by reverting the read); rehydrating a client-settings payload missing a nested key yields defaults, not `undefined` (the `loadTestCeilings` test at `client-settings-store.test.ts` is the pattern - generalize it); deleting a request drops its stored response.
- Dead-action removal needs only type-check + the suite staying green (`pnpm type-check`).

## Docs (same commit)

- `docs/app/state-management.md` - store inventories (coordinate with the docs sub-issue of #235, which is rewriting the file; this issue's edits are the code-side truth it should match).

## Out of scope

- `status:"pending"` visibility and `lastSavedAt`/`pendingSaveId` - #236. `useAddRunToCache`/`useEnvironmentQuery` dead exports - #238 owns the runs one; fold `useEnvironmentQuery` (zero consumers, `environments.ts:34-40`) into this issue's deletion pass.

## Sequencing

- Part of #235. Parallel-safe with every sibling and every other open issue (verified no file overlap; #236 touches save-store/tabs-store in different regions - the tabs `replaceActiveTab`/`closeAll` deletions here vs the eviction block there merge trivially, coordinate).
- Disjoint from PR #231.

## Acceptance criteria

- A disconnected engine shows *why* somewhere reachable without opening devtools.
- Zero exported store actions without production callers (proven by the deletion compiling clean).
- Response entries are dropped when their request or tab goes away.
- Every persisted store either versions its shape or provably survives a missing nested key.

---

athrvk:
## Description

The hygiene tail of the state-layer sweep (#235). Four items, all renderer-side.

**Plan.** The root cause across items 1-3 is the repo's documented written-but-never-read class: a field, an action or a map entry with no consumer on the other side. So the approach is to give each one a reader or delete it, rather than add surface. `engineError` gets a reader (the Dock's connection indicator, in a tooltip); the eight dead actions and `useEnvironmentQuery` are deleted, which makes the type-check the proof that nothing called them; the response map gets its eviction wired at the two seams that know an id is gone for good. Item 4 is a forward hazard rather than a live bug, and is fixed at the persist layer: nested preference objects are completed against their defaults on rehydrate, and every persisted store now carries the `migrate` a version bump needs. **Alternative rejected:** surfacing `engineError` as a toast rather than a tooltip. The health poll fires on an interval, so a down engine would re-raise it on every tick; dedup in `toast-store` collapses the copies but a boot-time engine that is merely slow would still throw a toast the user did not ask for. The Dock is where connection state already lives.

Two anchors had drifted from the issue's snapshot and are noted here:

- The dashboard `reset` comment the issue placed at `dashboard-store.ts:239` actually lives in `services/load-test-service.ts` - a `NOTE: do NOT call store.reset()`. Deleting the action turns that invariant into a compile error, so the comment is rewritten to state it positively and the mock-level regression test is replaced with the positive form (start only opens the stream).
- The issue expected `version: 1` on `client-settings-store` to need a `migrate` so existing users are not reset. Verified against zustand 5.0.14: it only compares versions when the payload *stamped* one (`typeof version === "number"`), and the payloads in users' localStorage carry none, so they pass through either way. The `migrate` is still added, for the same reason as the tabs/session stubs - a *stamped* mismatch is discarded outright - and the tests pin both facts separately.

### 1. `engineError` has a reader

Written on every failed health poll (`queries/health.ts`), read by nothing: "Disconnected" said the same thing for ECONNREFUSED, a timeout and a TLS failure, and the difference only existed in devtools. The Dock's indicator becomes `EngineStatus`; when the engine is down and a reason was recorded, the label is a focusable tooltip trigger carrying that text, with an `Info` glyph so there is something to say it can be hovered. The text wraps rather than truncating - the tail is what identifies the failure.

### 2. Eight dead actions deleted (plus `useEnvironmentQuery`)

engine-store `setPendingRestart` / `reset`; dashboard-store `reset` / `getLatestMetrics` / `getMetricsWindow`; tabs-store `replaceActiveTab` / `closeAll`; toast-store `dismissAll`; and `useEnvironmentQuery`, folded in per the issue. `pnpm type-check` passing is the proof that none had a caller.

### 3. The response map is evicted

`clearResponse(requestId)` now runs from `useDeleteRequestMutation.onSuccess` (the delete is what makes a response unreachable, tab or no tab) and from `tabs-store`'s `closeTabsForEntities` (where the collection cascade already knows every descendant id). Eviction sits *before* that function's "nothing matched" early return, so a deleted request with no open tab is still cleared.

### 4. Persistence versioning

`client-settings-store` gains `version: 1`, a `migrate`, and a `merge` that completes each stored nested object against its defaults - the general form of the `loadTestCeilings` defense that was written out by hand. The live hazard: `notifications.maxVisible` feeds `slice(-maxVisible)` in `toast-store`, and `slice(-undefined)` is `slice(NaN)`, so a payload predating a new key would silently stop capping the toast queue. `tabs-store` and `session-store` gain the pass-through `migrate` a future bump needs (both also refuse a payload of the wrong shape rather than handing a non-array to every reader).

### Deliberate deviation

`response-store.clearAll` is kept rather than wired to Clear-history, which the issue offered as optional. Clearing run history deletes *runs*; the response map is keyed by *request* id, so evicting there would take away responses the user can still see and whose requests still exist. It remains the reset seam three test files use. Flagged rather than silently skipped.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- `cd app && pnpm test` - **248 files, 2154 tests passed** on this branch. 3 test files and 18 cases are new here, and one existing case was rewritten in place rather than added. (I ran the suite on the branch only, not on the base commit, so those are the measured numbers and the baseline is inferred from them.)
- `cd app && pnpm type-check` - clean. This is what proves item 2: the deletions compile.
- `npx prettier --check` and `npx eslint` on every touched source file - clean. Docs were already not prettier-clean on master and are left unformatted per CLAUDE.md.
- Engine untouched, so no engine build or ctest run: the issue is app-only ("No engine or API change").

New/changed tests, each mutation-checked (fix reverted, failure confirmed, restored):

| Test | Guards | Reverting the fix |
|---|---|---|
| `Dock.engine-status.test.tsx` (4, new) | the reason renders, is keyboard-reachable, is not truncated, and does not follow the engine back up | 2 fail - no trigger to focus |
| `client-settings-store.test.ts` (+4) | unversioned payloads survive the bump; a *stamped* older version is translated; a nested object is completed; **every** object-valued pref is completed (enumerated, so a fifth one that nobody wired fails here) | dropping `migrate` fails 1; dropping `merge` fails 2 |
| `persist-migrations.test.ts` (4, new) | tabs and session survive a stamped version bump; a wrong-shaped payload degrades to defaults | dropping either `migrate` fails 1 each |
| `tabs-store.test.ts` (+4) | single delete, collection cascade, delete with no open tab, and the empty-input no-op | all but the last fail |
| `collections.delete-response.test.tsx` (2, new) | the mutation drops the response; a *failed* delete keeps it | 1 fails |
| `load-test-service.test.ts` | replaces the `reset` mock assertion with the positive form | n/a - the mistake is now a compile error |

## Checklist

## Related Issues

Closes #240

Docs updated in the same commit per CLAUDE.md: `docs/app/state-management.md` (tabs / session / engine / response / dashboard / toast / client-settings inventories, eviction and persistence notes) and `docs/app/COMPONENTS.md` (the Dock's ambient status row).

### DIFF
--- a/app/src/components/layout/Dock.engine-status.test.tsx
+++ b/app/src/components/layout/Dock.engine-status.test.tsx
@@ -1,0 +1,96 @@
+/**
+ * @vitest-environment jsdom
+ */
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * A disconnected engine has to say why.
+ *
+ * `queries/health.ts` writes `engineError` on every failed poll - ECONNREFUSED,
+ * a timeout, a TLS failure - and nothing in the app read the field. The two
+ * connection surfaces render `isEngineConnected` alone, so all three causes
+ * printed the same word and the difference was only visible in devtools. That
+ * is the write-only defect class CLAUDE.md calls this codebase's most repeated.
+ *
+ * Rendered, not source-scanned: the reason arrives through a store binding, and
+ * a scan cannot see a value that is not a literal in the file (the badge-hover
+ * guard missed both real instances for exactly that reason).
+ */
+
+import { describe, it, expect, beforeEach, vi } from "vitest";
+import { render, screen, cleanup, fireEvent, waitFor } from "@testing-library/react";
+import { TooltipProvider } from "@/components/ui";
+import { Dock } from "./Dock";
+import { useEngineStore } from "@/stores";
+
+// The Dock prints the app version, which Vite `define`s at build time; vitest
+// does not, so without this the component throws before it renders anything.
+vi.stubGlobal("__VAYU_VERSION__", "0.0.0-test");
+
+const renderDock = () =>
+	render(
+		<TooltipProvider>
+			<Dock />
+		</TooltipProvider>
+	);
+
+beforeEach(() => {
+	cleanup();
+	useEngineStore.setState({ isEngineConnected: false, engineError: null });
+});
+
+describe("the engine status indicator", () => {
+	it("says Disconnected with no affordance when no reason was recorded", () => {
+		renderDock();
+		expect(screen.getByText("Disconnected")).toBeTruthy();
+		// Nothing to hover for, so nothing pretends there is.
+		expect(document.querySelector("[tabindex='0'] .lucide-info")).toBeNull();
+	});
+
+	it("carries the health-poll error, reachable by keyboard", async () => {
+		useEngineStore.setState({
+			isEngineConnected: false,
+			engineError: "fetch failed: ECONNREFUSED 127.0.0.1:9876",
+		});
+		renderDock();
+
+		const trigger = screen.getByText("Disconnected").closest("[tabindex='0']");
+		expect(trigger).toBeTruthy();
+
+		fireEvent.focus(trigger!);
+		// Radix renders the content twice - the visible bubble and a
+		// visually-hidden copy carrying `role="tooltip"` for screen readers - so
+		// this counts "at least one", not "exactly one".
+		await waitFor(() => {
+			expect(
+				screen.getAllByText("fetch failed: ECONNREFUSED 127.0.0.1:9876").length
+			).toBeGreaterThan(0);
+		});
+	});
+
+	it("does not truncate the reason the transport produced", async () => {
+		// The full text is what identifies the failure - a port, a path, a TLS
+		// error. The tooltip wraps; nothing here clips it to a fixed length.
+		const long =
+			"request to http://127.0.0.1:9876/health failed, reason: connect ETIMEDOUT after 5000ms";
+		useEngineStore.setState({ isEngineConnected: false, engineError: long });
+		renderDock();
+
+		fireEvent.focus(screen.getByText("Disconnected").closest("[tabindex='0']")!);
+		await waitFor(() => expect(screen.getAllByText(long).length).toBeGreaterThan(0));
+	});
+
+	it("keeps the connected state a plain label", () => {
+		// A stale error must not follow the engine back up: `health.ts` clears it
+		// on success, and the indicator only offers the tooltip while down.
+		useEngineStore.setState({ isEngineConnected: true, engineError: null });
+		renderDock();
+		expect(screen.getByText("Connected")).toBeTruthy();
+		expect(screen.getByText("Connected").closest("[tabindex='0']")).toBeNull();
+	});
+});
--- a/app/src/components/layout/Dock.tsx
+++ b/app/src/components/layout/Dock.tsx
@@ -5,7 +5,7 @@
  * LICENSE file in the "app" directory of this source tree.
  */
 
-import { FolderOpen, Clock, Braces, PanelRight, Settings } from "lucide-react";
+import { FolderOpen, Clock, Braces, Info, PanelRight, Settings } from "lucide-react";
 import { cn } from "@/lib/utils";
 import { formatChord } from "@/lib/platform";
 import { useLayoutStore, useEngineStore, useSaveStore, type DrawerView } from "@/stores";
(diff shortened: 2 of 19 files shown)


athrvk:
- fix merge conlicts
- Settings page has a pending restart capability where when a setting is change which needs restart a banner is shown, make that is not affected

athrvk:
Both done, pushed as `17380c0`.

**1. Merge conflicts.** Merged master (six commits, #244-#246). Three conflicts, all where this branch and master touched the same seam:

- **`session-store.ts`** - the real collision. #246 removed `activeCollectionId` and landed a genuine v1 -> v2 migration; this branch had added a pass-through migrate for the same store. I kept master's semantics and folded the shape check into it: the migration now rebuilds the payload from the fields v2 knows rather than `delete`-ing the one key, so a stale `activeCollectionId` is dropped at every version (not only below 2), a future field reusing the name cannot inherit it, and a non-string id degrades to `null` instead of reaching the environment switcher and every `/compose` payload. Master's `session-store.test.ts` passes unchanged against it.
- **`load-test-service.test.ts`** - kept master's new report-race guard *and* this branch's replacement for the `store.reset()` assertion (the store has no `reset` any more, so that mistake is a compile error rather than a mock check).
- **`docs/app/state-management.md`** - kept master's v2 session narrative and put the migration/shape-check note under it.

One semantic conflict the merge could not flag: master gave `closeTabsForEntities` a `type` filter, and Clear-history plus the run-delete path now call it with **run** ids. Response eviction is now scoped to calls that can name a request (`type === undefined || type === "request"`) - a run id cannot key a response, so walking them would only look like it meant something. New test covers it.

**2. The restart banner is unaffected - and now has a test.** You were right to ask, because nothing was proving it. The deleted action was `setPendingRestart`, which had no callers; the banner's actual path is `addRestartRequiredKey` (it latches `pendingRestart` itself) -> banner -> `clearRestartRequired`, and neither of those was touched. But every existing `SettingsMain` test mocks `@/stores` wholesale and hands the component a frozen `pendingRestart: false`, so the state machine behind the banner had no coverage at all - a deletion one action over would have been caught by nobody.

`SettingsMain.restart-banner.test.tsx` uses the **real** engine store and the real save path: change a setting whose label carries `(Requires Restart)`, save, and the banner appears naming it with the parenthetical stripped; Dismiss puts it away and clears the keys; a key saved twice is not listed twice. Mutation-checked by stopping `addRestartRequiredKey` from latching the flag - three of the four cases fail.

**Verification on the merged tree:** `pnpm test` **259 files, 2216 tests passed**; `pnpm type-check` clean; prettier and eslint clean on every touched file. CI is running.

---

Issue #226, opened 2026-07-31
engine,app: consolidate `{{variable}}` interpolation and inherit-auth resolution into the engine (backlog A1)


athrvk:
> Verified against master `3baf220` (2026-07-31) by a four-agent sweep over the renderer copy, the MCP copy, the engine side, and the docs anchors. Every factual claim below was confirmed in code or corrected in place; corrections from that pass are folded in and marked *(corrected)* where the original text was wrong. Drift already found between the two client copies is filed as #227 (since slimmed and sequenced after this issue - its semantic items are decided here instead).
>
> **Owner prioritized this issue on 2026-07-31.** The "do not start without explicit ask" gate is satisfied; see Sequencing.

## Problem

Preparing a request before it executes is **duplicated across every engine client**, because the engine composes a request only partway. `docs/plans/pending-backlog.md` → **A1** has tracked this with no issue number (the file has no per-item dates; its header says last updated 2026-07-20); this is that issue.

Two steps remain client-side:

- **`{{variable}}` interpolation** into URL / headers / body / form fields / auth config.
- **`inherit` auth resolution** by walking the collection ancestor chain.

There is no `{{}}` handling anywhere in `engine/` - *(corrected)* verified: every `rg '\{\{'` hit is prose or a fixture, zero executable interpolation. The hits are Monaco autocomplete doc-strings in `routes/scripting.cpp` (:261, :487, :514), C++ comments in `script_engine.cpp` (:2715, :2823), and test comments/fixtures - not only `routes/scripting.cpp` as originally stated.

### The duplication, precisely

| Copy | Files |
|------|-------|
| **Renderer** | `hooks/useVariableResolver.ts` (the map + `resolveString`), `modules/request-builder/utils/auth-resolution.ts` (the inherit walk), and the substitution itself inlined at the send sites |
| **MCP** | `app/electron/mcp/resolve.ts` - a faithful main-process port (`buildVariableMap`, `makeResolver`, `buildCollectionChain`, `composeAuth`), because it shares no module graph with `app/src/` |

[...]

PR #251, opened 2026-08-01
feat(app): complete variable names inside pm.* accessors in script editors

athrvk:
## Description

The script editors could already complete the dotted `pm.*` surface (engine-served, `useScriptCompletionProvider`), but never the one thing only the workspace knows: **the variable names**. Those were visible in the panel *above* the editor, which you had to read off and retype into `pm.environment.get("...")` by hand. A misremembered name is silent - the accessor returns `undefined` and the script fails somewhere else entirely.

The `{{name}}` list stays deliberately unregistered for `javascript` (braces are not the syntax in a script). Instead the names are offered where a script actually writes them: the **string argument of an accessor** - `pm.environment.get` / `pm.globals.get` / `pm.collectionVariables.get` / `pm.variables.get`, plus the `set` / `has` / `unset` spellings.

Two rules keep the offered set equal to what the call can really read:

- **The accessor picks the scope.** `pm.environment.get` lists environment variables only; a global offered there would be a name that resolves to `undefined`. Only the merged `pm.variables.get` lists across scopes.
- **Generators belong to `replaceIn` alone**, which interpolates its argument. `pm.variables.replaceIn` gets brace-style completion including `{{$guid}}`; `pm.variables.get("$guid")` is not a lookup that resolves, so no generator is offered there.

The `pm.*` list now **yields inside a string literal**. Its chain regex matched the word under the caret wherever it sat, so `pm.environment.get("ba` offered the whole dotted surface, and accepting an item replaced the half-typed variable name with `pm.response.body`. Both providers register against `javascript`, so who yields where is a property of the pair.

### Known scope: environment + global, not collection

Collection scope on the resolver is **explicit-only**, and a Monaco completion provider is registered once per *language* rather than per editor - so it has no request to take a `collectionId` from. The live list is therefore environment + global. The `{{name}}` provider beside it is scoped the same way for the same reason, so whatever gives one of them a request context gives both. That is also the point at which decision D2 starts to matter here (the script's collection scope is the immediate parent, while the resolver merges the chain); the hook and `docs/app/variable-resolution.md` both record that as the thing to re-derive.

An earlier revision of this PR carried a D2 ancestor filter keyed on `useSessionStore().activeCollectionId`. That field was deleted on master in the `vayu.session` v2 migration - it had a reader and never a writer - and with no collection variables in the map there was nothing for the rule to narrow, so it came out rather than being repointed.

### Docs correction

`docs/engine/scripting.md` listed optional chaining and nullish coalescing as unavailable "ES2021+" features. Both are **ES2020 and both work**, as do `Array.prototype.at`, `Object.hasOwn` and `String.prototype.replaceAll` - verified by linking a probe against the vendored quickjs-ng build in this repo:

```
OK    ({a:1})?.a                       -> 1
OK    null ?? 'dflt'                   -> dflt
OK    [1,2,3].at(-1)                   -> 3
OK    Object.hasOwn({a:1},'a')         -> true
OK    'abc'.replaceAll('b','X')        -> aXc
FAIL  structuredClone === undefined    -> ReferenceError
OK    String(globalThis.setTimeout)    -> undefined
```

What a script lacks is **host globals**, not modern syntax. The section is rewritten to say that, so workarounds written around the old claim can go.

## Type of Change
- [x] New feature
- [x] Bug fix (the `pm.*` list appearing inside string arguments)
- [x] Documentation update

## Testing

- `app/src/lib/script-variable-completion.test.ts` - the pure rule (scope per accessor, string-literal scanning incl. escapes, first-argument-only, `replaceIn` brace mode).
- `app/src/hooks/script-completion-split.test.tsx` - the wiring, driving both providers the way Monaco does (build the provider, call `provideCompletionItems` with a model stub) rather than scanning source: the scope filter and the string guard are both invisible to a grep.
- **Mutation-checked** (revert the guard, confirm red, restore) on all three surviving guards - the scope filter, the string-literal suppression, and generators-only-in-`replaceIn`. Each produces failures; restoring returns green.
- Full app suite after rebase: **261 files, 2255 tests passing**. `pnpm type-check` clean. `eslint` reports zero problems in the touched files. `mkdocs build --strict` passes.

## Checklist
- [x] I have added tests

## Related Issues

Relates to #226 (decision D2 is referenced but not yet reachable, per the scope note above).

## Follow-up worth considering, deliberately not in this PR

Full "LSP" in the script editors - Monaco's in-browser TypeScript service via `addExtraLib` + diagnostics - would add hover docs, signature help, and squiggles for typos like `pm.response.staus`. It is feasible (the TS worker is already wired in `monaco-setup.ts` and unused for this). The reason to hold: the `pm.*` surface is **already owned by the engine** as a single source of truth (`engine/src/http/routes/scripting.cpp`, ~1000 lines of hand-maintained completions). A hand-written `pm.d.ts` in the app would be a second copy that drifts from it - the exact defect class `CLAUDE.md` warns about. The right shape is to *generate* the `.d.ts` from that table (or have the engine serve it), which is its own piece of work. Note also that the compiler `target` must be ES2022-ish, not ES2020: targeting ES2020 on the old docs' advice would have flagged `Object.hasOwn` and `.at()` as errors when they actually run fine.

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -22,6 +22,7 @@ import { useActiveEnvironmentGuard } from "./hooks/useActiveEnvironmentGuard";
 import { useAppearance } from "./hooks/useAppearance";
 import { useScriptCompletionProvider } from "./hooks/useScriptCompletionProvider";
 import { useVariableCompletionProvider } from "./hooks/useVariableCompletionProvider";
+import { useScriptVariableCompletionProvider } from "./hooks/useScriptVariableCompletionProvider";
 import { useMenuActions } from "./hooks/useMenuActions";
 import { useSaveStore } from "./stores/save-store";
 
@@ -54,6 +55,10 @@ function App() {
 	// `{{variable}}` completion in the body editors. Global per language, so one
 	// registration covers every editor instance - same as the line above.
 	useVariableCompletionProvider();
+	// Variable names inside `pm.environment.get("…")` and its siblings - the
+	// script editors' equivalent of the line above, since a script names
+	// variables through the accessors rather than through braces.
+	useScriptVariableCompletionProvider();
 
 	// Bridge native menu items (Preferences…/Settings) to in-app navigation
 	useMenuActions();
--- a/app/src/hooks/script-completion-split.test.tsx
+++ b/app/src/hooks/script-completion-split.test.tsx
@@ -1,0 +1,212 @@
+/**
+ * @vitest-environment jsdom
+ */
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * The two script-editor completion lists, and the line they divide.
+ *
+ * `useScriptCompletionProvider` owns the dotted `pm.*` surface; this hook's
+ * sibling owns variable names. Both register against `javascript`, so Monaco
+ * merges whatever they return - which makes "who yields where" a property of
+ * the pair, not of either file, and the reason they are tested together.
+ *
+ * Driven the way Monaco drives them (build the provider, call
+ * `provideCompletionItems` with a model stub) rather than by scanning source:
+ * the scope filter and the string-literal guard are both invisible to a grep.
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { renderHook } from "@testing-library/react";
+
+interface Suggestion {
+	label: string;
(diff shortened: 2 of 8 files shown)


athrvk:
Closing in favour of #252, which is these exact commits (`c6c0d6a`) on a branch named for what they do: `claude/script-variable-completion`.

This branch was named `claude/console-not-defined-k4rkl6` after an unrelated bug report that the work then moved away from. GitHub can't repoint an open PR's head branch, and the rename API wasn't reachable from the session, so the only way to correct the name was a new PR. No content difference between the two.


---

PR #252, opened 2026-08-01
feat(engine,app): a language service for the script editors, scoped variable completion, and console levels

athrvk:
## Description

Supersedes #251, which carried the first commit on a branch misnamed after an unrelated bug report (`claude/console-not-defined-k4rkl6`).

The branch started as one completion fix and grew into the editor experience around scripts: the engine now serves its `pm.*` surface as **types** as well as completions, which turns Monaco's TypeScript worker on and brings diagnostics, quick fixes, rename and go-to-definition with it. Alongside that, the completion lists learned the collection they are in, and the Console tab learned what level a line was logged at.

The unifying constraint throughout: **the engine owns the script surface, so the app never re-declares it.**

---

### 1. Variable names inside `pm.*` accessors

The editors could complete the dotted `pm.*` surface but never the one thing only the workspace knows: the variable names. Those were visible in the panel *above* the editor, to be read off and retyped into `pm.environment.get("...")` by hand - and a misremembered name is silent, since the accessor returns `undefined` and the script fails somewhere else entirely.

The `{{name}}` list stays deliberately unregistered for `javascript` (braces are not the syntax in a script). The names are offered where a script actually writes them: the **string argument of an accessor**, plus the `set` / `has` / `unset` spellings.

- **The accessor picks the scope.** `pm.environment.get` lists environment variables only; a global offered there would be a name that resolves to `undefined`. Only the merged `pm.variables.get` lists across scopes.
- **Generators belong to `replaceIn` alone**, which interpolates its argument. `pm.variables.get("$guid")` is not a lookup that resolves, so no generator is offered there.

The `pm.*` list now **yields inside a string literal**. Its chain regex matched the word under the caret wherever it sat, so `pm.environment.get("ba` offered the whole dotted surface and accepting an item replaced the half-typed name with `pm.response.body`.

**Both lists are now scoped to the active collection.** A collection's own variables previously never reached either editor: both providers called `useVariableResolver()` with no options, and collection scope is explicit-only. `useActiveCollectionId` derives that context from the active tab - the same source `ContextBar` already reads - rather than the session-store field earlier commits removed, which never had a writer. The `{{name}}` list offers globals + the whole ancestor chain + the active environment, matching compose-time resolution; the script list narrows back to the immediate parent, which is **decision D2 becoming reachable exactly as the docs predicted** - the engine fills a script's single collection scope from the immediate parent, so offering an ancestor's variable would offer a name that returns `undefined`.

**Accepting a completion no longer duplicates an existing call.** At `pm.variables.rep|("$guid")`, accepting `replaceIn` left `pm.variables.replaceIn("template")("$guid")`. The replace range is not the cause - it ends at the cursor - the snippet's own parens are. When the text after the cursor already opens a call, only the callee is inserted, the same call VS Code makes for `javascript.suggest.completeFunctionCalls`.

### 2. `GET /scripting/types` - the surface as TypeScript declarations

A completion list can only fill a dropdown. Hover over an existing call, signature help while typing arguments, and diagnostics all come from Monaco's TypeScript worker, which wants a `.d.ts`.

Hand-writing `pm.d.ts` in the app would have been a second declaration of a surface the engine owns, drifting the first time a method landed in one and not the other. So the endpoint **generates** it from the same table `/scripting/completions` serves. That works because a completion entry already carries its type: a function's `detail` is its signature, a field's `detail` is its type - all 74 functions parse.

Two things the table genuinely cannot express, both named in `script_types.cpp` rather than guessed, and both guarded:

- **Chain vocabulary.** A getter that *continues* a chain (`.to.not`, `.and`) and one that *performs* an assertion (`.to.be.true`) are identical as entries - a non-function whose `detail` restates its own name. A test asserts each name is still in the table, so a rename fails loudly instead of degrading that member to `void`.
- **Unparseable parameter lists.** Two entries document an overload in prose (`upsert({ key, value }) | (name, value)`); those fall back to `(...args: any[])` rather than emitting a file that will not compile and taking every other declaration with it.

### 3. Diagnostics, quick fixes, rename, references

Semantic validation is on, which is what squiggles `pm.response.staus` with a did-you-mean - and, through the same analysis, what makes quick fixes, rename, find-references, go-to-definition and highlights work. Those providers all default to `true` in Monaco and produce nothing without it; they are set explicitly so a changed default cannot remove them silently.

Validation was off because a *correct* script here produced errors. Rather than guess which, a realistic fragment was compiled with `tsc --checkJs`. Exactly two, both because the editor holds a fragment while the engine runs something larger:

| Code | Message | Why it is wrong here |
|------|---------|----------------------|
| `1108` | `return` outside a function body | The engine wraps every script in an IIFE |
| `2304` | Cannot find name | A collection-level part is joined to the request's before the engine runs it |

Suppressing `2304` would normally cost the best diagnostic of all - `fetch` and `setTimeout`, which the sandbox lacks. It does not, because the engine **declares the globals it lacks** as `never` with the reason as documentation: calling one is *not callable* rather than *cannot find name*, and hover explains why.

A hand-written list of what a runtime *doesn't* have rots silently, so it isn't trusted: the test runs a `typeof` probe in the real script engine for every entry, reading the names back out of the generated file rather than restating them. **It caught `queueMicrotask` on its first run** - quickjs-ng does provide it - and that entry is gone.

Worth stating plainly: this is Monaco's TypeScript service fed engine-generated declarations, **not** a Language Server Protocol implementation. There is no server process and nothing another editor could connect to.

### 4. Console levels reach the Console tab

`console.log`, `.info`, `.warn` and `.error` were indistinguishable in the response pane, and no styling could have fixed it: all four were bound to one C function that pushed a bare string, so the level was destroyed at the QuickJS binding.

The engine binds the four through `JS_NewCFunctionMagic` with the level as the magic argument, and `consoleLogs` becomes `{source, level, message}`. `source` as a field retires the `"[pre] "` text prefix, which was indistinguishable from a script logging those six characters itself. A bare string is still decoded the way the engine that sent it meant, so a renderer talking to an older sidecar keeps working.

`warn` and `error` take `--status-warning-text` / `--status-error-text` - a line's level is state, unlike the section colour above it, which is identity. `log` and `info` share the plain foreground, as Chrome does, and a gutter label keeps this from being a colour-only distinction.

### 5. Tab primitive fixes (`components/ui/tabs.tsx`)

Reported against Collection Detail, but reaching every tab strip. Radix's `forceMount` means "always present", not "present but hidden" - so Collection Detail's four draft-holding panels painted on top of whichever tab was selected. And the focus ring was clipped to two vertical strokes, because a trigger fills its list's height exactly while the three scrolling strips are `overflow-y-hidden`; `focus-visible:ring-inset` fixes every strip at once.

### Docs correction

`docs/engine/scripting.md` listed optional chaining and nullish coalescing as unavailable "ES2021+" features. Both are **ES2020 and both work**, as do `Array.prototype.at`, `Object.hasOwn` and `String.prototype.replaceAll` - verified against the vendored quickjs-ng build. What a script lacks is **host globals**, not modern syntax; the editor now enforces exactly that (`lib: ["es2022"]`, no `dom`, `target: ESNext`).

## Type of Change
- [x] New feature
- [x] Bug fix (string-literal completion, call duplication, collection scope, forceMount panels, clipped ring)
- [x] Documentation update

## Testing

**Engine - 760/760 passing.** `script_types_test.cpp` guards the derivation: chain vocabulary still in the table, every listed member reaching the output, the signature shapes actually used (including the two that cannot parse), JSDoc that cannot close early, and the absent-globals list probed against the real script engine.

**App - 2299/2299 passing across 267 files.** Providers are driven the way Monaco does (build the provider, call `provideCompletionItems` with a model stub) rather than scanned as source - the scope filter, the string guard and the callee-only rule are all invisible to a grep.

**Mutation-checked** (revert, confirm red, restore) on the guards in the work I authored - the scope filter, string-literal suppression, generators-only-in-`replaceIn`, the diagnostic suppression list and the semantic-validation flag. Each produces failures; restoring returns green.

**The generated declarations were compiled with the app's own `tsc`** under `--strict` with no DOM or Node types - Monaco's configuration:

```
correct fragment   raw: 2   after suppression: 0
mistake file       raw: 13  still reported:    9
  'staus' -> Did you mean 'status'?     'eqaul' -> Did you mean 'eql'?
  replaceIn(42) - number not assignable to string
  setTimeout / fetch / require - not callable;  new URL(...) - not constructable
```

`pnpm type-check` clean, eslint and prettier clean on touched files, `mkdocs build --strict` passes. CI green on all three platforms, including Engine and C++ Tests on Windows and macOS - the first exercise of the new C++ outside Linux GCC.

## Checklist
- [x] I have added tests

## Related Issues

Relates to #226 - decision D2 is no longer just referenced; the script list now implements it.

## Known limitations

- **`pm.response.json()` is `any`**, so everything downstream of a response body is untyped. Typing it from an observed response is the largest remaining win.
- **`2304` is suppressed**, so a genuine typo in a *local* variable name goes unreported. That was the price of not false-flagging correct collection-script usage; it can come off once collection script text is threaded into the editor's context.
- **`pm.response.headers` carries an index signature**, so `headers.gett()` is not caught - the cost of keeping the documented `headers['content-type']` legal.
- Diagnostics, quick fixes and rename are verified through the type layer and the provider configuration, **not** by driving the running editor UI.


athrvk:
Pushed `eedd995`, which takes up the follow-up this PR's description deliberately deferred - and takes it up in the shape that description argued for: **generated from the engine's table, not hand-written in the app.**

`GET /scripting/types` returns the `pm.*` surface as TypeScript declarations, derived from the same table `GET /scripting/completions` serves. The derivation works because a completion entry already carries its type - every function's `detail` is its signature, a field's `detail` is its type. All 74 functions parse, so there is no second copy to drift.

Two facts the table genuinely cannot carry are named in `script_types.cpp` rather than guessed, and both are guarded by tests:

- **Chain vocabulary.** A getter that *continues* a chain (`.to.not`, `.and`) and one that *performs* an assertion (`.to.be.true`) are indistinguishable as entries - both are non-functions whose `detail` restates their own name. A test asserts each name is still in the table, so a rename fails loudly instead of quietly degrading that member to `void`.
- **Unparseable parameter lists.** Two entries document an overload in prose (`upsert({ key, value }) | (name, value)`). Those fall back to `(...args: any[])` rather than emitting a file that will not compile and taking every other declaration with it.

App side, the worker is configured for the sandbox rather than a browser: `lib: ["es2022"]` with **no `dom`**, so the editor cannot offer `fetch`/`setTimeout` that do not exist at runtime, and `target: ESNext`, since ES2020 would flag `Object.hasOwn` and `.at()` - the same correction this PR already made to `docs/engine/scripting.md`, now enforced by the editor.

**Semantic diagnostics stay off**, deliberately and with the reason in the hook: a script editor holds a fragment the engine wraps in an IIFE, and a request's two scripts are separate models that cannot see each other's declarations, so validation would squiggle correct scripts. Hover, signature help and completion detail are all produced without diagnostics.

### Verification beyond the unit tests

The generated output was compiled with the app's own `tsc` under `--strict`, no DOM and no Node types - Monaco's configuration:

```
pm.gen.d.ts                     tsc exit 0   (compiles clean)
realistic script using the API  tsc exit 0   (no false errors)

pm.response.staus     TS2551  Did you mean 'status'?
pm.enviroment.get     TS2551  Did you mean 'environment'?
pm.variables.replaceIn(42)  TS2345  number is not assignable to string
setTimeout / fetch    TS2304  Cannot find name
```

One honest gap: `pm.response.headers.gett(...)` is *not* flagged, because `headers` carries an index signature so `headers['content-type']` keeps working. That tradeoff is documented.

757/757 engine tests, 2268/2268 app tests, `pnpm type-check` and `mkdocs build --strict` all pass.


---

athrvk:
Pushed `94a9757`, which turns on the three things the declarations made possible but the previous commit left switched off: **diagnostics, quick fixes, and rename / find-references / go-to-definition.**

The last two turned out not to be separate work. Every provider (`codeActions`, `rename`, `references`, `definitions`, `documentHighlights`) already defaults to `true` in Monaco - they simply produce nothing without semantic analysis. So enabling diagnostics enabled all of it. They are now set explicitly anyway, so a changed default cannot remove them silently.

### Why validation was off, and what actually changed

It was off because a *correct* script here produced errors. Rather than guess which, I compiled a realistic fragment with `tsc --checkJs`. Exactly two, both from the editor holding a fragment while the engine runs something larger:

| Code | Message | Why it is wrong here |
|------|---------|----------------------|
| `1108` | `return` outside a function body | The engine wraps every script in an IIFE, so a top-level `return` is legal |
| `2304` | Cannot find name | A collection-level script part is joined to the request's before the engine runs it, so a name declared there is undeclared to this model |

Suppressing `2304` would normally cost the best diagnostic of all - `fetch` and `setTimeout`, which the sandbox lacks. So the engine now **declares the globals it lacks**, typed `never` with the reason as documentation: calling one is *not callable* rather than *cannot find name*, so it survives the suppression, and hover explains why rather than the editor just shrugging.

A hand-written list of what a runtime *doesn't* have rots silently - nothing fails when the runtime gains one. So it isn't trusted: the test executes `typeof <name>` in the real script engine for every entry, reading the names back out of the generated file rather than restating them. **It caught `queueMicrotask` on its first run** - quickjs-ng does provide it - and that entry is gone.

### End-to-end verification

Against the actual generated output, with the app's own `tsc`:

```
A) correct fragment    raw: 2   after suppression: 0
B) mistake file        raw: 13  still reported:    9
     TS2551  'staus'  -> Did you mean 'status'?
     TS2551  'enviroment' -> Did you mean 'environment'?
     TS2551  'eqaul'  -> Did you mean 'eql'?      (on VayuExpectTo)
     TS2345  replaceIn(42) - number not assignable to string
     TS2349  setTimeout / fetch / require - not callable
     TS2351  new URL(...) - not constructable
```

The chain typo landing on `VayuExpectTo` is worth noting: the generated chain interfaces are carrying real weight, not just decoration.

Both new app guards are mutation-checked - widen the suppression list, flip validation off; each fails, restoring returns green. 758/758 engine, 2270/2270 app, `type-check`, eslint, prettier and `mkdocs --strict` all pass.


---

Issue #227, opened 2026-07-31
app: renderer resolver hygiene left after #231 - missing parentId cycle guard, dead hasUnresolvedVariables export, four copies of the variable regex


athrvk:
> **Rewritten 2026-07-31 after PR #231 merged** (master `9afeaa8`). The original issue listed seven drift items between the renderer and MCP composition copies. #231 deleted the MCP copy, decided the two disputed edge semantics engine-side (D17: absent `enabled` = enabled, non-string `value` = `""`, held by the cross-language conformance fixture), and routed the load-test path through `buildExecBody` + `POST /compose` - mooting items 1, 2, 3 and 5. What remains is the renderer-preview hygiene below, re-verified with fresh anchors at `9afeaa8`. All of it is low severity.

## Problem

1. **The renderer's collection-chain walk has no cycle guard.** `buildCollectionChain` (`app/src/hooks/useVariableResolver.ts:65-73`) follows `parentId` with no `seen` set, so a cyclic `parentId` pair loops forever inside a React memo and hangs the UI. The engine rejects cycles on write (#79; `engine/include/vayu/http/routes.hpp` cycle validation), so this fires only on a corrupted DB - but the failure mode is a hang and the guard is three lines. (The deleted MCP copy had the guard; its deletion removed the only guarded walk.)
2. **`hasUnresolvedVariables` is exported and never read in production.** Still true post-#231: defined and exported (`useVariableResolver.ts:53,247,267`), consumed only by the hook's own tests; unresolved-token painting reads the variable map directly. The repo's most repeated defect class. Delete the export (or wire it to a pre-send unresolved-variable warning); keep the absence-not-empty invariant documented on the map either way.
[...]

PR #253, opened 2026-08-01
docs(app): rewrite the drifted renderer state docs from the code

### LINKED ISSUE #241: docs(app): state-management.md has drifted wholesale - fabricated APIs, wrong query keys, phantom constants (plus architecture.md and api-integration.md spot drift)
Part of #235 (state-layer sweep). Every drift below verified line-by-line at master `b5db20d`. CLAUDE.md's rule - a stale line is worse than a missing one, because these docs are reference material future sessions trust - is the reason this is filed as a bug against the docs, not a nice-to-have.

## Problem

`docs/app/state-management.md` mismatches the code in 30+ places, ten of which would cause a wrong implementation decision outright:

1. **All four documented query-key shapes are wrong** (doc L497-505): `collections.all` is a constant not a function (`queries/keys.ts:18`); `collections.detail` is `['collections','detail',id]` not `['collections',id]` (:21-22); `requests.listByCollection` and `requests.detail` similarly carry the `'list'`/`'detail'` segment the doc omits (:28-32). Hand-writing a key from the doc produces a silent cache miss that presents as a data bug.
2. **Three documented APIs do not exist**: `useSSE()` (L543-560; streaming is a non-React singleton - `services/load-test-service.ts:25,35` → `sse-client.ts`), `useEngine().startLoadTest()` (L530; `useEngine.ts:14-26` has no such method), `useStopRunMutation()` (L481; stopping is `useEngine().stopLoadTest`). Also invented: `useCollectionQuery(id)` (L430).
3. **`HISTORICAL_METRICS_CAP = 3000` is asserted four times across two docs** (state-management L781, L827; architecture L143, L262) **and does not exist** (verified: `config/metrics.ts` exports only `METRICS_UI_THROTTLE_MS`). Actual retention is time-based (`liveWindowSeconds`, default 300s) with a 50,000-tick backstop (`dashboard-store.ts:168-191`, `constants/live-window.ts:113`) - and the doc contradicts its own L228-232, which describes the real mechanism.
4. **The doc teaches a reverted design**: per-view `drawerWidths` + `setDrawerWidth("collections", 300)` (L78, L94) were deliberately collapsed to a single width by the layout-store **v3 migration** (`layout-store.ts:25,93-99`); re-implementing from the doc re-introduces the fixed bug. The doc also says persistence is v1 (L97).
5. **Wrong numbers a tuner would act on**: health poll "5 seconds" (L517, also `api-integration.md` L537) vs actual 30s (`timing.ts:53`); run reports "1 minute" vs 5 (`config/cache.ts:30`); request-detail `staleTime: Infinity` unmentioned (`collections.ts:199`); autosave debounce "3 seconds" (L793) vs the 5s default the same doc states at L703 (`constants/client-settings.ts:77`); "Saved" indicator 2s vs 3 (`timing.ts:34`).
6. **Wrong directory throughout**: `app/src/services/queries/` (L425, L492) - everything lives in `app/src/queries/`.
7. `architecture.md` L144 describes response-store as "viewer state (expand/collapse, scroll position)" - it stores full response payloads keyed by requestId (`response-store.ts:30-61`); a dev could add view state to the wrong store. `api-integration.md` L525-527 claims the SSE client attempts reconnection - it deliberately does not (`sse-client.ts:152-166`).
8. Plus the long cosmetic tail (field renames, missing fields/methods/hooks, the undocumented `config/cache.ts` as the single home of every stale/gc number, `notifications`/`reducedMotion` missing from client-settings, toast severity floor and duration scale undocumented, etc.) - the full verified table is large; the implementer should regenerate sections from code rather than patch line-by-line.

## Fix / solution pathway

Rewrite the drifted sections of `docs/app/state-management.md` **from the code**, section by section, rather than patching claims: stores inventory (all 10 + toast/client-settings in the overview), query keys copied verbatim from `keys.ts`, cache policy quoting `config/cache.ts`/`timing.ts` constants by name instead of inlining numbers, save-model section describing the two documented models + the editor variants, and a streaming section describing `load-test-service`/`sse-client` as the singleton it is. Fix the two `architecture.md` lines and the two `api-integration.md` lines in place. Delete every reference to the four phantom APIs and the phantom constant.

Where a number must appear, prefer naming the constant (`QUERY_CACHE.RUNS_STALE_TIME_MS`) over restating its value - that is what kept `design-system.md` honest once `design-system-doc.test.ts` existed. **Consider a small guard test** in that spirit: assert the doc's named constants exist in code (a `state-management-doc.test.ts` scanning for backticked identifiers), so the next drift fails CI instead of accumulating. Scope it to identifiers, not prose.

## App changes

None - docs only, plus the optional guard test under `app/src/`.

## Tests

- If the guard test is added: `cd app && pnpm test` - and per repo convention it must assert it scanned something non-empty.
- `mkdocs build --strict` must stay green (`pip install -r requirements-docs.txt && mkdocs build -f .github/mkdocs.yml --strict`) - these files are published.

## Docs

This issue *is* the docs change. Note for coordination: #236-#240 each touch `state-management.md` for their own changes - see Sequencing.

## Sequencing

- Part of #235. **Land this rewrite first among the siblings, or explicitly last** - not interleaved. First is better: the siblings then edit accurate text for their deltas. If it lands last instead, regenerate against post-sibling code. Either way the rewrite must be done against whatever master is at pickup, not this issue's quotes.
- No file conflicts with anything else open (docs-only; #112 prunes `docs/engine/scripting.md`, a different file).

## Acceptance criteria

- Every named hook, key shape, constant, and number in the three docs exists in code with the stated value - spot-checkable by grep, ideally enforced by the guard test.
- The four phantom APIs and `HISTORICAL_METRICS_CAP` appear nowhere in `docs/`.
- `mkdocs build --strict` passes.

---

athrvk:
## Description

**Plan.** The root cause is that `docs/app/state-management.md` was written once and edited by hand as the renderer moved: hooks were renamed or deleted (`useSSE`, `useEngine().startLoadTest`, `useStopRunMutation`, `useCollectionQuery`), a constant was removed (`HISTORICAL_METRICS_CAP`), the query-key factory gained `list`/`detail` segments, layout-store's per-view `drawerWidths` was reverted by a v3 migration, and five durations moved. Every claim in the issue was re-verified against master at pickup before editing - all still true, with the line numbers drifted (the api-integration SSE-reconnect claim is now at L557-561, not L525-527). The approach the issue asks for is the one I took: regenerate the drifted sections **from the code** and name constants (`QUERY_CACHE.RUNS_STALE_TIME_MS`) rather than restating their values, since restating values is how the section drifted in the first place. The alternative I rejected was patching the wrong lines one by one - it fixes today's list and leaves the same failure mode, which is why the issue also suggests a guard, and why the guard is here.

Rewritten from source:

- **Query hooks** - the real inventory from `queries/index.ts` (`useMultipleCollectionRequests`, `useCollectionAncestors`, `requestDetailOptions`, `useSetActiveEnvironmentMutation`, `useRunTimeSeriesQuery`, the OAuth trio, `useImportMutation`), the directory corrected to `app/src/queries/`, and the key factory copied from `keys.ts` - a key one segment short is not an error, it is a silent cache miss.
- **Cache policy** as a table naming `QUERY_CACHE` / `TIMING` constants, replacing the five wrong numbers (health 5s -> 30s, reports 1m -> 5m, request detail's `staleTime: Infinity` unmentioned, autosave 3s -> the 5s user setting, "Saved" 2s -> `SAVED_STATUS_DURATION_MS`).
- **`useSSE()` section deleted**, replaced by what exists: `loadTestService` + `sseClient` as module singletons (the stream has to outlive the view), the `METRICS_UI_THROTTLE_MS` commit throttle, and why there is deliberately no reconnect.
- **`useEngine()`** at its real signature - compose / execute / stop; starting a load test is a call site in the request builder, not a hook.
- **Stores** - layout-store at v3 (one `drawerWidth`, `contextBarWidth`, the `settings` view), client-settings' full state incl. `reducedMotion` and `notifications`, the toast preferences (severity floor, duration scale, stack cap), retention as a time window rather than a 3,000-point cap, and the LRU/dirty rule stated as it actually behaves (nothing is flushed during eviction).
- **`architecture.md`** - session/response store descriptions, both data flows, the phantom metrics-cap and 2,000-point-downsampling bullets replaced with the bucketing that is really there.
- **`api-integration.md`** - the hooks box, the SSE step, the health poll interval, and the "reconnection attempted" claim the client's own comment contradicts.
- **`file-name-conventions.md`** - `useSSE.ts` was still an example filename; swapped for a hook that exists.
- **`CLAUDE.md`** - the doc-readers note now lists the second test that reads doc paths.

**Guard test** (`app/src/state-management-doc.test.ts`): every `useXxx` and `SCREAMING_CASE` name these three docs quote **in prose** must exist in the app source. Fenced blocks are blanked rather than stripped so reported line numbers still point at the doc. Two allowlists, both self-checking: `OWNED_ELSEWHERE` (engine-owned names like `DEFAULT_MAX_LIVE_TICKS`) asserts the owning file still declares them, and `REMOVED_ON_PURPOSE` (`STATS_STREAM`) asserts the name is still absent, so re-introducing one fails until it leaves the list. Scoped to identifiers, not prose - a hook existing says nothing about the sentence around it.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- `cd app && pnpm test` - **261 files, 2239 tests passing** (includes the new guard's 4).
- **Mutation-checked**: re-adding `` `useSSE()` ``, `` `HISTORICAL_METRICS_CAP` `` and `` `useStopRunMutation` `` to the prose fails the guard with the right file and line; removing them restores green. Verified the line numbers land on the doc line after the fenced-block handling was fixed (they were offset while blocks were stripped rather than blanked).
- `cd app && pnpm type-check` clean; `eslint` and `prettier --check` clean on the new file (only file touched under `app/`).
- `mkdocs build -f .github/mkdocs.yml --strict` builds clean.
- No engine change, so the engine suite was not run.

## Checklist

## Related Issues

Closes #241

## Notes / deliberate deviations

- The issue's item 8 asks the implementer to "regenerate sections from code rather than patch line-by-line", which is what happened - so this diff rewrites some prose the issue did not enumerate (the toast preference policies, the client-settings state block). Nothing was rewritten that was already accurate.
- Two drifted lines in `architecture.md` outside the issue's list are fixed because they are the same defect on the same screen: `session-store.ts: User session, theme preferences` (it holds the active environment id) and `Metrics Downsampling: Charts downsample beyond 2,000 points` (no such code exists; charts bucket by `chartBucketSeconds`).
- `useCollectionAncestors` walks `parentId` with no cycle guard, the same shape as the `buildCollectionChain` finding in #227 - left alone here, since #227 owns that fix.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JAEHP8f8mfe5otFW4ogZtL)_


athrvk:
master moved, verify and make sure the docs are uptodate

athrvk:
Rebased on master (`aab339a`, v0.14.0) and re-verified the docs against it.

One real drift arrived with the generated `pm.*` type declarations: `useScriptTypeDefinitionsQuery` (`queries/script-types.ts`) and its `scriptTypes` key family. Both are now in the hook inventory and the key-factory excerpt - which is the drift class this PR exists to stop, so it was worth catching before merge rather than after. Nothing else master changed contradicts the rewrite: the `consoleLogs` shape change is documented in `api-integration.md` by that PR itself and does not touch anything here, and the new completion hooks live under `hooks/`, which these docs do not enumerate.

Re-ran on the merged tree: app suite **268 files / 2303 tests passing**, `pnpm type-check` clean, `mkdocs build -f .github/mkdocs.yml --strict` clean, and the guard test still green (it reads the docs from disk, so it re-checked every quoted identifier against post-merge source).

---

PR #254, opened 2026-08-01
fix(app): guard the collection-chain walk, delete a dead export, and land one `{{name}}` matcher

### LINKED ISSUE #227: app: renderer resolver hygiene left after #231 - missing parentId cycle guard, dead hasUnresolvedVariables export, four copies of the variable regex

athrvk:
## Description

**Plan.** The root cause of all three items is the same one: the renderer's preview layer accumulated copies and exports that nothing holds to anything. `buildCollectionChain` walks `parentId` with no `seen` set, so a cyclic pair loops forever *inside a `useMemo`* - a synchronous hang of the render thread, not a wrong preview. `hasUnresolvedVariables` is exported and read only by its own tests, the repo's most repeated defect class. And the `{{name}}` regex is declared once per place that needs it, which is how the preview can drift from what `POST /compose` substitutes without anything failing. The approach the issue asks for is the one taken: guard the walk, delete the dead export, point every copy at `constants/variables.ts`. The alternative I rejected was fixing today's three copies and stopping there - #231 already showed that fails, having deleted one copy and introduced another while the count stayed at four. So the consolidation ships with a guard on the *shape*, and it immediately found copies the issue's grep had not.

Each item re-verified against master `db207f2` before editing; all three still true, line numbers drifted by two in `ScriptTab.tsx`.

### 1. The cycle guard

`buildCollectionChain` keeps a `seen` set and stops on a revisit. The comment says why it is a termination guard and not deduplication - a `seen` set that also skipped repeats would break override order rather than a cycle, so the acyclic case is tested alongside.

**The test is bounded, which took some care.** Without the guard the walk is a synchronous loop, so it never yields and vitest's per-test timeout can never fire - the naive test hangs the whole run rather than failing it. The fixture array therefore counts its own `find` calls (an own property shadowing `Array.prototype.find`) and throws past a budget no correct walk reaches. Mutation-checked: deleting the `seen` set fails two tests in **40ms** with `the parentId walk is not terminating`, and restoring returns green. A fourth test asserts the honest path stays well under the budget, so the cycle tests cannot pass for the wrong reason.

### 2. `hasUnresolvedVariables`

Deleted - interface entry, callback, and return field. Nothing in production called it; the unresolved-token painting (`VariableInput`) looks each name up in the variable map and falls back to the dynamic-variable table itself. The invariant the function depended on is real and is kept: a name whose every definition is disabled is **absent** from the map rather than present-and-empty, and the origins test now asserts that through `getAllVariables`, which is the thing the painting actually reads. Its two dynamic-variable assertions were rewritten against the same surviving contract rather than dropped.

### 3. One `{{name}}` matcher - six copies, not four

The issue counted four because it grepped the `/g` form. The guard found two more that a `/g` grep cannot see: `BodyPanel.tsx` and `ParamsPanel.tsx` each ran `/\{\{[^{}]+\}\}/.test(...)` as a "does this still contain a variable" check (in `ParamsPanel` it decides whether to `encodeURIComponent` a param, so it is on the wire path).

- `lib/variable-resolution.ts`, `ScriptTab.tsx` and `referenced-variables.ts` now import `VARIABLE_PATTERN`. `lib/` importing `@/constants` is not a new coupling: `lib/dynamic-variables.ts`, which `variable-resolution.ts` already imports, has always done it.
- The three `.test()` call sites use a new `containsVariableToken` in the canonical module - unanchored (unlike `isVariableToken`) and deliberately **non-global**, because `.test()` on a shared `/g` regex advances `lastIndex` and answers the next caller from the middle of a string.
- Two test doubles that hand-rolled a `\w+` variant now use the real pattern.

`constants/variable-pattern-single-source.test.ts` fails on a seventh declaration and on any stateful use of the shared object. It asserts it scanned a non-empty file set **and** that it still recognises the canonical declaration, so it cannot pass by failing to see what it was written to see. `services/importers/var-normalize.ts` is exempt with its difference named (identifier-only body, Postman prefixes - strictly narrower), and a test asserts every exemption still points at a file that exists.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- `cd app && pnpm test` - **262 files, 2244 tests passing** (2239 on master + 9 new, less the 4 the deleted export owned).
- **Mutation-checked**, both new guards: deleting the `seen` set fails the cycle tests fast; re-introducing a copy of the pattern in `lib/dynamic-variables.ts` fails the single-source guard with the file and line. Restoring each returns green.
- `cd app && pnpm type-check` clean.
- `eslint` and `prettier --check` clean on the 15 touched files; `useVariableResolver.ts` and the new guard were formatted after confirming they were prettier-clean at HEAD, nothing else was reformatted.
- `mkdocs build -f .github/mkdocs.yml --strict` builds clean.
- No engine change, so the engine suite was not run.

## Checklist

## Related Issues

Closes #227

## Notes / deliberate deviations

- **The issue said "Docs: none required".** Two were: `docs/app/variable-resolution.md` and `docs/app/state-management.md` both name `hasUnresolvedVariables`, and a doc naming a deleted export is exactly the drift #241 is open about. `variable-resolution.md` also gained the cycle-guard sentence and the single-matcher rule, since that page is where a future reader looks before writing the regex again.
- **Two extra copies fixed** (`BodyPanel.tsx`, `ParamsPanel.tsx`) beyond the issue's inventory, because the acceptance criterion is "one declaration, imported everywhere it is used" and they are the same declaration.
- **Left alone, adjacent:** `ScriptTab.tsx` still hand-rolls the whole of `referencedVariables()` - the same `PM_GET` regex and the same dedupe, in a module that has a tested extraction of exactly that logic. Only its `{{name}}` literal is fixed here. Worth an issue; it is not this one's scope.
- `docs/app/state-management.md` is also being rewritten by #253. This PR's edit there is a single line inside the fenced hook signature, so the two should merge cleanly, but #253 is the one to land first if git disagrees.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FpDmhVCWQjoh8MUaggZTyF)_


athrvk:
take care of the scripts tabs as well both pre and post

athrvk:
Pushed `d634439`, which takes up the script tabs and fixes the Windows failure.

### The script tabs, pre and post

`ScriptTab` is one component rendering both kinds, and it re-implemented the whole of `referencedVariables()` - the same `pm.*` regex, the same `{{name}}` regex, the same dedupe. So it was never just a duplicated pattern: the **empty-name filter that lives in the helper never reached either tab**, and a script containing `{{ }}` chipped an empty `{{}}` badge on both.

The helper moved to `app/src/lib/referenced-variables.ts` rather than being imported across module boundaries. It is now read by `collections/` and `request-builder/`, and a shared helper parked under one consumer is exactly what let the other keep its own copy - `lib/` is where the app's other pure text helpers already live (`variable-resolution`, `dynamic-variables`, `variable-cast`). `ScriptPanel` imports it from the new path; nothing else referenced it.

`ScriptTab.chips.test.tsx` drives both kinds through `describe.each(["pre", "post"])` - the duplicate was in the component they share, so a test of `pre` alone could not tell them apart. Mutation-checked: restoring the inline copy fails 4 tests (both kinds, both the ordering assertion and the empty-token one).

One thing worth reporting: the single-source guard flagged **my own new test**, which had used `getAllByText(/^\{\{.+\}\}$/)` to find the chips. That is a copy like any other, so the test uses a text function instead. The guard doing that unprompted is the argument for having it.

### The Windows failure

`globSync` returns the host's separator, so the guard received `constants\variables.ts` on Windows and reported the canonical file as a stray - green on Linux and macOS, red on the third runner. That is the platform-asserting defect CLAUDE.md warns about, in a test whose whole job is to be right everywhere.

Fixed by normalizing on **both** separators rather than on `path.sep`, which removes the host dependence instead of branching on it - and that is what makes it testable. A new case feeds the guard the Windows form from any runner; reverting the normalization to `path.sep` fails it on Linux, which is how it was verified without a Windows host.

The two failures you saw on `b38781d` are the merge commit, which predates this fix.

### Verification

Rebased onto the merged master (so this now includes #252) and re-run there: **270 files, 2315 tests passing**, `pnpm type-check`, eslint and prettier clean on the touched files. The guard passes against the new `script-variable-completion` code that arrived with #252.

---

Issue #174, opened 2026-07-28
engine,app: store response headers and body for sampled load-run requests


athrvk:
Load runs record a sampled request's status, latency and timing breakdown, but discard the response headers and body. The Samples tab therefore cannot ever show what the server actually returned, and two UI surfaces are already written to display that data and are permanently blank as a result - which is how this was found.

## Current behaviour

Verified against `61a3b3e` on a live daemon: 150 requests at 50 RPS, `save_timing_breakdown: true`, `success_sample_rate: 10`, and a `pm.test` script so the response-sampling path was active too. 15 rows were persisted, and the union of `trace` keys across every one of them was:

```
connectMs, dnsMs, downloadMs, firstByteMs, queueWaitMs, tlsMs, totalMs, wireMs
```

No `headers`, no `body`, no `request`, no `response`. At the DB level, of 16 `results` rows in that database exactly one mentioned headers - a design run fired at the same target for contrast:

```
15  load    run_726faf31...  headers/body: no
 1  design  run_e05f4344...  headers/body: YES
```

`engine/include/vayu/types.hpp:771` states it directly: `trace_data` is `// JSON (Headers/Body - only for Design Mode or Errors)`, and even the "or Errors" half is misleading - the load-run error writer stores an error envelope, not an exchange.

(Post-#263 note: the sampling path around these writers was reworked - see the anchor refresh in Dependencies and sequencing below. The finding itself, headers/bodies discarded for load runs, is unchanged.)

## Why this is not simply "do what design mode does"

`build_result_trace` (`engine/src/http/routes/execution.cpp:56`) stores the full exchange into `results.trace_data`, capped by `cap_trace_bodies` at `maxTraceBodyBytes` (5 MiB default, `constants.hpp:132`). Reusing that shape for load runs breaks the report path:

- `Database::get_results` (`engine/src/db/database.cpp:971`) is `get_all(where run_id == X)` with no limit - it loads every row for the run.
- `calculate_detailed_report` then JSON-parses **every** row's `trace_data` (`engine/src/utils/metrics_helper.cpp:73`) to accumulate error types, timing and slow counts.
- The handler serialises only the first 100 (`engine/src/http/routes/runs.cpp:609-631`).

[...]

PR #263, opened 2026-08-01
fix(engine): sample before serialising, capture outliers, retain across the whole run

### LINKED ISSUE #172: engine: load-run sampling path - test assertions cover only the run's opening, slow-request capture is dead, and per-request work is built then discarded
Part of #121. Base: `61a3b3e`.

Six defects in the path that decides which load-run completions get recorded. All sit in `load_strategy::handle_result` and `MetricsCollector`, on the event-loop worker thread. None is a crash; together they mean the sampling controls cost more than they should, one of them does nothing at all, and the records that survive describe only the opening of a run - including, per finding 6, the responses that post-run test assertions are evaluated against.

Threading note used throughout: per #129 finding 2, each `EventLoopWorker` invokes `data->callback(...)` inline on its own thread inside the `CURLMSG_DONE` drain (`engine/src/http/event_loop/event_loop_worker.cpp:359-360`), before `curl_multi_remove_handle`. So everything `handle_result` does is on the critical path of that worker's socket processing, not on a general-purpose pool.

### 1. Two config lookups per completion (low)

`engine/src/core/load_strategy.cpp:39-40`:

```cpp
int slow_threshold_ms = context->config.value ("slow_threshold_ms", 1000);
bool save_timing_breakdown = context->config.value ("save_timing_breakdown", false);
```

`context->config` is an `nlohmann::json` object, i.e. a `std::map<std::string, json>`, so each `value()` is a string-keyed map lookup constructing a `std::string` key. Both are read on every completion, and neither can change for the life of the run - `RunContext` already resolves four sibling fields once in its constructor (`engine/src/core/run_manager.cpp:194-202`).

**Failure scenario:** at 50k RPS this is 100k map lookups per second executed inline in the curl completion drain, to re-read two values that were fixed when the run started. No wrong output, but it is pure overhead in the one place where overhead feeds back into the numbers being measured (see finding 2).

### 2. The success trace is built for every completion, then discarded by the sampler (medium)

`engine/src/core/load_strategy.cpp:96-119` builds and serialises the trace before any sampling decision is made:

```cpp
std::string trace_data;
bool is_slow = latency >= static_cast<double> (slow_threshold_ms);

if (save_timing_breakdown || is_slow) {
    nlohmann::json timing_json = { ... eight fields ... };
    ...
    trace_data = timing_json.dump ();
}

context->metrics_collector->record_success (response.status_code,
latency, response.timing.queue_wait_ms, trace_data);
```

The 1-in-N gate is applied afterwards, inside `record_success` - `engine/src/core/metrics_collector.cpp:128-130`:

```cpp
if (config_.store_success_traces && !trace_data.empty ()) {
    size_t counter = success_sample_counter_.fetch_add (1, std::memory_order_relaxed);
    if (counter % config_.success_sample_rate == 0) {
```

So with the default period of 100 (`constants.hpp:146`, `DEFAULT_SUCCESS_SAMPLE_RATE`), 99 of every 100 constructed-and-dumped JSON strings are discarded immediately. The information needed to skip the work - the counter and the period - is available before the work, and `record_success` already owns both.

**Failure scenario:** a run with `save_timing_breakdown` on at 50k RPS performs ~50k `nlohmann::json` object constructions plus `dump()` per second, inline in each worker's completion drain, and drops ~99% of the resulting strings. Because that work sits ahead of `curl_multi_remove_handle` and the next `curl_multi_perform`, it delays socket processing for every other transfer on the same worker - so enabling a diagnostic option perturbs the latency distribution the run exists to measure. #170 has since made the dialog's sample control honest, so a user selecting "100% - everything" now sends a period of 1; the same code path then keeps everything but still pays the ordering cost at every other slider position.

### 3. Slow-request traces are built and then dropped unless the timing-breakdown toggle is on (medium)

`load_strategy.cpp:97-99` includes `|| is_slow` in the build condition specifically so an outlier is captured when `save_timing_breakdown` is off:

```cpp
bool is_slow = latency >= static_cast<double> (slow_threshold_ms);

if (save_timing_breakdown || is_slow) {
```

but `metrics_collector.cpp:128` gates *storage* on `store_success_traces` alone, which `RunContext` sets from `save_timing_breakdown` and nothing else (`run_manager.cpp:196`):

```cpp
if (config_.store_success_traces && !trace_data.empty ()) {
```

So whenever the toggle is off - its default, `DEFAULT_STORE_SUCCESS_TRACES = false` (`constants.hpp:148`) - every slow request's trace is built and then unconditionally discarded, and `slow_threshold_ms` has no observable effect anywhere. `isSlow` / `thresholdMs` are documented as a stored outcome in `docs/engine/db-schema.md` (the load-run success-sample writer row).

Verified against `61a3b3e` on a live daemon: 100 requests at 50 RPS with `slow_threshold_ms: 1` and `save_timing_breakdown: false`. Every request qualified as slow (avg 1.53 ms, p99 2.63 ms, both from the histogram-backed report fields). `GET /runs/:id/report` returned `results: []` - zero rows stored.

**Failure scenario:** a user sets a slow threshold to capture outliers, leaves the timing-breakdown toggle at its default, and gets nothing back - while still paying finding 2's build cost for every slow completion. On a saturated run, where most completions exceed the threshold, that is the maximum cost for zero benefit, at the worst possible moment. The report's `slow_requests_count` (`engine/src/utils/metrics_helper.cpp:96-101`) counts only stored rows, so it also reads 0 in the same configuration.

### 4. `max_success_results` is unreachable config, and the cap yields a prefix rather than a sample (medium)

`engine/include/vayu/core/metrics_collector.hpp:99` defaults the field to `DEFAULT_MAX_SUCCESS_RESULTS = 1000` (`constants.hpp:144`). `RunContext`'s constructor sets `success_sample_rate`, `store_success_traces`, `max_response_samples` and `response_sample_rate` from the run config (`run_manager.cpp:194-202`) but never `max_success_results`, and no route or config key reaches it either - so it is 1000 in production with no way to move it. Same shape as #129 finding 1's `max_errors`, reached by a different field.

The cap is also applied as a hard stop rather than a sample - `metrics_collector.cpp:132-133`:

```cpp
if (config_.max_success_results == 0 ||
success_results_.size () < config_.max_success_results) {
```

Once full, every later record is dropped, so the retained set is the *first* 1000 sampled completions, not 1000 drawn from the run. At the default period of 100 the cap fills around request 100,000. After #170 corrected the dialog's percentage-to-period conversion, "100% - everything" sends a period of 1, and the cap fills at request 1,000.

**Failure scenario:** a user drags the sample control to "100% - everything" on a 30-minute run and receives records covering its first 1,000 requests, with nothing from the remaining 99%+ and no signal that anything was dropped. The Samples tab, and the report's `error_types` and `slow_requests_count`, then describe only the opening seconds - which is also the least representative part of a run, before connection reuse and DNS caching settle.

### 5. `record_response_sample` acquires the mutex only to discover the buffer is full (low)

`metrics_collector.cpp:142-152`:

```cpp
void MetricsCollector::record_response_sample (const Response& response) {
    size_t counter = response_sample_counter_.fetch_add (1, std::memory_order_relaxed);
    if (counter % config_.response_sample_rate != 0) {
        return;
    }

    std::lock_guard<std::mutex> lock (response_samples_mutex_);
    if (response_samples_.size () < config_.max_response_samples) {
        response_samples_.emplace_back (response, now_ms ());
    }
}
```

`response_samples_` never shrinks during a run, so once it reaches `max_response_samples` it is full for the remainder. Every sampled completion after that still takes `response_samples_mutex_` to read `.size()` and return, and the copy on line 151 is a full `Response` (headers map plus body string) held under that same lock.

**Failure scenario:** a run with a test script and `response_sample_rate: 1` takes one lock acquisition per completion across all N worker threads for the entire run after the first 1000 responses, contending over a buffer that can no longer accept an element. The lock is held across a body-sized copy while it can still accept one, so the contention window is proportional to response size.

### 6. Post-run test assertions only ever see the opening of the run (high)

Same prefix behaviour as finding 4, but on the other buffer, and with a materially worse consequence: this one decides whether a run's tests pass.

`validate_scripts` evaluates the run's `tests` script against stored responses - `engine/src/core/run_manager.cpp:41` and `:90`:

```cpp
const auto& samples = context->metrics_collector->response_samples ();
...
for (const auto& sample : samples) {
```

That buffer is filled by `record_response_sample` (`metrics_collector.cpp:142-152`), which samples 1-in-`response_sample_rate` and hard-stops at `max_response_samples`. `max_response_samples` defaults to 1000 (`constants.hpp:150`) and `response_sample_rate` to 100 (`constants.hpp:152`), applied via `run_manager.cpp:199-202`, and the app sends neither - `app/src/modules/request-builder/index.tsx:454-456` sends only `success_sample_rate`, `slow_threshold_ms` and `save_timing_breakdown` - so in production they are always 100 and 1000. The buffer therefore fills at request 100,000 and admits nothing afterwards.

**Failure scenario:** a run issues 5 million requests. The target is healthy for the first minute and then starts returning 500s, or returning 200s with a truncated body. Every response the test script is evaluated against was captured before that point, so `testsPassed` is reported against a period during which nothing was wrong, and `testsFailed` is 0. The run's own status-code distribution and error rate - which come from the lock-free counters, not this buffer - simultaneously show the failures. Nothing in the report indicates that validation covered only the first 2% of the run: `TestsSampled` (`run_manager.cpp:61`) reports how many responses were tested, not which part of the run they came from.

## Fix / solution pathway

1. **Per-request config lookups:** resolve `slow_threshold_ms` and `save_timing_breakdown` once into `RunContext` alongside the four fields already resolved at `run_manager.cpp:194-202`, and read the plain members in `handle_result`.

2. **Build-then-discard:** move the sampling decision ahead of the serialisation. Either expose a cheap `should_sample_success()` on `MetricsCollector` that `handle_result` consults before building the trace, or pass a builder/callback into `record_success` that is invoked only on the sampled path. The counter must still advance for every completion so the period keeps its meaning.

3. **Dead slow-capture:** make the storage condition match the build condition. `record_success` needs to know the completion was slow (a bool parameter, or an already-non-empty `trace_data` treated as intent to store), so that `store_success_traces == false` no longer discards an outlier trace that was built precisely because it was an outlier. Note this interacts with item 2: once the gate moves ahead of the build, "slow" and "sampled" become two independent reasons to build, and both need to reach storage.

   Slow completions should **not** consume the 1-in-N budget - a user who sets a threshold has asked for those specifically, and spending the sampling budget on them would mean a slow run silently stops sampling normal traffic. That makes the slow path unbounded on its own, though: under saturation most completions cross the threshold, so it needs its own retention limit (and its own displaced/dropped counter) rather than inheriting the success budget. Sizing it alongside `max_success_results` in item 4 keeps one place to reason about total retention.

4. **Unreachable / prefix cap:** plumb `max_success_results` through `RunContext` from the run config the same way its four siblings are plumbed, and replace the hard stop with a reservoir (Algorithm R) so a full buffer keeps admitting later completions with uniform probability instead of dropping them. Count what was displaced or dropped and surface it, mirroring #129's acceptance criterion that excess errors are counted rather than silently truncated.

5. **Validation sees only the run's opening (finding 6):** apply the same reservoir to `record_response_sample` (`metrics_collector.cpp:150`), so the responses handed to `validate_scripts` are drawn from the whole run rather than its first 100,000 requests. A prefix is not defensible here even though the buffer is smaller: unlike the Samples tab, a biased selection does not merely look unrepresentative, it produces a pass/fail verdict that contradicts the run's own error rate. Since the reservoir keeps admitting responses for the run's full duration, the copy cost in item 6 stops being a first-1000-only concern and applies throughout - another reason to move the `Response` copy out of the lock.

   Consider also reporting the span the tested responses covered (first and last sample timestamp are already carried on `ResponseSample`), so a report can distinguish "tested across the run" from "tested at the start" without the reader having to know the sampling internals.

6. **Lock to find it full:** publish the size as a relaxed atomic on insert and check it before taking the lock; keep the in-lock check as the authoritative one. Move the `Response` copy out of the critical section - with item 5's reservoir the buffer keeps accepting for the whole run, so this stops being a cost paid only until the buffer fills.

## App changes

None required. All six anchors are engine-side, and every fix preserves the wire shape of `POST /runs` and `GET /runs/:id/report`.

Two follow-on notes, not part of this issue:

- Fixing finding 3 means slow-request rows begin appearing for anyone who has set `slow_threshold_ms` with the timing-breakdown toggle off. That is the documented intent, but it is a behaviour change and the load-test dialog's copy may want to say so.
- Finding 4's fix makes `max_success_results` settable engine-side; whether the dialog should expose it is a separate product question. `app/src/constants/load-test.ts` (`successSamplePeriod`, added in #170) is where a new control would convert units.

## Tests

Engine gtest, mutation-checked (revert each fix individually, confirm the matching test fails, restore):

- **Config hoist:** no direct behavioural test; covered by the existing load-strategy suite staying green. Optionally assert `RunContext` exposes the two resolved fields with the documented defaults.
- **Build-then-discard:** a test that records N successes with a period of 100 and asserts the trace builder ran ~N/100 times, not N - e.g. by counting invocations through a builder callback, or by asserting `record_success` never receives a non-empty trace it then drops. This is the one finding where the fix is invisible to output, so the test has to pin the call count or the ordering directly.
- **Dead slow-capture:** record a success whose latency exceeds `slow_threshold_ms` with `store_success_traces == false`, and assert one record is stored carrying `isSlow`. Reverting the fix leaves the collector empty. Add the mirror case (fast completion, breakdown off) asserting nothing is stored, so the fix cannot become "store everything".
- **Unreachable cap:** assert `RunContext` propagates a `max_success_results` from the run config into `MetricsCollectorConfig`, and that the default is still 1000 when absent.
- **Prefix vs sample:** record far more sampled successes than the cap and assert the retained set contains records from the final decile, not only the first - the current hard stop fails this. Assert the retained count never exceeds the cap, and that a drop/displacement count is available.
- **Slow bypasses the sampling budget:** record a long run of fast completions interleaved with a few slow ones under a period large enough that none of the fast ones is sampled, and assert every slow completion is still stored. Then record more slow completions than the slow-path limit and assert the retained count stops at that limit with a non-zero dropped count - so "always keep slow ones" cannot become unbounded under saturation.
- **Validation covers the run (finding 6):** fill `record_response_sample` well past `max_response_samples` and assert the retained responses include ones recorded late in the sequence. The sharpest version drives `validate_scripts` end to end with a script asserting `status == 200`, feeding successes first and failures only after the buffer would have filled, then asserts `testsFailed > 0` - which fails on today's prefix behaviour and is the actual user-visible defect.
- **Lock-free full check:** assert `record_response_sample` past the cap does not grow the buffer and, if the size atomic is exposed, that it agrees with `response_sample_count()`. Contention itself is not directly assertable; the existing concurrent-recording tests guard correctness. Note this test's premise changes once item 5 lands - past the cap the buffer no longer stops accepting, it starts displacing, so assert a bounded size rather than a frozen one.

`engine/tests/metrics_collector_test.cpp` is the natural home for all but the first, and gained three tests in #171 for the reserve caps - keep the new cases beside those so the reserve and the retention limits stay legible together.

## Docs

- `docs/engine/db-schema.md` - the writers table currently says a load-run success sample stores timing "only when `save_timing_breakdown` is on or the sample crossed `slow_threshold_ms`". The second clause is finding 3's bug, not current behaviour; correct it in the same change that fixes the code so the doc describes what ships.
- `docs/engine/api-reference.md` - `success_sample_rate` and `slow_threshold_ms` semantics on `POST /runs`, plus `max_success_results` if it becomes settable.
- `docs/engine/scripting.md` - the `tests` script's evaluation model. Finding 6 means the documented behaviour and the actual coverage differ today; once fixed, state plainly that assertions run against a bounded sample drawn from the whole run, and that `TestsSampled` is that sample's size rather than the request count.
- `docs/engine/benchmarks.md` - if the sampling reorder measurably moves achieved RPS with `save_timing_breakdown` on, the numbers and the method note should reflect that this option is no longer a per-request cost.

## Sequencing

Part of #121, and it must land **after** the sub-issues already in flight, because it edits the same two functions:

- **#132** (`POST /runs` validation) fixes the `% 0` SIGFPE at what is now `metrics_collector.cpp:130` - the exact modulo finding 2 relocates. Rebasing onto it avoids re-deriving the guard.
- **#136** rewrites pacing in `load_strategy.cpp`.
- **#147** item 6 changes `handle_result`'s early return (`load_strategy.cpp:43-48`), the same function as findings 1-3.
- **#169** (#129) rewrites `record_success`'s neighbourhood - bounded stores and the atomic histogram calls - and bounds `max_errors`. Finding 4 is the `max_success_results` counterpart of #129 finding 1 and should adopt whatever plumbing and drop-counter shape #169 settles on, rather than inventing a second one.

So: strictly after **#132 -> #136 -> #147 -> #169** in #121's stated order. Nothing here blocks any of them.

This issue is deliberately scoped to the sampling path's cost and retention. Persisting response headers and bodies for sampled load-run completions is a separate feature and is not proposed here; it depends on findings 2-4 landing first, and on #128's response-body bound deciding how large a stored body could ever be.

## Acceptance criteria

- [ ] `handle_result` reads no `nlohmann::json` config field per completion; `slow_threshold_ms` and `save_timing_breakdown` are resolved once into `RunContext`.
- [ ] No success trace is serialised for a completion that the sampler then discards; the sampling counter still advances once per completion.
- [ ] With `save_timing_breakdown` off and `slow_threshold_ms` set, a completion slower than the threshold is stored with `isSlow`, and the report's `slow_requests_count` is non-zero. The empirical repro in finding 3 returns a non-empty `results` array.
- [ ] `max_success_results` is settable from the run config through `RunContext`, defaulting to 1000 when absent.
- [ ] With more sampled completions than the cap, the retained set is drawn from the whole run rather than its prefix, never exceeds the cap, and the number displaced or dropped is available to the report.
- [ ] A completion slower than `slow_threshold_ms` is stored without consuming a 1-in-N slot, and the slow path has its own retention limit and dropped count so a saturated run cannot grow it without bound.
- [ ] `validate_scripts` evaluates responses drawn from the whole run, not its first `max_response_samples`. A run whose target begins failing after the buffer would previously have filled reports `testsFailed > 0`.
- [ ] `record_response_sample` does not acquire `response_samples_mutex_` on the fast path once the buffer is at capacity, and does not hold it across the `Response` copy.
- [ ] Docs listed above updated in the same change; `ctest --preset linux-dev` green.

athrvk:
## Description

All six findings of #172, verified still present at `6d4bbbc` at the issue's anchors (shifted by a few lines).

**The plan.** The six defects share one root cause: the sampling path decides *after* it has already paid. It serialises a trace and then asks whether to keep it; it builds a slow-request trace because the completion was slow and then discards it because a different toggle was off; it takes a mutex to discover a buffer has been full since request 100,000; and it treats "the store is full" as "the run is over", so what survives describes the run's opening. So the fix moves each decision in front of the work it authorises - a cheap gate (`should_sample_success`) before the build, a claimed slot before the copy - and replaces every hard stop with reservoir retention so a full store keeps admitting.

The alternative I rejected for retention was raising the caps. It buys a longer prefix, not a representative one: at 52k RPS the default buffers fill ~1.9 seconds into a 60-second run, and no ceiling a user would accept moves that from "the connection-establishment burst" to "the run". Algorithm R makes the retained set uniform over the whole stream at the same memory cost, which is the property the report and the test verdict actually need. For the slow path I rejected letting outliers share the 1-in-N budget (a degrading target would silently stop sampling ordinary traffic) and rejected leaving it unbounded (under saturation most completions cross the threshold), so it gets its own budget and its own drop count.

**Blast radius traced.** `handle_result` and `MetricsCollector` are engine-internal; the changed `record_success` signature has one production caller (`load_strategy.cpp`) and the tests. `flush_to_database` now walks both success stores into the same `results` table, so the report's `slow_requests_count` / `timingBreakdown` / `error_types` readers are unchanged - a stored row is still a stored row, told apart by the trace's `isSlow`. The one new wire field is `sampling` on `GET /runs/:id/report`, read back from a new `sampling` object in `runs.summary`; `runs_route_test.cpp` round-trips the pair like the other summary keys. `POST /runs` gains three validated fields and rejects nothing any client sends today (the dialog's slow-threshold range, 0-60000, sits inside the engine's 0-86400000).

**Consumers checked, both clients:** the renderer sends `success_sample_rate`, `slow_threshold_ms` and `save_timing_breakdown` (`request-builder/index.tsx:445`) and reads no retention field; MCP sends none of them. `rg "max_success_results|max_slow_results|successTracesDropped|responseSamplesDropped" app/src app/electron` is **empty** - verified, not assumed.

### The findings

1. **Two config lookups per completion.** `slow_threshold_ms` is resolved once into `RunContext` and read as a plain member.
2. **Build-then-discard.** `should_sample_success()` advances the period and answers before anything is serialised. The counter still ticks for every completion, so the period keeps meaning "1 in N of the run".
3. **Dead slow capture.** Storage no longer gates on `save_timing_breakdown`: a completion past the threshold is stored with `isSlow` whatever the toggle says. The issue's empirical repro (100 requests, `slow_threshold_ms: 1`, breakdown off) now returns a non-empty `results` array.
4. **Unreachable / prefix cap.** `max_success_results` reaches the collector from the run config like its four siblings, and the cap is a reservoir with a drop count.
5. **Lock to find it full.** The response-sample slot is claimed before the `Response` copy; a refusal past capacity takes no lock and copies nothing, and the copy never happens inside the critical section.
6. **Validation saw only the opening.** The same reservoir on `record_response_sample`, so `validate_scripts` grades responses drawn from the whole run.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

Not marked breaking: nothing that worked stops working. Two behaviour changes are deliberate and documented - a run with `slow_threshold_ms` set and the breakdown toggle off now *does* store outlier rows (that is finding 3), and `slow_threshold_ms: 0` now disables outlier capture rather than marking every completion an outlier.

## Testing

```bash
python build.py -e -t
cd engine && ctest --preset linux-dev --output-on-failure   # 785/785 passed
```

**Engine: 785/785 pass**, up from 760 on `master` - **+25**: 5 in the new `engine/tests/reservoir_test.cpp`, 10 in `metrics_collector_test.cpp`, 4 in `load_strategy_test.cpp`, 4 in `run_config_validation_test.cpp`, 2 in `runs_route_test.cpp`. No pre-existing failures on the base commit.

App suite not run and no app file touched - `rg` over `app/src` + `app/electron` finds no reference to any changed field (above), and nothing on the renderer's request or report path changed shape.

**Mutation-checked** per CLAUDE.md - each fix reverted, the failure observed, then restored:

| Reverted | Fails |
|---|---|
| slow storage re-gated on `store_success_traces` | `SlowTraceIsStoredWithTimingBreakdownOff`, `SlowCaptureIsBoundedAndCountsWhatItDropped`, `SlowCompletionIsCapturedWithTimingBreakdownOff` |
| reservoir → hard stop at capacity (both stores) | `SampledSuccessesAreDrawnFromTheWholeRun`, `ValidationSamplesCoverTheWholeRun` |
| gate authorises every completion (build first, sample after) | `TheSamplingGateSelectsOneInN`, `OnlySampledCompletionsAreStored` |
| `slow_threshold_ms` no longer resolved from the run config | `SlowCompletionIsCapturedWithTimingBreakdownOff`, `RunContextResolvesTheSamplingConfigOnce` |

**One thing I will not overclaim.** Finding 2's *ordering* is invisible from outside - a trace that is never built leaves nothing to assert on, and pinning it would mean a production counter that exists only for a test. What is pinned instead is the gate the caller consults (1000 completions at a period of 100 authorise exactly 10 traces, on the period) plus the storage outcome, and the test comments say so rather than implying a check they do not perform.

Formatting: `clang-format` in full on the files that were clean on `origin/master` (`constants.hpp`, `run_config_validation_test.cpp`, and the two new files); `git clang-format` for added lines only elsewhere. `clang-tidy -p build` on every changed `.cpp` reports nothing new - the two `sign-conversion` warnings in `metrics_collector.cpp` are pre-existing, in untouched lines (`record_status_code` / `status_code_distribution`), exactly as #169 noted. `mkdocs build --strict -f .github/mkdocs.yml` clean.

## Checklist

## Docs

- **`docs/engine/db-schema.md`** - the writers table said a load-run success sample stores timing "only when `save_timing_breakdown` is on or the sample crossed `slow_threshold_ms`". The second clause was finding 3's bug described as behaviour; it now describes what ships, including the two separate budgets.
- **`docs/engine/api-reference.md`** - `POST /runs` gains the three new accepted ranges and a **What a run stores, and for how long** table (which budget each store fills from, what bounds it, that an outlier never consumes a sampling slot, that retention is a reservoir); `GET /runs/:id/report` documents the `sampling` object and that its absence is not a claim of "nothing dropped".
- **`docs/engine/scripting.md`** - the load-test evaluation model: assertions run against a bounded sample drawn from the **whole run**, and `samplesTested` is that sample's size rather than the request count.
- **`docs/engine/benchmarks.md` deliberately untouched.** The issue asks for it *if* the reorder measurably moves achieved RPS. I did not run `bench-compare.sh` (a shared container is not a load-generation baseline), so I have no measurement to write down; publishing an unmeasured claim would be worse than the silence.

- **`save_timing_breakdown` is not a second field on `RunContext`.** It is resolved once, in the same constructor, into `MetricsCollectorConfig::store_success_traces` - which is where the sampling gate reads it. A copy on `RunContext` would have no reader, which is this codebase's most repeated defect. `slow_threshold_ms` *is* on `RunContext`, because `handle_result` needs it per completion. The acceptance criterion this bends is "resolved once into `RunContext`"; the property it protects (no `nlohmann::json` field read per completion) holds.
- **`max_slow_results` is settable from the run config**, not only a constant. The issue asks the slow path for "its own retention limit ... sized alongside `max_success_results`"; since that sibling becomes settable per the acceptance criteria, leaving this one hardcoded would have split one decision across two mechanisms.
- **The tested-span report (first/last sample timestamp) is not included.** The issue offers it as a "consider", and its only consumer would be a UI surface the issue puts out of scope - so it would ship as a written-never-read field. `sampling.responseSamplesDropped` already tells a reader the sample was thinned, which is the part the acceptance criteria require.
- **`errors_dropped` is surfaced too**, though the issue scopes retention reporting to the new counters. It was an accessor with no reader since #169, it is one line in the same object, and a report that says what three of four stores dropped is worse than one that says all four.
- **Untouched:** response headers/body capture for sampled completions (#174, which this unblocks), the load dialog's copy about slow rows now appearing (the issue's own follow-on note), and `bench-compare.sh` numbers.

## Related Issues

Closes #172

Part of #121. Unblocks #174, which the issue lists as depending on findings 3, 4 and 6 landing first.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MkoAqZQ3N37ZhVctvDuMQR)_


athrvk:
update with latest master and fix merge conflicts

athrvk:
Do these changes/fixes require UI changes?

athrvk:
Merged `origin/master` (`f0fc04c`) in and pushed - three conflicts, all resolved by keeping both sides:

- **`run_manager.cpp`** and **`runs.cpp`** - #268's `http_version_downgraded` and this branch's `retention` / `sampling` land on the same two structs and the same two summary-write sites. Additive on both sides, no interaction.
- **`api-reference.md`** - both added a paragraph after the same `latency.*` sentence under `GET /runs/:id/report`. Both kept, downgrade first.

Verified the merge rather than trusting it: #268's downgrade counter still fires ahead of the error/success split in `handle_result` (it sits above the branch this PR rewrites), and both fields reach the summary at the completed-run *and* crashed-run sites. **800/800 engine tests pass** on the merged tree (785 from this branch + 15 from master's new suites).

---

**On UI changes: none required, and the interesting part is why.**

The dialog already promises what this PR implements. `LoadTestConfigDialog/index.tsx:549` has read, all along:

> *Requests slower than this are flagged and always saved, whatever the sample rate.*

Both halves of that were false until this branch. Nothing was saved unless **Save timing breakdown** happened to be on (finding 3), and there was no sense in which a slow request was exempt from the sample rate. So the UI needed no change because the UI was never wrong about the intent - the engine was wrong about the behaviour, and the copy is now true.

The wire shape is unchanged for everything the app reads: `POST /runs` takes the same fields (three new ones are optional and no client sends them), and `GET /runs/:id/report` only *gains* a `sampling` object. `rg "max_success_results|max_slow_results|successTracesDropped|responseSamplesDropped" app/src app/electron` is empty.

Two things a future UI change could do, both deliberately out of scope here:

1. **Surface the retention counts.** `sampling.successTracesDropped` / `responseSamplesDropped` are exactly the numbers a Samples tab needs to say "showing 1,000 of 30,000" instead of implying the list is everything. That belongs next to #174's Samples work rather than in a fix PR.
2. **Say that "always saved" has a ceiling.** Outlier retention is bounded (`max_slow_results`, default 1000) so a saturated run cannot grow it without limit, and the dialog cannot set that bound today. The hint over-promises only in that corner - a run with more than 1,000 outliers keeps a uniform sample of them, not all of them.

I left both alone rather than editing app copy inside an engine fix, but (2) is a one-line hint change if you want it in this PR - say the word and I will push it.

---

athrvk:
Two things a future UI change could do- do them right now

athrvk:
Both done, pushed as `868ec63`.

**1. The retention counts are now on screen.** `SampleRetentionNote` - shared, built on `Callout` (`severity="info"`) rather than a hand-rolled muted row - renders under every surface that displays a bounded set:

| Surface | Budget it reads |
|---|---|
| Dashboard → Sampled Requests | `successTracesDropped + slowTracesDropped` |
| Dashboard → Test Validation | `responseSamplesDropped` |
| History → Samples tab | `successTracesDropped + slowTracesDropped` |

> **Bounded retention** - 29,000 further samples were displaced by this run's retention limit. The 100 shown are drawn uniformly from the whole run, not from its opening.

Three surfaces, one component: written out three times it would have drifted, which is the defect the note itself is about. The two trace budgets are summed rather than shown separately - which store a given row was charged to is an engine detail, and making the reader reconcile two numbers against one list helps nobody.

It is **silent in two cases, deliberately**. A run that displaced nothing gets no notice (a "0 dropped" banner is noise). A run whose summary predates the counts also gets none - `sampling` absent means *we cannot tell*, not *nothing was dropped*, and turning that into a completeness assurance is the same class of lie the note exists to prevent. That mirrors the absent-vs-zero rule `httpVersionDowngraded` already follows.

While wiring it I found the badges themselves overstate: **`results` is capped at 100 by the report route irrespective of retention**, so "100 captured" was wrong even before the stores were bounded. Both badges now say **shown**.

**2. The dialog no longer promises more than the engine delivers.**

> ~~Requests slower than this are flagged and always saved, whatever the sample rate.~~
> Requests slower than this are flagged and saved whatever the sample rate, up to a retention limit. 0 turns the capture off.

Two corrections in one line: outlier capture is bounded (`max_slow_results`, default 1000, because under saturation nearly every request crosses the threshold), and `0` disables the capture rather than flagging everything - which is new behaviour from this PR and had no UI expression at all.

**Testing.** App suite **2364/2364 across 277 files** (up from 2352/273 - +12: 5 for the component, 4 for the dashboard wiring, 3 for the history wiring). `pnpm type-check` clean, ESLint clean on every touched file, Prettier run only on the four new files. Engine untouched by this commit.

Mutation-checked both halves:

| Reverted | Fails |
|---|---|
| note fires even when nothing was displaced | 4 tests - the two "stays silent" cases plus both surfaces' quiet-run cases |
| `SampleRetentionNote` removed from the history Samples tab | `says how much the bounded stores displaced` |

Docs in the same commit: `docs/app/api-integration.md` (what the renderer reads from `report.sampling`, and the absent-vs-zero rule) and `docs/app/COMPONENTS.md` (the new shared component and why it is shared). `mkdocs build --strict` clean.

---

**On the earlier red CI** - it was not a real failure, and needs no fix. You clicked *Update branch* at 16:26:03, which pushed merge commit `1067074` while run 610 was still building; `pr-tests.yml` sets `cancel-in-progress: true`, so all three engine jobs were cancelled mid-`Build engine` and the gate reported `engine: cancelled <-- failed`. No test ran, let alone failed. Run 611 picked up `1067074`, and this push supersedes it in turn.

---

PR #264, opened 2026-08-01
fix(app,ci): clear the ESLint debt and enforce lint in the PR matrix

### LINKED ISSUE #189: app: burn down the ESLint debt (100 errors) and enforce lint in the PR CI matrix

athrvk:
## Description

**Plan.** The root cause is that `pnpm lint` has never run in CI - `rg "lint" .github/workflows` matched only the PR-title linter - so a rule the repo states in `CLAUDE.md` ("no `any`") was unenforced and the count drifted to **106 errors and 8 warnings** (the issue's snapshot of 100/8 at `4136c66`; re-measured on current master before the first edit). The approach is the issue's two steps: fix every finding at its source, then add `pnpm lint` and `pnpm format:check` to the app CI job so a regression fails the PR. **The alternative I rejected is widening the rule config** - `allowExportNames` for the fast-refresh warnings, or downgrading `no-explicit-any` - which makes the number go to zero without changing anything the rules exist to catch; the same debt would simply stop being reported.

Verified the problem still exists as described before starting: `pnpm lint` fails, and the breakdown matches the issue apart from `vayu/app` (that plugin is no longer in `eslint.config.mjs`) and `no-require-imports`, which the issue did not list.

### `@typescript-eslint/no-explicit-any` (84)

Most of them are the four import parsers, `importers/shared.ts` and the two wire transformers walking JSON that came from outside the app. They now walk it through shared narrowing helpers - **`src/lib/json-node.ts`**: `asRecord` / `asArray` / `asStr` / `prop` - so each hop states what it expects instead of `any` making every property read look safe. One copy, because all four parsers walk the same kind of document (per `CLAUDE.md`: a hand-rolled copy of a primitive does not receive the primitive's fixes). `mapKeyValues` / `toVarRecord` take `unknown` and narrow internally rather than each caller shaping a row for them.

`RequestTransformer.toFrontend` / `CollectionTransformer.toFrontend` take a `RawRequest` / `RawCollection` - a bag of unknowns, which is what the engine actually sends. `api.ts` had those responses typed as `Request` / `any`; a wire row carries **numeric** timestamps and may predate a column, which is the reconciliation these transformers exist to do, so the old typing was a lie in the direction that hides bugs.

Two casts remain, each with the reason on the line: the synthesised `SanityResult` in `useEngine` (no exchange happened, so headers/body/timing do not exist and zeroed timings would read as measured), and the transformers' `new Date(raw.createdAt as string | number)` which keeps the existing "invalid date throws" behaviour rather than quietly inventing one.

**Blast radius traced:** the parsers' consumers are the import orchestrator and `ImportModal` (drafts only - shapes unchanged); the transformers' only callers are the seven `apiService` methods above. Typing the graphql test mocks surfaced a **stale fixture**: `introspect.test.ts` built `preRequestScripts` entries with `source`, where `ScriptPart` has `origin` - it type-checked before only because the object was cast to `any`.

### `react-hooks/*` (13) - fixed, not disabled

- **`variable-popover.tsx` (3 × `set-state-in-effect`)** - seeding the edit buffer and clearing the secret reveal belong to the open/close *event*, so both moved into `openPopover()` / `closePopover()`. The seeding effect was keyed on `isOpen` and carried an `exhaustive-deps` suppression precisely because it re-ran and **reset `editValue` under a user mid-type** whenever the parent passed a fresh `varInfo` object; that suppression is gone with it.
- **`RequestBuilderProvider.tsx` (2)** - the request-identity reset and the backend restore now adjust state during render (React's documented "adjust state when a prop changes"), which also drops an `exhaustive-deps` suppression. The mount pass is exactly what the `useState` initialisers already do. The half that writes to the response store stays in an effect, because that is a write to something outside React.
- **`SettingsMain.tsx` (1)** - the category reset adjusts during render; the flush of pending edits stays in the cleanup, where the ref still holds the *outgoing* category's values.
- **`dashboard/index.tsx` (1 + `preserve-manual-memoization`)** - the report fetch runs from an async function with a cancel flag (the shape the sibling retry effect in the same file already uses), so a late answer cannot write state for a run the user left. The memo now depends on a hoisted `runConfiguration`, which is the dependency the compiler infers.
- **`CollectionTree.tsx`, `request-builder/index.tsx` (`exhaustive-deps`)** - `navigateToRequest` / `navigateToCollection` memoised; the missing `showToast` dependency added.

**Two suppressions remain**, each justified on the line: `preload.ts`'s `require()` (a preload script is loaded as CommonJS in the isolated world - an ESM import fails at runtime), and `VariableTableEditor`'s row seed plus its `Date.now()` stamp. That seed is gated on `hasPendingChangesRef`, a flag written synchronously so a save's echo cannot overwrite keystrokes; deriving the rows during render would mean *reading that ref during render*, and moving the guard into state would make it one render late - i.e. reintroduce the bug the guard exists to prevent. Left alone deliberately rather than rewritten inside a lint sweep.

### `react-refresh/only-export-components` (6 warnings)

`--max-warnings 0` means these fail the run too. The non-component exports moved out rather than being allow-listed: `badgeVariants` / `buttonVariants` / `toastVariants` into sibling `*-variants.ts` modules (still re-exported from `components/ui`, so no caller changes), and `formatRelative` / `fmt` into `format.ts` files beside their `shared.tsx`. `request-builder/index.tsx` no longer re-exports `useRequestBuilderContext` - every consumer already imports it from `./context`.

### Enforcement

`pnpm lint` and `pnpm format:check` run in the app job, on `ubuntu-latest` only: both are platform-independent, and the three-OS matrix exists for the suite's platform assertions. The app tree is already Prettier-clean, so `format:check` starts green (only files touched here were formatted - never repo-wide, per `CLAUDE.md`).

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

All run on this branch:

- `cd app && pnpm lint` - **exits 0** (was 106 errors + 8 warnings).
- `cd app && pnpm format:check` - clean.
- `cd app && pnpm type-check` - clean (all three tsconfigs).
- `cd app && pnpm test` - **2332 passed, 273 files** (+8 tests, +1 file: `src/lib/json-node.test.ts`). Baseline on the same tree before the change: 2324 passed, 272 files - no pre-existing failures.

Engine untouched, so no `ctest` run: no C++ file is in the diff and no test there could change colour.

New coverage is the narrowing helpers, which every parser now depends on: `asRecord` rejecting an **array** is the case worth a test, since `typeof [] === "object"` is exactly what the obvious inline check gets wrong, and `asStr` never coercing is what keeps a numeric field from arriving as a `string`-typed number.

**Mutation-checked**: dropping `!Array.isArray(v)` from `asRecord` fails 2 of the new tests **and** 2 importer tests (`Test Files 1 failed | 15 passed`); restored, all green.

The behaviour-preserving claim rests on the existing suites: the importer/fixture-parity tests (190) cover the four parsers end to end, and the request-builder (492) and settings suites cover the effect rewrites.

## Checklist

## Related Issues

Closes #189

Docs updated in the same commit, per the issue: `CONTRIBUTING.md` (lint and formatting are CI-enforced, and how to justify a suppression), `docs/plans/pending-backlog.md` (**N2** removed as shipped), `app/CLAUDE.md` (the "no `any`" rule now names its enforcement), and `docs/app/COMPONENTS.md` (the `*-variants.ts` and `format.ts` splits, so the next reader is not surprised by a module that only exists for fast refresh).

---
_Generated by [Claude Code](https://claude.ai/code/session_01TVhq5B7fqLZcVStef6KPWR)_


athrvk:
merge with latest master and eslint is fixed and added into CI and forced by CI

athrvk:
Merged latest `master` (through #263, `a2bc8f3`) into the branch and pushed - no conflicts, and nothing in the merge needed a fix.

The interesting part is what the merge proves rather than what it changed: everything that landed on master since this branch was cut - the sample-retention notes, the MCP resources tests, the protocol-downgrade viewer, the libsodium swap's app-side edits - passes the newly enforced gate untouched. `pnpm lint` exits 0 and `pnpm format:check` is clean on the merged tree, so the rule set is not about to fail somebody else's already-merged work.

Re-verified on the merge commit:

- `cd app && pnpm lint` - 0 errors, 0 warnings
- `cd app && pnpm format:check` - clean
- `cd app && pnpm type-check` - clean (all three tsconfigs)
- `cd app && pnpm test` - **2372 passed, 278 files** (was 2332/273 before the merge; the difference is master's new tests, all green)

Engine untouched by this branch, so no local `ctest` run - CI covers it on all three platforms.

---

PR #265, opened 2026-08-01
refactor(engine): replace the hand-written hash and encoding layer with libsodium

### LINKED ISSUE #232: engine: replace the hand-written hash and byte-encoding layer (and vendored picosha2) with libsodium
Base: `cef24ca` (PR #230 branch). Filed out of review discussion on #230 - the question there was "why are we hand-rolling this, can't we use something industry-trusted?", and the answer is that we should. This issue is that work, kept out of #230 so the signing feature and the dependency swap stay reviewable separately.

## Problem

The engine hand-maintains its entire hash and byte-encoding layer, plus a vendored third-party hash:

| Thing | Where | What it is |
|---|---|---|
| `sha256` | `include/vayu/utils/sha256.hpp` | wrapper over vendored picosha2 |
| `hmac_sha256` | `include/vayu/utils/sha256.hpp` | RFC 2104, written in-repo (#230) |
| `base64_encode` | `include/vayu/utils/encoding.hpp` | written in-repo |
| `base64url_encode` | `include/vayu/utils/encoding.hpp` | written in-repo |
| `base64_decode` | `include/vayu/utils/encoding.hpp` | written in-repo (#230) |
| `hex_encode` | `include/vayu/utils/encoding.hpp` | written in-repo (#230) |
| picosha2 | `engine/vendor/picosha2/` | vendored single-header, MIT |

Every one of these has real callers, so this is not dead code being tidied:

- **PKCE** - `include/vayu/http/pkce.hpp:26,27,42` (`sha256`, `base64url_encode`)
- **OAuth 2.0 client auth** - `src/http/oauth_client.cpp:112` (`base64_encode`)
- **Basic auth** - `src/http/auth_resolver.cpp:163` (`base64_encode`)
- **Script sandbox** - `src/runtime/script_engine.cpp:1706,1719,1814-1818,1861,1885` (all six, via `pm.crypto` and `btoa`/`atob`)

**Why now.** One hand-written base64 encoder for a Basic auth header was a defensible trade when that was the whole requirement - adding a dependency cost more than it saved. #230 pushed the surface to a hash, an HMAC, three encoders and a decoder, which flips that calculus: this is now a maintenance surface that belongs to a library.

## What was already ruled out - please don't re-litigate these

All four were checked against the actual toolchain and the actual installed headers during #230 review, not from memory.

**The C++ standard library.** There is no base64, SHA-256 or HMAC in C++20, C++23 or C++26, and no adopted proposal on track. C++23 brought `std::expected` / `std::print` / `std::mdspan`; C++26 brought reflection, senders/receivers, contracts. `std::text_encoding` (C++26) is the IANA *charset-name* registry, not a codec. The only relevant standard facility is `std::format("{:02x}")` for hex, which is **already available at our current C++20** (compiled and confirmed: `ba78160f`), so raising `CMAKE_CXX_STANDARD` buys nothing here and would raise the compiler floor on three CI platforms.

**cpp-httplib.** Has `detail::base64_encode` and no decoder. It is in `detail::`, it is introduced at `httplib.h:4588` by a link to a StackOverflow snippet - so it is a hand-rolled base64 too, just someone else's - and `vayu_core` deliberately does not link httplib (only `vayu-cli`, `vayu-engine` and `vayu_tests` do).

**curl.** No public base64 - `Curl_base64_encode` is internal. `curl_easy_escape` / `unescape` are URL encoding only.

**OpenSSL.** Two independent disqualifiers.

1. *It is not in the Windows build.* vcpkg's curl port selects the TLS backend per platform - `openssl` on `uwp | !windows`, `sspi` (Schannel) on `windows & !uwp` - so the "OpenSSL is already a transitive dependency" premise holds only on Linux and macOS. `engine/vcpkg.json` never declared it directly either.
2. *Its APIs for these two jobs are wrong.* Measured against the installed OpenSSL 3.6.2:

```
EVP_DecodeBlock("Zg==") -> returned len=3, bytes: 66 00 00   (true payload is 1 byte)
EVP_DecodeBlock("Zm8=") -> returned len=3                    (true payload is 2)
EVP_DecodeBlock("Zm9$") -> -1                                (rejects invalid input, good)
OPENSSL_buf2hexstr({ba,78,16}) -> "BA:78:16"
```

`EVP_DecodeBlock` zero-pads and reports the *padded* length, so `atob('Zg==')` would yield `"f\0\0"` unless the caller counts trailing `=` itself. `OPENSSL_buf2hexstr` returns colon-separated uppercase when a signature needs `ba7816…`. Adopting OpenSSL would keep the wrappers *and* add a heavy dependency.

## Proposed: libsodium

`libsodium` 1.0.22, **ISC** licensed, in vcpkg with first-class Windows (MSBuild) support. It is the only candidate that covers all four needs with APIs that are correct as-called. Verified against the real headers (1.0.18 via apt; the API is unchanged in 1.0.22):

| Current | Replacement |
|---|---|
| `sha256` (picosha2) | `crypto_hash_sha256` |
| `hmac_sha256` (in-repo RFC 2104) | `crypto_auth_hmacsha256_init/update/final` - **`_init` takes an arbitrary `keylen`**, so RFC 2104 key padding and key-hashing move inside the library |
| `base64_encode` | `sodium_bin2base64`, `sodium_base64_VARIANT_ORIGINAL` |
| `base64url_encode` | `sodium_bin2base64`, `sodium_base64_VARIANT_URLSAFE_NO_PADDING` - exactly PKCE's format, no post-processing |
| `base64_decode` | `sodium_base642bin` - reports the **true** decoded length, validates, and takes an `ignore` set for whitespace |
| `hex_encode` | `sodium_bin2hex` - lowercase, no separators |

Net effect: a **vendored dependency is removed**, roughly 200 lines of hand-written encoding/hash code are deleted, and one well-known library is declared. It also brings `sodium_memcmp`, which closes the constant-time-comparison gap documented in `sha256.hpp` today - relevant the moment anything wants to *verify* a signature rather than produce one.

Signatures for reference:

```c
int    crypto_hash_sha256(unsigned char *out, const unsigned char *in, unsigned long long inlen);
int    crypto_auth_hmacsha256_init(crypto_auth_hmacsha256_state *state,
                                   const unsigned char *key, size_t keylen);
char  *sodium_bin2base64(char *b64, size_t b64_maxlen,
                         const unsigned char *bin, size_t bin_len, int variant);
int    sodium_base642bin(unsigned char *bin, size_t bin_maxlen,
                         const char *b64, size_t b64_len, const char *ignore,
                         size_t *bin_len, const char **b64_end, int variant);
char  *sodium_bin2hex(char *hex, size_t hex_maxlen, const unsigned char *bin, size_t bin_len);
```

## Work

- `engine/vcpkg.json` - add `libsodium`.
- `engine/CMakeLists.txt` - `find_package(unofficial-sodium CONFIG REQUIRED)`, link `unofficial-sodium::sodium` on `vayu_core`, **delete** the `vendor/picosha2` SYSTEM include block.
- `engine/include/vayu/utils/sha256.hpp` - reimplement both functions; drop `#include <picosha2.h>`.
- `engine/include/vayu/utils/encoding.hpp` - reimplement the four base64/hex functions. **Leave `url_encode` / `url_decode` / `form_encode` alone** - they are URL/form encoding, libsodium does not do them, and they are out of scope.
- `rm -rf engine/vendor/picosha2`.
- `sodium_init()` must run before any other libsodium call (it seeds the RNG and selects CPU-dispatched implementations). Prefer a `std::call_once` guard invoked by the wrappers over per-entry-point calls, so `vayu-engine`, `vayu-cli`, `vayu_tests` and the next entry point cannot forget. A WIP `sodium_init.hpp` doing exactly this is in the #230 discussion.
- Watch for `std::string_view{}.data()` being null on empty input - libsodium marks buffer parameters `nonnull`, and hashing `""` is a published vector, so route empty views through a non-null pointer.

## Behaviour to preserve - the existing tests are the spec

**No test in `encoding_test.cpp`, `pkce_test.cpp` or `script_engine_test.cpp` should need its expectations changed.** They are known-answer vectors (SHA-256 FIPS 180-4, HMAC RFC 4231 cases 1/2/3/6, base64 RFC 4648 §10 both directions, PKCE RFC 7636 Appendix B), so they are exactly the harness that proves this swap is behaviour-preserving. If one goes red, the replacement is wrong - do not adjust the vector.

Two places where libsodium may legitimately differ, to be checked and then decided deliberately:

1. **Unpadded base64.** `sodium_base64_VARIANT_ORIGINAL` requires padding, but `Encoding.Base64DecodeAcceptsUnpaddedAndWrappedInput` pins that `"Zg"` decodes to `"f"` (web `atob` accepts unpadded). Likely fix: try `ORIGINAL`, fall back to `ORIGINAL_NO_PADDING`.
2. **Trailing-bit strictness.** `Encoding.Base64DecodeRejectsMalformedInput` pins that `"Zm9vYmF"` is rejected because its trailing bits are non-zero. Confirm libsodium agrees; if it is more permissive, that is a real behaviour change to `atob` and belongs in the PR description rather than a silently edited test.

## Docs (same commit)

- `CLAUDE.md:22` - the vendor list still reads `quickjs-ng, picosha2 (PKCE/SHA-256), hdrhistogram`.
- `docs/engine/architecture.md` - the Script Engine section names picosha2 and states "no new dependency, and OpenSSL is still not linked"; both need rewriting.
- `docs/engine/building.md` - vcpkg dependency list, if it enumerates them.
- `engine/vendor/picosha2/README.md` goes with the directory.

## Sequencing

**After #230 merges.** It rewrites the same two headers and adds the tests this issue is validated by; doing it first would mean rebasing the vector suite onto itself. Nothing else currently open touches `utils/encoding.hpp` or `utils/sha256.hpp`.

Independent of the `pm.*` series (#180, #187, #188) in file terms - this is the layer underneath them.

## A note for whoever picks this up

If you are working in a sandboxed environment, check you can actually fetch the dependency before starting: vcpkg downloads libsodium from `github.com/jedisct1/libsodium/archive/…`, which some egress policies block with a 403 (this is what stopped it landing in #230 directly). CI runners are unrestricted. `apt-get install libsodium-dev` gives the same API for local development if the vcpkg fetch is unavailable, but note that then the CMake wiring itself is only exercised in CI.

## Acceptance criteria

- [ ] `libsodium` is declared in `engine/vcpkg.json` and linked to `vayu_core`; `engine/vendor/picosha2/` no longer exists and nothing references it.
- [ ] `sha256`, `hmac_sha256`, `base64_encode`, `base64url_encode`, `base64_decode` and `hex_encode` contain no hand-written hashing or alphabet tables.
- [ ] The full existing vector suite passes **unmodified** - or any deliberate behaviour change is called out in the PR with its reasoning, not absorbed into an edited expectation.
- [ ] Engine, app and C++ test jobs are green on ubuntu, macOS **and** Windows - Windows is the load-bearing one, since it is where OpenSSL was absent and where libsodium is genuinely new.
- [ ] `CLAUDE.md` and `docs/engine/architecture.md` no longer describe a vendored picosha2 or claim the engine links no crypto library.

athrvk:
## Description

**Plan.** The root cause is scope creep in a trade that was once fine: one hand-rolled base64 for a Basic auth header cost less than a dependency, but #230 pushed the surface to a hash, an HMAC, three encoders and a decoder, plus a vendored third-party hash - a maintenance surface that belongs to a library. The approach is the issue's: declare `libsodium` in `engine/vcpkg.json`, reimplement the six functions in `utils/sha256.hpp` and `utils/encoding.hpp` over `crypto_hash_sha256`, `crypto_auth_hmacsha256_*`, `sodium_bin2base64`, `sodium_base642bin` and `sodium_bin2hex`, and delete `engine/vendor/picosha2`. **The alternative I rejected is keeping the wrappers and reaching for OpenSSL**, for the two reasons the issue already established and which I re-checked rather than re-litigated: it is absent from the Windows build (vcpkg's curl selects Schannel there), and `EVP_DecodeBlock` reports the padded length, so `atob('Zg==')` would need the caller to count `=` itself - adopting it would keep the hand-written layer *and* add a heavy dependency.

I verified the problem still exists as described on current master before starting. Two drifts from the issue's file list, adapted rather than followed:

- The vendor list it cites as `CLAUDE.md:22` has since moved to **`engine/CLAUDE.md:14`** (the root file was restructured by #259). Updated there.
- `.github/workflows/codeql.yml` also named picosha2 in its `paths-ignore` rationale; the issue did not list it. Updated, since it is a comment describing a directory this PR deletes.

**What landed:**

- `utils/sha256.hpp` - `sha256` over `crypto_hash_sha256`; `hmac_sha256` over `crypto_auth_hmacsha256_init/update/final`. `_init` takes an arbitrary key length, so RFC 2104's key schedule (hash keys over 64 bytes down first, zero-pad shorter ones) is now libsodium's to get right. libsodium's own test is `keylen > 64`, matching the boundary `HmacSha256.KeyLengthsAroundTheBlockBoundary` pins.
- `utils/encoding.hpp` - `base64_encode` / `base64url_encode` share one `sodium_bin2base64` body parameterised by variant, so PKCE's URL-safe-no-padding format comes straight out of the library instead of being post-processed from the standard alphabet. `base64_decode` over `sodium_base642bin`, `hex_encode` over `sodium_bin2hex`. `url_encode` / `url_decode` / `form_encode` are untouched, as scoped.
- `utils/sodium_init.hpp` (new) - the one-time `sodium_init()` and the nonnull guard.
- `engine/vendor/picosha2/` deleted, along with its SYSTEM include block. **Net: a vendored dependency removed, ~200 lines of hand-written hashing and alphabet tables gone, one library declared.**

**Three design points worth a reviewer's eye:**

1. **The two libsodium divergences the issue flagged both resolved in our favour, and I checked rather than assumed.** Unpadded input: `VARIANT_ORIGINAL` requires the `=`, so a failed padded decode is retried with `ORIGINAL_NO_PADDING`, which is what keeps `atob('Zg')` = `"f"`. The fallback cannot launder a malformed group, because `=` is outside the no-padding alphabet - `"Zg=v"` and `"Zm9v="` fail both attempts. Trailing-bit strictness: libsodium rejects non-zero leftover bits itself, so `atob('Zm9vYmF')` still throws and **no behaviour change reaches the sandbox**. Passing `b64_end = nullptr` makes libsodium reject input it did not consume in full, so trailing junk after a valid prefix is an error rather than a short decode.
2. **The init guard is a function-local static, not `std::call_once`.** The issue preferred `call_once`; a magic static carries the identical guarantee (initialiser runs once, concurrent callers block until it completes) with less machinery, and it lives in the wrappers rather than at each entry point, so `vayu-engine`, `vayu-cli`, `vayu_tests` and the next entry point cannot forget. A failed init throws rather than returning a status a caller can drop - a libsodium that could not initialise makes every digest after it untrustworthy.
3. **Empty input reaches libsodium as a valid pointer.** libsodium marks its buffer parameters `nonnull` and an empty `std::string_view` may carry a null `data()`; hashing and encoding `""` are both published vectors, so `detail::sodium_bytes` routes empty views through a pointer to a zero byte. Verified under ASan/UBSan, not just by the vectors passing.

**Blast radius traced.** All six functions have live callers and every one still compiles and passes: PKCE (`http/pkce.hpp` - `sha256`, `base64url_encode`), OAuth 2.0 client auth (`http/oauth_client.cpp:112`), Basic auth (`http/auth_resolver.cpp:168`), and the script sandbox (`runtime/script_engine.cpp` - all six, via `pm.crypto`, `btoa` and `atob`). The link is `PUBLIC` on `vayu_core` because both headers include `<sodium.h>` in their interface. **No app changes**: `rg "picosha|sha256|base64" app/src app/electron` finds no reference to the engine's C++ helpers - the sandbox surface the app knows about is `pm.crypto`, whose behaviour is unchanged.

**One deliberate non-change:** `.github/release-notes/v0.13.0.md` still says picosha2 backs `pm.crypto`. That is a record of what shipped in v0.13.0, and it was accurate then; rewriting a published release note to match today's tree would falsify it. The swap belongs in the next release's notes instead.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update
- [x] Dependency change / refactor

## Testing

- `cd engine && ctest --preset linux-dev --output-on-failure` - **760 passed, 0 failed**. No test count change and **no expectation edited**: the vector suite is the spec for this swap, and it is byte-identical to master's. The 35 tests that directly cover the swapped functions (`Encoding.*`, `Sha256.*`, `HmacSha256.*`, `Base64Url.*`, `Pkce.*`, plus the sandbox's `pm.crypto` / `btoa` / `atob` cases) all pass unmodified, including the two divergence risks - `Base64DecodeAcceptsUnpaddedAndWrappedInput` and `Base64DecodeRejectsMalformedInput`.
- The encoding and PKCE vectors additionally compiled and run under `-fsanitize=address,undefined` and under `-Wconversion -Wsign-conversion -Werror`, both clean. The `-Werror` run is why `sodium_base64_ENCODED_LEN` is handed an unsigned variant: the macro mixes its argument into unsigned arithmetic, and the plain `int` constant warns - which would be an **error** on the MSVC `/W4 /WX` build rather than a warning.
- No app change, so no app suite run: no TypeScript test could change colour.
- No pre-existing failures observed on the base commit.

**Mutation-checked** (applied the mutation, rebuilt, confirmed red, reverted):

| Mutation | Result |
|---|---|
| Drop the `ORIGINAL_NO_PADDING` retry from `base64_decode` | `Encoding.Base64DecodeAcceptsUnpaddedAndWrappedInput` fails |
| `base64url_encode` uses `VARIANT_URLSAFE` (padded) instead of `..._NO_PADDING` | 3 fail: `Base64Url.NoPaddingUrlSafeAlphabet`, `Pkce.Rfc7636AppendixBChallenge`, `Pkce.RandomTokenLengthAndAlphabet` |

The second is the one that matters: it confirms the PKCE challenge format is pinned by a published vector rather than by the encoder agreeing with itself.

**How the vcpkg wiring was verified, and how it was not.** This sandbox's egress blocks `github.com/jedisct1/libsodium/archive/…` with a 403 - exactly the case the issue's closing note anticipated - so `vcpkg install libsodium` fails here. Everything above was therefore built and run against the **same libsodium API** from `apt` (1.0.18; the API is unchanged in vcpkg's 1.0.22), with `CMakeLists.txt` temporarily pointed at the system library. The committed wiring - `libsodium` in `vcpkg.json`, `find_package(unofficial-sodium CONFIG REQUIRED)`, `unofficial-sodium::sodium` - is restored in the commit but is **first exercised by CI**, which is unrestricted. The Windows job is the load-bearing one, since that is where libsodium is genuinely new. Flagging it plainly rather than implying a green local run covered it.

## Checklist
- [x] I have added tests (if applicable - none added deliberately; the existing known-answer vectors are the spec this swap is validated against, and adding to them would obscure that they passed unchanged)

Docs updated in the same commit: `engine/CLAUDE.md` (vendor list), `docs/engine/architecture.md` (the Script Engine signing bullet, the PKCE hashing line, and the dependency list), `docs/engine/building.md` (the vcpkg dependency table and the manual `vcpkg install` line), `.github/workflows/codeql.yml` (the vendored-directory comment). No page was added, moved or renamed and no link or anchor changed, so `mkdocs --strict` has nothing new to resolve (mkdocs is not installed in this environment; the docs CI job gates it).

## Related Issues

Closes #232

Follows #230, which added the `pm.crypto` surface and the vector suite this swap is validated by. Independent of the `pm.*` series (#180, #188) in file terms.

---
_Generated by [Claude Code](https://claude.ai/code/session_01B4FgZrNA2njkoKPXWFjTg8)_


athrvk:
The first run went red on ubuntu and macOS, both for the same reason, and it is worth flagging because it adds a prerequisite the issue did not anticipate.

vcpkg has no prebuilt libsodium: on Linux and macOS it builds the port from source through the **autotools** path, which runs `autoreconf` before configure. Neither runner carries `autoconf-archive`, so `vcpkg install` failed before a single Vayu source file compiled:

```
libsodium currently requires the following programs from the system package manager:
    autoconf autoconf-archive automake libtoolize
```

This is not a code failure - it is upstream of compilation entirely, which is why the ubuntu job showed 10 compile requests and the macOS job's 912 were the *other* dependencies (curl, openssl), not libsodium.

Pushed `948ca42`, which installs `autoconf`, `autoconf-archive`, `automake` and `libtool` in every workflow that builds the engine (PR matrix, release, CodeQL) and in `build.py --setup`. Windows needs nothing - the port builds libsodium with MSBuild there. brew installs `glibtoolize` rather than `libtoolize`, which vcpkg accepts, so the macOS check looks for that name.

**The part that needs your judgement rather than mine:** this makes autotools a real prerequisite for every Linux and macOS contributor, not just CI. `python build.py --setup` installs them, and it is documented in `docs/engine/building.md` and the root prerequisites line - the failure mode reads nothing like a missing build tool, so it needed writing down. If that trade is not one you want for this dependency, say so and I will close this rather than leave it half-landed.

---

Issue #88, opened 2026-07-23
engine(db): never sleep holding the global DB mutex; delete dead transaction helpers; deduplicate the busy-retry machinery


athrvk:
## Problem

Three related defects in the `Database` layer's concurrency posture (`engine/src/db/database.cpp` / `engine/include/vayu/db/database.hpp`):

### 1. Retry loops sleep while holding the global mutex

Every `Database` method serializes on one `std::recursive_mutex`. Four methods contain hand-rolled SQLITE_BUSY retry loops that **sleep inside that lock**:

| Method | Location | Worst-case sleep under lock |
|---|---|---|
| `update_run_status_with_retry` | `database.cpp` ~720-752 | 100+200ms = 300ms (3 attempts) |
| `add_metric` | ~771-801 | 50+100ms = 150ms |
| `add_metrics_batch` | ~805-848 | 100+200+300+400ms = 1s (5 attempts) |
| `add_results_batch` | ~889-932 | 1s |

If BUSY ever fires during a load run, the "recovery" stalls **every** endpoint - `/health`, live SSE ticks, the UI's 5s runs poll - for up to a second, because they all need the same mutex. The cure is worse than the disease.

For context on how reachable BUSY is (verified): the lock file guarantees a single engine process (`docs/lock-file-handling.md`), `daemon.cpp:152` creates the only `Database` instance, all access is serialized by the mutex, and `init()` sets `busy_timeout` to 10s. So in-process BUSY should be effectively unreachable; the realistic remaining source is an **external writer** on `vayu.db` (a user's sqlite3 CLI, backup tooling). Keep a retry as belt-and-braces, but it must not hold the lock while sleeping.

### 2. The public transaction helpers are a dead footgun

`begin_transaction()` / `commit_transaction()` / `rollback_transaction()` (`database.hpp:76-78`, `database.cpp:953-966` as of PR #92's merge) each take the mutex **only for the duration of the single call**, so any transaction spanning them is interleaved by other threads' writes - the API cannot be used safely. Verified: **zero callers** anywhere in `engine/src`, `engine/include`, or `engine/tests`. Delete them (header + implementation). The safe pattern already in use - `impl_->storage.transaction([&]{...})` under the lock, as in `add_metrics_batch` - stays internal.

### 3. The retry machinery is four hand-rolled copies

[...]

PR #266, opened 2026-08-02
refactor(engine): delete the legacy EAV metrics read path

### LINKED ISSUE #177: engine(db): delete the legacy EAV metrics read path once pre-#146 runs have aged out
## Problem

PR #146 (#90) stops writing the EAV `metrics` table and moves the time series to `metric_ticks` + `runs.summary`. It deliberately **keeps** the table and both readers so runs recorded before the upgrade still render - exactly as #90 specified ("kept intact so pre-upgrade runs still render charts and reports. Mark the legacy path clearly for deletion in a future minor once old histories age out"). Three code comments and `docs/engine/db-schema.md` all say the path is deletable later, but nothing tracks the deletion, so "later" is nobody's job. This issue is that job.

**The path self-obsoletes on a clock nobody is watching.** Retention defaults are `maxRunsRetained` 200 / `runRetentionDays` 30 (`database.cpp:919-921`, `prune_runs`), and `prune_runs_configured()` runs at startup and after every terminal run. A user who upgrades therefore stops having any run with EAV rows within ~30 days of normal use - after which every line below is unreachable in practice while still being compiled, tested and maintained.

**What is dead once the last legacy run is pruned** (line numbers against PR #146's head, `30ec26d` - re-verify after it merges):

| Site | What |
|---|---|
| `engine/src/db/database.cpp:279` | the `metrics` table in `make_storage`, and its index at `:195` |
| `engine/src/db/database.cpp:929-982` | `add_metric`, `add_metrics_batch`, `get_metrics`, `get_metrics_since`, `get_metrics_paginated`, `count_metrics` - the writers already have **zero** callers in `src/` |
| `engine/include/vayu/db/database.hpp:120-125` | the matching declarations |
| `engine/include/vayu/types.hpp:548-583` | `enum class MetricName` (30 values) plus `to_string` (`:586`) and `parse_metric_name` (`:623`) - the wide payload uses plain JSON keys, so the enum has no other reader |
| `engine/include/vayu/types.hpp:760-767` | `struct Metric` |
| `engine/src/http/routes/metrics.cpp:107-215` | the legacy bucket-grouping branch of `run_time_series_response`, reached only when `count_metric_ticks == 0` |
| `engine/src/http/routes/metrics.cpp:422-517` | the EAV fallback arm of the deprecated `/stats/:runId` SSE loop, including the now-unwritten `MetricName::Completed` completion check (new runs already terminate via the run-status branch) |
| `engine/src/http/routes/runs.cpp:182-249, 410` | `apply_legacy_metrics` and its call site in `run_report_response` |

That is ~250 lines of engine code plus a table, kept alive for data that deletes itself.

**Removing it also retires a documented trap.** `apply_legacy_metrics` is the last home of the labels-based percentile disambiguation - `!m.labels.empty()` guards `p50/p95/p99` but not `p75/p90/p999/min/max/avg`, the exact fragility #90 was opened over. It cannot bite a summary-backed report (pinned by `RunsRouteTest.StoredSummaryWinsOverStrayPerTickPercentileRows`), but while the branch exists the trap is still in the tree and still has to be read and understood by anyone touching the report route.

## Fix / solution pathway

One PR, all of it deletion:

1. Drop the `metrics` table and `idx_metrics_run_id` from `make_storage`. `sync_schema()` is called without `preserve`, so removing a table is handled - **but confirm** it drops rather than silently ignores, and decide whether an explicit one-shot `DROP TABLE IF EXISTS metrics` in `Database::init` is wanted so existing DB files actually reclaim the space (a user who upgraded with 12k rows/run of history should not keep paying for it).
2. Delete the six `Database` metric methods, their declarations, `struct Metric`, `enum class MetricName`, `to_string(MetricName)` and `parse_metric_name`.
3. Delete the legacy branch in `run_time_series_response` - the tick path becomes unconditional, so `count_metric_ticks` is just the pagination total and the `tick_count > 0` fork disappears.
4. Delete `apply_legacy_metrics`; `run_report_response` calls `apply_run_summary` and, when it returns false, reports from the sampled `results` alone (state that behaviour explicitly - a run with no summary now means "engine died before completion", not "old engine").
5. Delete the EAV arm of the `/stats/:runId` SSE loop, leaving the tick poll plus the run-status completion check.
6. Remove the `metrics` cascade line from `remove_run_cascade_locked` (`database.cpp:826`).

**Do this only after the deprecation window.** The gate is a release boundary, not a date: the deletion must ship at least one minor after the release carrying #146, so anyone upgrading across two versions at once still has their history pruned by then. A DB that predates the window loses its old charts/reports (the runs themselves survive) - acceptable for history that retention was going to delete anyway, but it is the one user-visible cost and belongs in the release notes.

## What NOT to change

- `metric_ticks`, `runs.summary`, and the response shapes of `GET /runs/:id/metrics` and `/runs/:id/report` - unchanged, and the same "no app source changes" acceptance test applies.
- The `results` table and the live SSE topic (`/runs/:id/live`).
- The `/stats/:runId` endpoint itself - it stays a deprecated alias; only its EAV arm goes.

None expected. Verify and state it in the PR: `rg "metric_name|metricName" app/src app/electron` (expected empty - the app has never seen the EAV shape, which is why #146 needed no app changes).

## Tests

Deletion, not addition - the point is that the suite shrinks:

- `engine/tests/stats_route_test.cpp` - drop the legacy-only cases (`SeededRunGroupsIntoTickBuckets`, `HonorsLimitAndOffsetForPagination`, `RequestsFailedIsDerivedRegardlessOfRowOrderWithinATick`) and the legacy halves of the two comparison tests (`StoredTickMatchesTheLegacyBucketByteForByte`, `StoredTickAndLegacyRowsAgreeOnTheFailedCount`). **The byte-for-byte contract lock is the one real loss** - once there is no second implementation to compare against, replace it with a fixture pinning `build_metric_tick_payload`'s exact key set and value types, so a dropped or retyped field still fails loudly.
- `engine/tests/runs_route_test.cpp` - drop `ReportFallsBackToLegacyMetricRows` and `StoredSummaryWinsOverStrayPerTickPercentileRows`; rewrite `MalformedSummaryFallsBackToLegacyMetricRows` to assert the new no-summary behaviour (report from sampled results, not an exception).
- `engine/tests/db_test.cpp` - drop the `Metric` round-trip and pagination cases and the `get_metrics` assertions inside the prune/cascade tests (the `count_metric_ticks` assertions beside them already cover the cascade). Remove `idx_metrics_run_id` from `EXPECTED_INDEXES`.
- `engine/tests/db_concurrency_test.cpp` - this file is built entirely on `add_metrics_batch`/`get_metrics`/`count_metrics` as its representative write path. **Retarget it to `add_metric_tick`/`get_metric_ticks_since`/`count_metric_ticks` rather than deleting it** - the mutex/retry contract it guards (#88) is still live and would otherwise lose its only test.
- Add one case asserting a DB carrying an old `metrics` table opens and serves normally after the drop (upgrade path).

## Docs (same commit)

- `docs/engine/db-schema.md` - remove the `metrics (legacy, read-only)` section and the two-flavour percentile explanation; drop `metrics` from the retention/cascade prose and the index list.
- `docs/engine/api-reference.md` - drop the "runs recorded before `metric_ticks`" sentences under `GET /runs/:runId/metrics` and `/runs/:runId/report`, and the legacy-fallback note under `GET /stats/:runId`.
- `docs/engine/architecture.md` - only if it still names the EAV write path.
- Release notes: a `### Removed` bullet naming the lost pre-upgrade history, per the Keep a Changelog format in CLAUDE.md.

## Sequencing

**Blocked on PR #146 (#90) merging** - the legacy path does not become legacy until it does; re-verify every line reference above against master afterwards.

Coordinate with #112, which independently proposes dropping the `TestsValidating` metric: #146 already stopped writing it, and this issue removes the whole `MetricName` enum including it. Whichever lands second drops that item as already done - and if #112 lands first, it only removes the enumerator, leaving the rest of the enum to this issue.

Nothing else depends on this. It is pure debt paydown with a release-boundary gate, so it can sit until the window passes.

## Acceptance criteria

- No `metrics` table, `Metric` struct, or `MetricName` enum anywhere in the engine; `rg "MetricName|add_metrics_batch|count_metrics" engine/` is empty.
- `GET /runs/:id/metrics`, `/runs/:id/report` and `/stats/:id` behave identically for runs recorded since #146; a run with no summary reports from its sampled results rather than erroring.
- An existing DB file with a populated `metrics` table opens, drops it, and serves normally.
- `metric_ticks` payload keys are pinned by a fixture that replaces the retired contract-lock test.
- `db_concurrency_test.cpp` still guards the write-path mutex/retry contract, retargeted rather than deleted.
- Engine suite green; docs updated in the same commit; clang-tidy clean; no em-dashes (use ` - `).

athrvk:
## Description

**Plan.** The root cause is that PR #146 moved the time series to `metric_ticks` + `runs.summary` but kept the EAV `metrics` table and both its readers so pre-upgrade runs still rendered - and nothing tracked the deletion, so ~250 lines stayed compiled, tested and maintained for rows that retention deletes on its own (`maxRunsRetained` 200 / `runRetentionDays` 30, pruned at startup and after every terminal run). The approach is pure deletion: drop the table, the six `Database` metric methods, `struct Metric`, `enum class MetricName` and its two converters, the legacy branch of `run_time_series_response`, the EAV arm of the `/stats/:runId` SSE loop, and `apply_legacy_metrics`. The alternative I rejected was keeping the read path behind a config flag - that trades a dead branch for a dead branch plus a knob, and the issue's own gate is a release boundary, not a preference.

I verified the problem still exists as described. Every line reference in the issue was re-checked against master `6d4bbbc`; the only drift was line numbers, and one item the issue did not list - `vayu::json::serialize(const db::Metric&)` in `utils/json.cpp` plus its `json_test.cpp` case - which had to go with the struct.

**Two decisions the issue left open, and how I settled them:**

1. **`DROP TABLE IF EXISTS metrics` in `Database::init()` - yes.** I confirmed `sync_schema()` only syncs the tables the storage still *declares*; it does not drop one that was removed from it. Without the explicit drop, every upgraded database keeps the table and its rows forever. The call is idempotent, so it needs no one-shot migration flag.
2. **No `VACUUM`.** The drop returns the pages to SQLite's freelist for reuse, but the file does not shrink. A `VACUUM` would rewrite the whole database under a write lock, and startup is the worst moment to pay that. Documented in `db-schema.md` rather than left implicit.

**Behaviour change worth naming:** a run with no usable summary now reports from its sampled `results` alone. That state used to mean "old engine"; it now means only "the engine died before the run reached a terminal status". `apply_run_summary` returns `void` instead of `bool`, since no caller branches on it any more (a retained-but-unread return is this repo's most repeated defect).

**Deliberate deviation from the issue spec:** the issue asks for a release-notes `### Removed` bullet in the same commit. Release notes here are authored wholesale in the `chore(release): x.y.z` commit (confirmed: `v0.14.0.md` landed in a single release commit, and the `releasing` skill writes `.github/release-notes/vX.Y.Z.md` at bump time). Adding a speculative `v0.15.0.md` would preempt that process and collide with the next bump, so I did not. **The next release needs this bullet:** databases that predate the deprecation window lose their pre-upgrade charts and reports; the runs themselves survive.

**Timing, for your call - please read before merging.** The issue's gate is "at least one minor after the release carrying #146". #146 (`7d3d2c7`, 2026-07-29) shipped in **0.13.0** (2026-07-31), and **0.14.0** has since shipped, so this landing in 0.15.0 satisfies the stated gate with a margin. But the gate's *stated reason* is that ~30 days of retention age pre-#146 runs out, and 0.13.0 is three days old - a user who upgraded this week still has EAV history that this change discards. The literal rule is met; the intent behind it is not yet. Holding this until the retention window has actually elapsed is entirely reasonable, and the branch will keep.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update
- [x] Technical debt removal (deletion only; no user-facing feature change)

## Testing

Ran the issue's stated engine commands on Linux:

```
python build.py -e -t
ctest --preset linux-dev --output-on-failure
```

- **749/749 engine tests pass**, no failures, no skips. No pre-existing failures to report - the suite was green on the base commit too.
- **clang-tidy clean** on all four changed sources (`db/database.cpp`, `http/routes/metrics.cpp`, `http/routes/runs.cpp`, `utils/json.cpp`), via the `scripts/pre-commit` invocation (`-p engine/build`).
- **`mkdocs build --strict` passes** (both changed docs are under `docs/`).
- **clang-format**: only `engine/tests/db_concurrency_test.cpp` was format-clean at HEAD and would not have stayed so; it is the only file formatted. The others were already non-clean before this change and were left alone, per CLAUDE.md.
- **No em-dashes** in any changed file.

**App suite not run, deliberately.** No file under `app/` changed, so no app test could go from green to red (CLAUDE.md: "if no test could possibly go from green to red, do not run tests"). CI runs the full matrix regardless.

**Mutation-checked**, not just written: reverting `drop_table_if_exists("metrics")` makes `DatabaseTest.DropsTheLegacyMetricsTableFromAPreUpgradeDatabase` fail, and restoring it makes it pass. That test also guards its own guard - it asserts the legacy table really was recreated before asserting it was dropped, so it cannot pass against a database that never had one.

### Test changes

The suite shrinks, which is the point - but two things were preserved rather than deleted:

- **`db_concurrency_test.cpp` is retargeted, not removed.** It was built entirely on `add_metrics_batch`/`get_metrics`/`count_metrics`, but the #88 mutex/retry contract it guards is still live. It now drives `add_metric_tick` / `get_metric_ticks_since` / `count_metric_ticks` as the issue directs, **plus** `add_results_batch` - because `add_metric_tick` is a bare retried insert and the retired `add_metrics_batch` was a *transaction inside* the retry. Dropping to ticks alone would have quietly lost coverage of that second shape.
- **The byte-for-byte contract lock is replaced, not just deleted.** `StoredTickMatchesTheLegacyBucketByteForByte` compared two independent implementations of one object; with the EAV reader gone there is no second implementation to compare against. `StoredTickPayloadKeysAndTypesArePinned` pins `build_metric_tick_payload`'s exact key set (count asserted, so an *added* field fails too) and each key's JSON value type, plus a value round-trip.

Dropped as legacy-only: `SeededRunGroupsIntoTickBuckets`, `HonorsLimitAndOffsetForPagination`, `TicksWinOverLegacyRowsForTheSameRun`, `ReportFallsBackToLegacyMetricRows`, `StoredSummaryWinsOverStrayPerTickPercentileRows`, the `Metric` round-trip/pagination/serialization cases, and the two `MetricName` enum round-trips. The three `RequestsFailedIs*` cases went too - they exercised the read-time `error_rate% x requests_completed` derivation that only the EAV branch performed; the tick path carries the producer's own count, which `StoredTickCarriesTheProducersFailedCount` now asserts. `MalformedSummaryFallsBackToLegacyMetricRows` is rewritten as `MalformedSummaryReportsFromSampledResults`, joined by `RunWithNoSummaryReportsFromSampledResults` for the no-summary case.

## Checklist

## App changes

**None - verified empty, as the issue asked.** `rg "metric_name|metricName" app/src app/electron` returns nothing (exit 1). The app never saw the EAV shape, which is why #146 needed no app changes either. `GET /runs/:id/metrics`, `/runs/:id/report` and `/stats/:id` keep their response shapes.

## Docs

- `docs/engine/db-schema.md` - the `metrics (legacy, read-only)` section and the two-flavour percentile explanation are gone, along with the `idx_metrics_run_id` row and the cascade/retention mentions. A short note under `metric_ticks` records that the table was removed, why `init()` drops it explicitly, and why there is no `VACUUM`.
- `docs/engine/api-reference.md` - dropped the "runs recorded before `metric_ticks`" sentences under `GET /runs/:runId/metrics`, `/runs/:runId/report` and `GET /stats/:runId`, and rewrote the `requests_failed` paragraph, which described the deleted read-time derivation.
- `docs/engine/architecture.md` - checked; it names only the `metric_ticks` write path, so no change was needed.

## Acceptance criteria

- [x] No `metrics` table, `Metric` struct or `MetricName` enum anywhere in the engine - `rg "MetricName|add_metrics_batch|count_metrics" engine/` is empty
- [x] `GET /runs/:id/metrics`, `/runs/:id/report` and `/stats/:id` behave identically for runs recorded since #146; a run with no summary reports from its sampled results rather than erroring
- [x] An existing DB file with a populated `metrics` table opens, drops it, and serves normally
- [x] `metric_ticks` payload keys pinned by a fixture replacing the retired contract lock
- [x] `db_concurrency_test.cpp` still guards the write-path mutex/retry contract, retargeted rather than deleted
- [x] Engine suite green; docs updated in the same commit; clang-tidy clean; no em-dashes
- [ ] Release notes `### Removed` bullet - see the deviation above; belongs in the next release commit

## Related Issues
Closes #177

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lg1AW6Eu3GedVkFv9jANqA)_


athrvk:
merge latest master and fix the merge conflict

athrvk:
Merged `master` (14 commits) and resolved the conflict - pushed as `83acd1b`.

**The conflict, and how I resolved it.** One file: `engine/tests/runs_route_test.cpp`. #268 added `httpVersionDowngraded` assertions to `ReportFallsBackToLegacyMetricRows`, and this branch deletes that test along with the legacy read path it exercises. I kept the deletion. The assertion #268 added there was specifically that the *legacy* path reports `0` downgrades ("none recorded, not none happened"), which is moot once that path no longer exists.

**The new field does not lose coverage.** `ReportReadsTheStoredSummary` came across from master intact and round-trips it on the summary path - sets `http_version_downgraded = 4` on the way in, asserts `httpVersionDowngraded == 4` on the way out. That is the path every run recorded since #146 actually takes, and it is where the field is read (`runs.cpp` `read_number (summary, "http_version_downgraded", ...)`). Everything else auto-merged.

Re-verified after the merge rather than assuming: `rg "MetricName|db::Metric|get_metrics|count_metrics|add_metrics_batch|apply_legacy_metrics" engine/` is still empty, and the `DROP TABLE IF EXISTS metrics` in `Database::init()` is intact - now sitting next to master's new retired-config sweep for `requestBatchSize`/`contextPoolSize`, which is the same shed-it-on-upgrade idea applied to config rows.

**One thing you should know about the verification on this push.** I could not run the engine suite locally. #265 made libsodium a build dependency, and this sandbox's egress policy returns `403` for `https://github.com/jedisct1/libsodium/archive/1.0.22-RELEASE.tar.gz`, so `vcpkg install` fails before anything compiles. That is an environment limit, not a repo problem - CI builds it fine (that is what the autotools install in the workflows is for). What I could verify locally, I did: all eight changed engine sources and tests pass `g++ -std=c++20 -fsyntax-only` against the merged headers, so the conflict resolution is not just textually clean but actually compiles. The 749/749 suite run and clang-tidy pass reported above stand for the pre-merge commit; **CI on this merge commit is the authority for the merged tree**, and I am watching it.

---

Issue #301, opened 2026-08-04
engine(scripting): `pm.cookies` needs a cookie jar the engine does not have (`pm.response.cookies` has shipped)


athrvk:
Split out of #188 (item 2). Anchors originally verified at `6b5cc15`; **body refreshed 2026-08-04 at `10c8193`** after step 1 shipped. **Closed 2026-08-06: the jar and the read half shipped via PR #336 (merged as `f703eaa`), closure-verified at acceptance-criteria level. The write half (`pm.cookies.set/unset/clear`) is #337, filed as this issue's Fix pathway prescribed.**

> **Status.** Step 1 (`pm.response.cookies`, read-only) shipped via PR #305: the parser lives in `engine/include/vayu/http/set_cookie.hpp`, pinned by `engine/tests/set_cookie_test.cpp` plus the cross-language fixture `engine/tests/fixtures/set-cookie-conformance.json` read by both the gtest and the renderer's `parse-set-cookie.conformance.test.ts` (the fixture route this issue asked for). PR #306 additionally fixed repeated response headers collapsing last-wins (#307), so multiple `Set-Cookie` headers all reach both parsers. `pm.response.cookies` is advertised in completions (`engine/src/http/routes/scripting.cpp:144-177`) and documented as supported (`docs/app/pm-api-compatibility.md:22`, `:281-287`). **Step 2 (the jar + `pm.cookies.get/has/toObject`) shipped via PR #336**: per-environment in-memory jar of libcurl Netscape lines behind one mutex (`engine/include/vayu/http/cookie_jar.hpp`), `GET`/`DELETE /cookies`, the Settings Cookies card, `pm.sendRequest` sharing stated and tested, and the load path untouched by design.

## Problem (remaining)

`pm.cookies` - the jar API (`get/has/toObject`, then the write half) backed by a cookie jar the engine keeps across requests. libcurl has one; nothing surfaces it. The response the script sees carries `Set-Cookie` in `Response::headers` (`engine/include/vayu/types.hpp:244`) and can now be *read* per-response, but nothing persists a session across requests, so "log in once, reuse the session" is impossible.

## The decisions a jar forces

None of these are plumbing, which is why this issue is sized above #300 (`pm.info`):

- **Scope and lifetime.** Per what - process, environment, collection, run? A jar shared across environments leaks a session cookie from staging into production. A jar per run makes "log in once, reuse the session" impossible, which is the use case that justifies the feature.
[...]

Issue #300, opened 2026-08-04
engine(scripting): `pm.info` is absent - no `requestName`, `requestId`, `eventName` or iteration in a script


athrvk:
Split out of #188 (item 3), which scoped four missing capabilities and said each should graduate to its own anchored issue when picked up. This is the first of the four and the one #188 called "the cheapest".

Anchors verified at master `6b5cc15` (2026-08-04).

## Problem

`pm.info` is `undefined`. `setup_pm_object` (`engine/src/runtime/script_engine.cpp:3411-3448`) binds exactly `test`, `expect`, `response`, `request`, the three variable scopes, `variables` and `crypto` - nothing else. A script has no way to ask what request it is attached to or which hook it is running in.

Postman's `pm.info` carries five fields; here is what each would cost, verified rather than assumed:

| Field | Where the value already lives | Gap |
|---|---|---|
| `requestId` | `Run::request_id` (`engine/include/vayu/types.hpp:696`), set on `/execute` from the payload's `requestId` (`engine/src/http/routes/execution.cpp:605-607`) | plumbing only |
| `requestName` | `db::Request::name` (`engine/include/vayu/types.hpp:654`) - reachable by id via `Database::get_request` | **the payload never carries it** (see App changes) |
| `eventName` | `execute_prerequest` vs `execute_test` (`script_engine.cpp:3678` / `:3687`) already distinguish the hook | plumbing only |
| `iteration` / `iterationCount` | see below - **not** what Postman means | design decision |

**The iteration pair is the one that needs a decision, not plumbing.** Vayu's load runs do not execute a test script per iteration. `validate_scripts` (`engine/src/core/run_manager.cpp:47-160`) runs the `tests` script **once per sampled response, after the run has finished** (called at `run_manager.cpp:705-709`), against a `dummy_request` rebuilt from the run config (`:90-95`) - there is no pre-request script on the load hot path at all. So there is no per-iteration script execution to number. Reporting the sample index as `iteration` would be a plausible-looking lie: samples are a reservoir, not the first N iterations. The honest options are (a) omit both fields until a collection runner exists (#303), or (b) expose `iteration = 0`, `iterationCount = 1` for a design-mode `/execute` and leave them absent in a load run, which is at least checkable with `pm.info.iteration === undefined`.

Recommendation: **(a)**. #188's own rule applies - "a binding that cannot fail is worse than a missing one" - and a sample index dressed as an iteration number is exactly that.

## Fix / solution pathway

[...]

Issue #197, opened 2026-07-29
engine(perf): re-tune high-RPS defaults (backlog P1) and benchmark the post-#169 atomic histogram cost (backlog D8)


athrvk:
Base: master `ffafcd0` (2026-07-29); anchors re-verified 2026-08-06 at `f703eaa` - all eight file:line anchors exact, no drift. Tracks backlog entries **P1**, **D8**, and **D9** from `docs/plans/pending-backlog.md`, re-verified against current code; this issue owns refreshing that doc (see Docs).

## Problem

**Part 1 - the high-RPS defaults P1 documented are all still in place** (backlog's own 2026-06-05 investigation: tuned Vayu reaches ~45k req/s on loopback vs `wrk`'s ~51k, but the defaults collapse it to ~11-27k). Verified at base:

- `workers` still defaults to auto = `hardware_concurrency()` (`engine/src/http/event_loop/event_loop_impl.cpp:41`, seeded via `run_manager.cpp:627` `get_config_int("workers", 0)`). The investigation found throughput scales to ~ncpu-2/3 workers then **collapses at ncpu** (busy-poll workers starve the strategy producer and any co-located target).
- `MAX_PER_HOST = 200` (`engine/include/vayu/core/constants.hpp:71`) is still applied **per worker** (`event_loop_worker.cpp:248` sets it on each worker's multi handle), so effective per-host connections = `workers x 200` - far past the measured sweet spot.
- `maxInFlight` still defaults to `max(target_rps x 10, 1000)` (`engine/src/core/load_strategy.cpp:341-342`) - effectively unbounded backlog growth (the investigation observed 390k queued) instead of load shedding.

**Part 2 - the D8 trade-off was silently reversed by PR #169 and never measured.** The backlog documents an explicit decision NOT to use the atomic histogram API because it "adds a CAS per sample on the 60k+ RPS hot path (works against P1)", with the instruction to pair any change with a hot-path benchmark. PR #169 (issue #129) switched recording to `hdr_record_value_atomic` / `hdr_interval_recorder_record_value_atomic` (`engine/src/core/metrics_collector.cpp:249-251`). That fix was **correct and necessary** - it cures a writer-vs-writer lost-increment race that D8's rationale (reader-side torn reads, benign on shipped arches) never covered - but the per-sample CAS cost D8 warned about is now on the hot path with no measurement. The alternative #129 itself listed (per-worker sharded histograms merged at tick/report time) trades that cost for merge complexity.

[...]

PR #306, opened 2026-08-04
feat(engine): read the response's Set-Cookie through pm.response.cookies

athrvk:
## Description

Step 1 of #301: `pm.response.cookies`, read-only, over the `Set-Cookie` header the response already carries. **The cookie jar (step 2) is deliberately not in this PR** - see Deviations.

**The plan.** Root cause: the header was already on `Response::headers` (`types.hpp:238`) and nothing surfaced it, so a script could not see a cookie a response set. The approach is an accessor over state the engine already holds, with the parse shared with the renderer's existing `parse-set-cookie.ts` through a conformance fixture rather than re-derived - a second C++ parser that did not know the two cases the first one documents (a comma inside `Expires=Wed, 21 Oct ...`, `=` padding in a base64 value) would show one value in the Cookies tab and assert another in a script, which is exactly the drift `variable-resolution-conformance.json` exists to prevent. The alternative I rejected was porting the regex into C++ and testing each side separately: that is two tables of cases that agree today and diverge on the next edit, and the fixture costs one file to make divergence a failing test instead.

**Surface.** `pm.response.cookies` is an array of `{ name, value, attrs }` in wire order with `get()` / `has()` / `toObject()` over it:

```javascript
pm.response.cookies.get('session');   // value, or undefined
pm.response.cookies.has('session');   // boolean
pm.response.cookies.toObject();       // { session: 'abc' }
pm.response.cookies[0].attrs;         // ['Path=/', 'HttpOnly'] - raw chunks
```

Three decisions worth reviewing, all documented and tested:

- **A self-replacing getter, not an eager property.** A load run executes a test script per completed request; eager parsing would charge every one of them for a header no script asked about. First read parses and caches, so `pm.response.cookies === pm.response.cookies` stays true.
- **Cookie names are case-sensitive** (RFC 6265), unlike header names - `get('SESSION')` does not answer the `session` cookie. Answering anyway would be a wrong value dressed as a right one.
- **A repeated name answers with its last value** from `get()` / `toObject()` (what a browser's jar would keep) while the array still lists both, because that is what came off the wire.

Entries are not `postman-collection` `Cookie` objects - no `key`, `path`, `secure`, `expires`, because the engine keeps no jar and those fields would restate the attribute string rather than name state anything holds. `attrs` is that string split on `;`. The divergence is written down in `pm-api-compatibility.md` rather than half-emulated.

Bad input fails loudly: `get(1)`, `has()` and `get('')` throw a `TypeError` naming the argument instead of reading as "no such cookie".

Completions gain the three members (`pm.response.cookies` / `.get` / `.has` / `.toObject`), which is also where the editor's `.d.ts` comes from since `script_types.cpp` derives it from that table, plus a guard that every offered cookie member is callable in the runtime - the `#182` drift, one level down.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Testing

Re-run on the merge commit, after master moved (#300 / PR #305 landed in the same three files):

- `cd engine && ctest --preset linux-dev --output-on-failure` - **831 / 831 pass**.
- `cd app && pnpm test` - **2557 / 2557 pass** in 289 files, including the 18 new conformance cases.
- `cd app && pnpm type-check`, `pnpm lint`, `prettier --check` on the new file - clean.

(The pre-merge run had one failure, `HttpClientTest.UnnegotiatedConnectionReportsEmptyHttpVersion`: this container's `HTTPS_PROXY` makes libcurl route `http://127.0.0.1:1/` through the proxy instead of failing fast with connection-refused, so it hits the 2 s timeout. It passed on the re-run, and passes with the proxy variables unset either way. Environmental, untouched by this diff.)

New tests (9 in the engine, 18 in the app):

- `set_cookie_test.cpp` - drives all 17 fixture cases, plus the boundary rule stated on its own so a later case cannot be "fixed" by loosening it back into the naive split. Both assert the fixture scanned something non-empty.
- `parse-set-cookie.conformance.test.ts` - the same fixture against the renderer's parser.
- `script_engine_test.cpp` - six over the script surface: the shape and both corruption cases end-to-end, an empty list without the header, the repeated-name rule, case-sensitivity, the throw on an unusable name, and list identity (which is the observable half of the caching getter - revert the cache and it goes red).
- `script_completions_test.cpp` - every offered cookie member is a function at run time.

Mutation-checked: reverting the boundary rule to `split(",")` fails the expiry cases on both sides; reverting the value slice to `split("=")` fails the base64 cases; dropping the reverse search fails the repeated-name test.

## Checklist
- [x] I have added tests

Docs in the same commit: `docs/engine/scripting.md` (the `pm` surface and the three rules above), `docs/app/pm-api-compatibility.md` (moved out of "Not (yet) supported" with the divergences from Postman spelled out; `pm.cookies.*` stays listed as absent, now saying the *jar* is what is missing).

- **Step 2 (the cookie jar) is not here, deliberately.** The issue says to consider stopping after step 1 if the jar's questions are unsettled, and they are: scope / lifetime / persistence / thread-safety have no answer yet, a jar is credential-grade state that should reuse whatever auth storage already does rather than open a new path, and the load path it would touch (`event_loop_worker.cpp`) is what #197 is currently re-tuning. Shipping a jar to make a deadline is exactly the stopgap `CLAUDE.md` says not to accept. `pm.cookies.*` therefore stays in "Not (yet) supported" and #301 stays open for it.
- **No app UI change**, per the issue: with only step 1 landing there is no jar to inspect or clear, so the app change is the doc line and nothing else. The MCP question (does a jar cross `run_request` calls) is a step-2 question and is untouched.
- **Anchor drift found while reading.** #301 says `pm-api-compatibility.md` no longer carries the unread-cookie-attributes defect - correct, `attrs` is rendered, and I did not go looking for it.
- **Adjacent defect, filed rather than fixed here**: `header_callback` does `headers[key] = value` in both HTTP paths, so a response sending **two** `Set-Cookie` headers keeps only the last - the parser handles the comma-joined form the renderer documents, but the engine never produces it from separate headers. That is a `Response::headers` semantics change with a blast radius across every header consumer, so it does not belong in this diff. Filed as #307.

## Related Issues

Refs #301 (step 1 of two; the jar keeps it open), part of #188.

### DIFF
--- a/app/src/modules/request-builder/components/ResponseViewer/parse-set-cookie.conformance.test.ts
+++ b/app/src/modules/request-builder/components/ResponseViewer/parse-set-cookie.conformance.test.ts
@@ -1,0 +1,61 @@
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * Cross-language conformance: this parser against the engine's copy
+ * (`engine/src/http/set_cookie.cpp`, behind `pm.response.cookies`), over the
+ * fixture both suites read.
+ *
+ * The two parse the same `Set-Cookie` off the same response - one for the
+ * Cookies tab, one for a script - so a case answered two ways is a user reading
+ * one thing on screen and asserting another in a test. Adding a case here fails
+ * whichever side does not handle it, the same arrangement
+ * `variable-resolution.conformance.test.ts` uses for `{{variable}}` resolution.
+ *
+ * Read from the engine tree on purpose: one file, two readers, no copy to drift.
+ */
+
+import { describe, it, expect } from "vitest";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+import { parseSetCookie, type ParsedCookie } from "./parse-set-cookie";
+
+const here = dirname(fileURLToPath(import.meta.url));
+const fixturePath = join(
+	here,
+	"..",
+	"..",
+	"..",
+	"..",
+	"..",
+	"..",
+	"engine",
+	"tests",
+	"fixtures",
+	"set-cookie-conformance.json"
+);
+
+interface ConformanceCase {
+	name: string;
+	header: string;
+	expected: ParsedCookie[];
+}
+
+const fixture = JSON.parse(readFileSync(fixturePath, "utf8")) as {
+	cases: ConformanceCase[];
+};
+
+describe("set-cookie conformance fixture", () => {
+	it("scanned a non-empty fixture (guards the scan itself)", () => {
+		expect(fixture.cases.length).toBeGreaterThan(10);
+	});
+
+	it.each(fixture.cases.map((c) => [c.name, c] as const))("%s", (_name, c) => {
+		expect(parseSetCookie(c.header)).toEqual(c.expected);
+	});
+});
--- a/docs/app/pm-api-compatibility.md
+++ b/docs/app/pm-api-compatibility.md
@@ -19,6 +19,7 @@ intent is that the most common Postman scripts paste in and run unchanged.
 | Core                | `pm`, `pm.test(name, fn)`, `pm.expect(value)`                                    |
 | Response            | `pm.response.code`, `.status`, `.responseTime`, `.headers`, `.json()`, `.text()`, `.reason()`, `.size()` |
 | Response headers    | `pm.response.headers.get(name)`, `.has(name)` - case-insensitive              |
+| Response cookies    | `pm.response.cookies` (array of `{ name, value, attrs }`), `.get(name)`, `.has(name)`, `.toObject()` - read-only, see below |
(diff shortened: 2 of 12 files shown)


athrvk:
fix merge conflict

athrvk:
Conflict resolved and pushed (`8e9c833`). #300 / PR #305 landed in the same three files; both conflicts were additive, so both sides are kept:

- `engine/CMakeLists.txt` - `script_info_test.cpp` and `set_cookie_test.cpp` both listed.
- `docs/app/pm-api-compatibility.md` - master's `pm.iterationData` / `pm.info.iteration` wording kept, next to the narrowed `pm.cookies.*` entry that now says the *jar* is what is missing and points at `pm.response.cookies` for the response half.
- `script_engine.cpp`, `scripting.cpp`, `scripting.md` and `script_engine_test.cpp` auto-merged; `setup_pm_info` and the cookies getter are both intact.

Re-verified on the merge commit: engine **831/831**, app **2557/2557**, type-check and lint clean. The proxy-bound `HttpClientTest.UnnegotiatedConnectionReportsEmptyHttpVersion` that failed on the first run passed this time - it was this container's `HTTPS_PROXY`, not the diff.

---

PR #308, opened 2026-08-04
fix(engine): keep every value of a repeated response header name

### LINKED ISSUE #307: engine(http): duplicate response headers are last-wins, so two `Set-Cookie` headers collapse to one
Found while implementing #301 step 1 (`pm.response.cookies`, PR #306). Filed rather than folded into that PR: the fix changes `Response::headers` semantics for **every** header consumer, which is a different blast radius from a read-only accessor.

Anchors verified at master `dc23aaa` (2026-08-04).

## Problem

`Headers` is a `std::map` (`engine/include/vayu/types.hpp:107`), and both header callbacks assign into it:

- `engine/src/http/client.cpp:143` - `response->headers[key] = value;`
- `engine/src/http/event_loop/curl_callbacks.cpp:121` - `data->response.headers[key] = value;`

So when a response sends the same header name twice, **only the last one survives**. For most headers that is merely lossy. For `Set-Cookie` it is wrong in a way a user will hit: a login response that sets a session cookie and a CSRF cookie as two headers - the normal way servers do it, since `Set-Cookie` is the one header RFC 7230 §3.2.2 forbids comma-folding - reaches the app and the script with one of them gone.

The gap is visible in the two parsers that read it. The renderer's `parse-set-cookie.ts` documents "Multiple Set-Cookie headers arrive joined by ', '" and its boundary regex exists to recover them; the engine's `parse_set_cookie` (added in PR #306) handles the same joined form and is pinned to it by `engine/tests/fixtures/set-cookie-conformance.json`. Both handle a shape **the engine never produces from separate headers** - they only see the joined form if an upstream proxy folded it. The parsers are right; what feeds them drops data first.

## Fix / solution pathway

Append instead of overwrite, in both callbacks:

```cpp
auto [it, inserted] = response->headers.try_emplace (key, value);
if (!inserted) {
    it->second += ", " + value;
}
```

That is the RFC 7230 §3.2.2 equivalence for comma-list headers, and for `Set-Cookie` it produces exactly the joined form both parsers already recover correctly - which is why the fix is four lines and not a `Headers` type change. A multimap would be more faithful and is the alternative worth weighing in the PR, but it changes the type every consumer indexes.

Trace the consumers before picking: `pm.response.headers` (indexing and `get`/`has`), `pm.response.to.have.header`, `pm.response.size()`'s header-byte figure, response capture, the app's headers table and `headers-format.ts`, the raw-request/response views, and `debug_redact`. Most of these get strictly more data; `size()` gains the folded bytes, which is closer to the wire, not further.

## Tests

- Mutation-check style, at the callback: a mock response sending two `Set-Cookie` headers and two of some ordinary header. Revert the append and both go red.
- Both HTTP paths need it - `client.cpp` (single send) and `curl_callbacks.cpp` (load path) have separate copies of the callback, so a fix to one is a fix to half.
- A case in `set-cookie-conformance.json` is **not** the place for this: the fixture is about parsing a header value, and this defect is upstream of the value.
- Worth adding to `mock_server.hpp` if it cannot currently send a duplicate header at all.

## Docs (same commit)

- `docs/engine/scripting.md` - the `pm.response.headers` section says the object is keyed by lower-cased name; note what a repeated name reads as.
- `docs/app/pm-api-compatibility.md` - the same, next to the header-methods divergence.
- `docs/engine/api-reference.md` if the response shape's description mentions header handling.

## Sequencing

**Not blocking anything, and nothing blocks it.** Independent of #301 step 2 (the cookie jar) - a jar would be fed by libcurl, not by this map - but landing this first makes `pm.response.cookies` and the response Cookies tab honest for the multi-header case, so it is worth doing before anyone leans on either. Parallel-safe with #197 (touches the callbacks, not the worker loop).

## Acceptance criteria

- [ ] A response sending the same header name twice keeps both values in `Response::headers`, on **both** the single-send and load paths.
- [ ] Two separate `Set-Cookie` headers reach `pm.response.cookies` and the app's Cookies tab as two cookies.
- [ ] Tests fail if either callback reverts to assignment.
- [ ] Docs state what a repeated header name reads as.
- [ ] `python build.py -t && cd engine && ctest --preset linux-dev --output-on-failure`, `cd app && pnpm test && pnpm type-check` all green.

athrvk:
## Description

`Response::headers` dropped every value but the last when a response sent the same header name twice. Anchors in #307 verified accurate when this branch was cut; still accurate after merging master.

**The plan.** Root cause: `Headers` is a `std::map` (`types.hpp:107`) and both curl header callbacks did `headers[key] = value` - `client.cpp:143` (single send) and `curl_callbacks.cpp:121` (load path). For most headers that is lossy; for `Set-Cookie` it is wrong in a way a user hits, because a login response setting a session cookie and a CSRF cookie sends **two** `Set-Cookie` lines - the normal way, since `Set-Cookie` is the one header RFC 7230 §3.2.2 forbids comma-folding - and one of them never reached the app or a script. The approach is to fold a repeated name with `", "`, which is §3.2.2's equivalence for comma-list headers and, not by coincidence, exactly the shape the renderer's `parse-set-cookie.ts` boundary regex was written to recover ("Multiple Set-Cookie headers arrive joined by ', '"). Both parsers already handled a shape the engine never produced from separate headers; this makes what feeds them honest rather than changing either parser.

The alternative I rejected is a `multimap` for `Headers`. It is more faithful to the wire, but it changes the type **every** consumer indexes - `pm.response.headers` and its `get`/`has`, `pm.response.to.have.header`, `pm.response.size()`, response capture, `run_manager`, `metrics_collector`'s `content-type` lookup, the CLI printer, the app's headers table - to buy a distinction only `Set-Cookie` currently needs, and one both `Set-Cookie` parsers already recover from the folded form. Folding gets every consumer strictly more data with no type change; `size()`'s header-byte figure gains the folded bytes, which is closer to the wire, not further.

**One copy, not two.** Rather than patching four lines twice, both callbacks now call a single `ingest_header_line` in `curl_utils` - where `validate_transferable` and `apply_method_and_body` already live "so the two cannot drift apart again" after those same two callbacks had drifted before. That deletes the duplicated parse rather than duplicating the fix.

Behaviour that is deliberately unchanged: headers are still cleared at each status line, so a followed redirect does not accumulate hops' headers into the fold; the key is still lower-cased; a line with no colon is still ignored.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Re-run on the merge commit, after #301 step 1 (PR #306) landed:

- `cd engine && ctest --preset linux-dev --output-on-failure` - **837 / 837 pass**.
- `cd app && pnpm test` - **2557 / 2557 pass** in 289 files.

No pre-existing failures reproduced on this base.

New tests (6), all in `curl_transfer_test.cpp` - the file that exists because these two paths had already drifted, so every property that must hold on both is asserted on both:

- `EventLoopKeepsEveryValueOfARepeatedHeaderName` / `ClientKeepsEveryValueOfARepeatedHeaderName` - a new `/repeated-headers` mock route sends two `Set-Cookie` lines *and* two of an ordinary header, so the rule reads as being about repeats rather than about cookies.
- `TwoSetCookieHeadersReachTheCookieParserAsTwoCookies` - the seam: the value a real two-header transfer produced, fed to `parse_set_cookie`, must come back as two cookies. Each half was already pinned on its own (the fold here, the boundary rule in `set_cookie_test.cpp`); nothing asserted they agreed, and agreeing is the point.
- `IngestHeaderLine.FoldsInArrivalOrder` - the fold format and order, on input the test controls. The end-to-end cases accept either order because httplib stores a response's headers in an `unordered_multimap`; this one pins it.
- `IngestHeaderLine.SingleOccurrenceIsStoredVerbatimUnderALowerCasedName`, `…ALineThatIsNotAHeaderFieldIsIgnored` - the unchanged behaviour, so folding cannot be "fixed" later by loosening the parse.

**Mutation-checked per path**, since a fix to one callback used to be a fix to half: inlining the old `headers[key] = value` back into `curl_callbacks.cpp` alone reddens only `EventLoopKeeps…` (client stays green), and into `client.cpp` alone reddens only `ClientKeeps…`. Both restored after.

## Checklist
- [x] I have added tests

Docs in the same commit, all three the issue named: `docs/engine/scripting.md`, `docs/app/pm-api-compatibility.md` and `docs/engine/api-reference.md` all state what a repeated header name reads as. The first two also point a reader at `pm.response.cookies` rather than at hand-splitting a folded `Set-Cookie` value - a hand-rolled split is the copy-of-a-primitive that never receives the primitive's fixes.

`clang-format` was applied to `curl_transfer_test.cpp`, which was clean before. `client.cpp` and `curl_utils.cpp` carry pre-existing format drift in regions this diff does not touch (51 and 59 lines), so they were left alone per the repo rule about not reformatting files that were not clean. All three docs files were already prettier-dirty before this branch, so they were not reformatted either.

- **No case added to `set-cookie-conformance.json`**, as the issue instructs - the fixture is about parsing a header *value*, and this defect is upstream of the value.
- **`mock_server.hpp` was not touched.** The issue flags it as "worth adding if it cannot currently send a duplicate header at all" - it can (httplib's `set_header` emplaces into a multimap), and the natural home for these cases is `curl_transfer_test.cpp`'s own `MethodEchoServer`, which is where both paths are already exercised side by side. `SlowMockServer` serves the timing/shutdown tests and has no reader for a header route.
- **The app needed no code change.** `ResponseCookies.tsx` already reads `headers["set-cookie"]` through `parse-set-cookie.ts`, whose boundary regex splits the folded form correctly today - so the response Cookies tab shows both cookies on the strength of the engine fix alone.

*(An earlier revision of this body listed the `pm.response.cookies` criterion as untestable on the old base, which was true before PR #306 merged. It is tested now - see `TwoSetCookieHeadersReachTheCookieParserAsTwoCookies` above.)*

## Related Issues

Closes #307

### DIFF
--- a/docs/app/pm-api-compatibility.md
+++ b/docs/app/pm-api-compatibility.md
@@ -37,6 +37,13 @@ case-insensitive the way HTTP header names are. Indexing is not, so
 `headers['Content-Type']` reads back `undefined` while
 `headers.get('Content-Type')` works (see [Header methods](#header-methods)).
 
+A header name the server sent **twice** reads as a single value with the two
+folded together by `", "` (RFC 7230 §3.2.2) - Postman's `HeaderList` would give
+you two entries. That matters most for `Set-Cookie`, which servers routinely
+send once per cookie: both cookies are there, in one string, and a caller that
+splits it must split on a comma followed by `name=` rather than on every comma,
+since an `Expires=` value contains one.
+
 Variable writes persist to the scope they target (environment / collection / globals) and
 participate in [variable resolution](./variable-resolution.md). Calling `set(name, value)`
 on a variable that already exists updates only its value - the existing `secret` flag,
--- a/docs/engine/api-reference.md
+++ b/docs/engine/api-reference.md
@@ -1157,6 +1157,15 @@ run-shaped way of stating the same field, not a second store.
 }
 ```
 
+**`headers` is keyed by the lower-cased header name, and a name the response
+sent more than once holds every value folded with `", "`** - the RFC 7230 §3.2.2
+equivalence for comma-list headers. So two `Set-Cookie` lines read back as one
+`"session=abc; Path=/, csrf=xyz; Path=/"` entry rather than the last one alone.
+A client splitting a folded `Set-Cookie` must split on a comma followed by
+`name=`, since an `Expires=` value contains a comma of its own. The same holds
+for the response headers stored on a design run's `trace_data.response` and on
+captured load-run samples, which come off the same parse.
+
 **`consoleLogs` entries carry their own source and level.** `source` is which of
 the request's two scripts wrote the line (`"pre"` for the pre-request script,
 `"test"` for the post-request one) and `level` is the `console.*` method that was
--- a/docs/engine/scripting.md
+++ b/docs/engine/scripting.md
@@ -168,6 +168,16 @@ works. Prefer the methods unless you know the exact key.
 The response object has **no** `add`/`upsert`/`remove` - the response has
 already arrived, so a mutator there would only appear to change something.
 
+**A name the server sent twice reads as one value, folded with `", "`.** Two
+`Set-Cookie` lines arrive as
+`"session=abc; Path=/, csrf=xyz; Path=/"`, and `get()` returns that whole
+string - there is no list form, because `headers` is a plain object. Folding is
+the RFC 7230 §3.2.2 equivalence for comma-list headers; before it, only the
+**last** value of a repeated name survived at all. If you need the parts, split
+on `", "` - except for `Set-Cookie`, whose values may contain commas of their
+own (`Expires=Wed, 21 Oct ...`), so split on a comma that is followed by
+`name=`.
+
 ### Response Assertions
 
 ```javascript
--- a/engine/include/vayu/http/event_loop/curl_utils.hpp
+++ b/engine/include/vayu/http/event_loop/curl_utils.hpp
@@ -11,6 +11,7 @@
 
 #include <optional>
 #include <string>
+#include <string_view>
 
 #include "vayu/http/event_loop.hpp"
 #include "vayu/types.hpp"
@@ -103,6 +104,26 @@ Response error_response (const Error& error);
  * Requires `validate_transferable(request)` to have passed.
  */
 void apply_method_and_body (CURL* curl, const Request& request);
+
+/**
+ * @brief Record one "Key: Value" response header line into `headers`.
+ *
+ * The key is lower-cased so every consumer can index one spelling, and the
+ * value keeps everything past the first colon with its leading spaces trimmed.
+ * A line carrying no colon is not a header field and is ignored.
+ *
+ * **A name that arrives twice keeps both values, folded with ", "** - the
+ * RFC 7230 §3.2.2 equivalence for comma-list headers. Both callbacks used to
+ * assign into the map, so the second `Set-Cookie` of a login response (the
+ * normal way servers set two cookies, since `Set-Cookie` is the one header
+ * §3.2.2 exempts from folding) silently replaced the first. Folding is what
+ * both `Set-Cookie` parsers - the renderer's `parse-set-cookie.ts` and the
+ * engine's - already recover cookie boundaries from, so it hands them the
+ * shape they were written for instead of half the data.
+ *
+ * Shared by both clients so the two cannot drift apart again.
+ */
+void ingest_header_line (std::string_view line, Headers& headers);
 
 /**
  * @brief libcurl's cumulative phase timers, in seconds, for one transfer.
--- a/engine/src/http/client.cpp
+++ b/engine/src/http/client.cpp
@@ -91,6 +91,9 @@ size_t write_callback (char* ptr, size_t size, size_t nmemb, void* userdata) {
  *
  * On followed redirects, curl emits one status block per hop in order,
  * so overwriting each time leaves status_text reflecting the final hop.
+ *
+ * A header name that arrives more than once in the same block keeps every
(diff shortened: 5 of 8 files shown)


athrvk:
Fix merge conflicts

athrvk:
Conflict resolved and pushed (`5c8be64`). #301 step 1 / PR #306 landed in the two docs files this branch also edits:

- `docs/engine/scripting.md` - the real conflict. Both sides were additive and both are kept: my "a repeated name folds with `", "`" paragraph closes the **Reading response headers** section, master's new **Reading response cookies** section follows it.
- `docs/app/pm-api-compatibility.md` - auto-merged.

One deliberate change while resolving, rather than a mechanical keep-both. Both notes used to tell the reader how to split a folded `Set-Cookie` themselves (on a comma followed by `name=`). `pm.response.cookies` is on master now, so they point at it instead - a hand-rolled split is exactly the copy-of-a-primitive that never receives the primitive's fixes.

That also closes the gap the PR body called out as untestable on the old base. The acceptance criterion "two separate `Set-Cookie` headers reach `pm.response.cookies` as two cookies" now has a test: `TwoSetCookieHeadersReachTheCookieParserAsTwoCookies` feeds the value a **real** two-header transfer produced into `parse_set_cookie` and asserts two cookies come back. Each half was already pinned on its own - the fold here, the boundary rule in `set_cookie_test.cpp` - but nothing asserted the two agreed, and agreeing is the whole point. `docs/engine/api-reference.md` needed no resolution.

Re-verified on the merge commit: engine **837 / 837**, app **2557 / 2557** in 289 files, `pnpm type-check` and `mkdocs build --strict` clean.

---

Issue #276, opened 2026-08-02
app(electron): corrupt window-state.json crashes the main process at import time, on every launch


athrvk:
Part of the 2026-08-02 Electron main-process sweep (parent tracking issue links all sub-issues). Base: master `a2bc8f3`. All anchors verified at that commit; adversarially re-verified by an independent pass, including the electron-store/conf source in `app/node_modules`.

## Problem

The `electron-store` `Store` is constructed at module scope (`app/electron/window-state.ts:30-32`) without `clearInvalidConfig`, and `window-state.ts` is imported by `main.ts` at module load (`app/electron/main.ts:13`), before `app.whenReady`. electron-store 11.0.2 delegates to conf 15.1.0, which defaults `clearInvalidConfig: false` and eagerly reads and parses the config file in its constructor, rethrowing `SyntaxError` for invalid JSON. A corrupt `window-state.json` therefore throws during main-process module evaluation: Electron shows the "A JavaScript error occurred in the main process" dialog and the app never starts - recurring on every launch until the user finds and deletes a hidden config file. `loadWindowState`'s own error handling (`window-state.ts:41` onward) never runs.

Plausible corruption sources: disk-full, a hand edit, antivirus interference, and conf's own non-atomic fallback write paths (plain `fs.writeFileSync` under Snap, and the Windows EXDEV fallback - both visible in conf's `_write`). Additionally there is no shape validation: syntactically valid JSON with wrong types (`width: "abc"`) flows straight into the `BrowserWindow` constructor.

Low severity (rare trigger, loud symptom) but disproportionate blast radius: a settings file whose only job is window geometry can brick startup for a non-technical user.

Ride-along hygiene, owned here: `saveWindowState` (`window-state.ts:69`) is exported with zero external importers - only called within its own module (`:107`, `:114`); no test file exists, so not even a test-only reader (verified). The real public surface is `loadWindowState`/`trackWindowState` (`main.ts:102`, `:151`).

## Fix / solution pathway

1. Pass `clearInvalidConfig: true` to the `Store` options - one line; conf then falls back to an empty store on corrupt JSON. Losing window geometry is the correct failure mode.
2. Validate shape in `loadWindowState`: `width`/`height` must be finite positive numbers (and `x`/`y` finite if present) before spreading into `BrowserWindow` options; fall back to defaults otherwise.
3. Drop the `export` on `saveWindowState`.

[...]

PR #328, opened 2026-08-05
fix(mcp): a corrupt mcp-config.json must not cost the user their window

### LINKED ISSUE #315: mcp: a corrupt mcp-config.json bricks app startup - the MCP store repeats the window-state crash #276 fixed
Part of the 2026-08-04 MCP tool-surface sweep (parent tracking issue links all sub-issues). Base: master `10c8193`. Surfaced by the sweep's completeness critic; anchors confirmed by direct read.

## Problem

`app/electron/mcp/store.ts:30` constructs `new Store<McpStoreShape>({ name: "mcp-config" })` **without `clearInvalidConfig: true`** - the exact omission #276 fixed for `window-state.ts`. electron-store 11 delegates to conf 15, whose constructor eagerly reads and parses the config file and rethrows `SyntaxError` on invalid JSON when `clearInvalidConfig` is false.

The store is touched during startup, before the window exists: `startMcp()` calls `loadMcpEnabled()` (`main.ts:487`) and `loadPersistedSafety()` (`main.ts:497`), both of which call `getStore()`. `startMcp()` is `await`ed (`main.ts:831`) before `createWindow()` (`main.ts:834`), inside the `app.whenReady().then(async ...)` handler (`main.ts:782`) which has no `.catch`. So a corrupt `mcp-config.json` throws at store construction, the whenReady handler rejects, and **the window is never created - while the engine sidecar is already running headless** (started earlier at `main.ts:828`). The app appears to launch (dock icon, engine process) and then shows nothing, on every launch, until the user finds and deletes a hidden file.

`loadMcpEnabled()` at `main.ts:487` additionally sits *outside* `startMcp`'s try block (which begins at `:491`), so even the MCP-disabled preference read is unguarded.

Corruption sources are the same as #276's: disk-full, a hand edit, antivirus, conf's own non-atomic write fallbacks. Rare trigger, but the blast radius is worse than #276 (no window at all, plus an orphan-looking headless engine), which is why this is medium rather than low.

## Fix / solution pathway

1. Construct the store with `clearInvalidConfig: true` (`store.ts:30`) - conf then falls back to an empty store on corrupt JSON, and `loadPersistedSafety`/`loadMcpEnabled` already resolve safe defaults from an empty store (`resolveSafetyConfig(sanitizeSafetyInput({}))`, `enabled` defaults true). Losing a hand-corrupted MCP config is the correct failure mode; the safety defaults are the safe ones.
2. Defensive depth (do both): wrap the `startMcp()` await and `loadMcpEnabled()` so an MCP-subsystem failure can never prevent `createWindow()` - MCP is an optional feature and must not gate the window. A `.catch` on the whenReady chain (or moving `createWindow()` ahead of `startMcp()`) closes the class beyond just this store.

## App changes

`app/electron/mcp/store.ts` and `app/electron/main.ts` (startup ordering / guard). No renderer, no engine. Mirror #276's `window-state.ts` fix exactly for the `clearInvalidConfig` half.

## Tests

New `app/electron/mcp/store.test.ts` (vitest, exercising the real electron-store constructor as `window-state.test.ts:17-20` does, not a mock - that is what pins the actual crash site), mutation-checked:

- A corrupt `mcp-config.json` yields default safety + enabled=true, no throw. Revert `clearInvalidConfig` and it throws.
- A `main.ts`-level test (or the store test plus a source-scan asserting `createWindow` is not gated on `startMcp` success) that an MCP init failure does not prevent the window - following #273's renderer-recovery extraction pattern if a testable seam is needed.

## Docs

None (no doc describes the MCP config file's failure behavior; verify and state in the PR).

## Sequencing

Parallel-safe: `store.ts` is edited by no other sub-issue; the `main.ts` startup-ordering touch is disjoint from the M5 `mcp:getSafety` IPC handler region (different part of `main.ts`) but flag it in whichever lands second. Independent of the engine track. `pnpm test`/`pnpm lint` (CI-enforced since #189). Cross-reference #276 as the sibling fix to copy.

## Acceptance criteria

- [ ] A corrupt `mcp-config.json` yields a normal launch with default MCP safety and the server's stored enabled state defaulting on; the window opens.
- [ ] An MCP subsystem init failure never prevents `createWindow()`.
- [ ] New tests pass, mutation-checked; `pnpm test`, `pnpm type-check`, `pnpm lint` green.

---

### LINKED ISSUE #327: mcp(config): master is red - #312's `maxIterations` never reached the CLI env table, breaking `pnpm type-check` and a config test

athrvk:
## Description

**Plan.** Root cause: `store.ts` built its electron-store without `clearInvalidConfig`, the exact omission #276 fixed for `window-state.ts`. conf parses the config file inside the constructor and rethrows the `SyntaxError`, and this store is first touched by `startMcp()` during startup - which was `await`ed *before* `createWindow()`, inside an `app.whenReady` handler with no `.catch`. A corrupt `mcp-config.json` therefore rejected that handler before any window existed, on every launch, with the engine sidecar already running headless behind it. The approach is two layers, because the flag alone only closes this one file: the flag itself, plus making the window's creation depend on no MCP work at all. The alternative I rejected was a `try`/`catch` around the `await startMcp()` call site - `startMcp` already swallows everything it can throw, so that guard is unreachable today and reads as protection it does not give; moving the call after `createWindow()` makes the ordering itself the guarantee, and a source-scan test can hold it.

Verified against `f55acb4`: the anchors had shifted (`store.ts:29`, `main.ts:487/831`) but the defect is exactly as described - `new Store({ name: "mcp-config" })` with no flag, `loadMcpEnabled()` sitting one line above `startMcp`'s `try`, and the whenReady chain unguarded. Blast radius traced: `getStore()` has four callers, all in this file; the four exported functions are read by `main.ts` only (`startMcp` and the `mcp:*` IPC handlers), re-exported through `mcp/index.ts` for the barrel and imported by no test before this one. The stdio CLI never touches the store (it builds its config from the environment), so nothing outside the desktop app is affected.

### `clearInvalidConfig: true` (`store.ts`)

An empty store is the correct failure mode here, because both readers already resolve safe values from one: `loadPersistedSafety` returns `resolveSafetyConfig(sanitizeSafetyInput({}))` - the locked-down defaults, empty allowlist and no writes - and `loadMcpEnabled` defaults to `true`. Losing a hand-corrupted MCP config costs the user their allowlist, not their safety posture.

### The window no longer waits on MCP (`main.ts`)

`await startMcp()` moves to the end of the `whenReady` handler - after `createWindow()`, `initAutoUpdater` and the `activate` listener - so an MCP failure that escaped `startMcp`'s own guard would cost the user MCP and nothing else. `loadMcpEnabled()` moves *inside* that guard; it was the one MCP config read the existing `try` did not cover. Ordering note for reviewers: `mcp:status` now answers `running: false` for the few hundred ms between the window appearing and the server binding. Nothing reads it in that window - the Settings panel queries on open - and the previous ordering delayed the window by exactly that interval instead.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

`cd app && pnpm test` - **291 files, 2602 tests, 1 failure**, and that one failure is master's, not this branch's (see below). `pnpm lint` and `pnpm format:check` clean. `pnpm type-check` fails only on the same master breakage. No engine change and no rebuild - this is app-only TypeScript.

New `app/electron/mcp/store.test.ts`, 13 tests. It drives the **real** electron-store against a temp `userData` directory rather than a mock, the way `window-state.test.ts:17-20` does, because the defect is in what the library does with a file we hand it - a mocked store would have "passed" against the bug.

Mutation-checked, each guard independently:

| Reverted | Fails |
|---|---|
| `clearInvalidConfig` | 5 tests - all three corrupt-file shapes, the enabled-preference read, and the replaced-on-next-save case |
| `createWindow()` moved back after `startMcp()` | *creates the window before starting MCP* |
| `loadMcpEnabled()` moved back outside the `try` | *reads the enabled preference inside startMcp's try, not ahead of it* |

The valid-config cases matter as much as the corrupt ones: *a saved override survives* fails if `clearInvalidConfig` is ever mistaken for "always start empty". The malformed-but-syntactically-valid cases (a string cap, a negative cap, a non-array allowlist, a null safety block) cover the half of the guard the flag cannot see - only the sanitizer can - so they stay green under all three mutations above.

The two `main.ts` assertions are source scans, since main.ts creates windows and starts the engine at import time; both extract a named region first and assert it was found (`renderer-recovery.test.ts` sets the precedent). A scan that read an empty string would fail on the region assertions, not pass vacuously.

### Pre-existing failure, reproduced on the base commit

`electron/mcp/config.test.ts > reports nothing when every variable is well-formed` fails, and `pnpm type-check` reports `TS1360` at `electron/mcp/config.ts:163` - both are #327 (`maxIterations` missing from `SAFETY_ENV_VARS`), already red on master at `f55acb4` before this branch existed. Not fixed or hidden here: #326 carries the fix and closes #327, so CI on this PR goes green once that merges or this branch is rebased onto it. Say the word and I will rebase.

## Checklist

No docs change. I grepped `docs/` for `mcp-config`, `window-state.json`, `clearInvalidConfig`, `startMcp` and `createWindow`: nothing describes the MCP config file's failure behaviour or the startup ordering, so there is no doc to keep in step - matching the issue's own "Docs: none" note.

## Related Issues
Closes #315

---
_Generated by [Claude Code](https://claude.ai/code/session_01BFqUoYmUhgsdnbhx9gZgTT)_

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -484,11 +484,13 @@ async function startEngine() {
 }
 
 async function startMcp() {
-	if (!loadMcpEnabled()) {
-		console.log("[Main] MCP server disabled by preference; not starting.");
-		return;
-	}
 	try {
+		// Inside the try, not ahead of it: this is the first thing in the whole app
+		// to touch the persisted MCP config, so it is where a store failure lands.
+		if (!loadMcpEnabled()) {
+			console.log("[Main] MCP server disabled by preference; not starting.");
+			return;
+		}
 		mcpService = new VayuMcpService({
 			engineBaseUrl: `http://${ENGINE_HOST}:${ENGINE_PORT}`,
 			host: MCP_HOST,
@@ -827,10 +829,11 @@ app.whenReady().then(async () => {
 	// Start the engine
 	await startEngine();
 
-	// Start the MCP server (best-effort; never blocks app startup)
-	await startMcp();
-
-	// Then create the window
+	// Then create the window. Nothing below this line may sit between the engine
+	// and the window: the MCP server used to, and because it reads a config file
+	// at that point, a corrupt one left the app with a headless engine and no
+	// window at all. MCP is an optional convenience with no UI depending on it,
+	// so it starts after everything the user can see is up and wired.
 	createWindow();
 
 	// Start checking for updates once a window exists to receive events. The
@@ -844,6 +847,11 @@ app.whenReady().then(async () => {
 			createWindow();
 		}
 	});
+
+	// Start the MCP server last (best-effort; it never blocks app startup). It
+	// swallows its own failures, and starting it here means even one that escaped
+	// that guard would cost the user nothing but MCP.
+	await startMcp();
 });
 
 app.on("window-all-closed", () => {
--- a/app/electron/mcp/store.test.ts
+++ b/app/electron/mcp/store.test.ts
@@ -1,0 +1,205 @@
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * An optional subsystem's config file must never be able to stop the app from
+ * starting - the same rule window-state.ts is held to, for a worse blast radius.
+ *
+ * conf parses the file inside the Store constructor and rethrows a SyntaxError
+ * unless told otherwise, and this store is first touched by main.ts's `startMcp`
+ * during startup. A corrupt mcp-config.json therefore rejected the `whenReady`
+ * handler before it ever reached `createWindow()`: no window, on every launch,
+ * with the engine sidecar already running headless behind it.
+ *
+ * These drive the real electron-store against a temp userData directory rather
+ * than a mock, because the defect is in what the library does with a file we
+ * hand it - a mocked store would have "passed" against the bug. `electron`
+ * itself is faked, since that is the part vitest cannot provide.
+ */
+
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { mkdtempSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { DEFAULT_MCP_SAFETY_CONFIG } from "./config.js";
+
+/** Mutable because the mock is hoisted above every test that retargets it. */
+const fake = vi.hoisted(() => ({ userData: "" }));
+
+vi.mock("electron", () => {
+	const api = {
+		app: {
+			getPath: () => fake.userData,
+			getVersion: () => "0.0.0-test",
+		},
+		ipcMain: { on: () => {} },
+		shell: { openPath: async () => "" },
+	};
+	// electron-store reaches for the default export.
+	return { ...api, default: api };
+});
+
+/** Seed the userData directory the next import will read, with raw file bytes. */
+function seedStore(contents?: string): void {
+	fake.userData = mkdtempSync(join(tmpdir(), "vayu-mcp-config-"));
+	mkdirSync(fake.userData, { recursive: true });
+	if (contents !== undefined) {
+		writeFileSync(join(fake.userData, "mcp-config.json"), contents);
+	}
+}
+
+function storeFile(): string {
+	return readFileSync(join(fake.userData, "mcp-config.json"), "utf8");
+}
+
+/**
+ * A fresh module instance, so the module's lazily-built `new Store(...)` runs
+ * against the directory just seeded rather than reusing a cached store.
+ */
+async function importStore() {
+	vi.resetModules();
+	return import("./store.js");
+}
+
+afterEach(() => {
+	if (fake.userData) rmSync(fake.userData, { recursive: true, force: true });
+	fake.userData = "";
+});
+
+describe("a corrupt mcp-config.json", () => {
+	it.each([
+		["truncated JSON", '{"enabled": tr'],
+		["not JSON at all", "this is not json at all"],
+		["an empty file", ""],
+	])("still resolves the default safety config when the file is %s", async (_label, contents) => {
+		seedStore(contents);
+
+		const { loadPersistedSafety } = await importStore();
+
+		expect(loadPersistedSafety()).toEqual(DEFAULT_MCP_SAFETY_CONFIG);
+	});
+
+	it("leaves the server enabled rather than throwing on the preference read", async () => {
+		seedStore('{"enabled": fal');
+
+		const { loadMcpEnabled } = await importStore();
+
+		expect(loadMcpEnabled()).toBe(true);
+	});
+
+	it("is replaced by a valid file on the next save", async () => {
+		seedStore('{"safety": {"maxRps": 5');
+
+		const { savePersistedSafety, loadPersistedSafety } = await importStore();
+		savePersistedSafety({ ...DEFAULT_MCP_SAFETY_CONFIG, maxRps: 42 });
+
+		expect(JSON.parse(storeFile())).toEqual({
+			safety: { ...DEFAULT_MCP_SAFETY_CONFIG, maxRps: 42 },
+		});
+		expect(loadPersistedSafety().maxRps).toBe(42);
+	});
+});
+
+describe("a readable mcp-config.json", () => {
+	it("is not discarded - a saved override survives", async () => {
+		seedStore(
+			JSON.stringify({
+				safety: {
+					...DEFAULT_MCP_SAFETY_CONFIG,
+					allowlist: ["api.example.com"],
+					maxRps: 25,
+				},
+				enabled: false,
+			})
+		);
+
+		const { loadPersistedSafety, loadMcpEnabled } = await importStore();
+
+		expect(loadPersistedSafety()).toMatchObject({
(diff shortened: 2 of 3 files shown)


athrvk:
CI is red here and **not because of this diff** - the `App on *` matrix dies in `pnpm type-check`, before a single test runs:

```
electron/mcp/config.ts(163,12): error TS1360: ... Property 'maxIterations' is missing
  in type 'Record<keyof McpSafetyConfig, string>'
```

That is #327, already red on master at `f55acb4` (this PR's base) before this branch existed - #312's `maxIterations` never reached #313's `SAFETY_ENV_VARS` table. Reproduced locally on the base commit: identical single type error, plus the one `config.test.ts` failure it causes. This branch's own suite is 2602 tests with that one master failure and nothing else, and its 13 new tests are green.

The fix is already written and sitting in #326 (`5cc9441`), which closes #327. I have deliberately not duplicated it here - a pre-existing base-branch failure belongs in one commit, not copied across every open branch. So this PR goes green the moment #326 merges.

If you would rather review this one first, say the word and I will either rebase this branch onto #326's head or cherry-pick that single commit - both are a one-command change. Otherwise I will re-run CI here once #326 lands.

---

athrvk:
fix ci

athrvk:
Done - #326 merged while this was open, so the fix was simply to rebase onto the new master (`05f6408`) rather than duplicate anything. Pushed `984de69`; CI re-runs against a base that carries `VAYU_MCP_MAX_ITERATIONS`.

Clean rebase, no conflicts - #326 touched `tools.ts` / `config.ts` / `config.test.ts` / `mcp.md`, this touches `store.ts` / `store.test.ts` / `main.ts`, and the sets are disjoint. Full local run on the rebased head is green end to end:

- `pnpm test` - **291 files, 2607 tests, 0 failures** (was 2602 with 1 failure on the old base; that one failure was #327's and is now gone from master, and 2607 - 2602 = the five `config.test.ts`/`tools.test.ts` cases #326 added)
- `pnpm type-check`, `pnpm lint`, `pnpm format:check` - all clean

The diff itself is unchanged from what you have already got in front of you; only its base moved.

---

PR #333, opened 2026-08-05
fix(mcp): size the /execute budget to the engine's own timeout, not a flat 35s

athrvk:
## Description

**Plan.** Root cause: `EngineClient` armed one flat `DEFAULT_TIMEOUT_MS = 35_000` for every call, including `POST /execute` - the only call that waits on a third-party server. The engine bounds that call by the payload's `timeout` or, absent one, the user-configurable `defaultTimeout` (`resolve_request_timeout_ms`, `execution.cpp:44`), which the config schema lets a user or an agent raise to 300s. When it exceeds 35s the client aborted a request the engine went on to complete - traffic sent, run row written - and `engineErrorResult` matched `/abort/i` alongside `ECONNREFUSED` and answered "engine not running, retry", talking the agent into re-firing a request that already ran. Approach: derive the budget for `/execute` the way the renderer already does for its proxied calls (`proxiedRequestTimeoutMs` in `app/src/services/api.ts`) - `max(payload timeout, configured defaultTimeout) + 10s grace`, falling back to the 300s ceiling when the config read fails - and give an expired budget its own error type carrying the number, so the message can name it. **Alternative rejected:** deriving the budget from the composed payload's `timeout` alone. It is a no-op in exactly this scenario - MCP-composed payloads carry no `timeout` (`payload_from_stored` emits none, `readRequestOverrides` reads none), so the config read is what does the work.

Anchors re-verified against master `e50d540` (the issue cites `10c8193`): the flat constant, the `/abort/i` branch and both dead methods were all still present as described. Blast radius traced: `EngineClient` has two construction sites (`index.ts:35`, `cli.ts:28`), neither passing `timeoutMs`, so both pick up the new behavior; `executeRequest` has two callers (`run_request`, `run_collection_smoke`) and both go through `engineErrorResult`; `getRequest`/`getGlobals` had zero production callers - type-check now proves it.

Changes:

- **`engine-client.ts`** - `request()` takes a per-call budget and distinguishes *its* timer expiring from the caller's cancellation, throwing the new `EngineTimeoutError` (carrying `timeoutMs`) only for the former. `executeRequest` awaits `proxiedTimeoutMs()`, which probes `GET /config` on a short 2s budget and falls back to the ceiling. `/compose` and `POST /runs` deliberately stay on the flat budget - both return once the engine has done local work. `getRequest` and `getGlobals` deleted.
[...]

Issue #321, opened 2026-08-04
MCP tool-surface review sweep (2026-08-04): safety-cap enforcement holes, config-store startup crash, secret-stripping, and the dead-surface batch


athrvk:
Deep review sweep of the MCP tool surface (`app/electron/mcp/**`, all 21 files, ~5k lines incl. tests), against master `10c8193` - the last subsystem not yet covered, deliberately held until the post-#231/#233/#305/#309 shape settled. Method: five parallel subsystem readers (tool surface + safety, server/transport lifecycle, engine-client contracts, resources/prompts, and a written-never-read audit), every candidate finding independently adversarially verified against the pinned commit, plus a completeness critic. 34 agents. 27 findings survived verification (26 from readers + 1 the critic surfaced), 2 refuted, 0 pure-security-hardening; grouped by conflict surface into implementer-ready sub-issues - originally nine, now **ten** after #316 was split on 2026-08-05 (Settings safety-config kept in #316, Windows/POSIX connect moved to #322). All sub-issue bodies re-verified against `10c8193` on 2026-08-05.

## Sub-issues by wave

**Wave M-core - correctness, start these first (all `component:app` + `area:mcp`):**

- #312 - safety caps silently fail to enforce: `startConcurrency` and iterations-mode escape the cap, an omitted duration escapes a sub-60s cap (medium; edits `safety.ts checkLoadCaps` + `start_load_run` schema, plus a small engine `validate_run_config` ride-along)
- #313 - the stdio CLI builds its safety config without the sanitizer, so a malformed cap env var silently disables that cap (medium; `cli.ts`, parallel-safe)
- #314 - `update_environment` strips a variable's secret/type metadata on overwrite, un-masking it in the UI (medium; `tools.ts`, parallel-safe)
- #315 - a corrupt `mcp-config.json` bricks app startup - the MCP store repeats the window-state crash #276 fixed (medium; `store.ts` + `main.ts` startup, sibling of #276)

**Wave M-robustness - IPC/renderer seam and contracts:**

- #316 - Settings shows and then wipes the persisted safety config when the server is off, swallows IPC errors, and re-derives the tool-category list (medium; `main.ts` MCP handler + `store.ts` + `McpSettingsPanel.tsx`)
- #322 - one-click connect fails silently on Windows (`.cmd` shims cannot spawn without `shell`), renderer connect handler has no catch (medium; `connect.ts` + the `handleConnect` hunk of `McpSettingsPanel.tsx`; split from #316)
[...]

Issue #320, opened 2026-08-04
mcp: written-never-read / dead-surface batch - unfired cancellation, drifted prose resource, unread flags, test-only exports, triplicated /mcp URL


athrvk:
Part of the 2026-08-04 MCP tool-surface sweep (parent tracking issue links all sub-issues). Base: master `10c8193`. Every zero-reader claim was independently re-searched (mcp/, main.ts, app/src, docs, tests, SDK registration) by an adversarial pass. Seven hygiene items, one PR - none urgent, together the repo&#39;s most-repeated defect class concentrated in one surface. Deliberately last in the sweep: it deletes and moves surface the other sub-issues touch.

## Problem

**1. Cancellation signal is dead in production (low).** `StaticResourceDef.read`, `RUN_REPORT_RESOURCE.read/listRuns`, and `McpPromptDef.build` all accept an optional `AbortSignal`, but `registerResources`/`registerPrompts` call them without one (`server.ts:148/158/180/201`), though the SDK hands the read, list, and prompt callbacks a `RequestHandlerExtra` whose `signal` fires on client cancellation. (The `complete.runId` callback at `server.ts:169` is excluded: the SDK's `CompleteResourceTemplateCallback` is `(value, context?)` with no `extra`, so completion cannot receive a signal - leave it as-is.) The only caller passing a signal is a test (`resources.test.ts:162`) - so the suite claims cancellation works when it cannot fire. A cancelled slow resource read keeps the engine request running detached until the 35s client timeout.

**2. Scripting resource description enumerates the sandbox in prose and has drifted (hygiene).** The `vayu://scripting/completions` description opens &#34;Every name the pre-request / test script sandbox provides:&#34; then lists a set omitting `pm.info`, `pm.sendRequest`, `pm.test` (`resources.ts:63-73`) - all of which landed via #305/#309 (`scripting.cpp:412-446`, `:641-668`). This is exactly the &#34;prose copy that drifts&#34; the file&#39;s own header (`:15-17`) says the resource exists to eliminate. (The resource *body* is engine-sourced and correct; only the human-facing description prose is stale.)

**3. `McpToolInfo.readOnly` is computed, shipped over IPC, typed, and read by nothing live (hygiene).** `toolCatalog()` copies `readOnly` into the IPC payload (`tools.ts:1248`), `domain.ts:948` re-declares it, `docs/engine/mcp.md:470` documents it - but no renderer code reads `.readOnly`. Only `tools.test.ts` reads it (test-only).

[...]

Issue #318, opened 2026-08-04
mcp(engine-client): dead methods and a fixed 35s timeout that misreports a slow request as "engine not running", plus a stale mcp.md tools table


athrvk:
Part of the 2026-08-04 MCP tool-surface sweep (parent tracking issue links all sub-issues). Base: master `10c8193`. Anchors verified and adversarially re-verified. The engine-client contract cluster: dead methods, a timeout that lies, and the doc table that drifted with them.

## Problem

**1. Fixed 35s client timeout undercuts longer requests and misreports the abort (low).** `EngineClient` hard-codes `DEFAULT_TIMEOUT_MS = 35_000` for every call (`engine-client.ts:38`, armed at `:65`); neither constructor site overrides it. MCP-originated `/execute` payloads never carry a per-request timeout (saved requests store none, `POST /compose` emits none via `payload_from_stored`), so the engine's `defaultTimeout` (`execution.cpp:631`, user-raisable to 300s, agent-raisable via `update_engine_config`) is the only live knob - and it can exceed 35s. When it does, a `run_request` step that legitimately runs past 35s is aborted client-side while the engine completes it server-side (side effects happen, run row recorded), and `engineErrorResult` (`tools.ts:145-149`) matches `/abort/i` and returns "engine not running, retry" - doubly wrong (engine is up; the request likely succeeded) with an explicit retry instruction that leads the agent to re-fire a request that already ran.

**2. `getRequest` and `getGlobals` are dead since #231 (low).** `EngineClient.getRequest` (`engine-client.ts:116`) and `getGlobals` (`:125`) have zero production callers; their only references are unused `vi.fn()` stubs in `tools.test.ts:50-51` that no test asserts on. PR #231 moved composition engine-side (`POST /compose`) and removed the last callers. Written-but-never-read; a future engine change to `GET /requests/:id` or `GET /globals` would rot these silently while tests stay green.

[...]

PR #334, opened 2026-08-05
fix(mcp): stop presenting truncated tool results as complete

### LINKED ISSUE #319: mcp(tools): silent truncation and unvalidated inputs - "All runs" is a page of 100, run_collection_smoke skips nested folders, get_live_metrics limit is unbounded
Part of the 2026-08-04 MCP tool-surface sweep (parent tracking issue links all sub-issues). Base: master `10c8193`. Anchors verified and adversarially re-verified. Five tool-result-correctness defects, mostly the repo's named silent-truncation shape - a tool returns a partial or wrong-shaped result while presenting it as complete.

## Problem

**1. `vayu://runs` says "All runs" but serves the first 100 (low).** The resource description is "All runs (single requests and load tests), newest first" (`resources.ts:37`), but its reader requests `GET /runs?limit=100&offset=0` (`engine-client.ts:146-148`) - the newest 100 only. `docs/engine/mcp.md:285` states the truth ("first page, 100"), so the description shipped to agents contradicts both the behavior and the doc. The resource content does include the engine's `{data, pagination}` envelope (`total`/`hasMore`, `runs.cpp:278-282`, stringified whole at `server.ts:148`), so the truncation is technically visible in the payload - but the description is what the agent reads first, and it lies. A workspace with >100 runs: an agent trusting the description does not find last week's baseline and concludes it was deleted.

**2. `run_collection_smoke` silently omits nested child collections (hygiene).** The handler lists via `GET /requests?collectionId=` (`tools.ts:890`), which returns direct children only (`engine/src/http/routes/requests.cpp:284-291`; proven by the DB query's strict `collection_id ==` filter, `engine/src/db/database.cpp:669-673` - the route's doc comment says "all requests belonging to a specific collection" without the word "direct"), while collections nest (`parentId`, `engine/src/http/routes/collections.cpp:100-106`) and the description promises "Execute every saved request in a collection". A smoke run on a parent returns a green matrix whose `total` counts only direct requests, with nothing signaling that descendant folders went untested. Related runtime concern (critic gap): the handler loops serially with the 35s default per request (see M7/#318), so an N-request collection against slow-but-allowlisted hosts runs N*35s+ with no progress reporting - fold a bound or caveat in here.

**3. `get_live_metrics` limit is unvalidated (hygiene).** The handler forwards any number (`tools.ts:1192`) into `ticks.slice(-limit)` (`engine-client.ts:290`): `limit: 0` gives `slice(0)` = every collected tick (the description promises "default 10"), and `limit: -3` gives `slice(3)` = all but the three oldest - the opposite of a limit. Schema is bare `z.number().optional()` (`tools.ts:1188`).

**4. `get_engine_health` non-object fallback violates its own outputSchema (hygiene).** The fallback (`tools.ts:510-512`) returns a text result, but the declared `outputSchema` (`:506`) makes SDK 1.29 reject any non-error result lacking `structuredContent`; the `McpError` is caught and returned as an `isError` result reading "Output validation error: Tool get_engine_health has an output schema but no structured content was provided" - blaming the tool's schema and hiding the payload. Reachable on a 200 with a valid-JSON non-object body (a mid-restart engine).

**5. Malformed JSON body surfaces as 500 / -32603 "Internal error" (hygiene).** A client POSTing invalid JSON to the Electron-hosted `/mcp` endpoint gets HTTP 500 with JSON-RPC `-32603` because `readJsonBody`'s rejection (`http.ts:144-148`) falls into the generic top-level catch (`:54-65`), bypassing the SDK's own 400/-32700 parse-error path. A client-side payload mistake looks like a Vayu crash. (stdio `cli.ts` unaffected.)

## Fix / solution pathway

1. Reword the resource description to match the doc ("The most recent 100 runs, newest first; see `pagination.total`/`hasMore` in the content for the full count"). The pagination envelope is already in the resource content - the description string is the entire fix.
2. Recurse into child collections, or append a caveat via `withCaveat` when the collection has children ("N sub-collections not included"). Consider a total request budget / progress note for the serial-runtime concern. Preference: the caveat (stating "direct children only" in the tool result and description) is the smaller, allowlist-safe change; recursion changes the smoke matrix shape and interacts with the serial-runtime bound owned by #318's coordination note. Default to the caveat unless the implementer has a reason to recurse.
3. `limit: z.number().int().positive().optional()`, or clamp `Math.max(1, Math.floor(limit))` in the handler.
4. Return `structuredResult({ status: "unknown", raw: value })` in the fallback (schema-satisfying), or make the outputSchema's `status` `.optional()`.
5. Catch `readJsonBody` rejection in `handle()` and answer 400 / -32700 (and 413 for the size limit), leaving -32603 for genuine handler failures.

## App changes

`app/electron/mcp/resources.ts`, `tools.ts`, `engine-client.ts`, `http.ts`. No engine change. No renderer.

## Tests

`app/electron/mcp/` tests, mutation-checked where a fix is invisible to shape:

- `run_collection_smoke` on a collection with children either includes them or emits the caveat (assert the caveat text / recursion).
- `get_live_metrics` with `limit: 0` and `limit: -3` returns a bounded, sensible slice.
- `get_engine_health` with a non-object 200 body returns a valid tool result, not an output-validation error (through the SDK path).
- Malformed JSON to `/mcp` returns 400/-32700 (`protocol.integration.test.ts`, or a new `http.test.ts`).
- The `vayu://runs` description matches the doc (a source assertion is fine; assert non-empty).

## Docs

`docs/engine/mcp.md` - the `vayu://runs` resource line already correct; the `run_collection_smoke` tool description should state whether it recurses. No new doc.

## Sequencing

Scattered across `tools.ts` handlers, `resources.ts`, `engine-client.ts`, `http.ts` - the `engine-client.ts` slice touch overlaps M7/#318's timeout work (coordinate; both small). Parallel-safe against the engine track. `pnpm test`/`pnpm lint` (CI-enforced since #189).

## Acceptance criteria

- [ ] No MCP tool or resource presents a truncated result as complete: `vayu://runs` says "100", `run_collection_smoke` discloses omitted sub-collections.
- [ ] `get_live_metrics` limit is bounded and positive.
- [ ] `get_engine_health` returns a valid result for any 200 body.
- [ ] Malformed JSON to `/mcp` is a 400 parse error, not a 500.
- [ ] New tests pass, mutation-checked; `pnpm test`, `pnpm type-check`, `pnpm lint` green.

---

athrvk:
## Description

**Plan.** Root cause is one shape repeated five times: a result that is partial, or that its own schema rejects, presented as whole. The description an agent reads first is part of the result - `vayu://runs` promised "All runs" while `EngineClient.listRuns` asks for `?limit=100`, and `run_collection_smoke` promised "every saved request in a collection" while `GET /requests?collectionId=` returns direct children only (the DB filters on `collection_id ==`) and collections nest via `parentId`. Approach: make each result state its own bounds - the resource description names the page size and points at the `pagination` envelope it already ships; the smoke matrix appends a note naming the sub-collections it left out; the two unvalidated inputs fail loudly instead of producing a plausible-looking answer; the malformed-body path answers the client's mistake as the client's. **Alternative rejected:** recursing into child collections in `run_collection_smoke`. It changes the matrix shape (`total` would stop meaning "requests in this collection") and multiplies the serial runtime that #318 owns; the issue names the caveat as the smaller, allowlist-safe change, and the caveat carries the same information without either cost.

Anchors re-verified against master `e50d540` (the issue cites `10c8193`): all five were still present as described. Blast radius traced - the `readOnly` scan aside, none of these fields are read elsewhere: the two descriptions have no consumer but the MCP client; `smokeResultSchema` is unchanged so the structured matrix keeps its shape for anyone parsing it; `getLiveMetricsSnapshot`'s `limit` has this one caller; `readJsonBody` has one caller.

Changes:

- **`resources.ts`** - `vayu://runs` says "the most recent 100" and points at `pagination.total` / `hasMore`.
- **`tools.ts`** - `run_collection_smoke` reads its child collections *before* any traffic flows (after the loop the lookup would race a cancellation the run itself raised) and appends a note naming them, distinguishing "no children" from "could not check"; its description states the direct-only scope and the serial runtime. `get_live_metrics` gains `optionalPositiveInt`, and its schema `.int().positive()`. `get_engine_health` wraps any body its `outputSchema` would reject as `{status: "unknown", raw}`.
- **`http.ts`** - `handle()` catches `readJsonBody`'s rejection: 400 / -32700 for a parse failure, 413 / -32600 for the 4 MB cap (a new `BodyTooLargeError` separates them), leaving -32603 to mean a genuine handler failure.
- **`docs/engine/mcp.md`** - the `vayu://runs` row, the `get_live_metrics` constraint, the smoke tool's non-recursion and serial runtime, and the transport's new status codes.

Deliberate deviation from the issue spec: none. The `get_engine_health` fix is slightly wider than the issue's wording - it wraps *any* body the schema would reject, not only a non-object one, because a 200 carrying a `status`-less object fails validation identically and the acceptance criterion asks for "any 200 body".

Coordination with #318 (PR #333, open): disjoint. This branch does not touch `engine-client.ts` at all, and in `tools.ts` it changes individual handlers, not `request()` / `engineErrorResult`. Per that PR's note, the smoke caveat quotes no timeout number, so it stays correct whichever lands first.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

19 new tests, each mutation-checked by reverting its fix and confirming the failure:

| Reverted | Fails |
|---|---|
| `vayu://runs` description back to "All runs…" | 2 resource-description tests |
| smoke result back to a bare `structuredResult` | 2 disclosure tests |
| `optionalPositiveInt` back to `typeof args.limit === "number"` | 3 rejection tests |
| health back to `typeof value === "object" ? … : jsonResult` | 5 wrapping tests |
| `handle()` back to an uncaught `readJsonBody` | the 400 / -32700 test |

- `tools.test.ts` (+13): the smoke matrix names its 2 sub-collections and no unrelated ones while `total` still counts only what ran; a collection with no children adds no noise; an unreadable collection list is disclosed as *unchecked* rather than as absent, and does not fail the run; `limit` of 0, -3 and 2.5 are refused without the engine being called, while 5 forwards and an omitted one defaults to 10; the input schema refuses the same three; five malformed 200 bodies (string, number, array, null, `status`-less object) each produce a result its own `outputSchema` accepts, with the body preserved under `raw`.
- `protocol.integration.test.ts` (+2): a truncated JSON body over a real socket returns 400 / -32700 and says nothing about an internal error; the server still serves a valid request afterwards.
- `resources.test.ts` (+2): the runs description states the page size, does not open "All runs", and names `pagination.total`.

All green on this branch:

- `cd app && pnpm test` - **292 files, 2659 tests passed** (2640 on master)
- `cd app && pnpm type-check` - clean
- `cd app && pnpm lint` - clean (zero warnings)
- `cd app && pnpm format:check` - clean

No engine change, so the C++ build and `ctest` were not run - nothing in this diff can change their colour. `mkdocs build --strict` could not be run (mkdocs is not installed here); the docs edit adds no links, headings or nav entries - two table cells, one bullet extended, and four lines inside an existing section. No pre-existing failures observed.

## Checklist

## Related Issues
Closes #319

Sub-issue of #321. #320 remains the sweep's last item by design.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BHFpWuZXLT9BouHZxECgBz)_

### DIFF
--- a/app/electron/mcp/http.ts
+++ b/app/electron/mcp/http.ts
@@ -112,15 +112,24 @@ export class McpHttpServer {
 			// error"), which reads as the server having fallen over. The SDK's own
 			// parse-error path is unreachable here because the body is parsed
 			// before the transport sees it, so the codes are answered directly.
-			const tooLarge = err instanceof BodyTooLargeError;
-			res.writeHead(tooLarge ? 413 : 400, { "Content-Type": "application/json" });
+			//
+			// Only the two *client-caused* failures are answered here, each with a
+			// message written at its throw site. `readJsonBody` also rejects with
+			// the socket's own error, which is neither the client's mistake nor
+			// safe to echo - that keeps falling through to the 500 path, where a
+			// dead socket belongs.
+			if (!(err instanceof BodyReadError)) throw err;
+			res.writeHead(err.status, {
+				"Content-Type": "application/json",
+				// The oversized case leaves unread bytes in flight; closing is how
+				// they are dropped without reading them. Harmless on the parse
+				// error, where the body was already consumed in full.
+				Connection: "close",
+			});
 			res.end(
 				JSON.stringify({
 					jsonrpc: "2.0",
-					error: {
-						code: tooLarge ? -32600 : -32700,
-						message: err instanceof Error ? err.message : String(err),
-					},
+					error: { code: err.rpcCode, message: err.message },
 					id: null,
 				})
 			);
@@ -144,11 +153,29 @@ export class McpHttpServer {
 }
 
 /**
- * Body exceeded {@link MAX_BODY_BYTES}. Distinguished from a parse failure so
- * `handle()` can answer 413 rather than 400 - the payload was never read, so
- * calling it malformed would send the client rewriting valid JSON.
+ * A body the *client* got wrong, carrying the status and JSON-RPC code that
+ * says which way. Every message is a literal written here, never an underlying
+ * error's text: the response goes to whoever can reach the endpoint, so it
+ * states what the client did wrong and nothing about this process.
+ *
+ * Size and syntax are separate codes on purpose - the oversized body was never
+ * parsed, so calling it malformed would send the client rewriting valid JSON.
  */
-class BodyTooLargeError extends Error {}
+class BodyReadError extends Error {
+	constructor(
+		message: string,
+		readonly status: number,
+		readonly rpcCode: number
+	) {
+		super(message);
+	}
+}
+
+const bodyTooLarge = () =>
+	new BodyReadError(`Request body exceeds the ${MAX_BODY_BYTES}-byte limit.`, 413, -32600);
+
+const bodyNotJson = () =>
+	new BodyReadError("Parse error: the request body is not valid JSON.", 400, -32700);
 
 /** Read and JSON-parse the request body, tolerating an empty body. */
 function readJsonBody(req: http.IncomingMessage): Promise<unknown> {
@@ -158,8 +185,14 @@ function readJsonBody(req: http.IncomingMessage): Promise<unknown> {
 		req.on("data", (chunk: Buffer) => {
 			size += chunk.length;
 			if (size > MAX_BODY_BYTES) {
-				reject(new BodyTooLargeError(`Request body exceeds ${MAX_BODY_BYTES} bytes.`));
-				req.destroy();
+				// Paused, not destroyed: destroying takes the socket down with it,
+				// and the 413 below is then written to a connection the client has
+				// already lost - it reads as a dropped request, which is what a
+				// documented status code is supposed to save it from. The response
+				// carries `Connection: close`, so the rest of the upload is
+				// discarded once it has been flushed.
+				req.pause();
+				reject(bodyTooLarge());
 				return;
 			}
 			chunks.push(chunk);
@@ -173,7 +206,7 @@ function readJsonBody(req: http.IncomingMessage): Promise<unknown> {
 			try {
 				resolve(JSON.parse(raw));
 			} catch {
-				reject(new Error("Parse error: the request body is not valid JSON."));
+				reject(bodyNotJson());
 			}
 		});
 		req.on("error", reject);
--- a/app/electron/mcp/http.ts
+++ b/app/electron/mcp/http.ts
@@ -119,13 +119,7 @@ export class McpHttpServer {
 			// safe to echo - that keeps falling through to the 500 path, where a
 			// dead socket belongs.
 			if (!(err instanceof BodyReadError)) throw err;
-			res.writeHead(err.status, {
-				"Content-Type": "application/json",
-				// The oversized case leaves unread bytes in flight; closing is how
-				// they are dropped without reading them. Harmless on the parse
-				// error, where the body was already consumed in full.
-				Connection: "close",
-			});
+			res.writeHead(err.status, { "Content-Type": "application/json" });
 			res.end(
 				JSON.stringify({
 					jsonrpc: "2.0",
@@ -180,24 +174,32 @@ const bodyNotJson = () =>
 /** Read and JSON-parse the request body, tolerating an empty body. */
 function readJsonBody(req: http.IncomingMessage): Promise<unknown> {
 	return new Promise((resolve, reject) => {
-		const chunks: Buffer[] = [];
+		let chunks: Buffer[] = [];
 		let size = 0;
+		let overflowed = false;
 		req.on("data", (chunk: Buffer) => {
+			// Past the cap the body keeps arriving and is thrown away chunk by
+			// chunk. Neither of the alternatives delivers the 413 it is meant to:
+			// destroying the socket, or closing it while the upload is still in
+			// flight, resets the connection before the client can read the
+			// response - it sees a dropped request, which is exactly what a
+			// documented status code exists to prevent. (This showed up as a
+			// macOS-only CI failure; on Linux the response happened to flush
+			// first.) Draining costs bandwidth already committed by the sender
+			// and protects what the cap is really for: this process's memory.
+			if (overflowed) return;
 			size += chunk.length;
 			if (size > MAX_BODY_BYTES) {
-				// Paused, not destroyed: destroying takes the socket down with it,
-				// and the 413 below is then written to a connection the client has
-				// already lost - it reads as a dropped request, which is what a
-				// documented status code is supposed to save it from. The response
-				// carries `Connection: close`, so the rest of the upload is
-				// discarded once it has been flushed.
-				req.pause();
+				overflowed = true;
+				chunks = [];
 				reject(bodyTooLarge());
 				return;
 			}
 			chunks.push(chunk);
 		});
 		req.on("end", () => {
+			// Already rejected with the 413; the drain finishing is not a result.
+			if (overflowed) return;
 			const raw = Buffer.concat(chunks).toString("utf8");
 			if (raw.trim() === "") {
 				resolve(undefined);
--- a/app/electron/mcp/http.ts
+++ b/app/electron/mcp/http.ts
@@ -103,7 +103,29 @@ export class McpHttpServer {
 			return;
 		}
 
-		const body = await readJsonBody(req);
+		let body: unknown;
+		try {
+			body = await readJsonBody(req);
+		} catch (err) {
+			// A client-side payload mistake is the client's, not a Vayu crash: the
+			// generic catch in `start()` would answer 500 / -32603 ("Internal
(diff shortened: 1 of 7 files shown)


athrvk:
Fixed the CodeQL finding in `3a40186`, and it turned out to be pointing at two real defects rather than one.

**The exposure.** The catch echoed `err.message` for every rejection. `readJsonBody` rejects three ways, and only two of them are the client's doing - the third is `req.on("error", reject)`, the socket's own error. That one is neither safe to hand to whoever can reach the endpoint nor accurate: it was being answered `400 -32700`, telling a client to fix JSON that was never the problem. Rejections now carry their own status and JSON-RPC code on a `BodyReadError`, with a message written at the throw site; a socket error is rethrown to the 500 path, where a dead connection belongs.

**What writing the test found.** The `413` was undeliverable. `readJsonBody` called `req.destroy()` on overflow, which takes the socket down before `handle()` can write anything - so a client sending an oversized body saw a dropped connection, never the status code documenting its mistake. A status code the caller cannot receive is the same written-never-read shape as the rest of this PR, so it is fixed here rather than deferred: the read is paused instead of destroyed, and the response carries `Connection: close` so the remainder of the upload is discarded once it has been flushed. The new test is its own mutation check - it failed with `expected +0 to be 413` (status 0 = dropped connection) against the old `req.destroy()`.

One note on the diff: `docs/engine/mcp.md` is edited by hand and not run through prettier. The file is not prettier-clean on `master` (it uses `*emphasis*` where the repo config wants `_emphasis_`), so per CLAUDE.md's "format only files you touched that were clean before" it is left alone - formatting it reflows 152 lines of table padding for an 18-line change.

Re-verified: `pnpm test` **2660 passed** (2659 before this commit, 2640 on master), `type-check`, `lint` and `format:check` clean.

---

athrvk:
CI went red on macOS and Windows (`App on macos-latest`, `App on windows-latest`, and the `CI gate` aggregate) - all three the same assertion, `expected +0 to be 413`, in the new oversized-body test. Fixed in `3718fd1`.

**The bug was mine, and it was in the fix, not the test.** Answering a request whose body is still uploading and then closing the socket is a race: the client has to read the response before its own write fails. `Connection: close` after pausing the read is what created it. Linux flushed first and passed - which is why my local run and the ubuntu job were green - macOS and Windows did not. Status 0 is a dropped connection, so the client saw exactly the thing the status code exists to prevent.

Past the cap the remaining body is now read and discarded chunk by chunk, and the connection completes normally. No socket teardown, so there is no race left to lose on any platform.

Two things worth your judgement, since this changes what the cap does:

- **The cap now bounds memory, not bandwidth.** Nothing is buffered after the limit (the accumulated chunks are released at that point), so what this process holds is still capped - but the sender's bytes are read off the wire rather than cut off. For a loopback endpoint that seems like the right trade: the alternative provably loses the response.
- **A peer could hold a connection open by streaming forever.** Previously it got destroyed. I did not add a drain bound, since the endpoint is loopback-only and a local process can open connections regardless - but if you would rather have one, say so and I will add a byte/time ceiling that falls back to destroying.

I also dropped the `Connection: close` line from the docs and replaced it with what actually happens.

Re-verified: `pnpm test` **2660 passed**, and the integration file run three times for flakiness (27/27 each). `type-check`, `lint`, `format:check` clean. Watching the new CI run.

---

athrvk:
Claiming this - `master` moved to `30abb30` after this branch was pushed, so I'm bringing the branch current and re-verifying. That commit also edits `docs/engine/mcp.md`, which this PR touches, so the merged doc gets a content check rather than just a clean-merge check.

---

athrvk:
Brought current with `master` (`30abb30`) in `50af754`. Clean merge, no conflicts.

**The doc check.** Both commits edit `docs/engine/mcp.md`. `30abb30` adds the CLI-resolution paragraph under **Connect** (~line 33) and rewords the `connect.ts` row in the file table; this branch edits the transport status codes (~line 82), the `get_live_metrics` row, the smoke non-recursion note and the `vayu://runs` resource row. Disjoint, and the merged text reads coherently.

**Re-verified on the merged tree:** `pnpm test` **2672 passed, 293 files** (2660 before the merge + the 12 `30abb30` brings in), `type-check` clean.

**Merge-order note vs #333** (also left on that PR). The two branches' *code* is disjoint as both bodies claim, but their `mcp.md` edits both rewrite rows of the same tool→endpoint table - this one the `get_live_metrics` row, #333 the four rows just below it. If git merges them without a conflict, the result is still wrong in one direction: this branch's copy of that table still carries `run_request | POST /execute`, which #333 corrects to `POST /compose` + `POST /execute`. Whichever lands second, the `POST /compose` rows are the ones to keep.

---

PR #345, opened 2026-08-06
fix(load-test): raise the engine's maxInFlight bound to 1,000,000 and pin UI/engine parity

### LINKED ISSUE #338: app/engine(load-test): raise the engine's maxInFlight bound to 1,000,000 and pin UI/engine parity - the UI advertises values the engine refuses
Follow-up from the #324 closure verification (2026-08-06, master `f703eaa`). #324 gave `maxInFlight` an engine-side range check; the renderer's advertised range was not reconciled and now promises values the engine refuses.

> **Decision (owner, 2026-08-06): Option B - raise the engine bound.** Rationale: Vayu's target is the Postman + JMeter + k6 combination, and on the k6/JMeter half `maxInFlight` is a *backpressure ceiling*, not a concurrency count - `limits::MAX_CONCURRENCY` (10,000) is the wrong constant for it. The engine's own default formula (`max(targetRps * 10, 1000)`) legitimately reaches 500,000 at the UI's 50k RPS maximum, so an explicit bound of 10,000 refuses configurations the engine happily runs implicitly. The bound becomes **1,000,000** - matching the UI's existing `MAX_IN_FLIGHT.MAX` and covering the default formula across the whole advertised RPS range - and the asymmetry disappears by construction.

## Problem

PR #332 (closes #324) added `{"maxInFlight", 1, limits::MAX_CONCURRENCY}` to the `validate_run_config` field table (`engine/src/http/routes/execution.cpp:594`), where `limits::MAX_CONCURRENCY` is 10,000 (`engine/include/vayu/core/constants.hpp:114`, `10 * event_loop::MAX_CONCURRENT`). The renderer still advertises `MAX_IN_FLIGHT: { MIN: 1, MAX: 1_000_000 }` (`app/src/constants/load-test.ts:66`), and that file's own doc comment states the invariant that UI ranges sit inside the engine's guards. `NumberField` passes min/max through as HTML attributes only - no clamping (`app/src/modules/load-test/components/LoadTestConfigDialog/index.tsx:134-145`) - so a user who types 20,000 into a field whose UI says "up to 1,000,000" gets a 400 from the engine at submit time. Verified at `f703eaa`: the mismatch is real and user-reachable from the load-test dialog.

The related asymmetry, resolved by the decision above: with `maxInFlight` omitted, the engine's default ceiling is `max(targetRps * 10, 1000)` (`engine/src/core/load_strategy.cpp:341-342`), which exceeds 10,000 whenever `targetRps > 1000` - at 50k RPS the implicit default is 500,000, yet explicitly requesting anything over 10,000 is a 400.

## Fix / solution pathway

1. **Engine**: introduce a dedicated limit constant for the backpressure ceiling (e.g. `limits::MAX_IN_FLIGHT = 1'000'000` beside `MAX_CONCURRENCY` in `engine/include/vayu/core/constants.hpp` - a named constant, not a literal in the table) and change the `validate_run_config` entry at `execution.cpp:594` to use it. Do NOT reuse `MAX_CONCURRENCY`; the two limits mean different things and the next reader should see that.
2. **MCP**: the `start_load_run` schema bound from #324 (`app/electron/mcp/tools.ts`, the `maxInFlight` field) must match the new ceiling - update it and its schema-rejection tests.
3. **UI**: `LOAD_TEST_LIMITS.MAX_IN_FLIGHT.MAX` stays 1,000,000 and is now genuinely inside the engine's guard, restoring `load-test.ts`'s stated invariant. No dialog change needed.
4. **Default formula**: unchanged here - `max(targetRps * 10, 1000)` now always sits inside the explicit bound (50k RPS max -> 500k < 1M). Whether the formula itself should shed load differently is #197's measured territory, untouched by this issue.

## App changes

`app/src/constants/load-test.ts` only if the parity test (below) wants a shared source; `app/electron/mcp/tools.ts` for the MCP schema bound. The dialog reads the constant - no component change.

## Tests

- Engine: bounds-inclusive gtest cases at the new ceiling in `engine/tests/run_config_validation_test.cpp` (mirror `StartConcurrencyBoundsAreInclusive`: accepted at 1 and 1,000,000, rejected at 0, -1, 1,000,001).
- App: a vitest case asserting `LOAD_TEST_LIMITS.MAX_IN_FLIGHT.MAX` equals the engine's documented bound (mutation-checked: change either side and it fails). If a constants-vs-engine parity test pattern already exists for the other limits, extend it; if not, this is the first and should cover the neighbouring fields too.
- MCP: the schema bound test updated to the new ceiling.

Commands: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`; `cd app && pnpm test && pnpm type-check`.

## Docs

`docs/engine/api-reference.md` - the `maxInFlight` range on `POST /runs` becomes 1-1,000,000 with the backpressure-ceiling rationale; note the default formula sits inside it. Same commit.

## Sequencing

Parallel-safe against everything open (#197/#287 perf harness, #303/#340 runner track, #337, #339, #134 parked). Touches the `execution.cpp` field table that #197's ride-along region neighbours - trivial rebase, flag in whichever lands second.

Out of scope: any change to the default-ceiling formula's behaviour under load tuning - that is #197's territory; this issue only makes the explicit bound honest.

## Acceptance criteria

- [ ] The engine accepts explicit `maxInFlight` up to 1,000,000 via a dedicated named limit constant; the UI's advertised range is inside the engine's guard again.
- [ ] The MCP schema bound matches; a dialog-accepted value is engine-accepted.
- [ ] Parity is pinned by a mutation-checked test; engine bounds cases are inclusive at both ends.
- [ ] `api-reference.md` documents the range and rationale.
- [ ] `pnpm test`, `pnpm type-check`, `pnpm lint`, engine suite all green.

---

athrvk:
## Description

**Root cause.** `validate_run_config` borrowed `run_config::MAX_CONCURRENCY` (10,000) for `maxInFlight`, but the two limits bound different things: the connection guard bounds an *eager* per-worker curl-handle pre-allocation, while `maxInFlight` bounds a counter of outstanding requests that pre-allocates nothing. The engine's own default ceiling is `max(targetRps * 10, 1000)` (`load_strategy.cpp:341`), which reaches 500,000 at the load dialog's 50k RPS maximum - so requesting explicitly what the engine picks implicitly was a 400, and the renderer's advertised `MAX_IN_FLIGHT.MAX` of 1,000,000 (with `NumberField` passing min/max through as HTML attributes, no clamping) promised a range the engine refused. **Approach:** a dedicated named constant `run_config::MAX_IN_FLIGHT = 1000000` beside `MAX_CONCURRENCY`, read by the field table, plus the same ceiling on the MCP schema and a test that reads the engine header so the three copies cannot drift apart again. **Rejected alternative:** lowering the UI ceiling to 10,000 (option A on the issue) - it would have hidden ceilings the engine already selects for itself when the field is omitted, making the explicit and implicit paths disagree in the opposite direction.

Verified at `f703eaa`: the mismatch was real and user-reachable. One anchor in the issue had drifted - the dialog is `app/src/modules/request-builder/components/LoadTestConfigDialog/index.tsx`, not `modules/load-test/...`. It needed no change either way (it reads the constant).

**Blast radius traced.** `maxInFlight` has four consumers: the validator (`execution.cpp:594`), the strategy that reads it (`load_strategy.cpp:341`), the dialog (via `limits.MAX_IN_FLIGHT`), and the MCP schema. `MAX_CONCURRENCY` keeps every other use - `concurrency` and `startConcurrency` still bound by it, unchanged and untouched.

**One deviation from the issue spec, with reason.** The issue suggested `app/src/constants/load-test.ts` as a shared source for the MCP bound. It cannot be: `tsconfig.node.json` deliberately withholds the `@/*` mapping so production code in `electron/` cannot reach into `src/`, and the import fails type-check (confirmed - the first draft did exactly this). The bound stays a named exported literal in `tools.ts`, and the chain is pinned from the test side instead, the way the dynamic-variable table already is: `tools.test.ts` (which *may* use `@/*`) ties the schema to the renderer's constant, and the new parity test ties that constant to the engine header.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- `python build.py -e -t` then `ctest --preset linux-dev --output-on-failure` - **888/888 passed**.
  - `MaxInFlightBoundsAreInclusive` now runs against `MAX_IN_FLIGHT`: accepted at 1 and 1,000,000, rejected at 1,000,001 (0 and -1 keep their own cases).
  - New `MaxInFlightAcceptsWhatTheDefaultFormulaReaches` pins 500,000 - what the default formula reaches at 50k RPS - as accepted, and asserts it is above `MAX_CONCURRENCY` so the case cannot quietly stop covering the asymmetry it was written for.
- `pnpm test` - **2700/2700 passed, 295 files**; `pnpm type-check`, `pnpm lint`, `prettier --check` on touched files all clean.
  - New `src/constants/load-test.engine-parity.test.ts` reads `constants.hpp` and `execution.cpp` directly: the advertised ceiling equals the engine constant, the route actually reads that constant, the backpressure ceiling stays above the connection guard, and the connection ranges stay inside it. It asserts the sources were non-empty first, per CLAUDE.md.
  - **Mutation-checked**: reverting both engine edits (constant back to 10,000, table back to `MAX_CONCURRENCY`) fails 4 of its assertions. The constant-reader evaluates expressions rather than matching literals, because `MAX_CONCURRENCY` is written as `10 * event_loop::MAX_CONCURRENT` and a guard that could not read it would not guard the field it names - this was caught by the test failing on first run.
  - `tools.test.ts` adds inclusive-ceiling cases (1 / 500,000 / 1,000,000 accepted, 1,000,001 rejected) and the MCP-to-renderer parity assertion.
- `mkdocs build --strict` was not run - mkdocs is not installed in this environment. The docs edits are prose inside existing tables and paragraphs; no page, heading, or link was added or moved.

## Checklist

Docs updated in the same commit: `docs/engine/api-reference.md` (the `maxInFlight` range becomes `1`-`1000000`, with the backpressure-ceiling rationale and the note that the default formula sits inside it) and `docs/engine/mcp.md` (the schema now carries the engine's ceiling). `load-test.ts`'s own doc comment is corrected - `MAX_IN_FLIGHT` sits *on* the engine's bound rather than under it, and that is deliberate.

## Related Issues
Closes #338

---
_Generated by [Claude Code](https://claude.ai/code/session_015Ke7NKKQ1RZWb4k96VCjw6)_

### DIFF
--- a/app/electron/mcp/tools.test.ts
+++ b/app/electron/mcp/tools.test.ts
@@ -7,9 +7,16 @@
 
 import { describe, expect, test, vi } from "vitest";
 import { z } from "zod";
-import { dispatchTool, toolCatalog, TOOLS, type ToolContext } from "./tools.js";
+import {
+	dispatchTool,
+	MAX_IN_FLIGHT_BOUND,
+	toolCatalog,
+	TOOLS,
+	type ToolContext,
+} from "./tools.js";
 import { resolveSafetyConfig, type McpSafetyConfig } from "./config.js";
 import { EngineRequestError, EngineTimeoutError, type EngineClient } from "./engine-client.js";
+import { LOAD_TEST_LIMITS } from "@/constants/load-test";
 
 /**
  * Default `composeRequest` fake: the engine's identity composition for an
@@ -1063,6 +1070,38 @@ describe("dispatchTool", () => {
 				parseArgs("start_load_run", { url: "https://api.example.com", [field]: 2.5 })
 			).toThrow();
 		}
+	});
+
+	test("the schema's maxInFlight ceiling is the engine's, inclusive", () => {
+		// The bound exists so an out-of-range value is named here rather than
+		// returned as an opaque 400 from POST /runs - which means it has to be
+		// the *same* bound, not a stricter guess. 500,000 is what the engine's
+		// own default formula reaches at 50k RPS, so refusing it would refuse a
+		// run the engine starts on its own.
+		for (const value of [1, 500_000, MAX_IN_FLIGHT_BOUND]) {
+			expect(() =>
+				parseArgs("start_load_run", {
+					url: "https://api.example.com",
+					maxInFlight: value,
+				})
+			).not.toThrow();
+		}
+		expect(() =>
+			parseArgs("start_load_run", {
+				url: "https://api.example.com",
+				maxInFlight: MAX_IN_FLIGHT_BOUND + 1,
+			})
+		).toThrow();
+	});
+
+	test("the maxInFlight ceiling is the one the load dialog advertises", () => {
+		// Third copy of the same number: engine header, renderer constant, this
+		// schema. `electron/` production code may not import `src/`, so the copy
+		// stays - and this is the half of the chain that keeps it honest, the
+		// renderer-to-engine half living in
+		// `src/constants/load-test.engine-parity.test.ts`. An agent and a human
+		// composing the same run must be accepted or refused identically.
+		expect(MAX_IN_FLIGHT_BOUND).toBe(LOAD_TEST_LIMITS.MAX_IN_FLIGHT.MAX);
 	});
 
 	test("start_load_run sends an explicit duration when the cap is under the engine default", async () => {
--- a/app/electron/mcp/tools.ts
+++ b/app/electron/mcp/tools.ts
@@ -401,6 +401,20 @@ async function composeLoadRunRequest(
  */
 const DEFAULT_LOAD_MODE = "constant_concurrency";
 
+/**
+ * The engine's `maxInFlight` guard (`run_config::MAX_IN_FLIGHT` in
+ * `engine/include/vayu/core/constants.hpp`), mirrored here so the schema
+ * rejects an out-of-range value by name instead of surfacing a 400.
+ *
+ * It is a literal rather than an import because production code in `electron/`
+ * may not reach into `src/` - `tsconfig.node.json` withholds the `@/*` mapping
+ * on purpose, so the import would not type-check. The copy is kept honest from
+ * the test side, the way the dynamic-variable table already is: `tools.test.ts`
+ * ties it to the renderer's `LOAD_TEST_LIMITS`, and
+ * `src/constants/load-test.engine-parity.test.ts` ties that to this header.
+ */
+export const MAX_IN_FLIGHT_BOUND = 1_000_000;
+
 // --- Shared input schema fragments ------------------------------------------
 
 /** Optional resolution scope shared by the ad-hoc execute/load tools. */
@@ -1182,12 +1196,17 @@ export const TOOLS: McpTool[] = [
 			targetRps: z.number().optional().describe("Target RPS (constant_rps)."),
 			// A pending-request ceiling read as a `size_t`, so `-1` - the natural
 			// "unlimited" spelling - removes the ceiling instead of tightening it.
+			// The upper bound is the engine's, so a value this schema accepts is
+			// one `POST /runs` accepts.
 			maxInFlight: z
 				.number()
 				.int()
 				.positive()
+				.max(MAX_IN_FLIGHT_BOUND)
 				.optional()
-				.describe("In-flight cap (constant_rps only)."),
+				.describe(
+					`In-flight cap (constant_rps only), 1-${MAX_IN_FLIGHT_BOUND}. Default max(targetRps * 10, 1000).`
+				),
 			requestId: z
 				.string()
 				.optional()
--- a/app/src/constants/load-test.engine-parity.test.ts
+++ b/app/src/constants/load-test.engine-parity.test.ts
@@ -1,0 +1,118 @@
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * The dialog promises a range; the engine enforces one. When they disagree the
+ * user finds out at submit time, as a 400 with no field on it.
+ *
+ * `load-test.ts` states the invariant - every advertised range sits inside the
+ * engine's guard - but nothing checked it, and `maxInFlight` drifted the moment
+ * the engine grew an explicit bound (#324): the dialog offered up to 1,000,000
+ * against an engine ceiling of 10,000, and `NumberField` passes min/max through
+ * as HTML attributes without clamping, so typing 20,000 into a field labelled
+ * "up to 1,000,000" was a rejected run.
+ *
+ * This reads the engine's own header rather than a hand-copied number, which is
+ * what makes it a drift guard instead of a restatement: move either side and it
+ * fails.
+ *
+ * The MCP schema holds a third copy of the `maxInFlight` bound and cannot be
+ * checked from here - `src/` may not import from `electron/`, and the type-check
+ * enforces it. That link is pinned from the other end, in
+ * `electron/mcp/tools.test.ts`, which may import `@/constants` and so ties the
+ * schema to the ceiling this file ties to the engine.
+ */
+
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { describe, expect, it } from "vitest";
+import { LOAD_TEST_CEILING_BOUNDS, LOAD_TEST_LIMITS } from "./load-test";
+
+const repoRoot = join(dirname(fileURLToPath(import.meta.url)), "..", "..", "..");
+const constantsHpp = readFileSync(
+	join(repoRoot, "engine", "include", "vayu", "core", "constants.hpp"),
+	"utf8"
+);
+const executionCpp = readFileSync(
+	join(repoRoot, "engine", "src", "http", "routes", "execution.cpp"),
+	"utf8"
+);
+
+/**
+ * One `constexpr <type> NAME = <expr>;` from the engine header, evaluated.
+ *
+ * The expression grammar is deliberately tiny - a product of integer literals
+ * and other constants in this header, with `static_cast<>` and namespace
+ * qualifiers stripped - because that is all the run-config guards use, and
+ * anything richer should fail loudly here rather than be guessed at. It has to
+ * evaluate rather than pattern-match a literal: `MAX_CONCURRENCY` is written as
+ * `10 * event_loop::MAX_CONCURRENT`, and a guard that could not read it would
+ * simply not guard the field it names.
+ */
+function engineConstant(name: string, seen: string[] = []): number {
+	expect(seen, `cyclic definition reaching ${name} in constants.hpp`).not.toContain(name);
+	const match = constantsHpp.match(
+		new RegExp(String.raw`constexpr\s+\w+\s+${name}\s*=\s*([^;]+);`)
+	);
+	expect(match, `${name} not found as a constexpr in constants.hpp`).not.toBeNull();
+	const factors = match![1]
+		.replace(/static_cast<[^>]*>/g, "")
+		.split("*")
+		.map((factor) => factor.replace(/[()\s]/g, ""));
+	return factors.reduce((product, factor) => {
+		if (/^\d[\d']*$/.test(factor)) return product * Number(factor.replace(/'/g, ""));
+		const referenced = factor.split("::").pop() ?? "";
+		expect(referenced, `unsupported expression "${match![1].trim()}" for ${name}`).toMatch(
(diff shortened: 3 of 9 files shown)


athrvk:
fix ci

athrvk:
CI is red, but nothing here ran - this is a runner-scheduling problem, not a failure in the diff.

**What happened on the first attempt.** Four workflow runs were created at 17:06:53. In `PR CI - run tests` (run `31122138607`), the `Detect changes` job sat queued for exactly 15 minutes and was cancelled at 17:21:55 with `runner_id: 0` and an empty `runner_name`, and its job logs 404 - it was never assigned a runner. The matrix jobs then skipped, `CI gate` queued for its own 15 minutes and was cancelled at 17:36:57, and that is what turned the run red. `PR title`, `Labeler` and `Docs site` were all cancelled at the same instant, 17:21:55. `CodeQL` is the only workflow that got runners; `Analyze JavaScript/TypeScript` ran 17:08:27-17:12:48 and passed.

**What I did.** Re-ran all four runs. `PR CI - run tests` restarted at 18:22:49 and has now been `queued` for over 15 minutes with **zero jobs created** - the same symptom, one step earlier.

So no job has executed a line of this branch, on either attempt, and there is nothing in the change to fix in response. This looks like Actions capacity or an account-level scheduling issue rather than anything repo-side; worth a glance at the Actions status/billing page, or simply waiting the incident out and re-running.

For the record, the same checks locally at `85d2aea`:

- `python build.py -e -t` then `ctest --preset linux-dev --output-on-failure` - 888/888 passed.
- `pnpm test` - 2700/2700 passed across 295 files; `pnpm type-check`, `pnpm lint`, and `prettier --check` on the touched files all clean.

I'm still subscribed and will re-run and report if the queue starts moving.

---

athrvk:
Following up on the previous comment with the actual cause, which is now confirmed rather than inferred: **GitHub Actions is in a declared major outage.**

From `githubstatus.com` as of 18:41 UTC:

- Incident opened **15:22:49 UTC**, escalated degraded -> partial outage -> **major outage**, still open.
- Stated impact: "workflow runs are failing or delayed in starting", "**some queued jobs may timeout**", hosted runners impacted, Actions API returning errors, webhook deliveries delayed.
- `Actions` and `Pages` are both listed as `major_outage`. Latest update 18:11:41 UTC - mitigations applied, fixes still rolling out.

That matches this pull request's symptoms exactly. The first attempt's `Detect changes` job was cancelled at the 15-minute mark with `runner_id: 0` and no runner name; the re-run requested at 18:22:49 has now been `queued` for 19 minutes with **zero jobs created**. `CodeQL` is the only workflow here that ever got a runner - it started at 17:08, roughly the last moment before the incident escalated, and passed.

**State of the change itself:** unaffected. `master` is still at `f703eaa`, this branch is one commit ahead of it with no drift, and the diff at `85d2aea` verified clean locally - 888/888 ctest, 2700/2700 vitest across 295 files, plus type-check, lint and prettier on the touched files.

**Recommended handling:** leave the queued re-run alone rather than cancelling it - it may schedule itself once capacity returns. If it does time out (the status page warns queued jobs can), one re-run after GitHub marks the incident resolved should be all this needs. There is nothing to change in the branch in response, so I have not pushed anything.

---

athrvk:
Update after the close/reopen: it did not produce new workflow runs, and the runner problem is not specific to this branch.

**The reopen created nothing.** The branch still has the same five runs. `PR CI - run tests` (`31122138607`) and `Docs site` (`31122138615`) have been `queued` since 18:22:5x with **zero jobs created**. The one job that did get created since - `Apply path labels`, attempt 2, at 18:47:20 - was cancelled at 19:02:23, again after exactly 15 minutes, again with `runner_id: 0` and an empty `runner_name`.

**It is happening on `master` too.** The most recent `master` run, `Docs site` `31113047952` from the 14:52:43 push, was also cancelled. The 04:43 runs that morning succeeded, so whatever this is started between those two points and predates this pull request entirely.

So the pattern is consistent everywhere: jobs are created, never assigned a runner, and cancelled ~15 minutes later. Nothing on this branch can influence that, and I have no lever left - I re-ran all four runs once and they queued indefinitely. Two things worth checking, in order: the Actions spending/usage limit on the account (a reached limit cancels queued jobs exactly like this, including for repositories that are otherwise free), and <https://www.githubstatus.com>.

The branch itself is unchanged and green locally at `85d2aea` - engine 888/888 ctest, app 2700/2700 vitest, plus type-check, lint and prettier. Once runners are flowing again, a re-run of `PR CI - run tests` is all this needs.

---

athrvk:
Picking this back up - runners are flowing again, and this branch is the only one still stranded.

Evidence: #346 and #347 both got real runners tonight (#347's engine matrix ran 23:13-23:32 on all three OSes and passed; #346's app matrix ran 23:05-23:23 and passed). This branch's runs are still the *original* 17:06:53 dispatch sitting in `queued` with zero jobs created - `PR CI - run tests` (`31122138607`), `Docs site` (`31122138615`) and `PR title` (`31122138589`). They were queued before the outage cleared and never got re-dispatched, so they are stale queue entries rather than a fresh attempt.

Cancelling those three and re-running them so they schedule against the recovered capacity. No change to the diff - `master` is still at `f703eaa` and this branch remains one commit ahead of it with no drift.

---

athrvk:
**CI is unstuck and running for real.** Everything that has reported is green.

**What was wrong.** The runs on this branch were the *original* 17:06:53 dispatch, made during the outage and left in `queued` with zero jobs created. They were not recoverable in place - `POST /actions/runs/{id}/cancel` returned `409 Cannot cancel a workflow re-run that has not yet queued` on all three, so re-running them was not an option either. What actually re-dispatches is a fresh `push`/`synchronize` event, which is why #346 and #347 recovered when they were pushed at ~23:00 and this branch, untouched since 17:06, did not.

**What I did.** Amended the single commit so it takes a new SHA (`85d2aea` -> `2870a48`) and force-pushed with lease. **The tree is byte-identical** - both commits point at tree `b3e60f3ca23b5caeed0d0d6c0ef9cb1e1513c6f4`, and the author date is preserved; only the committer timestamp moved. No content change, so the review state of the diff is exactly what it was. `master` is still at `f703eaa` and this branch remains one commit ahead of it.

It worked immediately - jobs were created and assigned real runners within seconds, against zero jobs on every prior attempt.

**Results so far at `2870a48`:**

| Check | Result |
|---|---|
| App on ubuntu-latest | ✅ (lint, prettier, type-check, vitest) |
| App on macos-latest | ✅ |
| App on windows-latest | ✅ |
| Docs site / Build | ✅ |
| Analyze JavaScript/TypeScript | ✅ |
| Detect changes / Conventional title / Apply path labels | ✅ |
| Engine on ubuntu / macos / windows | still building |

The three engine jobs are on a cold vcpkg cache (the cache step was a miss - it completed in 0s), so they are compiling openssl, curl and libsodium from source rather than restoring them. That is the slow path, not a hang.

**One note on reading this run's status.** The Actions API is serving badly stale data right now - it reported all three App jobs as `in_progress` for over an hour after they had in fact finished at 00:44-00:45, and the run object's own `updated_at` is still pinned at 00:40:32. So the engine jobs may well have finished already too. Trust the web UI over the API here.

**Independent check that the diff is not the cause of any slowness:** I re-ran the app suite locally at this exact tree - **2700/2700 passed across 295 files in 113 seconds**. The new `load-test.engine-parity.test.ts` does two `readFileSync` calls at module load and regex-matches them, with a cycle guard on the constant recursion; there is no globbing and nothing unbounded in it.

---

athrvk:
**Engine on ubuntu-latest passed** - `C++ Tests (ubuntu-latest)` ✅, job completed at 00:49:59.

Correcting one thing in my previous comment: I attributed the apparent slowness to a cold vcpkg cache. That was wrong. The ubuntu engine job ran **00:40:30 -> 00:49:59, about 9.5 minutes** - a completely normal time, in line with #347's engine jobs. Nothing was ever slow. The Actions API was simply reporting hour-plus stale state for this run, exactly as the staleness note in that comment suspected. The same is almost certainly true of the two jobs still showing `in_progress`.

So the whole run at `2870a48` looks like this:

| Check | Result |
|---|---|
| Engine on ubuntu-latest + C++ Tests | ✅ (9.5 min) |
| App on ubuntu / macos / windows | ✅ |
| Docs site / Build | ✅ |
| Analyze JavaScript/TypeScript | ✅ |
| Detect changes / Conventional title / Apply path labels | ✅ |
| Engine on macos / windows | reported `in_progress`, near-certainly stale |

Every check that has reported a conclusion is green, and nothing has failed. The re-dispatch is what this needed and it is done - the branch is one commit ahead of `f703eaa` with a byte-identical tree to the version you already had. Ready for your review whenever the last two check runs catch up in the UI.

---

PR #346, opened 2026-08-06
fix(context-bar): commit a variable edit to the definition the resolver picked

### LINKED ISSUE #341: app(context-bar): a variable edit can silently write into the wrong definition - commitValue re-derives its target instead of using the resolver's winner, and stale-snapshot commits revert concurrent edits
From the 2026-08-07 context-bar review sweep (4 readers, adversarial verification per finding, completeness critic; base master `f703eaa`). This issue is the correctness cluster: every defect in how `ContextBar.commitValue` picks its write target and builds its payload, plus the tests that pin the fixes. It is the prerequisite for the context-bar feature work (see the alive-context-bar issue), which keeps this component's variables section.

## Problem

**1. The collection-scope write target diverges from the resolver's winner (medium - the rare + silent + wrong-target class).** The resolver picks the displayed winner with `isEnabledDefinition` - absent `enabled` counts as enabled, the documented D17 rule (`app/src/lib/variable-resolution.ts:53-55`, applied at `app/src/hooks/useVariableResolver.ts:134`, winner = leaf-most enabled at `:167-176`). `commitValue` re-derives the target with a *truthy* test: `chain.find((col) => col.variables?.[name]?.enabled)` (`ContextBar.tsx:130`) skips definitions whose `enabled` is absent. Reachability is proven end to end: the engine stores variable blobs verbatim (`engine/include/vayu/http/routes.hpp:219`) and returns them verbatim (`engine/src/utils/json.cpp:201`), and the transformer passes them through (`app/src/services/transformers/collection-transformer.ts:43`), so an `enabled`-absent definition written via the raw engine API reaches the cache - exactly the blob D17's own comment names as real. Two concrete failures:
   - Leaf defines `apiKey` with no `enabled` key, an ancestor defines the same name with `enabled: true`: the bar displays the leaf's value, but blur **writes the edit into the ancestor's definition** - silently, cross-collection, and if the ancestor's def is `secret: true` the non-secret input just overwrote a secret. The input's key (`name:resolved.value`) is unchanged, so the typed text sits on screen looking committed while resolution never changed.
   - The `enabled`-absent leaf is the *only* definition: `find` returns nothing and `gone()` toasts "no longer defined in its collection scope" - false; the variable is defined, resolving, and displayed two inches away.

**2. The commit payload is a stale render-time snapshot, so concurrent edits revert each other (medium, critic finding).** Each scope branch spreads the *render-closure* copy of the whole variables map (`ContextBar.tsx:105`, `:117-120`, `:135-138`) and the transport replaces the map wholesale. Blur-commit variable A, then blur-commit variable B before A's `onSuccess` lands: B's payload was built from the pre-A cache and **re-sends A's pre-edit value** - last write wins, both mutations succeed, A's edit is silently reverted.

**3. Mid-edit context changes retarget or drop the commit (low, two routes).** The input's identity key is `${name}:${resolved.value}` (`ContextBar.tsx:204`) - no scope, no source. (a) An environment switch mid-edit: if the new resolution has the same value, the DOM node survives and blur writes into the *newly*-resolved scope (possibly a different environment or a collection); if the value differs, the input remounts and the typed text is silently discarded. (b) A keyboard tab switch (Ctrl/Cmd+1-9, `Shell.tsx:150`) fires without blurring: same mechanism against the *new request's* collection chain - two sibling collections both defining `baseUrl` with the same value routes the edit into the wrong one.

**4. The hand-rolled chain walker lacks the resolver's cycle guard (low).** `buildLeafFirstChain` (`ContextBar.tsx:31-41`) walks `parentId` with no seen-set; the resolver's `buildCollectionChain` (`useVariableResolver.ts:61-83`) carries exactly that guard with a comment explaining why (a cycle in a bad database otherwise hangs the renderer synchronously; the engine rejects cycles on write, so this only fires on an already-bad DB). The repo's named hand-rolled-copy defect: the copy never received the primitive's fix. The bar renders fine (guarded path) and freezes the window on the first blur (unguarded path).

**5. An in-flight commit is invisible to the quit-time flush (low, critic finding).** `onBeforeQuit` awaits only save-store contexts (`App.tsx:80-85`, `save-store.ts:186-191`); ContextBar registers none (verified: its only save-store call is `failSave`). The renderer can die mid-PUT after the input already showed the value as committed.

**6. Defensive asymmetry (hygiene).** The collection branch is the only scope path with no `else`: `if (request?.collectionId)` falsy means silent no-op - unreachable today (verified: a collection-scoped entry only renders when `collectionId` was truthy in the same render), but one refactor away from the documented silent-swallow defect. Add `else return gone();`.

## Fix / solution pathway

One design change dissolves items 1, 3, and 4 together: **commit against `resolved.sourceId`, which the resolver already emits on `ResolvedVariable` precisely to name the winning source** (`app/src/types/domain.ts:250-261` documents the move; `useVariableResolver.ts:195-203` populates it).

1. `commitValue` looks up the write target by `resolved.sourceId` per scope (environment: the source environment by id, not `activeEnvironmentId`; collection: the source collection by id - `buildLeafFirstChain` is **deleted**, which also removes the unguarded cycle walk). `gone()` fires when the source-by-id no longer holds the name.
2. The input key becomes `${name}:${resolved.scope}:${resolved.sourceId}:${resolved.value}` so any retarget remounts the node (kills 3a/3b's surviving-node route; the remount-discard route remains the documented lesser behavior - an abandoned edit is dropped, never mistargeted).
3. Payload freshness: read the variables map from the query cache at commit time (`queryClient.getQueryData`) rather than the render closure, and apply an optimistic cache update at mutate time (TanStack `onMutate` or `setQueryData` before `mutate`) so a second commit's fresh read includes the first - the two together close item 2 for the sequential-blur case, which is the reachable one.
4. Register the commit with the save-store machinery (`startSaving`/`completeSave` around the mutation, alongside the existing `failSave`) so `flushAll` awaits it at quit and the Dock's save status reflects the edit - item 5, and consistent with how the rest of the app reports saves.
5. `else return gone();` on the collection branch - item 6.

The uncontrolled-input + `failSave` rollback pattern is deliberate and stays (it has its own regression test); this issue changes where the commit goes, not how failure is shown.

## App changes

All in `app/src/components/layout/ContextBar.tsx` plus the three mutation call sites' optimistic-update wiring (whether in `ContextBar` or in `app/src/queries/globals.ts` / `environments.ts` / `collections.ts` `onMutate` - implementer's call, say which in the PR). No engine change (verified: the engine's verbatim blob storage is by design, D17 normalization is renderer policy).

## Tests (the sweep found these branches at zero coverage - all mutation-checked)

Extend `ContextBar.save-error.test.tsx` or add `ContextBar.commit-scope.test.tsx` (jsdom pragma):

- Per scope (global/environment/collection): resolver reports a winner with a known `sourceId`; assert the corresponding mutation's `mutate()` receives **that exact id** - revert the sourceId fix and it reddens.
- The D17 case: a collection variable with no `enabled` field displays and commits to the leaf the bar displayed, not an ancestor.
- Payload integrity: seed two globals, edit one, assert the payload contains the untouched sibling verbatim and the edited one keeps `enabled`/`secret`/`type` with only `value` changed (today only `toHaveBeenCalledTimes(1)` is asserted - a sibling-wipe regression passes the suite).
- Sequential-blur freshness: commit A, then commit B before A settles; assert B's payload carries A's new value.
- Escape discards (mutate never called, value restored) and Enter commits - the order dependence at `ContextBar.tsx:208-215` is untested today.
- Secret masking: a `secret: true` variable renders the bullet mask readOnly, plaintext not in the document, blur cannot mutate.
- `gone()` refusal paths for env and collection assert the toast fires and no mutate.

Commands: `cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check`.

## Docs

`docs/app/variable-resolution.md` - if it describes the bar's write-back, align it with the sourceId design; one line stating edits write to the resolved winner's source. Verify and state in the PR.

## Sequencing

**First of the context-bar wave**: the polish batch and the alive-context-bar feature both edit this component and land after this issue (same file, ordered). Parallel-safe against everything else open (#197/#287, #303/#340, #337, #338, #339, #134, the PanelResizeHandle Home/End issue). `pnpm lint`/`format:check` CI-enforced.

Out of scope: the bar's markup/a11y (polish batch issue), new sections (feature issue), true parallel-mutation serialization beyond the optimistic-cache fix (would need engine-side single-variable patch semantics - note it in the PR if it bites).

## Acceptance criteria

- [ ] An edit always writes to the definition the bar displayed (`resolved.sourceId`), pinned per scope by tests that fail on the truthy-`enabled` regression.
- [ ] `buildLeafFirstChain` is gone; no unguarded `parentId` walk remains in the component.
- [ ] A second blur-commit no longer reverts a settling first commit (sequential case pinned by test).
- [ ] Mid-edit environment/tab switches can drop an abandoned edit but can never mistarget it (key includes scope+sourceId).
- [ ] Quit-time flush awaits an in-flight commit; the Dock save status reflects it.
- [ ] All new tests green, mutation-checked; `pnpm test`, `pnpm type-check`, `pnpm lint`, `pnpm format:check` green.

---

athrvk:
## Description

**Root cause.** `ContextBar.commitValue` re-derived its own write target instead of using the winner `useVariableResolver` had already picked and displayed, and it used a different rule to do it: `chain.find((col) => col.variables?.[name]?.enabled)` wants `enabled` truthy, while the resolver's `isEnabledDefinition` counts an *absent* `enabled` as enabled (D17). The engine stores variable blobs verbatim, so an `enabled`-absent definition reaches the cache - and a leaf holding one displayed in the bar while an **ancestor's** definition took the write, silently and cross-collection. If that ancestor's definition was `secret: true`, a non-secret input overwrote a secret. When the `enabled`-absent leaf was the *only* definition, `find` returned nothing and the bar toasted "no longer defined in its collection scope" about a variable resolving two inches away.

**Approach.** Commit against `resolved.sourceId`, which the resolver emits precisely to name the winning source - one implementation of "which definition wins" instead of two that can drift. That dissolves the retarget and cycle-walk items together: `buildLeafFirstChain` (a copy of `buildCollectionChain` minus its cycle guard, so a bad database froze the window on the first blur) is deleted, and the component drops its own globals/collections/environments query reads along with it. Payload freshness is handled by reading the map from the query cache at blur time plus an optimistic patch, so a second commit's read already includes the first.

**Alternative rejected.** Fixing the walk in place - swapping the truthy test for `isEnabledDefinition` - would have made this instance agree again while leaving a second copy of the winner rule in the tree, free to drift the next time the resolver changes. The repo's named hand-rolled-copy defect is exactly that; the walk had to go, not be corrected. The optimistic update was also kept local to `ContextBar` rather than added to the three mutations' `onMutate`, since those mutations have other callers (`VariableTableEditor`, the collection tabs) that did not ask for the behaviour change.

Also in the commit path:

- **Payload freshness.** The payload is read from the query cache at blur time and patched into it optimistically rather than spread from the render closure. The transport replaces the map wholesale, so committing B before A settled re-sent A's pre-edit value and reverted it. A refused commit restores that one name rather than the whole snapshot, so it cannot revert a sibling's edit in turn.
- **Input identity.** The key carries scope and source, not just the value. An environment switch or a Ctrl/Cmd+N tab switch mid-edit that resolved to the same string kept the DOM node alive and let the blur write into the newly-resolved definition; it now remounts, so an abandoned edit is dropped - the documented lesser behaviour - and never mistargeted.
- **Quit-time flush.** Commits register one save context (`context-bar-variables`), so `flushAll` awaits an in-flight PUT and the Dock's status reflects it. `failSave` covered only the failure half.
- **Defensive asymmetry.** The collection branch's missing `else` is now `return gone()`.

**Blast radius, verified.** `ContextBar` is the only consumer of any of this - `buildLeafFirstChain` had no other caller, and nothing outside the component reads its commit path. `ResolvedVariable.sourceId` gains a second reader (the popover was the first); its producer is unchanged. No engine change: the verbatim blob storage is by design and D17 normalization is renderer policy.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Local, at the branch head:

- `pnpm test` - **2706 passed** across **295 files** (2700/295 before; the new file adds 14, the existing `ContextBar.save-error.test.tsx` keeps its 2).
- `pnpm type-check`, `pnpm lint`, `pnpm format:check` - all clean.
- `mkdocs build --strict -f .github/mkdocs.yml` - clean, for the three docs pages.

New `ContextBar.commit-scope.test.tsx` pins, all mutation-checked:

- the commit target per scope (global / environment / collection) by **exact source id**;
- the D17 case - an `enabled`-absent leaf commits to the leaf, not the enabled ancestor, and the ancestor's secret stays untouched;
- payload integrity - untouched siblings verbatim, `enabled`/`secret`/`type` preserved, only `value` changed;
- sequential-blur freshness - B's payload carries A's new value with A still in flight;
- rollback on refusal restores only the edited name in the cache;
- Escape discards / Enter commits; secret masking is read-only and cannot mutate;
- both `gone()` refusals (source environment deleted, source collection no longer holds the name) plus a non-global winner with no `sourceId` at all;
- the remount when the winner's source changes under an unchanged value;
- the quit-time flush waiting on an in-flight commit, and the save status reaching `saved`.

Mutation checks run explicitly - reverting the key to `${name}:${resolved.value}`, the collection target to a re-derived one, the optimistic patch, the save-context registration, and the `gone()` refusal each reddens exactly its own case. The flush test needed a macrotask wait rather than a microtask one: a shorter wait passed either way, which is how a missing registration would have shipped green.

`ContextBar.save-error.test.tsx` now seeds the query cache and wraps in a `QueryClientProvider` instead of mocking the read hooks, since the commit no longer reads them. Its two assertions are unchanged.

## Checklist

Docs updated in the same commit: `docs/app/variable-resolution.md` (edits write back to the resolved winner's source, and why a second winner rule is the bug), `docs/app/COMPONENTS.md` (the bar's editing behaviour), `docs/app/state-management.md` (the bar's save-context registration, beside the existing `failSave` note).

None. Item 3's optimistic-update location was left to the implementer; it lives in `ContextBar` for the reason given above. True parallel-mutation serialization stays out of scope as the issue states - it would need engine-side single-variable patch semantics; the sequential-blur case, which is the reachable one, is closed.

## Related Issues
Closes #341

---
_Generated by [Claude Code](https://claude.ai/code/session_01V57ge8dNDxNgkt5Njoe6rT)_

### DIFF
--- a/app/src/components/layout/ContextBar.commit-scope.test.tsx
+++ b/app/src/components/layout/ContextBar.commit-scope.test.tsx
@@ -1,0 +1,413 @@
+/**
+ * @vitest-environment jsdom
+ */
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * Where a context-bar edit lands, and what it carries when it gets there.
+ *
+ * The bar displays the definition the *resolver* picked and used to write back
+ * to one it re-derived itself, with a different rule: the resolver counts an
+ * absent `enabled` as enabled (D17), the bar's chain walk wanted it truthy. A
+ * leaf definition with no `enabled` key therefore showed on screen while an
+ * ancestor's definition took the write - silently, cross-collection, and with
+ * the typed text still sitting in the input looking committed.
+ *
+ * The payload had the matching problem in time rather than space: it was spread
+ * from the render-closure copy of the whole map, and the transport replaces the
+ * map wholesale, so a second blur re-sent the first edit's pre-edit value.
+ *
+ * Every case here is mutation-checked - each one reddens if the commit goes back
+ * to re-deriving its target or to the render-time snapshot.
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render, screen, fireEvent, act } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { ContextBar } from "./ContextBar";
+import { queryKeys } from "@/queries/keys";
+import { useToastStore } from "@/stores/toast-store";
+import { useSaveStore } from "@/stores/save-store";
+import type { Collection, Environment, ResolvedVariable, VariableValue } from "@/types";
+
+type Handlers = { onError: (e: unknown) => void; onSettled: () => void };
+type VariableMap = Record<string, VariableValue>;
+
+const globalsMutate = vi.fn();
+const environmentMutate = vi.fn();
+const collectionMutate = vi.fn();
+
+vi.mock("@/queries", () => ({
+	useRequestQuery: () => ({ data: { id: "req_1", collectionId: "col_leaf" } }),
+	useUpdateGlobalsMutation: () => ({ mutate: globalsMutate }),
+	useUpdateEnvironmentMutation: () => ({ mutate: environmentMutate }),
+	useUpdateCollectionMutation: () => ({ mutate: collectionMutate }),
+}));
+
+/** The resolver's verdict for this render - the winner, and which source it won from. */
+let resolved: Record<string, ResolvedVariable> = {};
+
+vi.mock("@/hooks/useVariableResolver", () => ({
+	useVariableResolver: () => ({ getAllVariables: () => resolved }),
+}));
+
+const layoutStore = {
+	contextBarOpen: true,
+	setContextBarOpen: vi.fn(),
+	contextBarWidth: 280,
+	setContextBarWidth: vi.fn(),
+};
+const tabsStore = {
+	openTabs: [{ id: "t1", type: "request", entityId: "req_1" }],
+	activeTabId: "t1",
+};
+
+vi.mock("@/stores", async () => {
+	const saveStore =
+		await vi.importActual<typeof import("@/stores/save-store")>("@/stores/save-store");
+	return {
+		useLayoutStore: () => layoutStore,
+		useTabsStore: () => tabsStore,
+		useSaveStore: saveStore.useSaveStore,
+	};
+});
+
+const def = (value: string, extra: Partial<VariableValue> = {}): VariableValue => ({
+	value,
+	enabled: true,
+	...extra,
+});
+
+const collection = (id: string, variables: VariableMap, parentId?: string): Collection => ({
+	id,
+	name: id,
+	description: "",
+	parentId,
+	order: 0,
+	variables,
+	auth: { mode: "noauth" },
+	preRequestScript: "",
+	postRequestScript: "",
+	createdAt: "",
+	updatedAt: "",
+});
+
+const environment = (id: string, variables: VariableMap): Environment => ({
+	id,
+	name: id,
+	description: "",
+	variables,
+	isActive: false,
+	createdAt: "",
+	updatedAt: "",
+});
+
+/**
+ * The leaf defines `apiKey` with **no `enabled` key** and the ancestor defines
+ * it enabled - the exact shape the engine stores verbatim and D17 resolves to
+ * the leaf. The old truthy-`enabled` walk skipped the leaf and wrote the
+ * ancestor. The cast is the point of the fixture: `enabled` is required by the
+ * type and absent in the stored blob, which is what D17 exists to describe.
+ */
+const collections: Collection[] = [
+	collection("col_root", { apiKey: def("ancestor-key", { secret: true }) }),
+	collection("col_leaf", { apiKey: { value: "leaf-key" } as VariableValue }, "col_root"),
+];
+
+const environments: Environment[] = [
+	environment("env_a", { token: def("a-token") }),
+	environment("env_b", { token: def("b-token") }),
+];
+
+function seed(client: QueryClient) {
+	client.setQueryData(queryKeys.globals.all, {
+		id: "globals",
+		updatedAt: "",
+		variables: {
+			host: def("example.com", { secret: false, type: "string" }),
+			port: def("8080", { type: "number" }),
+		},
+	});
+	client.setQueryData(queryKeys.environments.list(), environments);
+	client.setQueryData(queryKeys.collections.list(), collections);
+}
+
+function renderBar() {
+	const client = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+	seed(client);
+	const view = render(
+		<QueryClientProvider client={client}>
+			<ContextBar />
+		</QueryClientProvider>
+	);
+	return { client, ...view };
+}
+
+function editAndBlur(input: HTMLInputElement, value: string) {
+	act(() => {
+		fireEvent.change(input, { target: { value } });
+		fireEvent.blur(input);
+	});
+}
+
+function inputFor(displayValue: string): HTMLInputElement {
+	return screen.getByDisplayValue(displayValue) as HTMLInputElement;
+}
+
+beforeEach(() => {
+	globalsMutate.mockReset();
+	environmentMutate.mockReset();
+	collectionMutate.mockReset();
+	useToastStore.setState({ toasts: [] });
+	useSaveStore.getState().reset();
+	resolved = {};
+});
+
+describe("ContextBar - the commit target is the resolver's winner", () => {
+	it("writes a global edit into the globals map", () => {
+		resolved = { host: { value: "example.com", scope: "global" } };
+
+		renderBar();
+		editAndBlur(inputFor("example.com"), "example.org");
+
+		expect(globalsMutate).toHaveBeenCalledTimes(1);
+		expect(globalsMutate.mock.calls[0][0].variables.host.value).toBe("example.org");
+	});
+
+	it("writes an environment edit into the source environment, not the active one", () => {
+		// The winner came from `env_b`. Nothing in the commit path may consult the
+		// session's active environment - the two are the same id only by luck.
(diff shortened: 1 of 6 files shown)


athrvk:
CI has not run on this branch, and not because of anything in the diff: **GitHub Actions is in a declared major outage** and the push that opened this pull request created **zero workflow runs**.

From `githubstatus.com`, latest update 20:34 UTC (incident opened 15:22 UTC, still `major_outage`):

> Webhook triggers are currently throttled to help with recovery and we are processing approximately 15% of webhooks, so many events such as pushes and pull requests are not triggering workflow runs.

That is this pull request exactly - `list_workflow_runs` for `claude/eloquent-cray-rphoef` returns `total_count: 0`, and there are no check runs on the head commit. There is no re-run to request, since no run exists, and `pr-tests.yml` is `pull_request`-only by design, so there is no `workflow_dispatch` to force one either. PR #345 is stuck one step differently on the same incident - its re-runs have been `queued` since 18:22 with zero jobs created, and the API now refuses to cancel them (`409 Cannot cancel a workflow re-run that has not yet queued`).

**Verified locally instead**, at this branch head against base `f703eaa`:

- `pnpm test` - **2706 passed** across **295 files**.
- `mkdocs build --strict -f .github/mkdocs.yml` - clean, for the three docs pages.
- Every new test mutation-checked by reverting the corresponding fix; each reddens only its own case. That pass caught a weak assertion in the quit-flush test - a microtask wait passed even with the save-context registration removed, so it now waits a macrotask.

No engine change here, so the engine suite is untouched by this diff.

**Suggested handling:** nothing to change in the branch. Once GitHub marks the incident resolved, a push (or closing and reopening the pull request) will re-dispatch the `pull_request` event and CI should schedule normally. I am subscribed and will report when runs start appearing.

---

PR #347, opened 2026-08-06
fix(engine): build the raw-request view from the wire, not the composed headers

### LINKED ISSUE #339: engine(http): the raw-request view omits the Cookie header the jar actually sent - request_headers never sees libcurl's jar-attached cookies
Follow-up from the #301 closure verification (2026-08-06, master `f703eaa`). The cookie jar (PR #336) introduced the first case where the engine sends a header the raw-request view does not show.

> **Decision (owner, 2026-08-06): Option A - capture libcurl's actual outbound header frame - and the raw view shows the cookie value plainly.** Rationale: on the Postman half of Vayu's target, the raw view exists to show exactly what was sent, and Postman's console shows real headers unredacted; the value is already readable in Settings and belongs to the user's own machine. The verbose transfer log keeps its redaction (`debug_redact.hpp`) because logs get exported and shared - the two surfaces have different audiences and keep different policies, stated in the docs.

## Problem

The "raw request" string shown in the response viewer is built from `response.request_headers` (`engine/src/http/client.cpp:461`, the loop inside the raw-request builder around `:410-480`), and `request_headers` is populated from the composed request's own headers plus the injected `User-Agent` (`client.cpp:271`, `:296`). Jar cookies take a different path entirely: they are injected into the curl handle via `CURLOPT_COOKIELIST` (`client.cpp:154-157`) and libcurl synthesizes the `Cookie` header on the wire itself - `request_headers` never sees it. So after a design-mode send in an environment whose jar holds a session cookie, the wire carries `Cookie: session=...` and the raw-request view says nothing was sent. Pre-jar, nothing was auto-attached, so the view was accurate; the jar made it silently wrong, in a tool whose purpose is showing users exactly what their request was.

The verbose transfer log does show the attached cookie (redacted per `debug_redact.hpp`), and Settings' Cookies card and `pm.cookies` both show the jar contents - but none of those is the raw-request view the user actually reads.

Severity: low-medium - no data is corrupted and the information exists elsewhere, but it is silent misinformation in the primary debugging surface, and "why did this request succeed when the raw view shows no auth" is exactly the confusion a raw view exists to prevent.

## Fix / solution pathway

**Capture `CURLINFO_HEADER_OUT` and use it as the raw-request header block** (decision above). The `CURLOPT_DEBUGFUNCTION` callback already exists for verbose logging; retain the outbound header frame (the *last* `HEADER_OUT` of the transfer, so redirects show the final request) and build the raw-request string's header section from it verbatim instead of from `request_headers`. This fixes the Cookie omission and any future divergence (libcurl-added `Host`, `Accept`, `Content-Length`, h2 pseudo-header rendering) in one move - the view becomes what was sent, by construction.

Constraints:

- Keep `response.request_headers` itself unchanged - it feeds other consumers; grep before touching and state the consumer list in the PR.
- No redaction in the raw view (decision above); the verbose log's redaction stays as is.
- The rejected alternative, for the record: merging the jar's matched cookies into the view synthetically would re-implement libcurl's domain/path/secure/expiry matching and drift - the exact trap the jar's design avoided by letting libcurl own matching.

## App changes

None expected - the renderer renders the engine-provided raw-request string as-is. Verify with a grep for `request_headers` consumers in `app/src` and state the result in the PR ("none, verified" acceptable). MCP: `run_request` results include response data - verify whether the raw-request string is part of any MCP tool result shape and note it.

## Tests

- Engine gtest (beside the `CookieJarTransfer` suite): a request in an environment whose jar holds a matching cookie - the raw-request string contains the `Cookie` header that was sent, value visible; a non-matching cookie does not appear. Mutation-checked: revert the capture and the first assertion fails.
- A no-jar request's raw view: the new view is header-set equivalent or a deliberate superset of the composed headers (libcurl adds `Host`/`Accept` the old builder synthesized differently) - pin the new form and update the existing raw-request expectations with intent, not incidentally.
- A redirected request's raw view shows the final transfer's headers (the last `HEADER_OUT` frame).
- Existing raw-request/protocol tests stay green or are updated with intent.

Commands: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`; `cd app && pnpm test` if any renderer expectation moves.

## Docs

`docs/engine/architecture.md` - the Cookie Jar section's display story, one sentence, including the two-surfaces redaction policy (raw view plain, verbose log redacted). If the raw view's shape changes visibly, whatever doc describes the response viewer's raw tab. Same commit.

## Sequencing

Parallel-safe against #337 (the write half) - both touch the jar's neighbourhood but this is `client.cpp`'s raw-request builder while #337 is `script_engine.cpp`/`cookie_jar.*`; no shared hunks. Parallel-safe against #338, #197/#287, #303/#340, #134. If #337 lands first, its script-written cookies are covered by this fix automatically; if this lands first, same result - order-independent.

Out of scope: redaction policy for the verbose log (already shipped, unchanged), and any persistence of the jar (#301's closed scoping holds).

## Acceptance criteria

- [ ] After a send that attached jar cookies, the raw-request view shows the `Cookie` header as sent, value plain, sourced from the captured `HEADER_OUT` frame.
- [ ] A request with no matching jar cookies shows no `Cookie` line, pinned by test; redirects show the final transfer.
- [ ] `response.request_headers` consumers unchanged, verified and listed in the PR.
- [ ] Engine suite green; app suite and type-check green if any expectation moved; docs updated same commit.

---

athrvk:
## Description

**Plan.** Root cause: `response.request_headers` is the *composed* header map, and the cookie jar hands its cookies to libcurl through `CURLOPT_COOKIELIST` - libcurl synthesizes the `Cookie` header on the wire itself, so the raw-request builder (which read `request_headers`) could not see it and reported a request that was never sent. Approach, per the issue's decision: capture the transfer's last `CURLINFO_HEADER_OUT` frame in the debug callback and use it as the raw view's header block, so the view is the wire by construction - which also removes every future divergence (`Accept`, `Content-Length`, h2 rendered in HTTP/1 form) rather than just this one. The rejected alternative is the one the issue records: merging the jar's matched cookies into the view synthetically would re-implement libcurl's domain/path/`Secure`/expiry matching and drift from it - the exact trap the jar's design avoided by letting libcurl own matching.

Verified the problem still exists as described at `f703eaa`; the anchors in the issue (`client.cpp:154-157`, `:271`, `:296`, the builder at `:410-480`) were all accurate.

Changes:

- `debug_callback` now carries a `TransferDebug` through `CURLOPT_DEBUGDATA` and keeps the **last** outbound header frame verbatim. Last, not first, so a followed redirect describes the hop that produced the response beside it.
- The debug stream is always on; `ClientConfig::verbose` now gates only the *logging*. The capture is unredacted (the raw view exists to be exact, and the same value is already readable in Settings); the verbose log keeps `debug_redact` because logs get exported and shared. Two surfaces, two audiences, two policies - stated in the docs.
- The old builder is extracted as `synthesize_raw_request` and kept as the fallback for a transfer that never reached the wire (DNS failure, connection refused), which is all that ever existed for those.
- `raw_request_from_wire` re-normalizes the frame's terminator before appending the body, so the body cannot run into the last header line.

**`response.request_headers` is unchanged** - still the composed headers plus the injected `User-Agent`. Consumers, verified by grep: engine `utils/json.cpp:497` (`requestHeaders`), and on the app side `execute-mapping.ts`, `restore-response.ts`, `response-store.ts` and `HeadersViewer`/`ResponseHeadersPanel`. None of them read `rawRequest` for header data.

**App changes: none, verified.** `RawRequestResponse` renders the engine's string as-is (`ResponseViewer/index.tsx:350`). **MCP:** `run_request` returns the engine's `/execute` response unshaped (`app/electron/mcp/tools.ts`, `callEngine(... executeRequest ...)`), so `rawRequest` reaches MCP clients too and now includes the cookie line - noted rather than changed, since that is the same disclosure the decision made for the raw view.

**One divergence worth naming, not fixed here:** a *restored* trace (history) rebuilds its raw string client-side with `buildRawRequest` from the stored composed request, so it will still show no `Cookie` line. Closing that would mean persisting the sent cookie in `trace_data`, which is credential-grade material the jar deliberately keeps in memory only - an owner decision, filed separately rather than smuggled into this diff.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

CI could not run on this branch: GitHub Actions is in a declared major outage (incident opened 15:22 UTC, still `major_outage` at 22:18 UTC, webhook triggers throttled so pushes create no workflow runs). Verified locally at this branch head against base `f703eaa`:

- `python build.py -e -t` then `cd engine && ctest --preset linux-dev --output-on-failure` - **891/891 passed** (887 before this change).
- No app file changed and no renderer expectation moved, so the app suite is untouched by this diff.
- `clang-format --dry-run` on the three touched files reports only pre-existing violations on lines this diff does not touch; nothing was reformatted (repo rule).

Mutation-checked by forcing `synthesize_raw_request` on every path and rebuilding: `CookieJarTransfer.TheRawRequestViewShowsTheCookieThatWasSent`, `HttpClientTest.RawRequestIsTheHeaderBlockThatWentOnTheWire` and `HttpClientTest.RawRequestOfAFollowedRedirectShowsTheFinalHop` all fail (the cookie test prints the header block with no `Cookie` line - the bug itself). The POST test asserts libcurl's `Accept: */*` alongside the body join for the same reason: no composed header map contains it, so it cannot pass against a synthesized block.

New tests:

- `CookieJarTransfer.TheRawRequestViewShowsTheCookieThatWasSent` - a jar cookie appears in the raw view with its value; a scope with no matching cookie gains no `Cookie` line.
- `HttpClientTest.RawRequestIsTheHeaderBlockThatWentOnTheWire` - pins the new form (request line, `Host`, the composed header, the injected `User-Agent`, libcurl's `Accept`, terminated block).
- `HttpClientTest.RawRequestOfAFollowedRedirectShowsTheFinalHop` - `/redirect/1` → the view names `GET /get`.
- `HttpClientTest.RawRequestKeepsTheBodyAfterTheWireHeaders` - the body survives the join exactly once.

`UnreachableHostsRawRequestNamesWhatWasAskedFor` now also pins the synthesized fallback; its comment says so.

## Checklist

Docs updated in the same commit: `docs/engine/architecture.md` (Cookie Jar - the display story and the two-surface redaction policy) and `docs/engine/api-reference.md` (what `rawRequest` is, how it differs from `requestHeaders`, redirects, and the failed-transfer fallback).

## Related Issues
Closes #339

---
_Generated by [Claude Code](https://claude.ai/code/session_016XtsB6kAJ6AM7TWzdGaBit)_

### DIFF
--- a/docs/engine/api-reference.md
+++ b/docs/engine/api-reference.md
@@ -1243,6 +1243,17 @@ A client splitting a folded `Set-Cookie` must split on a comma followed by
 for the response headers stored on a design run's `trace_data.response` and on
 captured load-run samples, which come off the same parse.
 
+**`rawRequest` is the header block libcurl actually sent**, captured from the
+transfer's last outbound header frame and followed by the request body. So it
+carries what libcurl added on its own - the `Cookie` line the
+[cookie jar](architecture.md#cookie-jar) matched, `Accept`, `Content-Length`,
+and an h2 request rendered in HTTP/1 form - none of which appear in
+`requestHeaders`, which stays the *composed* headers. Values are not redacted:
+this field exists to say exactly what went out. On a followed redirect it is the
+final hop, matching the response beside it. A transfer that failed before
+sending anything (DNS failure, connection refused) has no frame to read, and
+falls back to a request synthesized from what was composed.
+
 **`consoleLogs` entries carry their own source and level.** `source` is which of
 the request's two scripts wrote the line (`"pre"` for the pre-request script,
 `"test"` for the post-request one) and `level` is the `console.*` method that was
--- a/docs/engine/architecture.md
+++ b/docs/engine/architecture.md
@@ -210,6 +210,13 @@ makes a session survive from one design-mode request to the next.
   actually share, and a load run repeats a single request anyway.
 - **Threading:** one mutex around the scope map; every accessor copies out, so
   no reference into the storage escapes to a caller.
+- **Shown as sent.** Because libcurl attaches the matching cookies itself, the
+  composed header map never sees them - so the raw-request view is built from
+  the transfer's last `CURLINFO_HEADER_OUT` frame (`client.cpp`) rather than
+  from `request_headers`, and shows the `Cookie` line the wire carried, value
+  plain. The verbose transfer log keeps its redaction (`debug_redact.hpp`): a
+  log gets exported and shared, the raw view is read on the machine whose
+  Settings already display the same value.
 
 `pm.sendRequest` shares the jar of the execute it runs inside, so a pre-request
 script can log in and leave the session where the real request will find it.
--- a/engine/src/http/client.cpp
+++ b/engine/src/http/client.cpp
@@ -45,9 +45,40 @@ namespace vayu::http {
 
 namespace {
 
+/**
+ * @brief What a transfer's debug stream leaves behind for the caller.
+ *
+ * `verbose` gates the *logging* only - the outbound header frame is captured on
+ * every send, because the raw-request view is built from it (issue #339). Since
+ * the cookie jar, libcurl synthesizes headers we never composed (`Cookie`, from
+ * CURLOPT_COOKIELIST), so the composed header map no longer describes what went
+ * on the wire and only libcurl can say what did.
+ *
+ * The captured frame is deliberately **not** redacted. It backs the raw-request
+ * view, which exists to show exactly what was sent and whose values the user can
+ * already read in Settings; the verbose log below keeps its redaction because
+ * logs get exported and shared. Two surfaces, two audiences, two policies.
+ */
+struct TransferDebug {
+    bool verbose = false;
+    /// The last CURLINFO_HEADER_OUT frame, verbatim. Last, not first, so a
+    /// followed redirect reports the request that produced the response the
+    /// caller is looking at.
+    std::string last_header_out;
+};
+
 int debug_callback (CURL* handle, curl_infotype type, char* data, size_t size, void* userptr) {
     (void)handle;
-    (void)userptr;
+    auto* debug = static_cast<TransferDebug*> (userptr);
+
+    if (debug != nullptr && type == CURLINFO_HEADER_OUT) {
+        debug->last_header_out.assign (data, size);
+    }
+    // Nothing else to do when the caller only wanted the capture: the frames
+    // below are logging, and DATA_OUT can be a whole request body.
+    if (debug == nullptr || !debug->verbose) {
+        return 0;
+    }
 
     std::string text (data, size);
     // Remove trailing newlines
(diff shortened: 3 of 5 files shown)


athrvk:
CI has not scheduled on this branch, and not because of the diff: **GitHub Actions is still in the declared major outage** that opened at 15:22 UTC (latest status update 22:18 UTC - runners draining, webhook triggers throttled). The push that opened this pull request created zero workflow runs, the same way #346's did; I also closed and reopened #345 and #346 at 22:45 UTC to re-dispatch their `pull_request` events and neither produced a run either.

Verified locally at `95b4728` against base `f703eaa` instead - full detail in the PR body:

- `python build.py -e -t` then `ctest --preset linux-dev --output-on-failure` - **891/891 passed** (887 on the base commit).
- Every new test mutation-checked by forcing the old synthesized builder back on: three of the four redden, the fourth pins the body join against libcurl's own `Accept`.

No app file is touched and no renderer expectation moved, so the app suite is unaffected by this diff.

Follow-up filed as #348: the *restored* (history) raw view still rebuilds from the composed request, so it shows no `Cookie` line. Closing that means persisting a sent session cookie to the request store, which the jar's design deliberately avoids - a decision for you rather than something to fold into this PR.

Once the incident is resolved, a push or a close/reopen here should re-dispatch and CI ought to run clean.

---

PR #349, opened 2026-08-07
feat(engine): the cookie jar's write half - pm.cookies.jar()

### LINKED ISSUE #337: engine(scripting): the cookie jar's write half (`pm.cookies.set/unset/clear`) - mutating a jar concurrent transfers are reading
Split out of #301 (step 2), which shipped the jar and its read half in PR #336. Filed rather than folded in, because the write half raises exactly one question the read half does not and it is not a small one.

Anchors verified at master `f703eaa` (PR #336 merged as that commit; re-pinned and closure-verified 2026-08-06).

> **Decisions (owner, 2026-08-06)** - the forks below are resolved; the rationale is Postman parity (Vayu's target is the Postman + JMeter + k6 combination):
> 1. **Shape: Postman's `pm.cookies.jar()` ships, whole.** URL-scoped `get(url, name, cb)` / `set(url, cookie, cb)` / `unset(url, name, cb)` / `clear(cb)`, callback-shaped with the synchronous-under-the-hood execution model `pm.sendRequest` already established (PR #309). No flat `set(name, value)` - the read half is flat because Postman's own `pm.cookies` CookieList reads are flat; the write half in Postman exists only on the jar, so `jar()` is the drop-in-compatible surface for scripts ported from Postman. Only one shape ships, as this issue required.
> 2. **`clear()` is allowed, scoped to the current environment's jar.** Compat wins: Postman scripts use `jar.clear()`, the jar is in-memory with process lifetime so the blast radius is a session reset (not data loss), the Settings Cookies card shows the result, and "reset my session" is a legitimate script need. Document it plainly in `scripting.md`; no confirmation gate for scripts.
> 3. **Interleaving: a script write is applied to the jar after the enclosing transfer's capture** (not into the live map mid-transfer), making write-then-capture-discard impossible for the writing script itself; the concurrent-`/execute` lost-update window below remains the documented pre-existing behavior for received cookies and is not widened.

## Problem

`pm.cookies.get` / `.has` / `.toObject` read the jar (`engine/src/runtime/script_engine.cpp`, the `pm.cookies` section; jar in `engine/include/vayu/http/cookie_jar.hpp`). The write half - Postman's `pm.cookies.set(name, value)` / `.unset(name)` / `.clear()`, and the `pm.cookies.jar()` object those really hang off in Postman - is **not bound**, deliberately. A script that reaches for one meets "not a function" today.

The blocker is not the binding, it is what a write means:

- **The mutation races concurrent transfers.** A jar write from a pre-request script lands while the enclosing `/execute` has not yet sent, which is fine. And `pm.sendRequest` is *not* the risk it first appears to be: it is synchronous (PR #309), the script is blocked while its auxiliary transfer runs, so a script cannot write during one. The real interleaving is **two concurrent `/execute` runs in the same environment** (a UI Send racing an MCP `run_request`): each seeds its curl handle from the jar at inject, and `capture_jar_cookies` **replaces** the scope's whole line list with what the finishing handle held (`engine/src/http/client.cpp:169-182` -> `CookieJar::store`, `engine/src/http/cookie_jar.cpp:174-184`) - correct for a lone transfer, but it silently discards whatever the other transfer captured, and would discard a script's write that landed in between. That lost-update window already exists today for *received* cookies; the write half must not widen it to deliberate writes - resolved by decision 3 above.
- **A written cookie has no `Set-Cookie` to come from.** The jar stores libcurl's Netscape lines, deliberately, so libcurl owns matching and expiry. A script-set cookie has to become one of those lines - and the decided `jar()` shape answers the domain/path question the flat shape could not: `set(url, cookie, cb)` carries the URL, exactly why Postman's API is URL-scoped. Fields the cookie object omits default from that URL through one shared helper (see pathway step 3).
- **`clear()` crosses a user-visible boundary.** Resolved by decision 2: allowed, environment-scoped, documented.

## Fix / solution pathway

1. Bind `pm.cookies.jar()` returning the jar object with `get` / `set` / `unset` / `clear` per decision 1, callback-shaped, against the current environment's jar. `jar().get` reuses the read half's matching; it exists for compat even though flat reads are available.
2. Implement the write-after-capture ordering per decision 3: a script write is queued on the `ScriptResult`/execution context and applied to the jar after the enclosing transfer's capture step, then visible to the next transfer. State the `pm.sendRequest` non-case (synchronous, no intra-script race) in the PR so the next reader does not re-derive it. Within one script, a `jar().set` followed by `pm.sendRequest` must see the write - the queue applies pending writes before an auxiliary transfer injects.
3. Build the Netscape line from the `set(url, cookie)` arguments through one helper beside `parse_cookie_line`, so a written cookie and a received one are the same kind of thing; URL-derived defaults for domain/path, explicit cookie fields win.
4. `clear()` empties the current environment's jar (decision 2); `unset(url, name)` removes what it names.
5. Bind the new methods in the completions surface too (`engine/src/http/routes/scripting.cpp`, the `pm.cookies` completion entries beside the read half's).

## App changes

None expected - `CookiesCard` reads `GET /cookies` and shows whatever is there, so a script-written cookie appears with no app change. Verify that rather than assume it; if a script-written cookie needs to be distinguishable from a received one in the panel, that is a new field and needs a reader.

## Tests

- A script `jar().set` followed by a request to a matching URL: the cookie is on the wire.
- The same cookie **not** sent to a URL it should not match (wrong host, wrong path, `Secure` over cleartext) - the write half must not be a way around the matching the read half respects.
- A `jar().set` followed by `pm.sendRequest` in the same script: the auxiliary transfer sends the written cookie (pending writes apply before inject).
- A `set` inside a `pm.sendRequest` callback: runs after the auxiliary transfer completes (synchronous model), so it is a sequential write - pin that with a test rather than assuming it.
- The write-after-capture ordering from decision 3, pinned by a test that stages a write during an enclosing transfer and asserts the write survives the capture.
- `unset` / `clear` against `GET /cookies` afterwards; `clear` empties only the current environment's scope.
- Completions: the new `jar`/`set`/`unset`/`clear` entries are offered - note the existing completions guard only checks offered-implies-callable, not callable-implies-offered, so forgetting the entries fails nothing today; assert the entries explicitly.
- The isolation test that already exists (`CookieJarTransfer.CookiesDoNotCrossAnEnvironmentBoundary`) must still hold with writes in play.

Commands: `python build.py -t && cd engine && ctest --preset linux-dev --output-on-failure`, and `cd app && pnpm test && pnpm type-check` if anything app-side moves.

## Docs (same commit)

- `docs/engine/scripting.md` - "The cookie jar (`pm.cookies`)" currently says the write half is not bound; that paragraph is the one to rewrite, including the `clear()` policy (decision 2) and the write-after-capture ordering (decision 3).
- `docs/app/pm-api-compatibility.md` - move `pm.cookies.jar()` and its methods out of "Not (yet) supported" exactly as far as what ships.
- `docs/engine/architecture.md` - the Cookie Jar section, if the write path changes the threading story.

## Sequencing

PR #336 (the jar) has merged as `f703eaa` - the prerequisite is satisfied. #301 is closed with this issue as its successor. Parallel-safe against everything else open (#197/#287 perf harness, #303/#340 runner track, #338, #339, #134 parked): it touches `script_engine.cpp` and `cookie_jar.*` only, and stays off the load path for the same reason the jar does. #339 (raw-request Cookie display) is independent; if it lands first with its Option A capture, script-written cookies are covered by it automatically.

## Acceptance criteria

- [ ] `pm.cookies.jar()` ships with `get`/`set`/`unset`/`clear` per decision 1; no flat write API is added.
- [ ] The write-after-capture ordering (decision 3) and the `pm.sendRequest` non-case are stated in the PR and pinned by tests, not left emergent.
- [ ] A script-written cookie is matched by the same rules a received one is - covered by a test that it is **not** sent where it should not match.
- [ ] `clear()` is environment-scoped and documented per decision 2.
- [ ] The completion entries for the new methods ship with the binding and are asserted by a test.
- [ ] Docs above updated in the same commit.
- [ ] `python build.py -t && cd engine && ctest --preset linux-dev --output-on-failure`, `cd app && pnpm test && pnpm type-check` all green.

---

athrvk:
## Description

The jar could be read but not written to. This ships Postman's `pm.cookies.jar()` whole - `get` / `set` / `unset` / `clear` - per the three decisions recorded on #337.

**The plan.** The root cause is not a missing binding, it is the ordering: `capture_jar_cookies` **replaces** a scope's contents with what the finishing handle held (`client.cpp`), so a write dropped into the live map beside an in-flight transfer is discarded by it. So a write is *staged* rather than applied where it is made - `jar().set` pushes a `CookieWrite`, and the execution's next transfer applies the staged list on top of the stored lines when it seeds its handle (`ClientConfig::cookie_writes`). The write therefore rides the request it was made before, and that transfer's own capture is what persists it; a write with no transfer left to ride (a post-request script's) is applied by the route through `CookieJar::apply`. Exactly once either way. **The alternative I rejected** was the simpler "route applies the writes after `client.send()` returns": it is one line shorter, but the write then does not ride the request it was made for (a pre-request `set` would not be on that request's wire, which is what a Postman user expects), and a staged `clear()` would be applied twice - once by the transfer's capture and again afterwards - wiping the cookies the response had just set.

**The `pm.sendRequest` non-case**, stated so the next reader does not re-derive it: `pm.sendRequest` is synchronous (PR #309), so a script cannot write *during* an auxiliary transfer. It drains the queue into that transfer, which is why a `set` before it is carried by it, and a `set` inside its callback is a sequential write that lands on the next transfer instead. Both are pinned by tests rather than left emergent.

**Shape.** No flat `pm.cookies.set(name, value)` - a written cookie needs a URL to take its domain and path from, which is why Postman's write half hangs off `jar()`. `set(url, name, value)` *is* accepted alongside `set(url, cookie)` because both are URL-scoped and it is the spelling half of Postman's own docs use; the property decision 1 required is preserved. `cookie_for_url` + `format_cookie_line` build the Netscape line beside `parse_cookie_line`, so a written cookie and a received one are the same kind of thing.

Deviations from the issue text, both deliberate:

- **`jar()` itself does not throw where there is no jar.** It is an accessor; the sentence explaining the absence belongs on the method the script actually called. This also keeps `typeof pm.cookies.jar().set` answerable, which is what lets the existing offered-implies-callable completions guard cover the new methods for free.
- **The Secure case is asserted through the read views, not the wire.** libcurl treats a loopback host as a trustworthy origin (RFC 6265bis §4.1.2.5), so it sends a `Secure` cookie to `http://127.0.0.1` by design and no local test can show otherwise. The test asserts the flag survives the write and that the matcher behind `pm.cookies` refuses it over cleartext, with the reason in a comment.

**Robustness.** Bad input is refused with an error naming the field rather than stored wrong: a non-string `value`, `secure: "yes"`, a date string in `expires`, an unparseable URL, or a field carrying a tab or newline - the Netscape format's own separators, which would corrupt the line and make the cookie vanish on the next read. `clear()` is scoped to the current environment's jar (decision 2).

## Type of Change
- [x] New feature
- [x] Documentation update

## Testing

`python build.py -e -t` then `ctest --preset linux-dev --output-on-failure` - **910/910 passed** on the merged tree (891 on the current base `1d7b6c0`; 19 new). No pre-existing failures.

New coverage in `cookie_jar_test.cpp` (16 wire/script tests, 2 pure-value tests) and `script_completions_test.cpp` (1):

- A `jar().set` is on the wire of the request it was set for, survives that transfer's capture, and appears in the raw-request view.
- It is **not** sent to the wrong host or the wrong path; the `Secure` flag survives the write and the read views honour it.
- The ordering itself: a write staged before a transfer whose response rewrites the whole jar (a `/logout` expiry) still survives.
- `pm.sendRequest` carries a write staged before it; a `set` inside its callback does not reach that transfer and does reach the next one.
- A read - both `jar().get` and the flat `pm.cookies.get` - sees this script's own staged write.
- `unset` removes only what the URL would have carried; `clear` empties only this environment's jar; a written cookie does not cross an environment boundary.
- Nine bad-input cases, each asserting the message names the problem, plus "no jar" and "nowhere to apply" both saying why.
- A script-written cookie appears in `GET /cookies`, which is what makes "no app change" a checked claim rather than an assumption.
- The four new completion entries are asserted by name - the existing guard is offered-implies-callable only, so forgetting them would have failed nothing.

**Mutation-checked**, both directions of the ordering: reverting the transfer's application of staged writes reddens 7 tests; reverting the `pm.sendRequest` drain and the staged-read view reddens 3 more. Restored and re-run green.

`mkdocs build --strict -f .github/mkdocs.yml` clean. `clang-format` clean on every file that was clean at the base commit, and unchanged-in-count on the four that were not (formatted my added lines only, per the repo's no-repo-wide-formatting rule).

**The app suite was not run**: no file under `app/` is touched by this branch and no renderer expectation moves, so no test there could change colour.

## Merge with master

Master moved while this was open (#345, #346, #347 all landed). Merged in: the only conflict was the Cookie Jar section of `docs/engine/architecture.md`, which #347 also extended. Both bullets belong, so both are kept - and since #347 makes the raw-request view read the transfer's own header frame, a script-written cookie now shows up there too. That is asserted rather than left as a doc claim. `client.cpp` merged cleanly: #347's wire capture and this branch's staged-write injection are different parts of the send path.

## Checklist
- [x] I have added tests

Docs updated in the same commit: `docs/engine/scripting.md` (the read-only paragraph rewritten, plus a "Writing to the jar" section covering the field defaults, the refusals, when a write takes effect and the `clear()` policy), `docs/app/pm-api-compatibility.md` (moved out of "Not (yet) supported", with the four divergences from Postman named), and `docs/engine/architecture.md` (the Cookie Jar section's threading/write story).

## Related Issues
Closes #337

---
_Generated by [Claude Code](https://claude.ai/code/session_01JDXiPXGaip7yQJVipoMeD8)_

### DIFF
--- a/docs/app/pm-api-compatibility.md
+++ b/docs/app/pm-api-compatibility.md
@@ -20,7 +20,7 @@ intent is that the most common Postman scripts paste in and run unchanged.
 | Response            | `pm.response.code`, `.status`, `.responseTime`, `.headers`, `.json()`, `.text()`, `.reason()`, `.size()` |
 | Response headers    | `pm.response.headers.get(name)`, `.has(name)` - case-insensitive              |
 | Response cookies    | `pm.response.cookies` (array of `{ name, value, attrs }`), `.get(name)`, `.has(name)`, `.toObject()` - read-only, see below |
-| Cookie jar          | `pm.cookies.get(name)`, `.has(name)`, `.toObject()` - the stored session for this URL, read-only, see below |
+| Cookie jar          | `pm.cookies.get(name)`, `.has(name)`, `.toObject()` - the stored session for this URL; `pm.cookies.jar()` for `get`/`set`/`unset`/`clear`, see below |
 | Response assertions | `pm.response.to.have.status(code)`, `.header(name)`, `.jsonBody()`, and the `pm.response.to.be.*` status classes below |
 | Request             | `pm.request.url`, `.method`, `.headers`, `.body`                                 |
 | Request headers     | `pm.request.headers.get/has(name)`, `.upsert({key, value})`, `.add({key, value})`, `.remove(name)` |
@@ -325,12 +325,32 @@ automatically. What these answer is matched against **this request's URL** -
 domain, path, `Secure` and expiry - so they report what will go on the wire and
 not everything stored.
 
+The write half is `pm.cookies.jar()`, Postman's own jar object:
+
+```javascript
+const jar = pm.cookies.jar();
+jar.set(pm.request.url, { name: 'session', value: token });  // or (url, name, value)
+jar.get('https://api.example.com/', 'session');              // value, or undefined
+jar.unset('https://api.example.com/', 'session');
+jar.clear();                                                 // this environment's jar
+```
+
+Every method is URL-scoped and takes an optional trailing `callback(err, value)`,
+invoked inline. A written cookie's `domain` and `path` default from the URL, and
+it is then matched by exactly the rules a received cookie is.
+
 Divergences from Postman:
 
-- **No `pm.cookies.jar()`, and no write half.** `set` / `unset` / `clear` are
-  not bound. Mutating a jar that in-flight transfers are reading needs its own
-  design; a binding that accepted a cookie and dropped it would be worse than
-  a missing one.
+- **No flat `pm.cookies.set(name, value)`.** Only the `jar()` form ships: a
+  written cookie needs a URL to take its domain and path from, which is why
+  Postman's own write half lives on the jar object.
+- **A write is applied after the transfer it was made before**, not the moment
+  it is called - so it rides that request and cannot be lost when the response's
+  cookies are captured. Details in
+  [scripting.md](../engine/scripting.md#writing-to-the-jar-pmcookiesjar).
+- **`jar.clear()` takes no URL** and empties this environment's jar. Postman's
+  clears per URL; Vayu's scope is the environment, so that is the unit here.
+- **`expires` is seconds since the epoch**, not a date string.
 - **Scope is the environment**, not the domain-with-permission model Postman
   uses. One jar per environment plus one for "no environment", in memory only,
   clearable in Settings → General → Cookies.
@@ -362,8 +382,8 @@ These Postman APIs are **not** implemented - scripts that rely on them will fail
 - `pm.iterationData.*` - data-file driven runs, and with them
   `pm.info.iteration` / `pm.info.iterationCount` (see
   [above](#script-identity-pminfo---three-fields-all-optional))
-- `pm.cookies.set(...)` / `.unset(...)` / `.clear()` / `.jar()` - the jar's
-  write half. The read half (`get` / `has` / `toObject`) is supported - see
+- `pm.cookies.set(...)` / `.unset(...)` / `.clear()` - the *flat* write half.
+  Writing goes through `pm.cookies.jar()`, which ships whole - see
   [above](#the-cookie-jar-pmcookies)
 - `pm.request.url.query` / `.path` / `.host` - and any other `url.*` accessor.
   `pm.request.url` is a writable **string** that the write-back requires to still
--- a/docs/engine/architecture.md
+++ b/docs/engine/architecture.md
@@ -203,16 +203,28 @@ makes a session survive from one design-mode request to the next.
 - **Storage: libcurl's own lines.** The jar holds what `CURLINFO_COOKIELIST`
   exports and hands it back through `CURLOPT_COOKIELIST`, so domain matching,
   path matching, `Secure` and expiry stay inside libcurl's cookie engine. The
-  parser and matcher in `cookie_jar.cpp` serve only the two *read* views
-  (`pm.cookies`, `GET /cookies`), which need an answer without a transfer.
+  parser and matcher in `cookie_jar.cpp` serve the *read* views (`pm.cookies`,
+  `pm.cookies.jar().get`, `GET /cookies`), which need an answer without a
+  transfer, and the URL-scoped removal `jar().unset` performs. They never
+  decide what goes on the wire.
 - **Not on the load path.** `EventLoop` never touches it: a shared jar across
   workers is either a lock on the hot path or per-worker jars that do not
   actually share, and a load run repeats a single request anyway.
 - **Threading:** one mutex around the scope map; every accessor copies out, so
   no reference into the storage escapes to a caller.
+- **Script writes are staged, not applied in place** (`pm.cookies.jar()`, issue
+  #337). `capture_jar_cookies` *replaces* a scope's contents with what the
+  finishing handle held, so a write dropped into the map beside an in-flight
+  transfer would be discarded by it. Instead a write becomes a `CookieWrite`
+  that the execution's next transfer applies on top of the stored lines when it
+  seeds its handle (`ClientConfig::cookie_writes`) - the request carries it, and
+  that transfer's own capture is what persists it. A write with no transfer left
+  to ride, a post-request script's, is applied by the route through
+  `CookieJar::apply`. Either path applies it exactly once.
 
 `pm.sendRequest` shares the jar of the execute it runs inside, so a pre-request
-script can log in and leave the session where the real request will find it.
+script can log in and leave the session where the real request will find it; it
+also carries any write staged before it.
 
 ### Auth Resolution & OAuth 2.0
 
--- a/docs/engine/scripting.md
+++ b/docs/engine/scripting.md
@@ -613,13 +613,77 @@ to the next one automatically, with no header to set by hand.
   `undefined`, which would read as "the cookie is gone". The jar is deliberately
   off the load path: sharing one across the event loop's workers would put a
   lock on the hot path, and a load run repeats a single request anyway.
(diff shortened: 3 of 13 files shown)


athrvk:
fix merge conflict

athrvk:
Conflict resolved and pushed (`93e001d`). The branch is now on top of `1d7b6c0`.

**The only conflict was `docs/engine/architecture.md`** - #347 extended the same Cookie Jar section this branch does. Both bullets belong, so both are kept: master's "Shown as sent" (the raw view reads the wire frame) and this branch's "Script writes are staged". They also turn out to compose - since #347 builds the raw-request view from the transfer's own `CURLINFO_HEADER_OUT` frame, a script-written cookie appears there for free, because it goes out on that same transfer. I added an assertion for that rather than leaving it as a doc claim, since it is a statement about two features that had never met before this merge.

`engine/src/http/client.cpp` merged cleanly on its own: #347's wire capture and this branch's staged-write injection sit at different points of the send path (`apply_jar_cookies` seeds the handle, the header-out capture reads what libcurl then wrote).

**Re-verified on the merged tree:** `ctest --preset linux-dev --output-on-failure` - **910/910 passed**, against 891 on the base commit. `mkdocs build --strict` clean, `clang-format` clean on the files it was clean on before. No app file is touched by this branch, so the app suite could not change colour and was not run.

---

Issue #357, opened 2026-08-07
engine(perf): phase 6 - load-mode scenarios: a per-VU state machine over the shared plan, per-VU cookies, per-step histograms


athrvk:
Phase 6 of the collection-runner design (#340) - the k6/JMeter half. Decisions
live in
[`docs/plans/collection-runner-design.md`](https://github.com/athrvk/vayu/blob/master/docs/plans/collection-runner-design.md)
section 5. This is the phase the whole design-first exercise existed to protect:
mode 1 was built so this needs no rework of it.

## Problem

Vayu's load path repeats a **single** composed request (`execute_load_test`,
`engine/src/core/run_manager.cpp:589`). k6 scenarios and JMeter thread groups
execute a multi-step sequence per virtual user per iteration - login, browse,
check out. "k6-level load tests" that cannot express "log in, then hit the API" is
the gap a serious evaluation hits first.

## Fix / solution pathway

**The plan is shared; the executor is not** (§5.1). A scenario load run does not
spawn a thread per virtual user. A VU is a small value:

```
VirtualUser { const ScenarioPlan* plan; size_t step; size_t iteration;
              VariableOverlay vars; CookieState cookies; }
```

`concurrency` becomes **the number of virtual users** - what k6 and JMeter mean by
it anyway. On each completion, the existing per-completion callback (`handle_result`,
`engine/src/core/load_strategy.cpp`) advances that VU's state machine and submits
its next step to the same worker. This is the existing closed-loop controller with
one substitution: `maintain_concurrency`'s refill (`load_strategy.cpp:249`) issues
"the next step of VU *k*" instead of "another copy of the one request". The pure
`compute_refill_deficit` primitive, the SPSC submission path and the
single-producer discipline are unchanged.

The plan being fully composed at resolution time (#352) is what makes this
possible - there is no SQLite access on this path.

**Per-VU cookie state, never the shared jar** (§5.2). Each VU owns a private cookie
list, empty at the start of each iteration; the environment jar is untouched. This
*strengthens* the existing constraint rather than fighting it - a shared jar across
workers is either a lock on the hot path or per-worker jars that do not actually
share (`engine/include/vayu/http/cookie_jar.hpp`). It is also the semantically
correct answer: 1,000 VUs are 1,000 users, and one session shared between them is
not the thing being measured.

**Scripts stay deferred and sampled** (§5.3), keyed **per step index** rather than
per run. Running QuickJS inline per step at 60k RPS is not viable, and that
[...]

Issue #356, opened 2026-08-07
engine/app(scenario): phase 5 - data-driven iterations: pm.iterationData over rows sent on the run payload


athrvk:
Phase 5 of the collection-runner design (#340), and **the second half of #303**.
Decisions live in
[`docs/plans/collection-runner-design.md`](https://github.com/athrvk/vayu/blob/master/docs/plans/collection-runner-design.md)
section 2.3.

## Problem

`pm.iterationData` is `undefined` (`setup_pm_object`,
`engine/src/runtime/script_engine.cpp`; listed under "Not (yet) supported" in
`docs/app/pm-api-compatibility.md`). #303 found it needs "a data file bound to a
run and an iteration index that means something" - #353 gives the iteration index,
and this issue gives it rows to read.

## Fix / solution pathway

**Rows come in on the run payload, not from a file the engine opens** (§2.3).
`data` is an inline JSON array of objects on the `scenario` block, and the **app**
owns the file picker and the CSV/JSON parsing.

Why, stated so it is not relitigated: the script sandbox has no filesystem access
by design, and handing the daemon a user-supplied path is a new trust boundary
(traversal, arbitrary reads) bought for a parsing job the app already does well -
it owns the Postman/Insomnia/OpenAPI import parsers. The cost is payload size,
which is what `maxScenarioDataRows` (#352, default 1,000) bounds.

**Binding.** Row `i % rows` binds to iteration `i` as `pm.iterationData`, exposing
Postman's read surface (`get(key)`, and `toObject()`). Read-only: a write has no
defined destination - the rows are a run input, not a scope - so a setter must
throw rather than accept and drop.

**Iteration count.** With `data` present and `iterations` absent, `iterations`
defaults to the row count (Postman's default). With both given, the explicit count
wins and the row index wraps. **The wrap is not silent**: every per-iteration
record carries `dataRowIndex`, and `pm.info.iteration` already reports the
iteration. An empty `data` array is a `400` (#352 already enforces this) - a data
set that binds nothing is a mistake, not an empty run.

**Non-object rows** are a `400` naming the row index. `pm.iterationData.get` on an
unknown key returns `undefined`, matching every other `pm` scope reader.

## App changes (required, same PR)

- A data-file picker on the runner surface (#354's), reading CSV or JSON into the
  row array. CSV parsing follows whatever the import path already uses - `rg` for
  it before adding a parser; a second CSV reader would not receive the first's
  fixes.
[...]

Issue #355, opened 2026-08-07
engine(scripting): phase 4 - flow control as a ScriptResult value: pm.execution.setNextRequest / skipRequest


athrvk:
Phase 4 of the collection-runner design (#340), and **the first half of #303**.
Decisions live in
[`docs/plans/collection-runner-design.md`](https://github.com/athrvk/vayu/blob/master/docs/plans/collection-runner-design.md)
section 4.

## Problem

`pm.execution` is `undefined` - `setup_pm_object` (`engine/src/runtime/script_engine.cpp`)
binds it nowhere, and `docs/app/pm-api-compatibility.md` correctly lists it under
"Not (yet) supported". #303 established that the blocker was never the binding but
the absence of an ordered sequence to deviate from. #353 delivers that sequence, so
the binding is now implementable against something real.

## Fix / solution pathway

**Flow control is a value the script returns, not a side effect it performs**
(§4.1) - the channel #303 prescribed, honoured unchanged:

```cpp
struct ScriptControl {
    enum class Kind { None, Next, Skip, EndIteration };
    Kind kind = Kind::None;
    std::string target;   // request name, for Kind::Next
};
// ScriptResult (engine/include/vayu/types.hpp:515) gains: ScriptControl control;
```

`pm.execution.setNextRequest(name)` and `pm.execution.skipRequest()` **record** an
intent on the result and reach into nothing. `ScriptResult` is already how a script
speaks to its caller, and only the caller knows what a sequence is. Last call wins
within one script (Postman's behaviour).

**`pm.execution` throws where there is no sequence** (§4.2). `ScriptContext` gains
`bool in_scenario`, set only by the scenario runner. With it false - a
`POST /execute` send, a load run's deferred `tests` script - both methods throw a
sentence naming why. A binding that cannot fail is worse than a missing one
(#188's standing rule): `setNextRequest("checkout")` silently ignored in a single
send is exactly the false success that rule exists to prevent.

**Every ambiguous or unbounded case fails loudly** (§4.3):

| Case | Behaviour |
|------|-----------|
| `setNextRequest(name)` naming no step in the plan | Step fails with a named error; the iteration ends |
| `setNextRequest(name)` naming a **duplicated** step name | Step fails - ambiguous target, both indices named. Plan resolution (#352) records the duplicate set, so the message can be precise |
| `setNextRequest(null)` | `Kind::EndIteration` - ends this iteration, continues to the next. Postman's convention |
[...]

Issue #354, opened 2026-08-07
app(scenario): phase 3 - the collection runner UI: run a folder, live per-step progress, per-step results


athrvk:
Phase 3 of the collection-runner design (#340). The engine side is #353; this is
the surface that makes it reachable. Design context:
[`docs/plans/collection-runner-design.md`](https://github.com/athrvk/vayu/blob/master/docs/plans/collection-runner-design.md)
sections 3.2, 3.6 and 4.4.

## Problem

After #353 the engine runs an ordered sequence, streams a `step` SSE event per
step, and stores a `results` row per step execution. Nothing in the app starts one
or shows one. Until this lands the feature exists only over `curl`.

## Fix / solution pathway

Deliberately unspecified in the design (§8): this issue owns the UI shape against
`docs/app/COMPONENTS.md`. What the design **does** fix:

- **Entry point:** running a collection/folder is the action; `recursive` and
  `iterations` are its options. The scenario is folder-derived (§2.1), so there is
  no scenario-authoring UI to build - do not invent one.
- **Live progress** comes from the `step` SSE event on `GET /runs/:runId/live`
  (`{iteration, stepIndex, name, outcome, statusCode, latencyMs}`), not from
  polling. The renderer already consumes this stream for load runs; reuse that
  client rather than adding a second one.
- **Per-step results** restore from the stored `trace_data` through the same path a
  design run's response pane uses
  (`app/src/modules/request-builder/utils/restore-response.ts`) - a hand-rolled
  copy of that restore never receives its fixes (`bodyTruncated`/`bodyBytes`, the
  `httpVersion` fallback).
- **`skipped` is never `passed`** (§4.4). Outcomes are
  `passed | failed | skipped | errored` and the summary counts four numbers, not
  two. A skipped step rendered as a pass is the false-pass class #180 exists to
  eliminate. Nothing produces `skipped` until #355 lands, so this must be built to
  render it correctly before it can occur - a test asserts it, not a future reader.
- **Thinned results must say so.** #353 bounds stored steps at
  `maxScenarioStoredSteps` with failures kept first. A run whose successes were
  thinned must disclose it, the way the Samples tab discloses its retention -
  otherwise the step list silently claims to be the whole run.

## App changes

Renderer, plus a new query hook beside `app/src/queries/runs.ts`. Placement per
`docs/app/COMPONENTS.md` - decide and follow the doc. Design-system and
surface/border rules apply (`docs/design-system.md`); `rg` for an existing
[...]

PR #358, opened 2026-08-07
docs(plans): decide the collection runner / scenario primitive, design-first

### LINKED ISSUE #340: engine: collection runner / scenario primitive - the ordered-sequence engine feature that completes the Postman half and unlocks k6-style multi-step load, design-first
**Direction approved by the owner, 2026-08-06**: Vayu's product target is the Postman + JMeter + k6 combination (the repo description already says as much: "Postman-style requests + k6-level load tests in one local app"). The single largest capability gap against that target, on both halves at once, is an engine-side ordered-sequence primitive - a collection runner. #188 and #303 recorded the gap and correctly refused to build a runner as a side effect of a scripting API; this issue is the product decision they were gated on, made explicitly. Anchors verified at master `f703eaa` (2026-08-06).

## Why this is the load-bearing feature for the vision

**The Postman half stops at single sends without it.** Postman's Collection Runner - run a folder in order, per-request scripts, `pm.iterationData` over a data file, `pm.execution.setNextRequest` - is the workflow that makes collections more than a folder tree. Vayu has nothing that executes an ordered sequence: `run_collection_smoke` is a client-side TypeScript loop over independent `POST /execute` calls (`app/electron/mcp/tools.ts`, the smoke handler; composition documented at `docs/engine/mcp.md:153`, the deliberate no-recurse note at `:187`), so the engine never sees a sequence, and a script inside execute #3 has nowhere to send an instruction about execute #4 (#303's core finding, still true).

**The k6/JMeter half repeats one request without it.** k6 scenarios and JMeter thread groups execute a *multi-step sequence* per virtual user per iteration - login, then browse, then checkout. Vayu's load path repeats a single composed request (`engine/src/core/run_manager.cpp`), with scripts deferred to post-run validation over sampled responses (`validate_scripts`, `run_manager.cpp:55`, called at `:747`). "k6-level load tests" that cannot express "log in, then hit the API" is the gap a serious evaluation hits first.

**One primitive serves both.** An engine-side scenario (ordered request refs + per-step scripts + iteration model) is the design-mode collection runner in mode 1 and the unit a load worker repeats in mode 2. Building it engine-side - where the executes happen, not in one client - is what makes the second mode reachable at all, and is why the client-side smoke loop was never extended instead.

## What this issue is - design first, per the repo's core principle

The deliverable of THIS issue is a design, not code. Long-term architectural decisions before any implementation:

1. **The sequence model.** What a scenario is (ordered request refs? folder-derived? explicit?), where it lives (derived at run start from a collection/folder vs a stored first-class entity - db-schema implications), and what an iteration means (once, N times, data-file-driven rows).
2. **The execution model, design mode.** Engine-side sequential execution with per-step pre-request/test scripts running inline (not deferred like the load path's `validate_scripts`), state that flows between steps (variables, the #301 cookie jar - which is per-environment and already gives "log in once, reuse the session" its storage), and a run record the app can render step by step.
3. **Flow control as a value, not a side effect.** `setNextRequest`/`skipRequest` return through `ScriptResult` (`engine/include/vayu/runtime/script_engine.hpp:235`) to the runner that owns the sequence - #303's prescribed channel, which this design must honor.
4. **The load-mode convergence path.** How a load run's config points at a scenario instead of a single request, what per-step metrics look like (per-step latency histograms vs whole-scenario), what the per-worker execution of a sequence does to the event-loop model (`event_loop_worker.cpp`) - **decided in the design even if shipped last**, so mode 1 is not built in a way mode 2 has to undo. #197's re-tune of that same hot path is in flight; the design must name its coordination points.
5. **What is explicitly out.** JMeter's full logic-controller zoo, k6's open/arrival-rate executors, distributed load - name them as non-goals or later phases so the design has a boundary.

The design lands as `docs/plans/collection-runner-design.md` (or a section in `docs/engine/architecture.md` if small enough - the design decides), is reviewed on a PR, and its acceptance decomposes this issue into phased implementation sub-issues (engine sequence execution; app runner UI; iterationData/data files; load-mode scenarios) - each implementer-ready by the repo standard, filed as sub-issues here.

## Relationship to existing issues

- **#303 (flow control, `pm.iterationData`) becomes a follow-on to this issue**, exactly as its own pathway prescribed: bindings land against the real runner, and its skip-only carve-out against the client-side smoke loop is **declined** - approved direction makes investing in the TS loop the wrong foundation.
- **#301's shipped cookie jar** is the session-state half this feature needs and was scoped per-environment partly for this reason; the design consumes it, not redesigns it.
- **#197 (load-path re-tune)** shares the phase-4 surface; the design phase itself touches no code and is parallel-safe with everything.
- **`run_collection_smoke` stays** - a smoke matrix over independent executes is a different tool from an ordered run and remains valid for MCP agents.

## App changes

None in the design phase. The design must specify the app surface (runner panel, per-step results, data-file picker) so the implementation sub-issues carry real App-changes sections.

## Tests

None in the design phase. The design names the test strategy per phase (engine gtest for sequence semantics and flow-control values; app vitest for the runner UI; the isolation and skip-vs-pass matrix rules #303 already states).

## Docs

The design doc itself, same PR. `docs/plans/pending-backlog.md` gains a pointer if the phases stretch; `pm-api-compatibility.md` stays untouched until something ships (#303's standing rule: a capability that is documented must exist).

## Sequencing

Design phase: parallel-safe against everything open (#197/#287, #337, #338, #339, #134 parked). Implementation sub-issues will carry their own conflict surfaces; the load-mode phase must land after or coordinate with #197's measured defaults. #303 is sequenced behind this issue and stays open as the flow-control follow-on.

## Acceptance criteria

- [ ] A reviewed, merged design covering points 1-5, each as a stated decision with rationale - not options left open.
- [ ] The load-mode convergence path is designed (even if scheduled last), so design-mode structures need no rework for load.
- [ ] This issue decomposed into phased, implementer-ready sub-issues per the design; #303 re-anchored to the phase that delivers flow control.
- [ ] No implementation shipped from this issue itself.

---

athrvk:
## Description

The design deliverable for #340: `docs/plans/collection-runner-design.md`, plus
its nav entry and a pointer from `pending-backlog.md`. No implementation, per the
issue's last acceptance criterion.

**Plan (root cause, approach, rejected alternative).** Nothing in the engine holds
an ordered sequence: `run_collection_smoke` is a client-side TypeScript loop over
independent `POST /execute` calls, so the engine never sees one, and load runs
repeat a single composed request with scripts deferred to post-run validation
(`validate_scripts`, `run_manager.cpp:55`). Both halves of the Postman + JMeter +
k6 target stop there. The approach is **one shared plan, two executors**: a
scenario resolves once, at run start, into an immutable vector of fully composed
steps; design mode walks it sequentially through `http::Client` (for the cookie
jar and inline per-step scripts, which the event loop deliberately lacks), and
load mode drives the *same plan* from a per-VU state machine on the existing
closed-loop refill. The alternative I rejected is a **stored `scenarios` table**:
the collection tree with its `order` columns already *is* an ordered sequence, so
a table would be a second source of truth needing its own reconciliation on every
collection edit - this repo's most repeated defect class, bought before anyone has
asked for the thing it enables. The seam is kept (`{"source": "collection"}` is a
discriminator), so a stored scenario can land later without changing anything
downstream of resolution.

The load-bearing constraint that falls out: **the plan must be fully composed up
front**, because a lazily composed step would put SQLite on the 60k-RPS hot path
in mode 2. That is what "design first so mode 1 is not built in a way mode 2 has
to undo" actually cashes out to here.

Decisions made, each with its declined alternative recorded in the doc:

| § | Decision |
|---|----------|
| 2.1 | Folder-derived sequence, not a `scenarios` table |
| 2.2 | Plan resolved exactly once, before the first send |
| 2.3 | Data rows on the payload; the engine never opens a file (the sandbox has no filesystem access by design, and a user-supplied path is a new trust boundary for a parsing job the app already does) |
| 3.1 | A third `runs.type` - reusing `design` would break the documented "a design run has exactly one `results` row" contract that `GET /runs/:runId` serves on |
| 3.2 | Async, run-record-backed, SSE-streamed - one lifecycle, two executors |
| 3.3 | Sequential through `http::Client`, not `EventLoop` |
| 3.4 | Variables loaded once and persisted once; the existing per-environment cookie jar unchanged |
| 3.5 | `pm.info.iteration` set only by the runner - **#300's ruling is preserved verbatim**, `validate_scripts` still sets neither |
| 3.6 | One bounded `results` row per step, failures kept first, thinning disclosed |
| 3.7 | The snapshot carries a step manifest with **stored** URLs, never the composed plan - a composed plan carries resolved `Authorization` headers, and `apikey`/`in: query` puts a live key in the composed URL, which is what `sanitize_config_snapshot` exists to keep out |
| 4 | Flow control as a `ScriptResult` value (#303's prescribed channel); `pm.execution` throws where there is no sequence; every ambiguous case is loud; `setNextRequest` cycles are bounded; skipped is never passed |
| 5 | Per-VU state machine over the shared plan; per-VU cookies (which strengthens `cookie_jar.hpp`'s "not on the load path" rather than fighting it); scripts stay deferred; closed-loop only; `maxInFlight` moot for scenarios |
| 6 | Nine explicit non-goals |

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Docs-only - every changed line is Markdown or a nav entry, so no test could go
from green to red (CLAUDE.md's "Concrete cases where the answer is simply
don't"). The one check that *can* fail on a docs change is the site build:

```
mkdocs build --strict -f .github/mkdocs.yml   # what CI runs
Documentation built in 3.02 seconds           # exit 0
```

No engine or app suite was run, deliberately. No pre-existing failures observed.

## Checklist

- [x] My code follows the style guidelines (no em-dashes; the one en-dash left in
      `pending-backlog.md` is pre-existing text in the **P1** section, which #197
      owns - not touched)
- [x] I have added tests (N/A - design document, no behaviour)
- [x] I have updated documentation (the doc *is* the change; nav entry added in
      the same commit per the docs-site rules, and `pending-backlog.md` gains the
      pointer #340's Docs section asked for. Only the backlog's "Last updated"
      date was touched beyond the new section - **P1/D8/D9/M1 are #197's to
      refresh** and were left alone)

## Related Issues

Closes #340.

Decomposed into six implementer-ready sub-issues, all linked under #340:

| Phase | Issue |
|-------|-------|
| 1 | #352 - `ScenarioPlan` resolution + `scenario` on `POST /runs` + snapshot manifest |
| 2 | #353 - design-mode sequential runner |
| 3 | #354 - app runner UI |
| 4 | #355 - flow control (**#303's first half**) |
| 5 | #356 - `pm.iterationData` (**#303's second half**) |
| 6 | #357 - load-mode scenarios (coordinates with #197) |

#303 is re-anchored to #355/#356 in a comment on the issue and closes when both
land.

**Deliberate deviations from the issue spec, with reasons:**

1. #340 offered "a section in `docs/engine/architecture.md` if small enough - the
   design decides". It is not small enough: five decision areas with declined
   alternatives, plus a phase table, would have doubled a document that describes
   *shipped* behaviour with one describing unshipped behaviour. A separate page
   under `docs/plans/` keeps the "architecture.md describes what exists" property.
2. #340 named four candidate sub-issues; the decomposition has **six**. Plan
   resolution (#352) is split from the runner (#353) because every later phase
   consumes the plan and it is testable with nothing executing, and flow control
   (#355) is split from `pm.iterationData` (#356) because they gate differently
   and `pm-api-compatibility.md` then moves in two clean steps rather than one
   tangled one.
3. The sub-issues are filed **now**, before this PR merges, since filing them is
   an acceptance criterion of the issue this PR closes. Each is gated on the
   design landing; if review changes a decision, the affected issue is edited
   rather than the design drifting from its children.

### DIFF
--- a/.github/mkdocs.yml
+++ b/.github/mkdocs.yml
@@ -119,6 +119,7 @@ nav:
   - Design Notes:
       - Request Storage: request-storage-design.md
       - Lock Files: lock-file-handling.md
+      - Collection Runner: plans/collection-runner-design.md
       - Deferred Work: plans/pending-backlog.md
 
 markdown_extensions:
--- a/docs/plans/collection-runner-design.md
+++ b/docs/plans/collection-runner-design.md
@@ -1,0 +1,445 @@
+# Collection Runner / Scenario Primitive - Design
+
+**Status:** approved design, not yet implemented.
+**Tracks:** issue #340. Flow control (`pm.execution`, `pm.iterationData`) is #303.
+**Anchors verified at master `d5899bb` (2026-08-07).**
+
+Vayu's product target is the Postman + JMeter + k6 combination. The single
+largest capability gap against that target, on both halves at once, is an
+engine-side **ordered-sequence primitive**: Postman's collection run, and the
+multi-step sequence a k6 virtual user repeats. This document is the design
+decision for that primitive, made once, before any code.
+
+Every section below states a **decision** and the alternative it declines. Where
+a decision constrains something that is not being built yet - load-mode
+scenarios, stored scenario entities - it says so explicitly, because the point of
+designing first is that mode 1 must not be built in a way mode 2 has to undo.
+
+---
+
+## 1. Why Vayu has nothing today
+
+Two things look like runners and are not:
+
+- **`run_collection_smoke`** is a **client-side TypeScript loop** over
+  independent `POST /execute` calls (the smoke handler in
+  `app/electron/mcp/tools.ts`; composition described at `docs/engine/mcp.md`).
+  The engine sees N unrelated design-mode executes and holds no object that
+  contains them, so a script running inside execute #3 has nothing to say about
+  execute #4 and nowhere to put the instruction.
+- **Load runs repeat a single composed request**
+  (`engine/src/core/run_manager.cpp`, `execute_load_test`). Test scripts there
+  run **deferred, once per sampled response, after the run finishes**
+  (`validate_scripts`, `run_manager.cpp:55`, called at `:747`), so even the
+  ordering that exists is not something a script participates in.
+
+So "log in, then browse, then check out" - the workflow that makes a collection
+more than a folder tree, and the unit a serious load evaluation reaches for
+first - cannot be expressed at all.
+
+---
+
+## 2. The sequence model
+
+### Decision 2.1 - a scenario is a folder-derived ordered sequence, resolved once into an immutable plan
+
+A **scenario** is not a new stored entity. It is a *resolution* of an existing
+one: a collection (folder), its direct requests ordered by `requests.order`,
+optionally including descendant collections ordered by `collections.order`,
+depth-first.
+
+At run start the engine resolves that into a **`ScenarioPlan`**: an ordered,
+immutable vector of **fully composed steps**. Each step is the exact payload
+`POST /execute` accepts - `{{variables}}` substituted, `inherit` auth walked and
+applied, the ordered script-part lists from the collection chain plus the
+request's own - produced by the composition the engine already owns
+(`compose_request_core`, `engine/src/http/request_composer.cpp:454`).
+
+```
+ScenarioStep {
+  index          size_t          // position in the plan, stable for the run
+  request_id     string          // the stored request this came from
+  name           string          // requests.name, the setNextRequest target
+  request        vayu::Request   // composed, execute-ready
+  pre_script     string          // joined parts (script_parts.cpp::read_script)
+  post_script    string          // joined parts
+}
+ScenarioPlan { vector<ScenarioStep> steps; ... }
+```
+
+**Why folder-derived rather than a `scenarios` table.** The collection tree with
+its `order` columns *already is* an ordered sequence. A parallel `scenarios`
+table would be a second source of truth for that ordering, needing its own
+answer to "a request in this scenario was deleted" and its own reconciliation on
+every collection edit. This repo's most repeated defect is state one layer
+records and another re-derives; a scenario entity in v1 would be exactly that,
+bought before anyone has asked for the thing it enables.
+
+**Declined: a first-class `scenarios` table (v1).** It becomes necessary the
+moment a user wants a sequence that is *not* a folder - steps drawn from several
+folders, one request appearing twice, a step disabled for one scenario but not
+another. The design must not preclude it, and does not: the run payload carries
+a `source` discriminator, and plan resolution is the seam.
+
+```jsonc
+"scenario": {
+  "source": "collection",          // the only value in v1
+  "collectionId": "col_...",
+  "recursive": false,              // descend into sub-collections
+  "iterations": 1,
+  "data": [ { "user": "a" }, ... ] // optional; see 2.3
+}
+```
+
+A future stored scenario is `{"source": "stored", "scenarioId": "scn_..."}` and
+resolves to the same `ScenarioPlan`. Nothing downstream of resolution changes.
+
+### Decision 2.2 - the plan is resolved exactly once, before the first send
+
+No step is composed lazily, and no execution path touches SQLite for request
+data after resolution. Two reasons, and the second is the load-bearing one:
+
+- A collection edited mid-run would otherwise change the sequence underneath
+  itself. A run is a record of what ran; it must be resolvable to one answer.
+- **Load mode cannot query SQLite per step per virtual user.** The plan being
+  fully composed up front is what makes decision 5 possible at all. Resolving
+  lazily in mode 1 would be a stopgap mode 2 has to undo.
+
+Resolution failures are loud and pre-run: an empty collection, a step whose
+composition fails, a plan exceeding `maxScenarioSteps` - all `400` with the
+offending request named, before any `runs` row exists. This matches the existing
+`POST /runs` discipline, where `validate_run_config` and
+`normalize_run_http_version` both run before `create_run`
+(`engine/src/http/routes/execution.cpp`).
+
+### Decision 2.3 - an iteration is one full pass over the plan; data rows come in on the payload
+
+`iterations` (default `1`) is how many times the plan is executed end to end.
+
+`data` is an **inline JSON array of objects** on the run payload. Row `i % rows`
+binds to iteration `i` as `pm.iterationData`. When `data` is present and
+`iterations` is absent, `iterations` defaults to the row count - the Postman
+default. When both are given, the explicit count wins and the row index wraps;
+the wrap is **not silent**, because every per-iteration record carries its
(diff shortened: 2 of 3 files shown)


athrvk:
this doc looks like a design doc and tracker doc it does not follow the .md docs format vayu already has. I am assuming that once the whole feature is done this doc will be deleted ? check the issues ?

athrvk:
Both fair. Fixed in b8950ca.

**Format.** You are right that it read as an issue. It opened with a
`Status:` / `Tracks:` / `Anchors verified at <sha>` block, used
`**Decision 2.1 - ...**` as headings, and closed with a phase table of open issue
numbers. Its peers under Design Notes - `request-storage-design.md`,
`lock-file-handling.md` - open with a plain `## Overview` and use descriptive
headings, so it now does too. Every decision and every declined alternative
survived; only the packaging changed.

**The tracker had to go for a second reason.** A table of open issue numbers in a
published doc is a stale line waiting to happen - closing #352 would silently
falsify it, and this repo treats a stale doc line as worse than a missing one.
The phase/dependency table now lives on #340 instead, where closing an issue
updates it for free. Issue numbers appear in the page only as provenance for a
prior ruling (#300's iteration index, #188's binding rule, #180's false-pass
class), the way `architecture.md` already cites them, never as a work list.

**On deletion - your assumption is right, and it is now written down.** It is a
plan, and plans get consumed: as each phase ships its decisions fold into
`docs/engine/architecture.md` beside the Cookie Jar and request-composition
sections, which is where rationale for *shipped* behaviour lives here, and the
page is deleted when the last phase lands. Same shape as a backlog entry leaving
`pending-backlog.md` once it ships. The page states this about itself, and
because a lifecycle nobody owns is a lifecycle that does not run, **#357 (phase 6)
now carries the deletion as a Docs task** - delete the page, drop the nav entry,
remove the `pending-backlog.md` pointer, and confirm first that no decision
exists only on the deleted page.

The page also now says plainly, up front, that nothing in it describes behaviour
the engine has - which the old format left ambiguous, and which matters because
sessions are told to trust `docs/`.

**Checked the issues, as you asked.** Two things needed doing and are done:

- The tracking the doc shed is now on #340 (phases, dependencies, and that #303
  closes when #355 and #356 land). Each sub-issue already carried its own
  Sequencing section, so nothing was only in the doc.
- The `§N.N` citations in #352-#357 pointed at the numbering I removed. #340's
  comment carries the old-number to section-name map rather than my rewriting six
  bodies over a formatting change; if you would rather have the bodies edited,
  say so and I will.

`mkdocs build --strict` passes, including the two new relative links into
`engine/` and the `#cookie-jar` anchor.

---

PR #375, opened 2026-08-08
fix(app/mcp): invalidate the renderer's queries when an MCP call writes

### LINKED ISSUE #363: app(mcp): MCP writes are invisible to the renderer until an unrelated refetch - no main-to-renderer invalidation channel exists
From the 2026-08-07 collection-tree review sweep's completeness critic (verified at `d5899bb`). Filed separately from the tree batches because the fix is an IPC channel, not a tree change.

## Problem

MCP write tools mutate the engine from the Electron main process (`app/electron/mcp/tools.ts:877-924` `create_request` and the other `category: "write"` tools call `ctx.client.*` straight at the engine), and the renderer never learns: `refetchOnWindowFocus` is globally off (`app/src/lib/query-client.ts:36-42`) and no invalidation event crosses the process boundary. An agent creates a request over MCP and **it does not appear in the collection tree** until the user happens to perform some unrelated renderer mutation whose blanket invalidation refetches the lists. Same for environment updates, deletes, and load runs started via MCP (run lists). Silent staleness in the primary navigation surface, and the gap grows with every MCP write tool.

This also matters for the drag-and-reorder track: its design must state how external writers reach the tree, and this channel is the answer.

## Fix / solution pathway

A minimal invalidation event, not a data push: after any successful MCP write, the main process emits `mcp:data-changed` with `{ entity: "request" | "collection" | "environment" | "run" | "cookie" | "config", id?, collectionId? }` over the existing webContents send path (the `mcp:status` channel's plumbing is the pattern to copy). The renderer registers one listener at app init that maps entities to targeted `queryClient.invalidateQueries` keys (`listByCollection(collectionId)`, `collections`, `environments`, `runs`, `cookies` - the keys module already centralizes these in `app/src/queries/keys.ts`). Emit from one chokepoint (the tool dispatch success path in `app/electron/mcp/tools.ts` or `server.ts`'s `dispatchTool` wrapper - `dispatchTool` is the single dispatch path since PR #335, use it) rather than per-tool, keyed off each tool's declared entity so new write tools cannot forget to emit.

Do not push data over IPC - invalidation only, the renderer refetches through its normal query layer.

## App changes

`app/electron/mcp/` (dispatch chokepoint + per-tool entity declaration beside `category`), `app/electron/main.ts` or the MCP service (the send), preload channel exposure (follow the existing `mcp:status` pattern - and note the #278-era hardening posture: one-way main-to-renderer, no renderer input), one renderer listener module + registration in `App.tsx`. No engine change.

## Tests

- Electron-side (the existing `tools.test.ts`/`protocol.integration.test.ts` harness): a successful write tool dispatch emits exactly one `mcp:data-changed` with the right entity; a failed dispatch emits none; a read tool emits none (mutation-checked - remove the emit and the test reddens).
- Renderer (jsdom): the listener maps each entity to the expected invalidateQueries keys (spy on the query client).

## Docs

`docs/engine/mcp.md` - one paragraph: UI reflects MCP writes live via the invalidation channel. `docs/app/state-management.md` - the listener and its keys. Same commit.

## Sequencing

Parallel-safe against everything open: the tree batches (#361/#362), the ordering foundation (#360), the DnD track, #344, #348, the scenario phases, the engine perf pair. The DnD parent issue references this as its external-writers answer - land in either order.

Out of scope: MCP transport auth (security session); any polling fallback.

## Acceptance criteria

- [ ] A request created over MCP appears in the tree without any user action; same for environment and run writes (each pinned by the entity-mapping test).
- [ ] The emit lives at the single dispatch chokepoint; a new write tool gets invalidation by declaring its entity, not by remembering an emit.
- [ ] All tests green, mutation-checked; docs updated same-commit.

---

athrvk:
## Description

The main-to-renderer invalidation channel #363 asks for. Every anchor in the issue reproduces on `e1b3592` exactly as described.

**The plan.** Root cause is that an MCP tool call mutates the engine from the Electron **main** process, so no renderer mutation runs and no query notices - and with `refetchOnWindowFocus` globally off (`lib/query-client.ts:36-42`) nothing ever catches up on its own. A request an agent created stayed invisible in the collection tree until some unrelated renderer mutation happened to blanket-invalidate the lists. The approach is the issue's: a minimal invalidation event at the single dispatch chokepoint, keyed off a per-tool declaration so a new write tool cannot forget to emit; the renderer maps the family to keys and refetches through its normal query layer. **The alternative I rejected** was pushing the written row over IPC - it is one call shorter, but it gives a row a second way to reach the UI, and the two paths drift the moment the engine's response shape and the renderer's normalization disagree. Also rejected: keying the emit off `category`, see below.

### The declaration is per tool, not derived from `category`

`McpTool` gains a **required** `invalidates: readonly McpDataEntity[]`. Deriving it from `category` would have been free and wrong: `run_request` and `run_collection_smoke` are `execute`, yet each writes a design-run row **and** refills the cookie jar (`queries/cookies.ts` states the jar changes exactly when a request is sent), and `start_load_run` / `stop_run` are `load` and write run rows. Required rather than optional is the point of the field - a default would let a new tool ship silently invisible to the UI, which is the bug being fixed. `[]` on the nine read tools is a statement, and a test asserts read tools declare it.

| Tool | Declares |
|---|---|
| `create_request` | `request` |
| `update_environment` | `environment` |
| `update_engine_config` | `config` |
| `run_request`, `run_collection_smoke` | `run`, `cookie` |
| `start_load_run`, `stop_run` | `run` |
| the nine read tools | *(nothing)* |

### The emit

`dispatchTool` sends one event per declared family after a handler that did **not** return an error result. A rejected call emits nothing - it may have reached the engine, but it cannot say so, and invalidating on every rejection is worse than the one stale list a genuinely-partial failure leaves. A throwing listener is logged and swallowed rather than turned into a tool error: notification failure must not fail a write the engine already applied, the rule `update_engine_config`'s best-effort read-back already follows.

The event's two scope hints (`collectionId`, `requestId`) are read from the call's own arguments at the chokepoint. Every tool in the registry spells those two the same way, which is precisely what keeps a new write tool from needing an emit of its own. A hint the call did not name is **absent**, not an empty string - pinned by a test, since `{ collectionId: "" }` would key an invalidation at a collection that does not exist.

### The renderer side

`useMcpDataInvalidation()` registers one listener in `App.tsx` (one per app - the events name a data family, not a screen, so a second subscriber only duplicates work) and maps through `lib/mcp-invalidation.ts`:

| Entity | Invalidates | Why that key |
|---|---|---|
| `request` | `requests.listByCollection(collectionId)`, else `requests.lists()` | The same narrowing `useUpdateRequestMutation` does; without a named owner the owner is unknowable here |
| `environment` | `environments.all` | Variables are read through the detail cache as well as the list |
| `run` | `runs.lists()`, `runs.allRuns()`, `+ runs.lastDesign(requestId)` | The history list polls on its own, so this is immediacy there - but Settings' count and a request tab's restored response are **not** polled and stayed stale indefinitely |
| `cookie` | `cookies.all` | One key for every jar - the engine reports them together |
| `config` | `config.all` | |

Per-run reports and time series are deliberately untouched: a new run cannot have changed an existing run's report, and those are the expensive fetches in that family. The map is a `Record` over the union, so a new entity fails to compile until it names a reader - the "written but never read" guard the issue's own framing asks for. An unknown entity arriving over IPC is warned and dropped rather than thrown: a main process newer than the renderer bundle can legitimately name a family this build has not heard of, and throwing inside an IPC listener buys an unhandled rejection on top of the same stale list.

**Two deliberate notes.**

1. **The entity union is duplicated on purpose.** Production code under `electron/` cannot import from `app/src` - `tsconfig.node.json` withholds the `@/*` mapping for stated reasons - so `MCP_DATA_ENTITIES` (main) and `McpDataEntity` (renderer) are written twice. A silent divergence is the worst possible failure here: main emits a family the renderer's map has no entry for, and the write stays invisible, which is the bug. `data-changed.conformance.test.ts` is the one place the two meet, following the pattern `tsconfig.electron-test.json` documents.
2. **Deviation from the issue spec, in two places.** The issue names a `collection` entity and an `id?` field. No tool in the registry creates or deletes a collection, so a `collection` branch would be a mapping with no emitter - dead wiring of exactly the kind this repo keeps finding - and it is left out until a tool needs it. `id?` is replaced by the two named hints `collectionId` / `requestId`, because "the entity's id" is ambiguous on a `run` event (the `requestId` there is the saved request the run was *linked to*, the key `runs.lastDesign` uses, not the run's own id) and both named fields have concrete readers.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`cd app && pnpm test && pnpm type-check && pnpm lint` - all green.

- **`pnpm test`: 312 files, 2851 tests passed.** Base `e1b3592`, measured on this machine: **308 files, 2823 tests** - no pre-existing failures.
- `pnpm type-check` (all three projects), `pnpm lint`: clean.
- `prettier --check` on the 13 files this PR touches: clean. Repo-wide formatting untouched.
- **Engine untouched, so no engine build or `ctest` run** - per CLAUDE.md, no C++ test could change colour.
- `mkdocs` is not installed in this environment. The docs edits add no page, no heading and no anchor - two cross-directory relative links between existing pages, in the style `docs/app/pm-api-compatibility.md` already uses - so nothing `--strict` gates structurally changed.

**New (28 tests, 4 files):**

- `electron/mcp/data-changed.test.ts` (13) - the registry declares its effects (every tool has `invalidates`, read tools declare `[]`, and the mutating set is spelled out so a new mutating tool must be added or the guard notices); and the emit itself: a successful write emits exactly one event with its scope, a guard-rejected dispatch emits none, an engine error emits none, a read tool emits none, a disabled tool emits none, a two-entity tool emits one event each in order, unnamed hints are absent rather than empty, a context with no notifier dispatches normally, and a throwing listener leaves the write successful.
- `electron/mcp/data-changed.conformance.test.ts` (2) - the two copies of the entity list name the same families, plus the non-empty assertion CLAUDE.md requires of a scanning guard.
- `src/lib/mcp-invalidation.test.ts` (9) - one per entity, both `request` branches, the `lastDesign` addition, the report family staying untouched, and an unknown entity reported rather than thrown.
- `src/hooks/useMcpDataInvalidation.test.tsx` (3) - an event from main reaches the query cache, the listener unsubscribes on unmount, and the hook is inert with no `electronAPI`.

Every guard is mutation-checked - each reverted individually, its tests confirmed red, then restored (verified byte-identical with `cmp`):

| Mutation | Result |
|---|---|
| drop the emit entirely | 6 failed |
| emit even on an error result | 2 failed |
| `request` ignores the collection scope | 2 failed |
| `run` drops `lastDesign` | 1 failed |
| main-process entity list drifts from the renderer | 1 failed |
| the hook never registers | 2 failed |

No existing test file needed an edit.

## Checklist
- [x] I have added tests

## Docs

Same commit. `docs/engine/mcp.md` - the `mcp:data-changed` channel beside the existing IPC table, that the UI now reflects MCP writes live, and why the families are declared per tool rather than derived from `category`. `docs/app/state-management.md` - the entity-to-key table above, in the Query Keys & Cache Invalidation section that already documents why the cascade delete invalidates coarsely and the single-request update narrowly.

## Related Issues

Closes #363

---
_Generated by [Claude Code](https://claude.ai/code/session_015G8iWhQVdNmPaP8kyLv4k1)_

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -30,6 +30,7 @@ import {
 	connectClient,
 	toolCatalog,
 	type McpConnectClient,
+	type McpDataChangedEvent,
 	type McpSafetyConfig,
 } from "./mcp/index.js";
 import {
@@ -497,6 +498,7 @@ async function startMcp() {
 			port: MCP_PORT,
 			version: app.getVersion(),
 			safety: loadPersistedSafety(),
+			onDataChanged: sendMcpDataChanged,
 		});
 		await mcpService.start();
 		console.log("[Main] MCP server listening at", mcpService.getUrl());
@@ -506,6 +508,23 @@ async function startMcp() {
 		console.error("[Main] Failed to start MCP server (continuing without it):", error);
 		mcpService = null;
 	}
+}
+
+/**
+ * Forward an MCP data change to the renderer so its query cache can invalidate.
+ *
+ * One-way, main-to-renderer, and it carries no data - only which family went
+ * stale (the #278 hardening posture: nothing here takes renderer input, and a
+ * push of engine data would be a second source of truth beside the queries).
+ *
+ * The window is read at send time, not captured: the MCP server starts before
+ * the window exists and outlives a renderer reload, so a captured reference
+ * would either be null forever or point at destroyed `webContents`.
+ */
+function sendMcpDataChanged(event: McpDataChangedEvent): void {
+	const contents = mainWindow?.webContents;
+	if (!contents || contents.isDestroyed()) return;
+	contents.send("mcp:data-changed", event);
 }
 
 async function stopMcp() {
--- a/app/electron/mcp/data-changed.conformance.test.ts
+++ b/app/electron/mcp/data-changed.conformance.test.ts
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * Keeps the two copies of the MCP data-entity list honest.
+ *
+ * The main process owns `MCP_DATA_ENTITIES`; the renderer needs the same union
+ * to type the IPC payload and its invalidation map, and production code under
+ * `electron/` cannot import from `app/src` (the reasoning is in
+ * `tsconfig.node.json`). So the type is written twice on purpose - and a
+ * divergence would be silent in the worst possible way: the main process would
+ * emit a family the renderer's map has no entry for, and the write would stay
+ * invisible, which is the bug the channel exists to fix.
+ *
+ * This file is the one place the two meet. It reaches across the boundary the
+ * way `tools.test.ts` already reaches for `@/constants/load-test`.
+ */
+
+import { describe, expect, test } from "vitest";
+import { MCP_DATA_ENTITIES } from "./tools.js";
+import type { McpDataEntity } from "@/types/domain";
+
+/*
+ * Exhaustive by construction: TypeScript rejects this literal if the renderer's
+ * union gains a member that is not listed, and rejects a listed member the
+ * union does not have. The runtime comparison below catches the other
+ * direction - the main process gaining an entity the renderer never heard of.
+ */
+const RENDERER_ENTITIES: Record<McpDataEntity, true> = {
+	request: true,
+	environment: true,
+	run: true,
+	cookie: true,
+	config: true,
+};
+
+describe("McpDataEntity mirror", () => {
+	test("the main process and the renderer name the same entities", () => {
+		expect([...MCP_DATA_ENTITIES].sort()).toEqual(Object.keys(RENDERER_ENTITIES).sort());
+	});
+
+	test("the list it scanned is not empty", () => {
+		// A comparison of two empty lists passes while proving nothing - the
+		// failure mode CLAUDE.md records for source-scanning guards.
+		expect(MCP_DATA_ENTITIES.length).toBeGreaterThan(0);
+	});
+});
--- a/app/electron/mcp/data-changed.test.ts
+++ b/app/electron/mcp/data-changed.test.ts
@@ -1,0 +1,239 @@
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * The `mcp:data-changed` emit at the dispatch chokepoint.
+ *
+ * `tools.test.ts` owns what each tool sends to the engine; this owns the one
+ * thing that happens *after* a call succeeds - the renderer being told what
+ * went stale. Kept in its own file because every test here needs the notifier
+ * in the context, which none of the tests over there want.
+ */
+
+import { describe, expect, test, vi } from "vitest";
+import {
+	dispatchTool,
+	MCP_DATA_ENTITIES,
+	TOOLS,
+	type McpDataChangedEvent,
+	type ToolContext,
+} from "./tools.js";
+import { resolveSafetyConfig, type McpSafetyConfig } from "./config.js";
+import type { EngineClient } from "./engine-client.js";
+
+/** Minimal engine client for the calls exercised below. */
+function fakeClient(overrides: Partial<Record<keyof EngineClient, unknown>> = {}) {
+	return {
+		health: vi.fn().mockResolvedValue({ status: "ok", version: "1.2.3" }),
+		listCollections: vi.fn().mockResolvedValue([]),
+		listRuns: vi.fn().mockResolvedValue([]),
+		createRequest: vi.fn().mockResolvedValue({ id: "req_1", name: "New" }),
+		updateConfig: vi.fn().mockResolvedValue({ entries: [] }),
+		getConfig: vi.fn().mockResolvedValue({ entries: [] }),
+		getEnvironment: vi.fn().mockResolvedValue({ id: "env_1", name: "Dev", variables: {} }),
+		updateEnvironment: vi.fn().mockResolvedValue({ id: "env_1", name: "Dev" }),
+		executeRequest: vi.fn().mockResolvedValue({ statusCode: 200 }),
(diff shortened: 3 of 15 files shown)


athrvk:
MCP does not have tools to do CRUD ops on collection ? 
was there a issue for this ? is there an issue for this ? 
If it's a small change of capability maybe expose it in this PR itself Or create a new issue

athrvk:
Correct - there are no collection CRUD tools, and there was no issue for it. Filed as #378.

The gap is wider than collections. The engine has full CRUD on both routes (`GET/POST/PUT/DELETE /collections` and `/requests`); MCP exposes a fraction:

| Entity | Create | Read | Update | Delete |
|---|---|---|---|---|
| Collection | - | `list_collections` | - | - |
| Request | `create_request` | `list_requests` | - | - |
| Environment | - | `list_environments` | `update_environment` | - |

So `create_request` takes a `collectionId` an agent has no way to obtain for new work, and every MCP write is one-way - an agent that creates the wrong request cannot remove it.

**Not folding it into this PR**, for one reason that isn't size: `delete_collection` cascade-deletes every descendant collection and every request in them, and the `write` toggle is the wrong gate for that. It's a single session-wide switch a user flips once to let an agent save a request, not consent to destroying a subtree. The precedent for "destructive enough to ask a human" is `start_load_run` - elicitation when the client supports it, a `confirmed: true` flag when it doesn't. Deciding that, and whether the prompt states the descendant counts first, is a safety-model call I'd rather you made than have arrive inside an invalidation-channel PR. #378 has it as the first acceptance criterion.

The mechanical part is genuinely small (five tools, five `EngineClient` methods, no engine change), and #378 sequences after this PR because it extends what this one adds: each new tool declares its `invalidates`, and the collection tools are what finally give the `collection` entity an emitter. That's the deviation noted in the PR body - I left `collection` out because nothing emitted it, and a mapping branch with no emitter is dead wiring. The renderer's map is a `Record` over the union, so #378 won't compile until it names a reader.

Worth noting the repo already ruled on this shape: #318 deleted `getRequest`/`getGlobals` as dead code, with the explicit finding that planned-but-unbuilt MCP surface belongs in an issue rather than as unreferenced client methods.

CI on this PR is green across all three platforms.

---

Issue #369, opened 2026-08-07
app(save-status): three callers still arm the unguarded idle timer that stomps an unrelated Dock status


athrvk:
Split out of the collection-tree correctness batch (#361), which fixed the two instances it owned and left the rest rather than widening its diff.

## Problem

`useSaveStore` holds one status for the whole app, and the Dock renders it. Four
call sites reported a successful save as `completeSave()` followed by a bare
`setTimeout(() => setStatus("idle"), TIMING.STATUS_RESET_MS)`. That timer fires
two seconds later regardless of what happened in between, so it clears whatever
status is current when it lands - an `error` another surface published (the Dock
goes blank next to a failure toast), or a `pending` from an edit made since ("Unsaved
changes" disappears while the changes are still unsaved).

`triggerSave` has always guarded its own reset (`if (get().status === "saved")`),
and the store documents the rule as **"only the context that set a status clears
it"** (`docs/app/state-management.md`). The hand-rolled copies never received it -
the repo's hand-rolled-copy-of-a-primitive shape.

#361 added `completeSaveThenIdle()` to `save-store.ts` (set `saved`, then reset to
`idle` only if that `saved` is still the status on screen), moved `runSave` and
both collection-tree rename paths onto it, and left these three:

- `app/src/hooks/useSaveManager.ts:134-137`
- `app/src/modules/settings/main/SettingsMain.tsx:220-228`
- `app/src/modules/variables/main/VariableTableEditor.tsx:285-286`

## Fix / solution pathway

Replace each `completeSave()` + `setTimeout` pair with `completeSaveThenIdle()`.
No new API - the store action already exists and is documented. `SettingsMain`
additionally tracks whether the `pending` on screen is its own; check that logic
still reads correctly against the shared action rather than around it.

Worth confirming while there: `ContextBar` calls `completeSave()` with no reset at
all (`ContextBar.tsx:115`), so a committed variable leaves "Saved" in the Dock
until something else changes it. Decide whether that is deliberate (it is the only
non-draft commit path) and either move it over or say why not, in the PR.

## App changes

The three files above, plus `ContextBar.tsx` if the answer there is "move it". No
engine change; MCP: none.

## Tests

Mutation-checked, one per call site: with a save completed, publish a `failSave`
from another context, advance fake timers past `TIMING.STATUS_RESET_MS`, and assert
the status is still `error`. Reverting a call site to the raw timer must redden its
[...]

PR #376, opened 2026-08-08
feat(app/context-bar): a per-tab section registry, and copy-as-curl/fetch from the composed request

### LINKED ISSUE #344: app(context-bar): make the context bar alive - a per-tab section framework with auth context, cookies-to-be-sent, last result, and copy-as-curl/fetch code generation
**Owner direction, 2026-08-07**: the context bar should be "alive - useful and smart, like an actual context bar", including copy-pastable curl/fetch/etc. snippets. This issue is the scoped design: every Phase-1 section is buildable from hooks that already exist in the query layer (inventoried at master `f703eaa`), so the feature needs **no engine work**. Prerequisites: #341 (write-back correctness of the variables section this keeps) and #342 (the structural/a11y batch whose shell this builds in) - both land first, same file.

## What the bar is today, and the gap

One section ("Variables in scope"), rendered only for `request` tabs, `null` for the other six tab types (`ContextBar.tsx:67`; `TabType` union at `app/src/stores/tabs-store.ts:14-21`). It is a variables quick-editor, not a context bar. Postman's equivalent panel is famously dead weight - this is a differentiation surface for the Postman + JMeter + k6 target.

## The design: a per-tab section registry

One framework decision, then sections as leaf components:

- A section registry keyed by `TabType`: each entry is `{ id, title, Component, appliesTo(tab) }`, rendered as collapsible sections (collapsed state persisted per section id in `layout-store`, beside `contextBarWidth`). The bar renders the sections whose `appliesTo` matches the active tab - so it has real content on request, collection, and run tabs, which also completes the toggle-coherence fix #342 starts (the pressed state mirrors "has content").
- Each section is a self-contained component over existing queries - no shared god-state. Sections mount only while the bar is open and expanded (collapsed sections must not run their queries - `enabled` gating, so the bar stays free when unused).
- Loading/empty states per section, design-system compliant (this lands after #342, so on the fixed shell: `aside` landmark, inner scroll wrapper, header row).

## Phase 1 - request tab (all hooks exist today)

1. **Variables in scope** - the existing section, kept as-is post-#341/#342.
2. **Auth context** - which auth applies and *where it came from* (own vs inherited from an ancestor collection - the request-builder already models inheritance; reuse its resolution, do not re-derive), plus OAuth2 token state with fetch/clear via `useOAuth2TokenStatusQuery` / `useFetchOAuth2TokenMutation` / `useClearOAuth2TokenMutation` (`app/src/queries/oauth.ts`). Answers "why 401" without leaving the tab.
3. **Cookies to be sent** - the #336 jar via `useCookiesQuery` (`app/src/queries/cookies.ts:29`), filtered to the request's host, with the per-scope clear mutation. **Fidelity caveat, stated in the UI**: host-filtering is an approximation - do NOT re-implement libcurl's domain/path/secure matching (the trap #339 documents); label the section "cookies for this host" and link the exactness story to #339's raw-request view.
4. **Last result** - `useLastDesignRunQuery(requestId)` (`app/src/queries/runs.ts:233`): status, duration, timestamp of the last design send.
5. **Code** - copy-pastable snippets, the owner-requested capability. Design constraints:
   - **Generate from the composed request, not the template**: `api.composeRequest()` (`app/src/services/api.ts:262`, `POST /compose`) returns the fully-resolved request (variables substituted, inherited auth applied) - Vayu's snippets are *what will be sent*, which Postman's template-based generator cannot promise. Compose on section expand / explicit refresh, not per keystroke.
   - **Secrets are the leak vector**: templated-vs-resolved toggle, secrets masked by default in resolved output with explicit reveal - the same policy family as #339's raw-view decision.
   - **Targets**: curl (POSIX quoting; a cmd.exe/PowerShell variant is a second target, not a flag - quoting is where generators go wrong, see the #322/#331 Windows quoting work) and JS `fetch` in Phase 1; httpie/Python-requests/PowerShell later behind the same interface.
   - **Architecture**: pure per-target generator functions over `ComposedRequest` in `app/src/services/codegen/`, mirroring the existing `app/src/services/curl/` *parser* (Vayu currently parses curl in but generates nothing out - this closes the symmetry). Mutation-checked unit tests against hostile bodies: quotes, newlines, unicode, multipart, empty bodies.
   - **Cookie fidelity**: the composed request does not carry jar cookies (libcurl attaches them at transfer). Phase 1 states that limitation in the section; once #339 ships its captured `HEADER_OUT` frame, add "copy last send as curl" generated from the captured block - perfect fidelity for free.
6. **Environment line** - active environment name + switcher entry point, since sections 1/3 are environment-scoped.

## Phase 2 - collection and run tabs

- Collection tab: variables defined at that level (edit via the same commit path #341 fixes), auth configured there, child counts, last smoke outcome.
- Run tab: run config summary (mode, target, caps), environment used, link to the source request.
- Welcome/dashboard/variables/settings tabs: no sections - and the #342 toggle rule handles the pressed state.

Phase 2 ships as a follow-up PR on the same framework; if scoping grows, split it out then - do not gold-plate Phase 1.

## App changes

All renderer: `ContextBar.tsx` (framework + sections), `app/src/services/codegen/` (new), `layout-store.ts` (collapsed-state persistence), section components under `app/src/components/layout/context-bar/` (or `modules/` placement per `docs/app/COMPONENTS.md` - decide and follow the doc). MCP: none (verified - no MCP surface reads the bar). Engine: none.

## Tests

- Framework: registry renders exactly the sections for the active tab type; collapsed sections do not fire their queries (assert the query hook's `enabled` gate); collapse state persists.
- Codegen: per-target unit tests incl. the hostile-body matrix, mutation-checked; secret masking on by default in resolved mode; templated mode never contains resolved secret values.
- Sections: auth section shows inherited-from source; cookie section filters by host and labels the approximation; last-result renders empty state cleanly.

## Docs (same commit)

- `docs/app/COMPONENTS.md` - the new section components and registry.
- `docs/app/architecture.md` - one paragraph on the section-registry pattern.
- `docs/app/state-management.md` - the collapsed-state addition to layout-store.
- User-facing docs page for the bar if `docs/` grows one for the app UI (check nav via the docs-site rules before adding).

## Sequencing

**After #341 and #342** (same file, ordered: correctness, then structure, then this). The Code section is independent of #339 but gains the "copy last send" option when #339 lands - either order, the tie-in is additive. Parallel-safe against #337 (jar write half - the cookie section reads via `GET /cookies` regardless), #338, #343, the engine track (#197/#287), #303/#340, #134. Phase 2's collection-tab variables editing inherits #341's commit path unchanged.

Out of scope: engine-side codegen; a full "copy as code" for load-test scripts (k6 export is its own future feature and belongs near #340's scenario design); re-implementing cookie matching; MCP exposure of snippets.

## Acceptance criteria

- [ ] The bar renders useful sections on request tabs (the six above) and Phase-2 content or a clean empty state elsewhere; the toggle never lights up over nothing.
- [ ] Copy-as-curl and copy-as-fetch produce runnable output for the hostile-body matrix, from the composed request, with the secret policy enforced and pinned by tests.
- [ ] Collapsed sections provably run no queries; collapse state survives restart.
- [ ] All listed docs updated same-commit; `pnpm test`, `pnpm type-check`, `pnpm lint`, `pnpm format:check` green.

---

athrvk:
## Description

Phase 1 of #344: the context bar becomes a frame plus a registry of sections, and the request tab gets five of them.

**The plan.** The root cause is that the bar had no framework - it was one hardcoded section rendered for one tab type, and "does the bar have content?" was answered twice, independently (`tabType === "request"` in `context-bar-content.ts`, a `return null` inside the bar). The approach is one ordered registry of `{ id, title, appliesTo(tab), Component }` that both the bar and the Dock's toggle read through `sectionsForTab` / `contextBarHasContent`, with sections as leaf components over the existing query layer and no bar-wide shared state. The alternative I rejected was keeping the sections inline in `ContextBar.tsx` behind conditionals: it is less code today, but it re-creates the drift the toggle fix already had to clean up once, and Phase 2's collection/run sections would each need edits in three places instead of one array entry.

**Sections (request tab).** Variables in scope (moved out of `ContextBar.tsx` unchanged, commit path and all), Auth, Cookies for this host, Code, Environment.

- **Auth** reports the effective mode and where it came from through the shared `resolveAuthSource`, so it cannot claim one source while execution uses another. It distinguishes "no ancestor configured any" from "an ancestor is set to No Auth, which terminates the walk" - collapsing those is what sends someone looking in the wrong collection. For OAuth 2.0 it embeds `TokenStatusRow`.
- **Cookies** filters the active environment's jar by host and **says in the UI that host filtering is an approximation**, pointing at the raw-request view for the exact answer. Per the issue, libcurl's matching is deliberately not re-implemented.
- **Code** is the substantive new capability, below.

**No "last result" section, deliberately.** #344 specced one; built as specced it was a strict duplicate of `ResponseStatusBar` - same status chip, duration and age, from the same stored run, in the response pane on the same screen (the builder restores that run into the pane whenever nothing is in memory). It was removed, and `registry.test.tsx` guards the id so it cannot come back quietly. The version that would earn the slot is a *trend* across recent sends, which needs the paginated `GET /runs` to carry each design run's result first - tracked in #380.

**Query cost.** A section's component is mounted only while its section is expanded, so a collapsed section registers no queries - the bar is open on every request tab, so that is the load-bearing property, and it is asserted as "the component was never called" rather than as a proxy. Collapse state persists per section id in `layout-store`, collapsed-by-exception (so a section added in a later release ships expanded) and as an array (a `Set` JSON-serializes to `{}`, so every collapse would last until the next launch).

**`services/codegen/`** closes a symmetry: Vayu parses curl *in* (`services/curl/`) and generated nothing out. Snippets are generated from **`POST /compose`'s output** - variables substituted, `inherit` walked, by the engine that will send it - so what you paste is what Vayu would put on the wire; a Templated mode generates from the stored request with references intact. Auth is flattened into headers/query/`-u` once in `prepare.ts`, because the engine keeps `auth` beside the request and applies it at send time - a snippet built from the composed *headers* alone would authenticate in Vayu and 401 in a terminal. Modes no static command can reproduce (OAuth 2.0, digest, AWS, NTLM) produce a note rather than a silent drop, and the section states that jar cookies are attached at transfer time and are not in the snippet. Resolved secrets are masked by default and revealing is explicit; masking runs **before** quoting, since after it a value containing a quote no longer matches itself.

**Five targets**, each a pure function behind one `CODE_TARGETS` entry. Quoting is per target, never a flag on another one:

| Target | String rule | Multipart vs urlencoded |
|---|---|---|
| curl | POSIX single quotes, `'` → `'\''` | `-F` vs `--data-urlencode` |
| JS fetch | `JSON.stringify` (it *is* the JS literal grammar) | `FormData` vs `URLSearchParams` |
| Python requests | `JSON.stringify` - JSON's escapes are a strict subset of Python 3's | `files=` vs `data=` |
| HTTPie | POSIX, shared with curl | `--multipart` vs `--form` |
| PowerShell | single quotes, `'` → `''` | `-Form` vs `-Body` |

Three of those distinctions fail *silently* if you get them wrong: HTTPie's `--form` alone is urlencoded, requests only goes multipart via `files=`, and a POSIX `'\''` inside PowerShell puts a backslash in the data and leaves the string unterminated. PowerShell emits `Invoke-RestMethod` rather than `curl`, which on Windows PowerShell is an alias for that cmdlet.

### Deliberate deviations from the issue

- **`TokenStatusRow` is now exported** from `components/shared/OAuth2Form`, where its index explicitly called it "an internal detail". The auth section needs exactly that control, and a local copy would be a second copy of the cache-key derivation and the interactive-authorize flow - the trap `CLAUDE.md` names. The module doc says why.
- **The Environment section links to the Variables drawer** rather than embedding a switcher. The TitleBar's switcher is inline markup, not a primitive, and copying it would copy its switch-confirmation flow - one of the few places a click has consequences the user cannot see.
- **The last-result section was dropped**, against the issue's spec - see above and #380.
- **The language picker is a `Select`, not a segmented control.** `ToggleGroup` is an `inline-flex` with no wrap and five target names do not fit the bar's 252px; the resolved/templated switch stays a ToggleGroup, where both choices being visible is the point.
- **`ContextBar.commit-scope.test.tsx`, `.save-error.test.tsx` and `.save-status.test.tsx` are renamed** to `context-bar/VariablesSection.*` and render the section directly; every assertion is preserved. The markup test's variable-row half moved to `VariablesSection.markup.test.tsx`, and the frame test stubs the registry.
- **Phase 2 (collection and run tabs) is not in this PR**, as the issue scopes it. Filed as #377.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

All green locally at `9891ec4`, and CI green on the same commit:

- `pnpm test` - **325 files, 3004 tests passed**
- `pnpm type-check` - clean
- `pnpm lint` - clean (zero errors, zero warnings)
- `pnpm format:check` - clean
- `mkdocs build --strict -f .github/mkdocs.yml` - clean (docs-touching change)

No engine change, so the engine suite was not run. No pre-existing failures observed.

New coverage, mutation-checked:

- `services/codegen/codegen.test.ts` - a hostile body matrix (single quotes, double quotes and backslashes, newlines, shell/PowerShell expansions, unicode, a lone quote) run against **all five targets**, asserted by *decoding* each language's emitted literal rather than by string equality; the multipart/urlencoded pair per target; auth flattening per mode; the query-param api key landing before the URL fragment; masking across url/header/body/auth at once, longest-secret-first, empty secrets ignored, and a secret containing a quote (which masking-after-quoting would miss). A shared block asserts every *registered* target masks secrets, carries the unreproducible-auth note and applies bearer auth, so a target that reached around `prepare.ts` fails there.
- `context-bar/registry.test.tsx` - the sections and their order, unique ids, that `contextBarHasContent` is *derived from* `sectionsForTab` rather than restating it, and that no `last-result` section exists.
- `ContextBar.markup.test.tsx` - collapsed sections' components are never called; collapse persists; `aria-expanded` per section; plus the frame cases (landmark, close-button box, inner-scroll wrapper, single left edge, push/overlay).
- `CodeSection.test.tsx` - composes for the active environment and uses the resolved payload (not the stored template); inherited auth reaches the snippet; every registered target is offered and generated from one compose; masked by default, revealed only on an explicit act; **templated output never contains a resolved secret**; compose failure says so instead of showing a stale snippet.
- `CookiesSection.test.tsx`, `AuthContextSection.test.tsx`, `VariablesSection.markup.test.tsx`, `layout-store.test.ts` (collapse persistence read out of `localStorage`, which is the only assertion that catches the `Set` trap).

## Checklist

Docs updated in the same commits: `docs/app/COMPONENTS.md` (the registry, the section table, the codegen quoting table, why there is no last-result section), `architecture.md` (the section-registry pattern, `services/codegen/`), `state-management.md` (`contextBarCollapsedSections` and why it is an array, the `queryKeys.compose` family, and the save-store references repointed at `VariablesSection` after the move).

## Related Issues

Closes #344

Follow-ups filed: #377 (Phase 2 - collection and run tab sections), #380 (recent-sends trend, engine half first).


athrvk:
fix merge conflict

athrvk:
Claiming this - merging latest `master` into this branch and resolving the conflicts this run.

---

athrvk:
Merged `master` (`bf3abee`) in and pushed as `b1e80f4`.

**The conflict was semantic, not textual.** #373 (issue #369) routed the variables commit's success through `completeSaveThenIdle` and **deleted** `completeSave` - inside the `usePendingCommits` block this branch moves to `context-bar/VariablesSection.tsx`. Taking either side alone loses something: ours drops the fix and would not even compile against the new store, theirs restores a block this branch deliberately emptied. So `ContextBar.tsx` keeps this branch's frame, and master's change is applied where the code now lives, with the reason recorded beside it.

Two follow-on edits the conflict implied:

- `ContextBar.save-status.test.tsx` arrived with that PR and renders the bar to exercise the commit path. It moves to `context-bar/VariablesSection.save-status.test.tsx` and renders the section - the same retargeting the three other variables tests already had in this PR. **Assertions unchanged**, and it still reddens if the success stops going through `completeSaveThenIdle`.
- `docs/app/state-management.md` auto-merged cleanly, but master's save-store section names `ContextBar` four times for code that is now in `VariablesSection`. Repointed, so the doc still describes where the code is.

Re-verified on the merge commit: **2931 tests pass** (322 files), `type-check`, `lint` and `format:check` clean, `mkdocs build --strict` clean. No engine change on this branch, so the engine suite was not run.

One thing I left alone, since it is not this PR's: `engine/test_script_scopes.db.bak` is committed on `master` (it arrived with #374) and looks like a stray test artifact rather than something intended to be tracked. Worth a separate cleanup if you agree.

---

athrvk:
Merge conflict resolved; branch is now on current `master` (`93a2f93`).

A concurrent run of the routine had already pushed a resolution (`b1e80f4`) while I was working on my own. Rather than force-push over it, I kept it and built on top - it reached the same conclusion on the conflict and carried a doc fix mine didn't (repointing master's `ContextBar` references at `VariablesSection`). Two commits on top:

**`d97d017`** - merges the master commit that resolution missed. It merged `bf3abee`, but master had moved one further to `93a2f93` (#375, the MCP data-changed work). Clean merge, no conflicts.

**`e7f2456`** - a gap that belonged to neither side of the merge alone, and so was invisible to both. #375 added `lib/mcp-invalidation.ts`, which maps an MCP write to the query keys that read it, and enumerated the readers that existed when it was written. This branch added `queryKeys.compose` - `POST /compose`, which the Code section builds its curl/fetch snippet from. `update_environment` is a live MCP write tool and compose substitutes exactly the variables it changes, so once both were in the tree an agent's environment write left every cached composition, and every snippet built from one, showing pre-write values. Nothing refetches compose on its own, so it stayed wrong until the tab was reopened.

Invalidated at the `compose.all` prefix rather than the written environment's key, since a request composed against a *different* environment still reads globals the write may have shadowed. `create_request` needs nothing - a request that didn't exist has no composition cached. Covered by a test and added to the invalidation table in `docs/app/state-management.md`.

For the record, the conflict itself: #373's save-status fix changed one line inside `usePendingCommits`, a block this branch had moved whole into `context-bar/VariablesSection.tsx` - so the hunk landed on a function no longer in the conflicted file. Resolution keeps this branch's `ContextBar.tsx` and applies `completeSaveThenIdle()` at the code's new home. `ContextBar.save-status.test.tsx` arrived with that fix and drives the behaviour through the bar, which now renders six registry sections rather than the variables editor inline; it moved to `context-bar/VariablesSection.save-status.test.tsx` and renders the section directly - the same retargeting the three other variables tests already got here. Assertions unchanged.

**Verification on the merged tree:** `pnpm test` 326 files / 2960 tests passed, `pnpm type-check` clean, `pnpm lint` clean, `prettier --check` clean on touched files. The new invalidation test is mutation-checked - reverting the added line fails that test and only that one. `mkdocs` isn't installed in this environment so `--strict` wasn't run; the docs change is a single table cell with no new links or anchors.

Unrelated find, left out of this diff and filed as #379: `engine/test_script_scopes.db.bak` (120KB of test SQLite) is tracked on master, and `.gitignore` has no `*.db` rule to stop the next one.

---

athrvk:
the code section has just two curl and js fetch, it needs more

athrvk:
Added three more, pushed as `4ef4e37`. The Code section now offers **curl, JS fetch, Python requests, HTTPie and PowerShell** - the three #344 named as fitting "the same interface", and they cost what that interface promised: one entry in `CODE_TARGETS` and one pure function each. `prepare.ts` already owned auth flattening and secret masking, so none of that is repeated.

**Quoting is per target, never a flag on another one** - the part that cannot be shared, and where generators go wrong:

| Target | String rule | Multipart vs urlencoded |
|---|---|---|
| curl | POSIX single quotes, `'` → `'\''` | `-F` vs `--data-urlencode` |
| JS fetch | `JSON.stringify` (it *is* the JS literal grammar) | `FormData` vs `URLSearchParams` |
| Python requests | `JSON.stringify` - JSON's escapes are a strict subset of Python 3's, and each means the same thing in both | `files=` vs `data=` |
| HTTPie | POSIX, shared with curl | `--multipart` vs `--form` |
| PowerShell | single quotes, `'` → `''` | `-Form` vs `-Body` |

Three details worth the review time, each of which fails *silently* if you get it wrong:

- **PowerShell emits `Invoke-RestMethod`, not `curl`.** On Windows PowerShell `curl` is an alias for that cmdlet, so a snippet saying `curl` runs something other than what the reader expects. Its escape is a doubled quote; a POSIX `'\''` here would put a literal backslash in the data and leave the string unterminated.
- **HTTPie's `--form` alone is urlencoded**, not multipart - it only goes multipart when a field carries a file, which a pasted snippet never does. So a multipart body gets `--multipart` explicitly, or HTTPie would quietly send a urlencoded body for a request Vayu sends as multipart. Headers are bare `Name:value` words and the body is `--raw`, so composed bytes go out unchanged rather than HTTPie building a *new* JSON object from `key=value`.
- **requests sends multipart via `files=`**; a multipart body passed as `data=` is silently urlencoded.

**UI change:** the language picker is now a `Select`. `ToggleGroup` is an `inline-flex` with no wrap and five target names do not fit the bar's 252px. The resolved/templated switch stays a ToggleGroup, where both choices being visible is the point.

**Tests:** the hostile-body matrix (quotes, backslashes, newlines, shell/PowerShell expansions, unicode, a lone quote) now runs against all five, decoding each language's emitted literal rather than comparing to one. A shared block asserts **every registered target** masks secrets, carries the unreproducible-auth note, and applies bearer auth - so a target that reached around `prepare.ts` fails there rather than in one target's own file. One case per target pins the multipart/urlencoded pair.

Re-verified: **3008 tests pass** (326 files, +77), `type-check`, `lint`, `format:check` and `mkdocs build --strict` all clean. `docs/app/COMPONENTS.md` gains the quoting table; `architecture.md` lists the targets.

Worth noting one thing my own test harness caught rather than the code: the PowerShell body assertion originally split the snippet on newlines, which silently truncated the multi-line body case and would have passed against a broken generator. It slices to the blank line now - a PowerShell single-quoted string may legitimately span lines.

---

athrvk:
You're right, and it's worse than "sort of useless" - it's a **strict duplicate of something on the same screen**. Removed in `9891ec4`.

**The check.** The section showed status chip, duration, relative age. `ResponseStatusBar` is, in its own file comment, "the bar above a response: status chip, elapsed time, payload size" - plus a relative age, painted with the same `StatusCodeBadge`. And it isn't only populated after a live send: `RequestBuilderProvider` restores the last design run into that pane whenever nothing is in memory, reading **the same `report.results[0]`** the section read. So there was no state in which the section said anything the pane wasn't already saying more fully, a few hundred pixels to the left. Not a thin summary - a poorer copy.

**What I considered instead of deleting.** The slot is worth something; the content wasn't. What the pane structurally *cannot* show is more than one send - "was that 500 the first one, has latency been drifting, when did this last work" - which is exactly what a request tab lacks and what costs a trip to History today.

I didn't build it here, for a concrete reason rather than a scope one. The paginated `GET /runs` serializes rows **without** the exchange: `attach_design_result` is called only on the single-run route (`runs.cpp:854`), while the list path calls plain `serialize(run)` (`:777`). So a row carries `status`/`startTime`/`endTime` but no status code and no latency, and last-N becomes one list call plus N report fetches - and the report path "loads every result row for a run and JSON-parses each `trace_data`" (`:322`). Running that on every expanded request tab isn't acceptable, and making it cheap is an **engine** change, which is precisely what #344's premise ("needs no engine work") rules out of this PR.

So: filed **#380** with the engine half first (attach the design result to list rows, keeping the existing guard that stops a load run's unbounded results being fetched), then the section. It also records two traps for whoever picks it up - the new query must not sit under `queryKeys.runs.lists()` (the delete-run patch walks that prefix as `InfiniteData`, which is why `runs.lastDesign` has its own family), and it must use a new section id.

**What changed here:** section and its test deleted, registry entry removed, `registry.test.tsx` gains a guard that `last-result` is not in the registry so the duplicate cannot come back quietly, and `COMPONENTS.md` records why the absence is deliberate rather than an oversight. The bar now has five sections: variables, auth, cookies, code, environment.

Re-verified: **3004 tests pass** (325 files), `type-check`, `lint`, `format:check` and `mkdocs build --strict` clean.

If you'd rather have the trend version in this PR than in #380, say so and I'll do the engine half here - it's a contained change to one serialization path.

---

PR #394, opened 2026-08-08
feat(app/context-bar): phase 2 - collection and run tab sections on the existing section registry

### LINKED ISSUE #377: app(context-bar): phase 2 - collection and run tab sections on the existing section registry
Phase 2 of #344, split out because #344's Phase 1 shipped the framework and the request tab and the issue says Phase 2 ships as a follow-up PR on the same framework. Anchors verified against PR #376.

Re-anchored 2026-08-08 at `d2fe130` after PR #376 merged.

## Problem

The context bar now renders five sections on a request tab, and nothing on the other six tab types. That is correct-and-incomplete rather than broken: `contextBarHasContent` is derived from the registry, so the Dock toggle is honestly dark on those tabs and the bar renders nothing rather than an empty shell. But a collection tab and a run tab both have context worth showing, and #344's design already names it.

## Fix / solution pathway

Each section is one entry in `CONTEXT_BAR_SECTIONS` (`app/src/components/layout/context-bar/registry.ts`) and one leaf component beside the existing five. **No framework change is expected** - if one turns out to be needed, say so in the PR rather than working around it, because that is the registry failing its purpose.

- **Collection tab** (`appliesTo: tab => tab.type === "collection"`): variables defined at that level (edited through the commit path #341 fixed - reuse `VariablesSection`'s scope machinery rather than a second copy), the auth configured there, child counts, and the last smoke outcome.
- **Run tab**: run config summary (mode, target, caps), the environment used, and a link to the source request.
- **welcome / dashboard / variables / settings**: no sections, deliberately. The toggle rule already handles the pressed state, so nothing is needed for these beyond not adding entries.

Watch the "written but never read" rule on the run summary: `RunSummary` declares `followRedirects` / `maxRedirects` that nothing renders today (see the note on the type in `app/src/types/domain.ts`), and the run tab section is a plausible home if you want one - but decide, do not add a field without a reader.

## App changes

`app/src/components/layout/context-bar/registry.ts` (entries), new section components in the same directory. No engine change. MCP: none - verified during Phase 1, no MCP surface reads the bar.

## Tests

- Registry: `registry.test.tsx` currently asserts the other six tab types have **no** sections - that case moves rather than being deleted, and the collection/run entries get their own ordering assertion.
- Sections: each one's loading and empty bodies, and that a collapsed section still runs no queries (the Phase 1 guarantee, per new section).
- Dock: the toggle is now pressed on a collection tab with the bar open - `Dock.context-toggle.test.tsx`'s `it.each` list currently asserts the opposite for `collection` and `run`, so those two cases move out of the list. That move is the mutation-check that the derivation works: it should be the *only* Dock edit needed.

## Docs (same commit)

`docs/app/COMPONENTS.md` - the section table gains the collection/run rows. The "Phase 1 is the request tab" wording lives in the registry header comment (`app/src/components/layout/context-bar/registry.ts:17-18`), not in COMPONENTS.md as merged - update it there. `docs/app/architecture.md` if the registry shape changes (expected: it does not).

## Sequencing

**Unblocked** - the registry is on master (PR #376 merged as `d4ca67b`, closing #344). Parallel-safe against everything else open - the files are new or are one array. Not blocking anything. Coordinate with #380 (recent-sends section): both add entries to the same registry array and both edit the request-tab ordering assertion at `registry.test.tsx:45` - trivial, land in either order and rebase.

Out of scope: any change to the request-tab sections; anything in `services/codegen/` (#390 owns the known quoting edges).

## Acceptance criteria

- [ ] Collection and run tabs render their sections; the Dock toggle lights up on them with no second edit beyond the moved test cases.
- [ ] The four remaining tab types still render nothing and still leave the toggle dark.
- [ ] Collection variables are edited through the existing commit path, not a copy of it.
- [ ] No field is added without a reader; suites green; docs updated same-commit.

---

athrvk:
## Description

Phase 2 of #344: the context bar drew five sections on a request tab and nothing on the other six tab types. This adds the collection and run sections as **five entries in `CONTEXT_BAR_SECTIONS` plus five leaf components** - no framework change, which is what the registry was for.

**Plan.** Root cause is simply that Phase 1 shipped the registry with request-tab entries only, so `appliesTo` was never exercised past the tab type. The approach is one entry per section and one leaf component beside the existing five, each gating on `tab.type` **and** `tab.entityId` (a collection or run tab open on nothing renders no pane either, so a section there would query nothing while lighting the Dock toggle over an empty bar). The alternative I rejected was giving the collection and run tabs one combined section each: fewer entries, but it would have made the collapse state all-or-nothing for content with very different cost - the variables list is editable and the contents count is a cheap line - and the registry's per-section mount gating is the whole cost model.

**Collection tab**

| Section | What it shows |
|---|---|
| `collection-variables` | The definitions this collection owns (not a resolved set), editable in place. A disabled definition stays on screen marked `off` - it is still defined here, it is simply not sent. |
| `collection-auth` | The mode this collection is set to, plus what a descendant set to Inherit picks up. One set to plain No Auth names the ancestor that answers instead, via the shared `resolveAuthSource` walk, so it cannot claim a source execution would not use. |
| `collection-contents` | Direct child counts (requests, sub-collections), matching what the tree shows under the folder. A subtree total would disagree with the rows the user can see. |

**Run tab**

| Section | What it shows |
|---|---|
| `run-config` | Mode, duration, target RPS, concurrency, iterations, ramp and requested protocol from the run's stored `configSnapshot`, worded by `loadTestModeLabel` / `formatConcurrency` so the same run is never described two ways. A design run has no `mode` key and reads as "Single send" rather than "nothing was recorded". |
| `run-source` | The environment the run used and a link opening the request it ran from. |

**Reuse over a second copy.** The row markup and the commit path moved out of `VariablesSection` into `VariableRow.tsx` and `variable-commit.ts` (`useVariableCommit`) rather than being duplicated - the remount key, the Escape restore, the uncontrolled input a rejected save depends on, and the read-the-cache-at-commit-time rule are the same act on both lists, and a hand-rolled copy would not receive their fixes. The collection list hands the commit a `ResolvedVariable` naming this collection as the source, so an edit lands exactly where a collection-scope edit from the request tab lands. `VariablesSection`'s behaviour is unchanged; its existing four test files pass untouched apart from the source-scan guard, which now scans both halves (scanning only the shrunken component would let the unguarded `parentId` walk back into the half that actually writes).

**A field that had no reader.** `Run.environmentId` was written by every send and rendered nowhere, so a run measured against Staging looked identical to one against Production. `run-source` is its first reader. A deleted environment is named as deleted rather than folded into "No environment" (the run kept the id either way), and a deleted *request* drops the link via `isRequestNotFound` - never for a transport failure, which is not a deletion.

**Deliberate deviations from the issue, both stated rather than quietly dropped:**

1. **No "last smoke outcome" section on the collection tab.** There is no such thing in the app to read: a collection has no runs of its own until the collection runner (#354) lands, and a section claiming one could only invent the number. Pinned by a registry test so it is a decision, not an omission.
2. **`RunSummary.followRedirects` / `maxRedirects` stay unrendered.** The issue offered the run section as a plausible home. I left them alone: they are list-row fields on a type that mirrors the wire, the run tab reads `configSnapshot` instead, and "mode, target, caps" is what the section was asked for. The existing note on the type still describes the situation accurately.
3. **The `run-config` section restates part of the load-test pane's config strip.** Accepted deliberately: a *design* run has no config view anywhere, and a bar that changed shape between the two kinds of run tab would be the worse answer. Both read the same helpers, so they cannot word one run differently.

**Dock:** no edit. The toggle lights up on collection and run tabs off `contextBarHasContent` alone; the two cases that moved out of the "not pressed" `it.each` are the mutation-check for that derivation, and two new cases assert it stays dark on an entity-less tab.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`cd app && pnpm test && pnpm type-check && pnpm lint` - all green.

- **Full app suite: 331 files, 3054 tests, 0 failures** (127s).
- New: `CollectionVariablesSection.test.tsx` (9), `CollectionAuthSection.test.tsx` (6), `CollectionContentsSection.test.tsx` (5), `RunConfigSection.test.tsx` (8), `RunSourceSection.test.tsx` (7).
- Updated: `registry.test.tsx` (ordering per tab type, the entity gate, the absent collection-run section), `Dock.context-toggle.test.tsx` (collection/run moved from "not pressed" to "pressed", plus entity-less cases), `VariablesSection.commit-scope.test.tsx` (the source-scan guard now reads both files, each asserted non-empty).

Mutation-checked by actually reverting, not by inspection:

| Mutation | Reddens |
|---|---|
| `contextBarHasContent` back to a hardcoded `tab.type === "request"` | 4 tests (both Dock pressed cases, both registry derivation cases) |
| The collection row's `sourceId` pointed at another collection | the commit-target test |
| `config.mode` printed raw / a hand-written concurrency unit | the two naming tests |

No engine change, so no engine build or ctest run - nothing there could change colour. `pnpm format:check` is clean for every file touched; `docs/app/COMPONENTS.md` was already not prettier-clean before this change and was left that way per CLAUDE.md. `mkdocs build --strict` was not run (mkdocs is not installed in this environment); the docs edits add table rows and prose to existing pages, with no new headings, anchors or relative links.

## Checklist

Docs in the same commit: `docs/app/COMPONENTS.md` (per-tab section tables, the entity gate, the shared row/commit note, the "no last run" decision) and `docs/app/architecture.md` (the bar is no longer request-tab-only). The registry's own "Phase 1 is the request tab" header comment is updated where it lives, in `registry.ts`.

## Related Issues
Closes #377

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dp51E9xWg1CHzeMKCUviKF)_

### DIFF
--- a/app/src/components/layout/Dock.context-toggle.test.tsx
+++ b/app/src/components/layout/Dock.context-toggle.test.tsx
@@ -11,13 +11,13 @@
 /**
  * The context-bar toggle reports what is on screen, not what is stored.
  *
- * The Dock button and Ctrl/Cmd+I flip `contextBarOpen` on every tab type, while
- * the bar renders null off a request tab - so on six of the seven types the
- * button lit up and nothing appeared, and because the flag is persisted the bar
- * then popped out later on the next request tab.
+ * The Dock button and Ctrl/Cmd+I flipped `contextBarOpen` on every tab type
+ * while the bar rendered null wherever it had no sections - so the button lit
+ * up and nothing appeared, and because the flag is persisted the bar then
+ * popped out later on the next request tab.
  *
  * Mutation-check: drive `active` from `contextBarOpen` alone again and the
- * non-request cases redden.
+ * section-less cases redden.
  */
 
 import { describe, it, expect, beforeEach, vi } from "vitest";
@@ -37,9 +37,11 @@ function renderDock() {
 	);
 }
 
-function openTabOfType(type: TabType) {
+const ENTITY_TYPES: TabType[] = ["request", "collection", "run"];
+
+function openTabOfType(type: TabType, entityId = ENTITY_TYPES.includes(type) ? `${type}_1` : null) {
 	useTabsStore.setState({
-		openTabs: [{ id: "t1", type, entityId: type === "request" ? "req_1" : null }],
+		openTabs: [{ id: "t1", type, entityId }],
 		activeTabId: "t1",
 	});
 }
@@ -66,7 +68,29 @@ describe("Dock - the context-bar toggle's pressed state", () => {
 		expect(toggle()).toHaveAttribute("aria-pressed", "false");
 	});
 
-	it.each<TabType>(["welcome", "collection", "dashboard", "run", "variables", "settings"])(
+	// `collection` and `run` moved out of this list when they grew sections, and
+	// the Dock was not touched to make that happen - which is the derivation
+	// working. Put a hardcoded `type === "request"` back in
+	// `contextBarHasContent` and the two cases below redden.
+	it.each<TabType>(["collection", "run"])(
+		"is pressed on a %s tab with the bar open, off the registry alone",
+		(type) => {
+			openTabOfType(type);
+			renderDock();
+			expect(toggle()).toHaveAttribute("aria-pressed", "true");
+		}
+	);
+
+	it.each<TabType>(["collection", "run"])(
+		"is not pressed on a %s tab that is not open on an entity",
+		(type) => {
+			openTabOfType(type, null);
+			renderDock();
+			expect(toggle()).toHaveAttribute("aria-pressed", "false");
+		}
+	);
+
+	it.each<TabType>(["welcome", "dashboard", "variables", "settings"])(
 		"is not pressed on a %s tab even with the flag set",
 		(type) => {
 			openTabOfType(type);
--- a/app/src/components/layout/context-bar/CollectionAuthSection.test.tsx
+++ b/app/src/components/layout/context-bar/CollectionAuthSection.test.tsx
@@ -1,0 +1,130 @@
+/**
+ * @vitest-environment jsdom
+ */
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * What a collection hands down, and to whom.
+ *
+ * The four answers below are genuinely different outcomes, and the one worth
+ * the section is the third: a collection that configures nothing does not mean
+ * "no auth" - a request under it inherits from further up, and naming that
+ * ancestor is the difference between a five-minute 401 and an afternoon one.
+ *
+ * The walk is `resolveAuthSource`, shared with the Auth tab's chain and with
+ * what is actually sent. Mutation-check: answer from `collection.auth` alone
+ * and the inheriting cases redden.
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render, screen } from "@testing-library/react";
+import { CollectionAuthSection } from "./CollectionAuthSection";
+import type { Collection, RequestAuth } from "@/types";
+
+let collections: Collection[] = [];
+
+vi.mock("@/queries", async () => {
+	const { walkAncestors } = await vi.importActual<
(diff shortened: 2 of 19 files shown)


athrvk:
Claiming this - master moved ahead (#391 merged), so I'm merging `master` into this branch and re-running the app suite before pushing.

---

Issue #384, opened 2026-08-08
app(graphql): editor batch - space triggers completion, Format deletes comments, the variables pane loses drafts on tab glance and silently omits non-strict JSON from the wire, {{variables}} complete but diagnose as errors


athrvk:
From the 2026-08-08 GraphQL review sweep (adversarially verified at master `93a2f93`). The Monaco/editor cluster - one PR over `app/src/lib/graphql/language-providers.ts`, `diagnostics.ts`, `variables-schema.ts`, the GraphQL panes in `BodyPanel`/`GraphQLBody`, and `graphql-body.ts`.

## Problem (all confirmed)

1. **`" "` (space) is in the completion `triggerCharacters`**, directly contradicting the provider's own comment ("Structural triggers only - NOT space/newline") - typing a space pops suggestions and Enter then accepts one instead of inserting a newline, the exact bug class the comment exists to prevent (medium).
2. **Format Document deletes every `#` comment** in the query (the formatter round-trips through an AST print that drops them) (medium).
3. **In-progress variables-pane text is destroyed by glancing at another request tab** when it is not-yet-valid JSON - the exact draft-loss bug class `body-drafts.ts` documents fixing for the body string; the variables pane never got the fix (medium).
4. **`{{variable}}` completion is offered inside the query pane while diagnostics mark every inserted token as an error** - the editor argues with itself; the engine resolves the token fine (medium). Reconcile: either suppress the GraphQL parse-error marker over `{{...}}` spans (a pre-diagnostic masking pass substituting a valid placeholder, the approach the JSON body took) or scope the completion provider out of GraphQL - masking is the right call, the engine genuinely supports the idiom.
5. **Non-strict JSON in the variables pane is silently omitted from the wire** - including the `{"limit": {{n}}}` idiom the engine's template resolution handles; the request sends without its variables and nothing says so (low, pairs with 4).
6. **A string-typed `variables` value garbles on round-trip** - the shape the Postman importer deliberately preserves renders as a JSON-escaped quoted blob in the pane (low).
7. **Completion kinds are flattened to `Field`** - arguments, enum values, variables, directives all get the same icon (hygiene).
8. **`language-providers.ts` is the only lib module with no test file** - the trigger set, severity mapping, marker lifecycle, and schema-change resubscription are unguarded; `diagnostics.ts`'s position offsets and warning branch are mutation-survivable; `applyVariablesSchema`'s Monaco half is untested (gaps T4, T5, T6).

## Fix / solution pathway

[...]

Issue #383, opened 2026-08-08
app(graphql): schema-cache lifecycle - failure detail and freshness are written but never read, the cache key hashes unresolved auth, refresh discards the last good schema, and the cache never evicts


athrvk:
From the 2026-08-08 GraphQL review sweep (adversarially verified at master `93a2f93`). The schema-cache cluster - one PR over `app/src/lib/graphql/schema-cache.ts` and the `GraphQLBody`/`BodyPanel` wiring.

## Problem (all confirmed)

1. **`SchemaEntry.error` and `fetchedAt` have zero readers** outside the store and its test (verified by rg; the test asserts fields production never reads - the signature written-never-read shape). Every failure mode - HTTP 401, introspection disabled, malformed response - collapses to one static badge title; the user cannot tell "endpoint has no introspection" from "your token expired" (medium).
2. **The cache key hashes the auth block as typed, not as resolved** - a credential change upstream of resolution (an env variable the auth interpolates, an ancestor collection's auth edit) leaves the key unchanged and serves the stale schema fetched with the old credential (medium).
3. **`refreshSchema` nulls the cached schema entering `loading` and again on `error`** - every manual refresh downgrades a working editor to syntax-only (marker flicker), and a failed refresh loses the last good schema entirely (low).
4. **`byKey` grows unbounded** - every (resolvedUrl, collection, environment, auth) permutation retains a fully parsed `GraphQLSchema` forever; no eviction, cap, or TTL (low).
5. **The active target is never cleared on unmount** (the `!active` branch is unreachable - `BodyPanel` hardcodes `active`), so the last GraphQL tab's schema stays "active" after closing it (hygiene).
6. **Introspection parses synchronously on the renderer main thread with no size guard** - `JSON.parse` of the whole response plus `buildClientSchema` inside the awaited store action; a GitHub-scale schema visibly janks (low, critic find).
7. **`docs/app/architecture.md` still describes the cache as "keyed by endpoint URL"** - the pre-#228 design, contradicting the four-part key three findings above hinge on (hygiene, critic find).

## Fix / solution pathway

1. Read what is written: the schema badge/tooltip surfaces `entry.error`'s message (classified: auth failure vs no-introspection vs network vs parse) and `fetchedAt` as "schema from Xm ago" with the refresh affordance beside it. This is also the freshness affordance the schema-explorer feature needs - build it here, once.
[...]

PR #399, opened 2026-08-08
fix(app/graphql): import and wire correctness - operationName survives an edit, imported requests send as JSON, Insomnia bare queries get an envelope

### LINKED ISSUE #385: app/engine(graphql): import and wire correctness - operationName is destroyed on first keystroke, imported requests send as x-www-form-urlencoded, Insomnia bare queries ship as non-JSON bodies
From the 2026-08-08 GraphQL review sweep (adversarially verified at master `93a2f93`). The import-to-wire cluster for GraphQL bodies - one PR over `graphql-body.ts`, the importers, and a small engine test ride-along.

## Problem (all confirmed)

1. **`operationName` is unrepresentable end to end (medium).** The Postman importer deliberately preserves it in the stored envelope - and the first keystroke in either GraphQL pane silently deletes it: `parseGraphQLBody`/`serializeGraphQLBody` round-trip only `{query, variables}`. A multi-operation document whose named operation was imported executes a **different operation** after any edit, silently. The serializer must carry `operationName` through, and the UI needs the minimal surface: when the document contains multiple operations, an operation picker (a small select above the query pane, GraphiQL-style) that writes `operationName`; single-operation documents need nothing.
2. **An imported GraphQL request goes on the wire with `Content-Type: application/x-www-form-urlencoded`** - libcurl's default - because the auto-add of `application/json` fires only on an interactive mode switch, never on import (medium). Most GraphQL servers 400 that; the user sees a failing request that looks identical to a working one.
3. **An Insomnia import whose `body.text` is a bare query document is stored and sent verbatim as a non-JSON wire body** while the editor's parse fallback makes it look healthy (medium). Normalize at import: wrap bare queries into the `{query}` envelope.
4. **Doc/discoverability riders (hygiene):** `api-reference.md`'s `POST /execute` example keys the body as `type` (which the engine ignores - a copying caller silently sends no body) and its bodyType list omits graphql; MCP tool docstrings never mention `graphql` as a bodyType, so agents cannot discover it (P7/P8).
5. **The engine has zero GraphQL coverage** (the string appears nowhere in `engine/tests`) - the graphql body's store/send round-trip is pinned only by app-side assumptions (T12). One gtest ride-along: a graphql body round-trips storage and hits the wire as the JSON envelope with the right Content-Type.

## Fix / solution pathway

1. `parseGraphQLBody`/`serializeGraphQLBody` preserve `operationName` (and unknown envelope keys generally - forward-compat); the operation picker renders only for multi-operation documents and writes the field. Both functions finally get their test file (gap T2 - the body-destruction bug lived exactly here).
2. Importers set the Content-Type header on graphql bodies at import time (matching the interactive mode-switch behavior; respect an explicitly imported header).
3. Insomnia bare-query normalization into the envelope, with the `application/graphql` branch finally covered (gap T9 - the whole Insomnia-graphql chain is unproven end to end).
4. The two doc fixes + MCP docstring mention.
5. The engine gtest above.

## App changes

`app/src/lib/graphql/graphql-body.ts` (+ new test), `postman.ts`/`insomnia-v4.ts` importers (+ fixtures), the operation-picker component in the GraphQL body UI, `app/electron/mcp/` docstrings. Engine: test only.

## Tests

- `graphql-body.test.ts`: operationName round-trip (mutation-checked: drop the field, redden), string-variables passthrough (shared contract with the editor batch - coordinate), unknown-key preservation.
- Importer fixtures: Postman graphql with operationName + Content-Type asserted; Insomnia `application/graphql` bare query normalized (both currently zero-covered).
- Component: picker appears only for multi-operation documents; selection serializes.
- Engine: the round-trip gtest.
- Commands: `cd app && pnpm test && pnpm type-check && pnpm lint`; `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`.

## Docs (same commit)

`docs/engine/api-reference.md` (the `type`-vs-`bodyType` example fix + graphql in the list), `docs/engine/mcp.md` (bodyType mention), `docs/app/import-collections/` per-format pages for the two importer changes.

## Sequencing

Parallel-safe against #382/#383 and everything else. Shares `graphql-body.ts` with the editor batch (#384) in different functions - land in either order, flag the overlap in whichever lands second. The explorer feature is independent of this issue.

Out of scope: subscriptions transport; form bodies (#381 owns the engine body union work - the gtest here touches only the existing graphql passthrough).

## Acceptance criteria

- [ ] Editing an imported multi-operation request preserves and sends `operationName`; the picker controls it.
- [ ] An imported GraphQL request sends `application/json` without any interactive touch.
- [ ] An Insomnia bare-query import produces the envelope and a working wire body.
- [ ] The engine round-trip gtest exists and passes; docs corrected; all tests green, mutation-checked.

---

athrvk:
## Description

All three defects reproduced on master at `43efd1b`, as #385 describes. Now rebased over `308e97a` (see **Merge** at the end).

**Plan.** The root cause of finding 1 is one thing: `parseGraphQLBody` / `serializeGraphQLBody` treated the GraphQL-over-HTTP envelope as *the two panes*, when the envelope is strictly bigger than them. `operationName` has no pane, so a keystroke rebuilt the body from `{query, variables}` and dropped it - and an imported multi-operation request began executing whichever operation the server chose, with nothing on screen saying so. The approach makes the converter carry the whole envelope (`operationName` because a picker drives it, every unmodelled key verbatim because it is not ours to discard) and gives the component **one** write path taking only what changed, so a caller cannot name one field and silently drop another. The alternative I rejected was threading `operationName` alongside `query` and `variables` as a third positional argument: it fixes the one key that was reported and leaves the shape that loses the next one - `extensions` is already legal on the envelope today.

Findings 2 and 3 are both "the import path never runs the interactive path's rules". Fixed at import, where the shape is known, rather than at display.

Per finding:

1. **The whole envelope round-trips.** `parseGraphQLBody` returns `{query, variables, operationName, extras}`; `serializeGraphQLBody` takes that record back. `extras` holds keys the editor does not model, and a modelled key always wins over a stale copy, so an envelope cannot end up with two `query`s. The picker renders when the document defines **more than one named operation** - by spec a valid multi-operation document names every one, so that is exactly when `operationName` has to be on the wire. An `operationName` the document no longer defines is offered as an option rather than dropped: it is still what will be sent, so the trigger shows the truth and nothing rewrites the body on its own.
2. **Both importers write the Content-Type.** Through `contentTypeToAdd`, the same rule the mode picker uses - not a copy - so an explicitly imported header wins (including a deliberate `application/graphql`) and a disabled row does not count as declaring one. Its parameter widens from `KeyValueItem` to `KeyValueEntry`; the rule only ever read `key` and `enabled`, and importer rows have no `id` yet.
3. **Insomnia bare-query normalization.** `toGraphQLEnvelope` wraps a bare document and passes an envelope through untouched, so a mislabelled JSON envelope is never double-wrapped.
4. **Docs/discoverability.** `POST /execute` and `POST /runs` keyed the example body as `type`; the engine reads `mode`, so a caller copying it sent no body at all.
5. **The engine gtest.** `graphql_body_test.cpp` pins the envelope through storage (byte for byte - a re-serialization that reordered the JSON would still parse and would still change bytes the user never edited) and onto the wire, including that the engine derives **no** Content-Type for a content mode. That last one is why the app-side header work in finding 2 is necessary at all, so it is worth a test rather than a comment.

**Blast radius traced.** `parseGraphQLBody`/`serializeGraphQLBody` had exactly one consumer (`GraphQLBody.tsx`; `body-drafts.ts` only names them in a comment) and now have three. `contentTypeToAdd` had one (`BodyPanel` via `switchContentType`) and now has two. `withRequiredContentType` is applied at the single draft-construction site in each importer, so every imported request goes through it, and `RequestDraft.headers` flows to `POST /import/apply` unchanged.

**Deviations from the issue spec, with reasons:**

- **String-typed `variables` passthrough is deliberately *not* here.** The issue lists it as a test under "coordinate with #384". Implemented alone it is a regression: showing the raw string in the pane makes it non-strict JSON, which `serializeGraphQLBody` drops on the next keystroke - so a lossless-but-garbled round trip becomes a lossy one. It is safe only together with #384's item 5 (non-strict variables reaching the wire via template resolution, with the non-strict badge). Left to #384, which owns that half.
- **`graphql-body.ts` was not at `app/src/lib/graphql/`**, the path the issue names - it was in the body panel. It is *moved* there, because it is now shared by the panel and both importers and `lib/graphql/` is where this codebase already keeps GraphQL infrastructure. **#384 also edits this file** and will get the conflict.
- **The MCP docstrings already mention `graphql`** (the issue predicted they did not - drift since it was written). The real remaining gap was `start_load_run`, whose `bodyType` carried no description at all; it now has the text its two sibling tools have had since they existed. `docs/engine/mcp.md` and the `body` union table already listed graphql, so neither needed the change the issue predicted.
- **`docs/app/COMPONENTS.md` gained an entry** the issue did not ask for: it documents `lib/graphql/` file by file and the content-type rule as panel-private, both of which this diff makes stale.

**Adjacent, deliberately not in this diff:** `graphqlContent` in the Postman importer builds an envelope by hand rather than through `serializeGraphQLBody`. It is correct and its behaviour is pinned, and converting it would widen this diff into #384's file for no user-visible gain.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`; `cd app && pnpm test && pnpm type-check && pnpm lint`. **All figures below are from the post-merge tree** (`718d226`), re-run after resolving.

- **Engine: 1048 tests, 0 failures** (91s). 5 new, all in `graphql_body_test.cpp`.
- **App: 3153 tests across 334 files, all passing.** 45 added by this PR. `pnpm type-check` clean, `pnpm lint` clean at zero warnings, `prettier --check` clean on every file touched.
- No pre-existing failures on the base commit.

New coverage: `graphql-body.test.ts` (25 - the file the sweep found missing, gap T2), `GraphQLBody.operation-picker.test.tsx` (10), and 10 importer cases (Postman +5, Insomnia +5 - the `application/graphql` chain was previously zero-covered end to end, gap T9).

The component tests drive the mocked editors' `onChange` rather than asserting on state, because the bug *was* a keystroke and that path exists only inside the component. The picker is asserted by presence and accessible name rather than by opening it - Radix Select does not commit a value in jsdom, which is the same reason `content-type.ts` keeps its rule outside the click handler.

Mutation-checked by actually reverting each fix and re-running, not by inspection:

| Mutation | Reddens |
|---|---|
| serializer drops `operationName` | 6 |
| parse drops unknown envelope keys | 3 |
| Postman: no Content-Type at import | 2 |
| Insomnia: no Content-Type at import | 1 |
| Insomnia: bare query not normalized | 2 |
| engine: `implied_content_type` returns `application/json` for GraphQL (rebuilt) | 1 |

Restoring each returned every suite to green.

One test was caught by its own mutation check rather than written right: "carries the variables draft into a later query edit" failed on the first run because the keystrokes were driven outside `act`, so the second read a stale render. That was the harness, not the panes - wrapping the interactions makes it test what it claims to.

`mkdocs build --strict` was not run (mkdocs is not installed in this environment); the docs change adds no page, no heading and no relative link - only prose and table cells inside existing sections. clang-format: `git clang-format --diff origin/master -- 'engine/*'` reports no drift.

## Merge

`master` moved 9 commits (#395, #396, #397, #398 all landed) and the branch went un-mergeable. Merged and resolved in `718d226`.

**One conflicted file, and it was the one this PR predicted:** #397 (issue #383) rewrote `SchemaStatusBadge` to take the store's `SchemaEntry` and dropped the `active` prop, in the same header row this branch adds the operation picker to. Both sides kept - the picker sits beside master's badge - and the picker test no longer passes `active`, since the schema lifecycle it used to gate now hangs off mount/unmount, which this branch does not touch. `graphql-body.ts`'s move merged cleanly: nothing on master still imported the old path.

Everything else auto-merged, including `engine/tests/form_body_test.cpp` (the echo server extracted to `echo_server.hpp`) and three docs files. Both suites were rebuilt and re-run against the merged tree, not assumed - the engine count rose to 1048 and the app to 3153 because #395/#396/#397/#398 brought their own tests.

## Checklist

## Related Issues
Closes #385

---
_Generated by [Claude Code](https://claude.ai/code/session_01UEsZWFzS2KN5aQDcH4RwB7)_


athrvk:
fix merge conflict

Issue #353, opened 2026-08-07
engine(scenario): phase 2 - the design-mode sequential runner (steps, iterations, per-step results, SSE step events)


athrvk:
Phase 2 of the collection-runner design (#340). Decisions live in
[`docs/plans/collection-runner-design.md`](https://github.com/athrvk/vayu/blob/master/docs/plans/collection-runner-design.md)
section 3. This issue executes the plan #352 resolves.

## Problem

#352 produces a `ScenarioPlan` and nothing runs it. This is the phase where "log
in, then browse, then check out" actually works.

## Fix / solution pathway

**A third `runs.type`: `"scenario"`** (design §3.1). It cannot reuse `"design"`:
a design run has **exactly one `results` row** and `GET /runs/:runId` serves it as
`result` on that assumption (`docs/engine/db-schema.md`, `results`). A scenario run
has one row per step execution, so overloading `design` breaks a documented reader.

**Lifecycle: identical to a load run** (§3.2). `POST /runs` answers `202 {runId}`,
`RunManager::start_run` spawns the worker, and the run is registered / retained /
swept / stopped / reported by the machinery that already exists. A 50-request
folder takes as long as its requests added together, so a synchronous endpoint
would hold a cpp-httplib worker thread for minutes.

**The executor** (§3.3) - per iteration, per step, this is the `POST /execute`
handler's body lifted out of `engine/src/http/routes/execution.cpp`:

```
pre-request script  (ScriptContext::for_prerequest - pm.request writes back)
  -> Client::send   (through the environment cookie jar)
  -> post-request script (ScriptContext::for_test)
  -> record step result
```

It must be `http::Client`, **not** `EventLoop`: design-mode fidelity needs the
cookie jar and inline per-step scripts, and the event loop deliberately has
neither (`engine/include/vayu/http/cookie_jar.hpp`, "Not on the load path").
Extracting the shared body into a testable core that both `POST /execute` and the
runner call is preferred over a copy - a hand-rolled copy of a primitive never
receives the primitive's fixes.

**State between steps** (§3.4):

- Variable scopes loaded **once** at run start (`load_script_variable_scopes`),
  mutated in memory across every step and iteration, persisted **once at run end**
  through the existing diff-based `persist_script_variables`. Per-step persistence
  would be N x M diff-and-write cycles against the DB mutex for a value only the
  run's own later steps read.
- Cookies: the existing per-environment jar, scope = `environmentId`, unchanged.
[...]

PR #400, opened 2026-08-08
feat(app/scenario): the collection runner UI - run a folder, live per-step progress, per-step results

### LINKED ISSUE #354: app(scenario): phase 3 - the collection runner UI: run a folder, live per-step progress, per-step results

athrvk:
## Description

Phase 3 of the collection-runner track. The engine has executed ordered sequences since #353 landed (`67b35b8`) - it streams a `step` event per step and stores a `results` row per step execution - and nothing in the app started one or showed one. Verified on master at `308e97a`: `HistoryDetail` routed `scenario` runs to `LoadTestDetail` by falling through the `isDesignRun` branch, and there was no way to start a run at all.

**Plan.** The root cause is that a scenario run's `results[]` are step executions of *different* requests, so every load-report reader is wrong about it - percentiles over one-request-repeated, a status distribution that mixes six endpoints. The approach gives the run type its own view over a pure step-list module (`scenario-steps.ts`) that both sources - the live `step` stream and the stored rows - collapse into, so the tab renders one list rather than two code paths that drift. The alternative I rejected was widening `LoadTestDetail` with a scenario branch: it would put a sequence's step list inside a component whose whole shape (`reportToDerived` → `DashboardDerived`, the mode-adaptive hero strip) is built for one repeated request, and the two have no numbers in common.

Per finding:

1. **Entry point.** `RunCollectionDialog`, opened from a collection row's ⋯ menu where every other collection action lives. Two options - Recursive and Iterations - and no scenario-authoring UI, per design §2.1: the sequence *is* the folder's ordering, and a step list authored here would be a second source of truth for it. Invalid iterations are refused in the dialog rather than sent; the engine's own `400` (which names the step that would not compose) renders in place, because that message is what the user has to act on.
2. **Live progress rides the existing SSE client**, which gains `step` beside `metrics` - one stream, one client, one `complete`. The `step` listener is registered only when a caller passes `onStep`, so a load run does not carry dead wiring. `parseStepEvent` **drops** a payload it cannot read rather than defaulting it: the list keys on `(iteration, stepIndex)`, so a defaulted `0:0` would collide with the real first step's row rather than merely say nothing.
3. **Out-of-order and replayed events do not duplicate rows.** Keying on `(iteration, stepIndex)` rather than arrival order is what makes a `Last-Event-ID` replay idempotent - an append-only list doubles every row a reconnect re-saw. `appendStepEvent` returns the same array reference when a replay changes nothing, so an idempotent replay does not re-render.
4. **Per-step results restore through `restore-response.ts`**, the path a design run's response pane uses, so the step list inherits its fixes (`bodyTruncated`/`bodyBytes`, the `httpVersion` fallback) instead of re-reading `trace_data`.
5. **`skipped` is never `passed`** (design §4.4). Four counts, four row treatments. Nothing produces `skipped` until #355, so it is built to render correctly before it can occur, and a test asserts it rather than a future reader.
6. **Thinned results say so.** `stepsExecuted` vs `stepsStored` with the reason - failures are kept first, so what is missing is always successes.

**Blast radius traced.** `sseClient.connect` had one caller (`load-test-service.ts`) and now has two; the new parameter is optional and appended, so the load path is byte-identical. `SampledExchange` had two callers (the dashboard's live sample list, the history Samples tab) and now has three; both new props are optional and the derived state is unchanged when they are absent, which the existing suites confirm. `HistoryDetail`'s report gate moved from `!isDesignRun` to `isLoadRun` - the load path is unchanged and the design path was already exempt. `RunResultTrace` gained five optional fields, all of which the engine already writes (`stamp_step_identity`); `RunReport.scenario` likewise (`runs.cpp:689`). Every new type has a reader in this diff.

**Deliberate choices worth calling out:**

- **`SampledExchange` was extended rather than copied.** A parallel step row would never receive the primitive's fixes, which is this codebase's named defect. What it lacked is real and not cosmetic: its state is derived from the status code, so a `failed` assertion over a `200` renders as a success and a `skipped` step as a connection failure. It gained an optional `state` (`success | error | slow | skipped`) and a `title` slot; absent, every existing caller resolves exactly as before.
- **There is no stop control, and therefore no `stopMonitoring` / `clear`.** The stream ends on its own when the engine sends `complete`, and the issue does not ask for a stop. A detach method with no caller is surface that cannot be verified, so both belong with the control that would need them.
- **The dialog is mounted only once a folder is chosen** and unmounted on close. That mount *is* the options reset - resetting in an effect is a cascading render for something React already models as identity, and ESLint's `react-hooks/set-state-in-effect` says so.
- **A running scenario run's tab is not gated on its report.** Waiting for one would hold the tab on a skeleton for the length of the run. It also reads the run's own status, not just the stream, so a tab reopened onto a still-executing run says "waiting" rather than "no steps recorded" - a gap the tests caught rather than one I anticipated.

**Adjacent, deliberately not in this diff:** `run_collection_smoke` stays as it is (design §6 - a smoke matrix over independent executes is a different tool). No engine change; #353 shipped everything this reads.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check` - **336 files, 3160 tests, all passing** (178s). 52 of them are new; the baseline was 3108. `type-check`, `lint` (zero warnings) and `format:check` are all clean.

**No engine build or test run, and that is the intended scope.** The diff touches no file under `engine/` - #353 shipped the runner, the `step` event and the per-step storage this reads - so no engine test could go from green to red. Per `CLAUDE.md`'s scale-the-verification rule, running the ~110s suite here would be latency, not coverage. There are no pre-existing app failures on the base commit.

New coverage:

| File | Tests | Covers |
|---|---|---|
| `scenario-steps.test.ts` | 17 | replay/out-of-order merge, outcome counts, stored-row mapping, thinning |
| `ScenarioRunView.test.tsx` | 14 | live advance, the four outcomes, restored responses, iterations, disclosure, empty vs loading |
| `RunCollectionDialog.test.tsx` | 8 | payload shape, invalid iterations, tab + stream after `202`, in-place rejection |
| `CollectionTree.run.test.tsx` | 3 | the menu action runs the row's own collection |
| `sse-client.test.ts` | +7 | `parseStepEvent` accept/reject |
| `HistoryDetail.states.test.tsx` | +3 | scenario routing, the ungated running case |

Mutation-checked by actually reverting each fix and re-running, not by inspection:

| Mutation | Reddens |
|---|---|
| append-only list (drop the replay dedup) | 3 |
| fold `skipped` into `passed` | 3 |
| an unstamped stored row counts as `passed` | 1 |
| render a load-run `results` row as step 0 | 1 |
| drop the thinning disclosure's `stepsDropped` guard | 2 |
| derive the row state from the status code (ignore `state`) | 1 |
| hand-roll the response instead of `responseFromRunResult` | 2 |
| accept any `step` payload (defaults instead of rejection) | 3 |

Restoring each returned every suite to green.

The component tests drive the store and the report rather than asserting on internals, because both sources are what the view exists to reconcile. `ScenarioStepCard` mocks only `ResponseBody` (Monaco), which is irrelevant to every question asked and costs a jsdom mount per expanded row.

`mkdocs build --strict -f .github/mkdocs.yml` passes (mkdocs installed for the check). The docs changes add four headings and no relative links.

## Checklist

## Related Issues
Closes #354

---
_Generated by [Claude Code](https://claude.ai/code/session_01THHG5v4mbhmjTfJ2QWYibn)_


athrvk:
- the context bar needs to be updated with this collection run feature's context 
- each request's accordian is empty
- the history view's UI should show a collection run correctly

athrvk:
Pushed `9d6b9e1` - all three, plus the root cause behind the second one, which was a real bug rather than a missing view.

**Each request's accordion is empty.** It would never have filled. The step list prefers stored rows because only a stored row carries an exchange, and the refetch that produces them never fired: the run tab mounts when the run starts and fetches the report immediately, at which point the engine has written no step rows yet (they go to SQLite in one batch at the end), so the cache holds a report with an empty `results[]` seconds old. `ScenarioRunService.handleClose` then asked for it through `fetchQuery` under the hook's five-minute `RUNS_STALE_TIME_MS`, which found that entry fresh and resolved from it without touching the network. So the view stayed on the live rows for the life of the tab - and a live row has no exchange to show. Fixed with `staleTime: 0` on that one fetch, with a test that reddens if it goes back (verified by reverting). A live row also now expands onto the reason it has nothing rather than onto an empty panel.

I reproduced this against a real engine before changing anything: a two-request collection, two iterations, and `GET /runs/:id/report` returns four fully-traced rows - the data was always there, the app just never asked for it a second time.

**The history view.** A collection run's row was a status and a timestamp: no url, no method, the load badge is load-only, and it announced itself to screen readers as a "request run". It has nothing to show because the engine's list-row `summary` is built from the snapshot's flat keys, which a scenario has none of - so `build_run_summary` now sends a `scenario` descriptor (`collectionId`, `iterations`, `recursive`, `stepCount`) on scenario snapshots only. `stepCount`, not the manifest: a row carrying every step's name, method and URL would undo the reason the compact summary exists. The row shows the collection's name (resolved in `HistoryList` from the loaded tree - one shared query for the page, id as the fallback for a folder deleted since), the plan's size, and iterations when there was more than one pass. The type filter gains the third run type; it compared its value to `run.type` directly, so a collection run could only ever appear under "All".

**The context bar.** It was describing a collection run as a load run that recorded nothing - `RunConfigSection` found no `mode`, `duration` or root `iterations` and printed "No configuration was recorded" of a run that recorded a whole plan, and `RunSourceSection` said "Request: Not saved", which is true of the field and wrong about the run. Both now read the snapshot's resolved manifest through one narrowing helper (`run-scenario.ts`, gated on `run.type` so a load run whose raw body happened to carry a `scenario` key is never described as one): mode, steps, iterations, sub-folders, and the collection linked the way the request row links its request. I left the collection tab without a "last run" section and replaced the registry's now-stale note about this issue with why: a collection's runs are not addressable - `GET /runs` filters by `requestId` and a scenario row links none - so finding them means a substring scan of every stored snapshot, per open bar, for a number History already shows. It earns the slot once the runs list can be filtered by collection.

Verified: 1050 engine tests, 3228 app tests, `pnpm type-check` and `pnpm lint` all green. Docs updated in the same commit - `api-reference.md` for the tenth summary key, `COMPONENTS.md` and `api-integration.md` for the two shapes a scenario run comes back in.

---

athrvk:
All three addressed, in `9d6b9e1` (+ a docs follow-up in `29c0a52`).

I went and drove a real engine against a mock server rather than reading the code, which is what found the actual cause of the empty accordions - it was not where I would have guessed.

**The empty accordions were a cache bug, not a data one.** The stored trace is complete: a passing step's `results` row carries `response.headers` and `response.body`, verified on a live run. What failed was the refetch. The run tab mounts the moment the run *starts* and asks for the report immediately - and a scenario run writes its step rows to SQLite in **one batch at the very end** - so the cache held a report with `results: []`, seconds old. `handleClose` then asked for it through `fetchQuery` under the hook's five-minute `RUNS_STALE_TIME_MS`, which found that entry fresh and resolved from it **without a request**. The list therefore stayed on the live rows for the life of the tab, and a live row has no exchange to expand into. Now `staleTime: 0` on that one fetch - it is the only fetch in the app that knows the data just changed. Pinned by a test that reddens if the `staleTime` comes back.

**The context bar** described a collection run as a load run with nothing recorded: its snapshot has no `mode`, no `duration` and no root `iterations`, so `RunConfigSection` rendered zero rows and said "No configuration was recorded" of a run that recorded a whole plan, while `RunSourceSection` reported "Request: Not saved" - true of the field, wrong about the run. Both now read the snapshot's step manifest through one narrowing helper, gated on `run.type` so a load run whose raw body happened to carry a `scenario` key is never described as one.

**The history row** was a status and a timestamp. A collection run has no url and no method, so every branch in `RunItem` left it blank. The engine's list-row `summary` gains a `scenario` descriptor (`collectionId`, `iterations`, `recursive`, `stepCount`) on scenario snapshots only - `stepCount` and not the manifest, since a row carrying every step's name, method and URL would undo the reason the compact summary exists. The row shows the collection's name, resolved once in `HistoryList`, and the type filter gains the third run type.

Verified: **1051 engine tests, 3228 app tests**, type-check, lint, `format:check` and `mkdocs --strict` all green. The engine change is covered by three new gtests over `get_runs_response` asserting the row JSON exactly, including that a load run does not grow the key.

One note on this branch's history: two sessions landed a fix for these three findings at nearly the same moment. I kept `9d6b9e1` - it is the more complete of the two (it also gates on `run.type` and extends the History type filter) - and discarded my duplicate rather than merging both, which would have left two implementations of the same three fixes in the diff.

---

Issue #386, opened 2026-08-08
engine(reorder): the batch endpoint's atomicity is batch-internal only - no lock spans validate+stage+write, and three in-repo claims overstate the guarantee


athrvk:
Follow-up from the #365 closure verification (2026-08-08 at master `93a2f93`). #365's checkbox criteria all verify and it stays closed; this issue owns the gap the verification found beneath them.

## Problem

`reorder_response` validates and stages via separate per-call DB-mutex acquisitions (`engine/src/http/routes/reorder.cpp:228`, `:326`, `:337`) and only `apply_reorder`'s write runs inside one transaction (`database.cpp:743-762`). cpp-httplib serves requests on a thread pool, and MCP + the renderer are genuinely concurrent clients, so between one request's validation/staging and its commit another request can move the ground:

1. **The cross-request reparent-cycle race survives**: two concurrent batches (A-under-B, B-under-A) each validate against pre-write state and both commit - the exact TOCTOU #365's pathway claimed to close; it is closed only *within* one batch.
2. **A concurrent create can land inside a batch's renumbered range**: the create's `max+1` computes between staging and commit. The covering test (`reorder_route_test.cpp:474-490`) is strictly sequential and does not stage this.
3. **`apply_reorder` uses `storage.replace`, so a row deleted by a concurrent cascade between staging and commit is resurrected** - the old PUT path's exposure, now behind an "atomic" contract.
4. **Three in-repo claims overstate the guarantee** and must be corrected or made true: `reorder.cpp:370-372` ("share a lock scope"), `docs/engine/api-reference.md:653-654` ("no concurrent create can land inside the range"), `reorder_route_test.cpp:25-27`.

Severity honest: single local engine, sub-second windows, requires two concurrent writers - rare; but silent and structural, and the contract *says* it cannot happen.

## Fix / solution pathway

Hold the DB mutex across the whole of validate+stage+write: give `Database` a `with_lock(callable)` seam (the recursive mutex already exists per-call; this scopes it once around the composite) and run `reorder_response`'s validation, staging, and `apply_reorder` inside it - reads and the transaction under one acquisition. Inside that scope, replace `storage.replace` with existence-checked updates (a staged row missing at commit fails the whole batch with a 409 naming the row - no resurrection). Then the three claims become true as written; update the doc line to state the now-real guarantee.

## App changes

[...]

Issue #365, opened 2026-08-07
engine/app(reorder): phase 1 - an atomic batch reorder/move endpoint, first-drop normalization, and the renderer mutation with optimistic caches


athrvk:
Phase 1 of the drag-and-reorder track (parent #364 carries the decided design; read it first). Base facts verified at `d5899bb`/`0f7d975`.

## Problem

A drop that displaces N siblings has no sane write path: the API is one row per PUT (`collections.cpp`/`requests.cpp` merge-patch), so a renumber is N sequential PUTs - non-atomic (crash mid-sequence half-renumbers a parent), racy against concurrent creates computing `max_order+1` between two PUTs and against `POST /import/apply`'s own transactional order slots, TOCTOU-exposed for concurrent reparent cycles (validate and write are separate lock scopes, `collections.cpp:202-213`), and an invalidation storm renderer-side. The engine's transactional precedents are `import_apply` and cascade delete (`database.cpp:688-720`, `642-648`) - the exact pattern to copy.

## Fix / solution pathway

1. **`POST /reorder`** (one endpoint, both entity kinds): body `{ moves: [{ type: "collection" | "request", id, order, parentId? | collectionId? }], normalize?: [{ type, parentId | collectionId }] }`. Semantics: validate everything first - per moved collection the cycle check against the post-move shape (this closes the TOCTOU race: validation and write share one transaction), per moved request the target-collection existence check (#360's rule) - then apply all writes in one `storage.transaction` under the DB mutex. Any validation failure is a 400 naming the offender; nothing is written.
2. **Normalization**: the `normalize` list names parents whose children should first be renumbered dense 0..n-1 **in the pinned display order** (#360's `order, created_at, id`) before `moves` apply - the first-drop-into-a-legacy-collection case from the parent's decision 5. Idempotent; callable alone.
3. **Renderer**: `useReorderMutation` in `queries/collections.ts` - `onMutate` optimistically rewrites the affected `listByCollection`/collections caches via `setQueryData` (never mutating the maps - `combine`'s referential stability is load-bearing, see the comment at `collections.ts:102-109`), rollback on error, single settle-time targeted invalidation, failures through `failSave`. Cross-parent request moves update source and target list caches plus `requests.detail(id)` (staleTime Infinity) for `collectionId`.
[...]

Issue #364, opened 2026-08-07
app/engine(collection-tree): drag-and-reorder for collections and requests - tracking issue (atomic batch reorder, pointer-event DnD, keyboard move, undo)


athrvk:
**Owner direction, 2026-08-07**: add drag-and-drop reordering for both collections and requests - reorder within a parent and move across parents - "as user friendly as it could be". This tracking issue carries the decided design, grounded in the 2026-08-07 tree review sweep (4 readers + verification + critic, at `d5899bb`/`0f7d975`); three phase sub-issues hang off it. Every design decision below is stated as decided - implementers inherit the why.

## Ground truth the design stands on (verified)

- The `order` column exists end to end but is **not yet real**: request creates seed 0, ties are unstable, moves unvalidated - #360 fixes that foundation and is a hard prerequisite.
- **This is the app's first drag interaction.** Nothing drag-reorders anywhere today (verified: the TabStrip's "drag" is window-drag styling; the only pointer-capture prior art is `PanelResizeHandle.tsx:61-75`, the only drop target is ImportModal's file drop). The plumbing - gesture recognition, indicators, auto-scroll - is built fresh.
- There is **no batch write**: one drop displacing N siblings would be N non-atomic PUTs, racy against creates and import, with invalidation storms. The engine's transactional precedents are `import_apply` and cascade delete.
- Subfolders and requests are **two separate ordered blocks** rendered folders-first (`CollectionItem.tsx:329-396`) - the order field cannot interleave a request between folders; drop semantics must respect the blocks.
- The roving-focus hook derives keyboard order from DOM order of `[role="treeitem"]` and walks a specific wrapper structure (`useRovingTreeFocus.ts:34-58`) - drop indicators must not insert treeitem nodes or break the nesting.
- The drawer-row hit-area contract is behaviorally tested (`drawer-row-hit-area.test.tsx`, the 41%-clickable fix) and **a completed drop's trailing click would open the row** - click-after-drag suppression is mandatory, and `RequestItem.test.tsx` pins that opening is synchronous, so discrimination must be movement-threshold, never a timer.

## Decided design

[...]

Issue #237, opened 2026-07-31
app: per-editor save wiring - VariableTableEditor clobbers concurrent edits and marks them clean, ContextBar fails silently, category switches discard drafts


athrvk:
Part of #235 (state-layer sweep). All anchors verified at master `b5db20d`.

## Problem

Each editing surface wires its own save path, and three of them lose or hide data at the edges.

1. **VariableTableEditor: in-flight resync clobbers concurrent edits AND false-cleans them (silent loss).** The row-init effect rebuilds rows whenever `dataVariables` identity changes (`VariableTableEditor.tsx:396-421`). Sequence: blur row A → save snapshots `variablesRef` → user types in row B → save lands → `onSuccess` writes the query cache (`globals.ts:38-41` / `environments.ts:69-81` / `collections.ts:255-262`) → new `dataVariables` → rows rebuilt from pre-B server state, **reverting B on screen** - and the same `onSuccess` calls `setHasPendingChanges(false)` (`VariableTableEditor.tsx:270,291,311`), so B's edit is also marked clean and never saves. The mechanism is acknowledged in-code at `VariableTableEditor.tsx:435-439` ("the backend re-sync would then revert the change"). Window = one mutation round trip; wider on a slow engine. Also: `saveTimeoutRef` is cleared in three places but **never assigned** (verified: zero `saveTimeoutRef.current =` matches) - dead scaffolding implying a debounce that does not exist.
2. **ContextBar variable edits fail with zero feedback.** `ContextBar.tsx:69-107` fires three mutations with no `onError`, no `isError` reader, no toast (verified: zero matches for `onError|isError|failSave|showToast` in the file), and there is no global `MutationCache.onError` (`lib/query-client.ts:17-32`). An engine 400 or connection refusal silently reverts the edit - the exact defect class that `shared.tsx:18-22` documents as the pre-fix state of the collection tabs.
3. **Settings: switching category silently wipes dirty edits** - `SettingsMain.tsx:140-142` resets `editedValues` keyed on `selectedCategory`, no save, no prompt; unmount likewise unregisters without flushing (`SettingsMain.tsx:222-224`). Additionally `SettingsMain.tsx:146-152` forces global save status to `idle` on mount, which can wipe another context's visible `error` state.
[...]

PR #403, opened 2026-08-08
feat(app/collection-tree): drag-and-reorder - pointer gesture, drop indicators, keyboard move, undo

### LINKED ISSUE #367: app(collection-tree): phase 3 - the drag interaction: pointer-event gesture, drop indicators, spring-loading, auto-scroll, keyboard move, undo
Phase 3 of the drag-and-reorder track - the user-facing feature. Parent #364 carries every design decision (pointer events, whole-row + threshold, 25/50/25 zones, two-block rule, dense renumber, undo scope); this issue is the implementation contract. Requires #365 (the batch endpoint), #366 (the decomposed tree), and #362 (aria hierarchy + live region). All three are closed as of master `d2fe130` - this issue is ready to start.

## Problem

No way to reorder or move anything in the tree by direct manipulation; after #365/#366 the data path and the component seams exist, and this issue builds the interaction.

## Fix / solution pathway

1. **Pure modules first** (node-env testable, the `tab-fit.ts` blueprint):
   - `drag-gesture.ts`: a state machine fed pointer coordinates - `idle -> armed(threshold ~4px) -> dragging -> dropped|cancelled(Escape)`; suppresses the trailing click after a beyond-threshold drag.
   - `drop-position.ts`: `(offsetYWithinRow, rowHeight=32, {isFolder, isExpanded, depth}) -> before|after|into`, enforcing the folders-first two-block rule (a request "between folders" resolves to head-of-requests-block).
   - Reuses #365's `reorder-math.ts` and #361's `isDescendant`.
2. **`useTreeDnd` hook** filling #366's context slot (`app/src/modules/collections/context/CollectionTreeContext.tsx:42` defines `interface CollectionTreeDnd`, `:106` carries `dnd: CollectionTreeDnd | null`, and `CollectionTree.tsx:108` passes the `dnd: null` placeholder): pointer capture on rows (excluding `[data-tree-menu]`/chevron), hit-testing via the existing `data-collection-id`/`data-request-id` anchors, indicator state (row id + position), spring-loaded expand (store `expandCollection` after ~700ms hover, fake-timer tested), auto-scroll of the DrawerPanel body in 24px bands, Escape cancels, drop dispatches `useReorderMutation` with the pure-math payload. Dragged-subtree targets disabled via `isDescendant`; `isDeleting`/`isRenaming` rows excluded both directions (row state is context-derived via `deletingRequestId`/`renamingRequestId`, not props).
3. **Indicators**: between-line as a 2px `--primary` absolutely-positioned bar at target depth; into-highlight reusing the selected-row ring classes. Classes on existing nodes only - no new `role="treeitem"` DOM (the roving-focus invariant).
4. **Keyboard move**: Alt+ArrowUp/Down/Left/Right per parent decision 8, via hidden `data-tree-move-*` controls; each move announced in #362's live region with position and target ("Moved Get Users to position 2 of 5 in Billing"). "Move to..." action added to the row menu as the discoverable fallback (a small picker dialog listing collections, `isDescendant`-filtered).
5. **Undo**: after a cross-parent move, a toast with Undo inverting via the captured prior placement through the same mutation. After-move reveal: focus and reveal follow the moved entity (so a move into a collapsed folder re-reveals - the critic's gap; note that `revealedSelectionRef` is now private inside `useRevealActiveSelection.ts:41` with no exposed API, so phase 3 must add a re-reveal method to that hook's return surface and call it after a move).
6. Auto-save coordination: a row whose open tab has an in-flight save is not a drag source (check the save-store pending state) - two writers on one row is the #237 clobber family; note it in the PR.

## App changes

`useTreeDnd.ts`, the two pure modules + tests, row wiring in `CollectionItem`/`RequestItem`, menu action, toast. No engine change (all engine work landed in #365). MCP: none.

## Tests

- Node: `drag-gesture.test.ts` (threshold, cancel, click-suppression transitions), `drop-position.test.ts` (zone math incl. block-boundary cases, collapsed-folder depth ambiguity).
- jsdom `CollectionTree.dnd.test.tsx`: a beyond-threshold drag never opens the row AND `drawer-row-hit-area.test.tsx` stays green (both assertions required - the trailing-click trap); drop dispatches the mutation with the pure-math payload; illegal targets carry the disabled state; spring-expand honours the reveal-ref discipline (fake timers); keyboard move announces (live-region text asserted) and refocuses the moved row; undo round-trips.
- Polyfill `setPointerCapture`/`releasePointerCapture` in test setup (jsdom lacks them - verified).
- Indicator classes are tokens (palette/radius test compliance).

## Docs (same commit)

`docs/app/COMPONENTS.md` (the hook + modules); a short user-facing note wherever the app documents the sidebar (verify whether `docs/` has one; state the result). `docs/app/state-management.md` if the save-store pending check adds surface.

## Sequencing

All prerequisites closed as of master `d2fe130`; ready to start. Still parallel-safe against everything else open. `pnpm lint`/`format:check` CI-enforced.

Out of scope (parent decisions): multi-select drag, touch, collapse-restoration, tab-strip DnD.

## Acceptance criteria

- [ ] Mouse: drag reorders within a parent and moves across parents for both entity types, with correct indicators, spring-loading, auto-scroll, and Escape cancel; a fumbled drag never opens a tab.
- [ ] Keyboard: Alt-arrow moves with live-region announcements and preserved focus; "Move to..." works without chords.
- [ ] Cross-parent moves offer Undo and it round-trips.
- [ ] All new pure modules are node-env tested; all listed jsdom tests green, mutation-checked; the full app suite green.

---

athrvk:
## Description

Phase 3 of the drag-and-reorder track (#364) - the interaction itself. Verified against master at `bf9f5d0`: the three prerequisites are closed and their seams are real, so this fills them rather than re-deriving anything. `CollectionTreeContext.tsx:106` carried `dnd: CollectionTreeDnd | null` and `CollectionTree.tsx:108` passed the `null` placeholder; `reorder-math.ts` and `useReorderMutation` were built and unused by any gesture; the live region shipped empty with no writer.

**Plan.** The root cause of "this is the app's first drag interaction" is that a drag is three unrelated problems wearing one name: when a press stops being a click, where a drop lands, and which rows a move rewrites. The third already had a pure module (#365), so this adds the other two as pure modules - `drag-gesture.ts` and `drop-position.ts` - and leaves `useTreeDnd` holding only what genuinely needs a browser: pointer capture, hit testing, spring-loading, auto-scroll and focus. The alternative I rejected was HTML5 drag-and-drop: it cannot auto-scroll without a drag-image workaround in Electron, it does not model an Escape cancel, and none of its decision points are reachable in jsdom - the whole reason the state machine is extracted is that jsdom has no layout, so a gesture there measures every row at 0px.

**One move, one path.** A drop, an Alt+Arrow, a "Move to..." and an Undo all end in `applyPlacement` → `writeMove`, so the batch, the announcement, the reveal and the undo offer are the same steps in the same order whichever gesture started them. Four copies of that sequence is how the keyboard path ends up not announcing, or the dialog path not revealing.

Per decision in the parent issue:

1. **Pointer events, whole-row handle, ~4px threshold.** No grip icon: the row already fought dead zones to become clickable everywhere (`drawer-row-hit-area.test.tsx`, the 41%-clickable fix), and a grip hands most of that back. `RequestItem.test.tsx` pins that opening is synchronous, so the discriminator is movement and never a timer. The completed drag swallows exactly one trailing click - the one that would open the row just dropped - and a fresh press clears a suppression whose click never arrived.
2. **Drop semantics.** 25/50/25 on a folder, 50/50 on a request. The two-block rule is enforced in the resolver rather than at the call site: a request "between two folders" lands at the head of that parent's requests, a folder over a request row is refused outright (a redirect there would put the folder somewhere the indicator never pointed), and a root folder's edges are refused because the root has no requests block.
3. **The indicator is drawn where the row will land, not where the pointer is.** It is derived from the resolved destination, so the between-folders case draws its line on the first request rather than on the folder the pointer is over. 2px `bg-primary` span inside the target row, indented to that row's depth - the depth is the only thing separating "after this collapsed folder", "into it" and "after its parent". No new `role="treeitem"` nodes: the roving-focus walk and the group nesting are read off the DOM.
4. **Spring-loading and auto-scroll.** 700ms hover expands a collapsed folder through the store's `expandCollection` (the reveal effect skips what is already open, so the two cannot fight). Auto-scroll moves the drawer body, which now names itself `data-drawer-body` rather than being found by class.
5. **Illegal targets are greyed live**, asked of the same resolver the drop uses - so what is refused and what is dimmed cannot disagree. The dragged folder's own subtree comes from the shared `isDescendant`; the engine's cycle guard stays the backstop.
6. **Keyboard parity.** Alt+Up/Down among siblings, Alt+Right into the folder rendered above, Alt+Left out to after the parent, through the established hidden-control pattern. Every move announces in the live region - including the refusals ("One is already first in Alpha"), which is the only feedback a keyboard user gets for a row that visibly went nowhere. "Move to..." in both row menus is the no-chord path.
7. **Undo on cross-parent moves only**, and it re-plans against the tree as it then stands rather than replaying an inverted batch - after a move out, the source folder has closed the gap, so putting the row back displaces its neighbour again. Within-parent misdrops are one drag back and raise no toast.

**Blast radius traced.** `CollectionTreeDnd` gained four members and every one has a reader in this diff (`useRowDnd`, both row types, `useRovingTreeFocus`). `useRevealActiveSelection` went from returning `void` to returning `revealEntity` - its one caller is `CollectionTree`; the two existing effects are untouched, and the new one clears a ref rather than state so it stays out of the render loop it observes. `DrawerPanel` gained one attribute, read only by the auto-scroll. Rows gained `data-parent-id` / `data-owner-id`, read by the hit test (which has an element and no model). The save-store registry gained a second reader.

**Deliberate choices worth calling out:**

- **The per-row half is shared, not copied.** `useRowDnd` / `rowDndClasses` / `RowDropIndicator` / `RowMoveControls` are one implementation both row types use. A parallel copy in `RequestItem` would stop receiving `CollectionItem`'s fixes, which is this codebase's named defect.
- **A request whose open tab has unsaved edits is not a drag source.** Two writers on one row - a save carrying the row's contents while a reorder rewrites its owner and order - is the #237 clobber family. It says so in a toast rather than refusing silently, which would read as "the press didn't register".
- **The reorder math's throw is caught and reported.** It throws when the row is not where the caller thinks it is - an Undo clicked after the tree moved under it. Loud (an error toast naming the row), but not a crash inside a toast's click handler.
- **"Move to..." is appended by the rows rather than built in `getCollectionActions`.** It belongs to the drag slice, which mounts after the CRUD slice; threading it backwards would put a `null` handler in the CRUD hook's dependency list for the sake of symmetry.
- **`docs/` has no user-facing sidebar page** - I checked. `index.md` covers install, import and MCP; the collection tree is documented for contributors in `docs/app/COMPONENTS.md` and `docs/design-system.md`, and both are updated here.

**Adjacent, deliberately not in this diff:** multi-select drag, touch affordances, collapse-restoration after spring-loading and tab-strip reordering are all parent-issue "out of scope". The batch endpoint's cross-request lock scope is #386.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check` - **337 files, 3188 tests, all passing** (124s). 35 of them are new; the baseline was 3153. `type-check`, `lint` (zero warnings) and `format:check` are clean. `mkdocs build --strict -f .github/mkdocs.yml` passes.

**No engine build or test run, and that is the intended scope.** The diff touches no file under `engine/` - #365 shipped `POST /reorder` and its validation - so no engine test could go from green to red. There are no pre-existing app failures on the base commit.

New coverage:

| File | Tests | Covers |
|---|---|---|
| `drag-gesture.test.ts` | 9 | threshold in every direction, measured from the press; cancel; click suppression spent exactly once |
| `drop-position.test.ts` | 11 | band boundaries incl. the zero-height row, the two-block rule in both directions, the root-folder refusal |
| `CollectionTree.dnd.test.tsx` | 15 | the whole interaction through real pointer events on real rows |

The jsdom file stubs only what jsdom does not have - `getBoundingClientRect` (rows stacked 32px apart) and `elementFromPoint`; everything between the press and the batch is the code under test. The reorder mutation is mocked but **applies the plan to the fixture**, because an Undo has to plan from where the row is *after* the move and an inert mock would test the opposite of the real thing. Pointer capture is polyfilled in `src/test/setup.ts` (jsdom has none - verified).

Mutation-checked by reverting each fix and re-running, not by inspection:

| Mutation | Reddens |
|---|---|
| drop the trailing-click suppression | 1 |
| a request between two folders lands on the folder | 1 |
| Alt+Arrow falls through to navigation | 2 |
| no spring-load timer | 1 |
| blocked state covers only the dragged row | 1 |
| focus is not returned to the moved row | 1 |
| undo offered for same-parent moves too | 1 |
| the drop index ignores that the moved row left its slot | 1 |
| folder rows split 50/50 like a request row | 2 |
| Escape does not cancel | 1 |

Restoring each returned the suite to green. Two of those started at zero and are the reason three assertions changed: the index test needed a three-row block (with two, the clamp hides the difference), and the focus test needed a move into a *collapsed* folder, since a reorder within one parent keeps the same DOM element and would have passed either way.

`drawer-row-hit-area.test.tsx` stays green, and the beyond-threshold drag never opening a row is asserted directly alongside a press-with-a-tremor that still does.

## Checklist

## Related Issues
Closes #367

---
_Generated by [Claude Code](https://claude.ai/code/session_01AQFx5Moex7nYN6j7V7uZAZ)_


athrvk:
fix the merge conflict

athrvk:
Merged `master` in and resolved it - `1f2d337`.

One conflict, in `CollectionTree.tsx`: #400 and this branch each added a dialog at the end of the panel, the collection runner's and the "Move to..." picker. They are unrelated, so the resolution keeps both rather than choosing.

The merge also brought in `CollectionTree.run.test.tsx`, which mocks `@/queries` wholesale - so it gains the `useReorderMutation` entry the tree now calls, the same one-line addition the other nine tree tests already took. Without it the three runner tests fail on mount rather than on anything they assert.

Verified after the merge: **343 files, 3264 tests, all passing**, plus `type-check`, `lint` and `format:check`.

Worth flagging separately: CI had not started on this PR at all before that push - the only check run was the labeler, which is `pull_request_target`. The `pull_request` workflows do not appear to fire for a pull request opened through the API token this session uses; the push is what triggered them. `pr-tests` is running now.

---

Issue #402, opened 2026-08-08
app/engine(data): data files - CSV/TSV/JSON/JSONL rows driving collection runs and load tests, with a reserved {{data.*}} namespace bound into load traffic


athrvk:
**Owner direction, 2026-08-09**: use CSV/TSV/JSON/JSONL files as input to collection runs and load tests (Postman data files, JMeter CSV Data Set, k6 SharedArray parity). Design grounded in a two-reader scoping pass at master `bf9f5d0` against the shipped scenario phases. Every fork is decided below; the boundaries with #356 and #357 are explicit so nothing is double-owned.

## What already exists (verified - build on, do not re-decide)

- The engine payload contract shipped with #352: `scenario.data` is a non-empty JSON array of objects, rows validated with index-naming 400s, capped by `maxScenarioDataRows` (default 1000, config max 1M) - `engine/src/core/scenario_plan.cpp:150-174`. Iterations default to row count, explicit count wins, index wraps as `i % rows` (`:178-194`). Rows are deliberately never persisted (manifest records `dataRowCount` only) and the engine never reads files from disk - both relitigation-proofed in the design doc.
- **#356 keeps the engine script half exactly as written**: `pm.iterationData` binding, wrap semantics, `dataRowIndex` on per-iteration records, the 400s. **This issue takes over #356's App-changes section** (the picker/preview it sketched) - agreed boundary, noted on #356.
- **#357 keeps the per-VU state machine**; this issue specifies the data semantics phase 6 currently lacks (its `VirtualUser` has no data field) - noted on #357.
- Renderer reality: no CSV parser exists anywhere (this feature adds the codebase's single tabular primitive); the file-read precedent is ImportModal's hidden `<input type="file">` + `FileReader` (no Electron dialog IPC - keep it that way); `ui/table.tsx` is the preview grid primitive; the runner UI (#354) just merged and is the design-mode host surface.

## The crux, and the decision that resolves it

In load mode, scripts run **deferred over sampled responses** (shipped #353/#357 design) - so `pm.iterationData` alone cannot make data rows influence actual traffic. A load-mode data feature that only feeds deferred scripts is decorative. Therefore:

**Decision - the `{{data.*}}` namespace.** Data columns are addressable in requests as `{{data.column}}` - a *reserved namespace*, deliberately disjoint from globals/collection/environment variables, which dissolves the Postman precedence question (no tier, no collision: `{{data.x}}` and `{{x}}` are different names). Binding:
[...]

Issue #379, opened 2026-08-08
engine(repo-hygiene): a 120KB test SQLite artifact is committed, and no gitignore rule stops the next one


athrvk:
Noticed while resolving the merge conflict on PR #376; deliberately left out of that diff since it is unrelated to the context bar.

## Problem

`engine/test_script_scopes.db.bak` is tracked on `master` - 122,880 bytes of SQLite, added by 67b35b8 (the design-mode scenario runner PR). It is a test artifact, not a fixture: `engine/tests/script_variables_test.cpp:325` writes `test_script_scopes.db` as its scratch database, and the `.bak` is the sidecar the engine's own backup path leaves beside it.

Nothing reads it. It is not referenced by any test, CMakeLists, or preset - the test creates its own database at run time, so deleting the file changes no behaviour.

The reason it got in is the second half of the problem: `.gitignore` has no `*.db` rule at all - line 7 is `Thumbs.db`, which is Windows shell debris and matches nothing here. Meanwhile `engine/tests/` names **at least ten** scratch databases (`test_collections_route.db`, `test_db_concurrency.db`, `test_globals_route.db`, `test_load_strategy.db`, ...). Every one of those is a file a contributor who runs `ctest` from inside `engine/` can `git add -A` by accident. This one is simply the first that did.

## Fix / solution pathway

Two lines, both needed - either alone leaves the hole open:

- `git rm --cached engine/test_script_scopes.db.bak` and delete it.
- Add the scratch-database patterns to `.gitignore`. Scope them to where the tests actually run rather than a bare global `*.db`: the engine writes these into the working directory of the test binary. Something like `engine/*.db`, `engine/*.db.bak` plus the build-tree equivalents, matched against where `ctest --preset linux-dev` actually drops them - verify by running the suite and checking `git status`, rather than by pattern-guessing.

Worth confirming while there: whether the `.bak` sidecar should be cleaned up by the test's own teardown. If `script_variables_test.cpp` removes its `.db` on exit but not the `.bak` the backup path created, the teardown has a gap and the gitignore rule is covering for it. Fix the teardown if so - the ignore rule is the backstop, not the answer.

## Engine changes

`.gitignore`, one file deletion, and possibly teardown in `engine/tests/script_variables_test.cpp`. No production code.

## Tests

No new test - this is repo hygiene, and there is no assertion that a file is absent worth writing. Verification is procedural:

[...]

PR #404, opened 2026-08-08
feat(engine/scenario): phase 5 - pm.iterationData over the run's data rows

### LINKED ISSUE #356: engine/app(scenario): phase 5 - data-driven iterations: pm.iterationData over rows sent on the run payload

athrvk:
## Description

Phase 5 of the collection runner (#340). Verified against master at `48fb3c2`: both prerequisites are closed (#353 gave the iteration index, #355 merged 20 minutes before this branch was cut), and the problem is still exactly as the issue describes it - `scenario.data` is validated and **counted** by `parse_scenario_request` (`scenario_plan.cpp:146`) and then discarded, `pm.iterationData` is bound nowhere in `setup_pm_object`, and `docs/app/pm-api-compatibility.md` lists it under "Not (yet) supported".

**Scope.** Per your boundary comment on the issue (2026-08-09), #402 now owns this issue's App-changes section - the data-file picker, the parser and the pre-run summary. Everything engine-side is delivered here as written, and #402 consumes it as its shipped contract. The one app change in this diff is not that picker: it is the reader for `dataRowIndex` (below), without which the field would be written and never read.

**Plan.** The root cause is narrow: the rows are validated and dropped, so the binding has nothing to read. The fix is therefore about *where the rows live* rather than about parsing. They ride `ScenarioExecution` - the object the run's worker holds behind a `shared_ptr<const>` and nothing persists - while `ScenarioRequest` keeps carrying the count alone, because that is the object `build_scenario_manifest` serializes into `runs.config_snapshot`. Rows are user data of unknown sensitivity, and the existing guarantee that they never touch disk becomes structural rather than a comment to remember. The alternative I rejected was hanging them off `ScenarioRequest` beside `data_row_count`: it is the shorter diff, and it puts the rows one `dump()` away from the snapshot, guarded only by the manifest builder listing its keys by hand.

**Absence is `undefined`, deliberately.** This is the opposite treatment to `pm.execution` and `pm.cookies`, and the issue asks for it: those are *capabilities*, and one that silently does nothing is #188's false success, so they are always bound and explain themselves. A data row is *data*, and "this run is not data-driven" is a fact a script may legitimately branch on - `pm.iterationData ? ... : ...` only works if absence is visible. The writers (`set` / `unset` / `clear`) are bound and throw, because those *are* the false-success case: the rows are a run input with no destination a write could land in, and the next iteration binds a different row regardless.

**Blast radius traced.** `ScriptContext::iteration_data` has one writer (`execute_exchange`, from `ExchangeInputs`) and one reader (`setup_pm_iteration_data`); `ExchangeInputs::iteration_data` has one setter, the scenario runner, exactly as `iteration` does. `ScenarioResolution::data_rows` is read only where the route builds the execution. `StepRecord::data_row_index` reaches two writers of wire shapes - `stamp_step_identity` (the stored trace) and `build_step_payload` (the SSE event) - and both of those have renderer readers, which is the app half below.

**The one app change: `dataRowIndex` gets a reader.** "Written but never read" is this repo's named defect, so rather than adding the field and leaving it for #402, the step list reads it: `RunResultTrace` / `ScenarioStepEvent` carry it, `parseStepEvent` and `stepRowFromResult` pass it through, and `ScenarioStepCard` renders "Iteration 4 · Row 1". That is also what makes the issue's "the wrap is not silent" true in practice - with `iterations` above the row count the iteration number alone cannot say which row a pass re-used. It is absent, never `0`, for a run that had no data set, on both paths, so a step cannot gain a row number when its stored copy arrives.

**Two details worth calling out:**

- **A stashed reference is refused.** QuickJS contexts are pooled and the global object survives with them, so a script can do `globalThis.saved = pm.iterationData` and a later script in that context can call what it stashed. It reads the *current* execution's row - none - and throws rather than answering about a run that has finished. Both halves of that leak are tested.
- **JSON values are dumped and re-parsed** rather than walked by hand. A row is arbitrary JSON and QuickJS's own parser accepts exactly that; a hand-written converter would be a copy that stops receiving the parser's fixes, and would be the place large integers quietly become doubles.

**Adjacent, deliberately not in this diff:** the file picker, the CSV/JSON parser and the pre-run summary (#402 phase A), `{{data.*}}` request binding (#402), and the load-mode shared cursor (#402 phase B with #357). `pm.iterationData` stays out of load mode by the same rule `pm.execution` does, and that is asserted rather than assumed.

While reading `engine/build/` I noticed a dozen `test_*.db.bak` artifacts the suite leaves behind - that is #379's scope, not touched here.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` - **1089 tests, all passing** (116s). `cd app && pnpm test && pnpm type-check && pnpm lint` - **340 files, 3233 tests, all passing** (153s); type-check, lint (zero warnings) and `prettier --check` on the touched files are clean. `mkdocs build --strict -f .github/mkdocs.yml` passes. No pre-existing failures on the base commit.

New coverage - 17 tests:

| File | Tests | Covers |
|---|---|---|
| `script_engine_test.cpp` | 8 | `get` reads the row and keeps JSON types (number, boolean, null, nested object); an unknown key is `undefined`; `toObject`; all three writers throw; `undefined` in a single send and in a load run's deferred `tests` script; a row does not survive into the next pooled execution; a stashed reference refuses once the row is gone |
| `scenario_runner_test.cpp` | 5 | rows drive the iteration count and row `i` reaches the **wire** on iteration `i`; an explicit count wraps 0,1,0,1,0 and every stored row names its index; a run without data binds nothing and stamps nothing; both scripts of a step read the same row; the SSE event carries the index on the same terms as the stored row |
| `scenario_plan_test.cpp` | 3 (2 new, 1 extended) | resolution carries the rows while the manifest still does not; no data resolves to no rows; a rejected row leaves none behind, not even the ones before it |

The runner tests assert the `X-Marker` header the **mock server saw**, not the script's return value: a row that binds but changes nothing sent would be decorative, and that is the failure mode worth catching.

Mutation-checked by reverting each behaviour and re-running, not by inspection:

| Mutation | Reddens |
|---|---|
| `iteration % rows` becomes `min(iteration, rows - 1)` | 3 |
| `dataRowIndex` stamped unconditionally (`value_or(0)`) | 1 |
| `pm.iterationData` not rebuilt per execution | 6 |
| the trace's `dataRowIndex` dropped in `stepRowFromResult` | 1 |
| `parseStepEvent` drops `dataRowIndex` | 1 |
| the card stops rendering the row label | 1 |

Restoring each returned the suite to green. The SSE-payload test stays green under the trace mutation and vice versa, which is what keeps the live and stored shapes pinned separately.

## Checklist

Docs updated in the same commit: `docs/engine/scripting.md` (a Data rows section - the read surface, the read-only rule, the wrap, and why absence is `undefined` here and not for `pm.execution`), `docs/app/pm-api-compatibility.md` (moved out of "Not (yet) supported", with the supported-surface table and a divergences section), `docs/engine/api-reference.md` (the binding on scenario runs, `dataRowIndex` on the step event and the stored row), `docs/engine/db-schema.md` (the sixth trace key), `docs/engine/architecture.md` (how rows reach the worker and why they stop there), and the design plan's §2.3 gains its "Shipped" marker.

**One doc correction made while there:** `pm-api-compatibility.md` still said `pm.info.iteration` / `iterationCount` were unsupported, in the same bullet this issue removes and in the `pm.info` section - stale since #353 shipped them. Both now describe what the engine does. That also renamed the `pm.info` heading ("three fields" → "five"); its only inbound link was the bullet being deleted.

## Related Issues
Closes #356

---
_Generated by [Claude Code](https://claude.ai/code/session_01BWYmgfCroNwze5WhNtRsYs)_


athrvk:
Claiming this - master moved (#403 merged, this branch is `behind`). Merging master in and re-running the suites now.

---

athrvk:
Merged `master` (`f14ab2a`, i.e. #403's drag-and-reorder) into this branch - `2ea0b7e`. Clean merge, no conflicts: master's four commits are app-only (`app/src/modules/collections`, `app/src/config`, docs) and this branch's app surface is the scenario step card, so the two do not overlap.

Re-verified on the merged tree: `pnpm test` - **343 files, 3268 tests, all passing**; `pnpm type-check` and `pnpm lint` clean. Engine is untouched by the merged commits, so the engine suite could not change colour from this merge; CI re-runs it on all three platforms anyway.

---

Issue #360, opened 2026-08-07
engine/app(ordering): make the order column real before the scenario phases consume it - request create appends, ties become deterministic and pinned, moves are validated


athrvk:
From the 2026-08-07 collection-tree review sweep (4 readers, adversarial verification per finding; verified at `d5899bb`, load-bearing anchors re-checked at `0f7d975`). This is the ordering-foundation cluster - and it is **urgent**: phase 1 of the collection runner (PR #359, closes #352) shipped `ScenarioPlan` consuming `requests.order`/`collections.order` straight from SQL whose tie order is unspecified and unstable, and #353 (the sequential runner) builds on it next. The drag-and-reorder feature (parent issue filed alongside) also depends on every item here.

## Problem (each verified)

1. **Request create defaults `order` to 0 instead of appending (medium).** Collections append engine-side (`max_order+1`, `engine/src/http/routes/collections.cpp:108-120`); requests get 0 (`requests.cpp:204`) and no renderer call site sends `order` on create. Consequence: every UI-created request ties at 0, the stored column does not encode the sidebar's sequence, and the moment explicit orders exist (first drag), **every newly created or duplicated request jumps to the top** of its collection. `handleDuplicateRequest` (`CollectionTree.tsx:365-389`) copies content but not order - same jump.
2. **Tie order is a different rule in every consumer and unstable engine-side (medium).** SQL orders by the single key (`database.cpp:597` collections, `:672` requests - re-verified single-key at `0f7d975`); among ties the implicit rowid decides, and `INSERT OR REPLACE` (`database.cpp:592,658`; TEXT primary key) **reassigns the rowid on every edit** - so editing a request can silently reshuffle its position among ties. The renderer tiebreaks by `createdAt` (private comparator, `app/src/queries/collections.ts:89-96`); the exported `compareRequestOrder` in `domain.ts:205-209` tiebreaks by `id` and is imported nowhere (verified: zero importers - the repo's written-never-read class). MCP `run_collection_smoke` and now `ScenarioPlan` consume the raw engine order. Net: **"run this folder" can execute in a different order than the sidebar shows, and that order can change after any edit.**
3. **PUT/POST `/requests` accept any `collectionId` with no existence check (medium).** `requests.cpp:157-222` (`:162-165` applies it); a move/create against a stale or mistyped collection id succeeds and **strands the row invisibly** - no per-collection GET lists it, cascade delete never reaps it. This is the exact endpoint a DnD cross-collection move uses.
[...]

PR #430, opened 2026-08-09
test(engine/app): extend the tree-order fixture with post-drag shapes

### LINKED ISSUE #364: app/engine(collection-tree): drag-and-reorder for collections and requests - tracking issue (atomic batch reorder, pointer-event DnD, keyboard move, undo)

### LINKED ISSUE #426: engine(tests): extend the tree-order conformance fixture with post-drag shapes - #364's fourth tracking criterion, unmet by the DnD wave
Surfaced by the #367/#364 closure verification (2026-08-09, verified at master `25f7fd6`). DnD phase 3 (#367) shipped and its own criteria verify, but the **parent #364's fourth tracking criterion is unmet and tracked nowhere**: "Sidebar, MCP smoke, and ScenarioPlan order agree after any sequence of drags (the #360 conformance fixture extended with post-drag shapes)." The fixture `engine/tests/fixtures/tree-order-conformance.json` is untouched since #360 (`e79ec42`) - its 7 cases cover the pre-drag ordering contract but no post-drag shapes. This issue owns that extension so #364 can close.

## Problem

The conformance fixture is the cross-consumer guarantee that the sidebar comparator, MCP `run_collection_smoke`, and `ScenarioPlan` order all agree - read by both `tree_order_test.cpp` (through the real SQL) and `tree-order.conformance.test.ts` (through the exported comparator). After a drag reorder or cross-parent move, the order column holds shapes the original 7 cases never exercise: dense `0..n-1` blocks post-renumber, post-normalization results (the first-drop-into-a-legacy-collection case #360/#365 handle), and cross-parent move landings. Nothing pins that both consumers still agree on *those* shapes - the exact regression surface the fixture exists to prevent, left open at the moment drags can now produce them.

## Fix / solution pathway

Add cases to `tree-order-conformance.json` representing states a drag produces (the reorder-math and batch-endpoint outputs are the source of truth for the shapes):
1. A collection renumbered dense `0..n-1` after an adjacent-swap reorder.
2. A legacy all-zeros collection after first-drop normalization (dense, in the pinned `order, created_at, id` display order).
3. A cross-parent move landing: the moved row's new order among its new siblings, source collection re-densified or gapped per the endpoint's actual behavior (read `apply_reorder` / `reorder-math.ts` for the real post-state - do not invent).
4. A mixed folders-first block after a request reorder that must not interleave with subfolders.

Both existing consumers pick the new cases up automatically (they iterate the fixture); the non-empty-scan guards already assert the fixture is read. If a shape reveals a genuine disagreement between the SQL order and the comparator, that is a real bug this issue surfaces - fix it, do not paper the fixture over it.

## App changes

None - `tree-order.conformance.test.ts` reads the fixture unchanged. Engine: the fixture file plus any comparator/SQL fix if a disagreement surfaces.

## Tests

The fixture cases ARE the test on both sides; add the non-empty count bump (the guards assert `cases_run >= N`). Commands: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`; `cd app && pnpm test`.

## Docs

None expected (the fixture is the contract). Verify no doc quotes the case count and state so.

## Sequencing

Parallel-safe against everything open. **Closing this satisfies #364's fourth criterion; #364 can then be verified and closed** (its other three tracking criteria verified met in the #367 closure pass). No open issue edits the fixture.

## Acceptance criteria

- [ ] The fixture carries post-drag shapes (renumber, normalization, cross-parent landing, folders-first block); both consumers agree on all of them.
- [ ] Any surfaced SQL-vs-comparator disagreement is fixed, not masked.
- [ ] Engine and app suites green; #364's fourth criterion is satisfiable.

---

athrvk:
## Description

**Plan.** The root cause is that `tree-order-conformance.json` was written for #360, before drags existed, so its seven cases pin the *pre-drag* contract - all-zero legacy blocks and hand-authored ties - and none of them is a shape a drop actually writes. The approach is the issue's: derive the new cases from the real post-states rather than invent them, by reading `app/src/modules/collections/reorder-math.ts` (which rows a drop rewrites) and `normalize_scope` / `stage_move` in `engine/src/http/routes/reorder.cpp` (what the renumber commits), then add the shapes those two produce. I rejected the alternative of generating cases from a simulated drag in the test itself - it would pin the fixture to the reorder implementation, so a bug in `reorder-math.ts` would produce cases that agree with the bug, and the fixture's whole job is to be a rule two implementations are checked *against*.

**The gap still reproduces as described, at `25f7fd6`.** The fixture is untouched since `e79ec42`, and the guards on both sides asserted only `>= 5` cases.

**What the post-drag shapes actually are - traced, not assumed.** Every drop leaves its block dense with distinct `order` values: an in-scope reorder writes only the run between the two indices (`planMove`'s splice loop), a cross-parent move closes the gap in the source (`i - 1`) and opens one in the destination (`i + 1`), and normalization renumbers `0..n-1` in the pinned display order before the moves land and win. So what these cases pin is not a new tie rule but the opposite: on exactly the rows a drag rewrote, **both former tiebreaks stop mattering**, and `createdAt` is now routinely out of sequence with `order`. That is what would break if either side reordered its keys.

The five cases, each named for the drag that produces it:

| Case | The drag behind it |
|---|---|
| adjacent swap, block still dense | `req-third` from index 2 to 1: only two rows written, `createdAt` now 1, 3, 2 |
| the display order a first drop renumbers into | the *input* to `normalize_scope`, with a `createdAt` tie inside the legacy block so all three keys resolve in one case |
| that same block after the drop lands | normalize gives zulu 0 / alpha 1 / bravo 2, then the move wins - creation time would put zulu first, the id tiebreak alpha first, and both are wrong |
| a cross-parent landing | the arrival keeps its own `createdAt`, newer than every row it lands among |
| the source block two drags deep | "after any sequence of drags": already scrambled by an earlier swap, then a row leaves and the gap closes |

**Mutation checks, run rather than reasoned about:**

| Reverted | Result |
|---|---|
| `createdAt` moved ahead of `order` in `compareTreeOrder` | 8 red, **4 of them the new post-drag cases** (3 of the original 7 catch it) |
| the id tiebreak reversed (`aId > bId`) | 4 red, including the new all-three-keys case |
| `CollectionItem` rendering requests before subfolders | the folders-first render assertion reddens |

**One thing the fixture structurally cannot express, so it is a test instead.** The issue's fourth shape - "a mixed folders-first block ... must not interleave with subfolders" - is not expressible as a fixture case: a case is one flat block, and the engine keeps collections and requests in *different tables*, so `get_requests_in_collection` cannot interleave a subfolder no matter what the fixture says. The risk is renderer-side, and it is real precisely because of the reorder math: both blocks are densified `0..n-1` independently, so a request and a subfolder in the same folder legitimately hold the **same** `order`, and only the render decides which comes first. `CollectionTree.folders-first.test.tsx` pins that with the request holding a *lower* order than the subfolder, so a merged sort would visibly lift it above the folder. The existing a11y test has both at order 0, where folders-first and an order-based merge are indistinguishable.

**Blast radius, traced.** The fixture has exactly two readers - `tree_order_test.cpp` (through the real SQL, as both requests and collections) and `tree-order.conformance.test.ts` (through `compareTreeOrder`) - and both iterate it, so the new cases are picked up without wiring. The new cases carry a `comment` key the readers ignore; the app's `ConformanceCase` interface does not declare it, which is fine for a `JSON.parse` cast and `pnpm type-check` is clean. `scenario_plan_test.cpp` only *references* the fixture in a comment. The guards move 5 -> 12 on both sides so a fixture that failed to load still cannot pass vacuously.

**No disagreement surfaced between the SQL and the comparator** on any of the new shapes - the criterion for that was "fix it, do not paper over it", and there was nothing to fix. All 12 cases were also checked against an independent Python implementation of the three-key byte-wise rule before the suites ran.

## Type of Change

- [x] Bug fix (a missing regression guard on shipped behaviour)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` - **1124/1124 passed**. (An earlier run against a still-linking binary failed ~15 `ReorderRouteTest` cases; it reproduces on a mid-link tree and is gone once the build completes.)

`cd app && pnpm test` - **3441 passed / 352 files** (3434 / 351 on the base commit, plus the 5 new fixture cases and 2 new render tests). `pnpm type-check` clean. ESLint and Prettier clean on the two app files touched, both of which were clean before.

## Checklist

- [x] I have added tests
- [x] I have updated documentation (none needed - see below)

**Docs: none, verified.** The issue asked to confirm no doc quotes the case count. Neither reference does: `docs/engine/api-reference.md:168` and `docs/app/state-management.md:712` both name the fixture as the pinning mechanism without a count, and the rule they state is unchanged. `reorder-math.ts`'s existing docstring already documents the separate-blocks rationale the new render test pins.

**Pre-existing, not fixed.** `engine/tests/tree_order_test.cpp` carries 48 clang-format violations on the base commit; this PR changes three integer literals in it and adds none, so per the repo rule the file is left as it was.

## Related Issues

Closes #426

Satisfies #364's fourth tracking criterion for the fixture half. **It does not fully close #364**, and I would rather say so than let it look covered: verifying that criterion turned up a genuine cross-block disagreement that is *drag-independent*, so it is out of this issue's scope. The engine's recursive scenario walk (`collect_requests`, `scenario_plan.cpp:89-101`) is pre-order - a collection's own requests run **before** its subfolders' - while the sidebar renders **subfolders above requests** (`CollectionItem.tsx:400-425`). For the tree in `RecursiveDescentIsDepthFirstByCollectionOrder` the sidebar shows `a_1, g_1, b_1, root_a, root_b` and the run executes `root_a, root_b, a_1, g_1, b_1`. That is the #360 defect class one level up, it is true with zero drags, and which side should move is your call - filed separately rather than changed here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QcvongSNyLRV1en9X8sZ8A)_

### DIFF
--- a/app/src/modules/collections/CollectionTree.folders-first.test.tsx
+++ b/app/src/modules/collections/CollectionTree.folders-first.test.tsx
@@ -1,0 +1,113 @@
+/**
+ * @vitest-environment jsdom
+ */
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * A folder's subfolders and its requests are two separately ordered blocks, and
+ * the tree renders every subfolder above every request (issue #364's fourth
+ * tracking criterion, #426).
+ *
+ * This is what keeps a drag honest. `reorder-math.ts` plans against one block at
+ * a time and renumbers it dense `0..n-1`, so after a reorder a request and a
+ * subfolder in the same folder hold the *same* `order` values - the column
+ * cannot say which of the two comes first, and only the render decides. Sorting
+ * the two blocks together on `order` is therefore not a near-miss but a
+ * different tree: here the request holds a lower `order` than the subfolder, so
+ * a merged sort would lift it above the folder, and a drop planned against the
+ * block the user saw would land somewhere else.
+ *
+ * The engine cannot express this - collections and requests are different
+ * tables, so `tree-order-conformance.json` pins one block at a time and this
+ * pins the relationship between them.
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { TooltipProvider } from "@/components/ui";
+import { useTabsStore } from "@/stores";
+import { useCollectionsStore } from "./collections-store";
+import CollectionTree from "./CollectionTree";
+
+// The post-drag collision: inside "acme", the request sits at order 0 and the
+// subfolder at order 1. Both blocks are dense, which is exactly what a reorder
+// leaves behind - the numbers agree only because they count different things.
+const collections = [
+	{ id: "acme", name: "Acme", order: 0 },
+	{ id: "billing", name: "Billing", parentId: "acme", order: 1 },
+];
+const requests = new Map([
+	["acme", [{ id: "r-ping", collectionId: "acme", name: "Ping", method: "GET", order: 0 }]],
+]);
+
+vi.mock("@/queries", () => ({
+	useReorderMutation: () => ({ mutate: vi.fn(), isPending: false }),
+	useCollectionsQuery: () => ({
+		data: collections,
+		isLoading: false,
+		isError: false,
+		error: null,
+		refetch: vi.fn(),
+	}),
+	useMultipleCollectionRequests: () => ({ requestsByCollection: requests }),
+	useCreateCollectionMutation: () => ({ mutateAsync: vi.fn(), isPending: false }),
+	useUpdateCollectionMutation: () => ({ mutateAsync: vi.fn(), isPending: false }),
+	useDeleteCollectionMutation: () => ({ mutateAsync: vi.fn(), isPending: false }),
+	useCreateRequestMutation: () => ({ mutateAsync: vi.fn(), isPending: false }),
+	useDeleteRequestMutation: () => ({ mutateAsync: vi.fn(), isPending: false }),
+	useUpdateRequestMutation: () => ({ mutateAsync: vi.fn(), isPending: false }),
+}));
+
+function renderTree() {
+	return render(
+		<QueryClientProvider
+			client={new QueryClient({ defaultOptions: { queries: { retry: false } } })}
+		>
+			<TooltipProvider>
+				<CollectionTree />
+			</TooltipProvider>
+		</QueryClientProvider>
+	);
+}
+
+beforeEach(() => {
+	Element.prototype.scrollIntoView = vi.fn();
+	useCollectionsStore.setState({ expandedCollectionIds: new Set(["acme"]) });
+	useTabsStore.setState({ openTabs: [], activeTabId: null });
+});
+
+describe("subfolders and requests are separate blocks, folders first", () => {
+	it("renders a subfolder above a request that holds a lower order", () => {
+		renderTree();
+
+		const folder = document.querySelector<HTMLElement>('[data-collection-id="billing"]')!;
+		const request = document.querySelector<HTMLElement>('[data-request-id="r-ping"]')!;
+		expect(folder).not.toBeNull();
+		expect(request).not.toBeNull();
+
+		// DOCUMENT_POSITION_FOLLOWING: the request comes after the folder in the
+		// rendered tree, despite ordering ahead of it on the column alone.
+		expect(
+			folder.compareDocumentPosition(request) & Node.DOCUMENT_POSITION_FOLLOWING
+		).toBeTruthy();
+	});
+
+	it("keeps both rows in the same group, so the split is render-order only", () => {
+		renderTree();
+
+		// If the blocks were merged and sorted, or split into two groups, the
+		// numbering below would not be a single run of two.
+		const folder = document.querySelector<HTMLElement>('[data-collection-id="billing"]')!;
+		const request = document.querySelector<HTMLElement>('[data-request-id="r-ping"]')!;
+		expect(folder).toHaveAttribute("aria-posinset", "1");
+		expect(request).toHaveAttribute("aria-posinset", "2");
+		expect(folder).toHaveAttribute("aria-setsize", "2");
+		expect(request).toHaveAttribute("aria-setsize", "2");
+	});
+});
--- a/app/src/types/tree-order.conformance.test.ts
+++ b/app/src/types/tree-order.conformance.test.ts
@@ -59,7 +59,7 @@ function toRow(row: ConformanceCase["rows"][number]): OrderedTreeRow {
 
 describe("tree-order conformance fixture", () => {
 	it("scanned a non-empty fixture (guards the scan itself)", () => {
-		expect(fixture.cases.length).toBeGreaterThanOrEqual(5);
+		expect(fixture.cases.length).toBeGreaterThanOrEqual(12);
 		for (const c of fixture.cases) {
 			expect(c.rows.length).toBeGreaterThan(1);
 			expect(c.expected).toHaveLength(c.rows.length);
--- a/engine/tests/fixtures/tree-order-conformance.json
+++ b/engine/tests/fixtures/tree-order-conformance.json
@@ -1,5 +1,5 @@
 {
-	"description": "The one sidebar/scenario ordering rule, read by both suites. Rows sort by `order` ascending, ties by `createdAt` ascending, remaining ties by `id` ascending under byte-wise (SQLite BINARY) comparison. Engine side: engine/tests/tree_order_test.cpp inserts each case's rows into the collections and the requests table and reads them back through Database::get_collections / get_requests_in_collection, so the SQL ORDER BY is what is being asserted. Renderer side: app/src/types/tree-order.conformance.test.ts sorts the same rows with compareTreeOrder. A rule that changes on one side and not the other fails here or in ctest, never in a user's sidebar.",
+	"description": "The one sidebar/scenario ordering rule, read by both suites. Rows sort by `order` ascending, ties by `createdAt` ascending, remaining ties by `id` ascending under byte-wise (SQLite BINARY) comparison. Engine side: engine/tests/tree_order_test.cpp inserts each case's rows into the collections and the requests table and reads them back through Database::get_collections / get_requests_in_collection, so the SQL ORDER BY is what is being asserted. Renderer side: app/src/types/tree-order.conformance.test.ts sorts the same rows with compareTreeOrder. A rule that changes on one side and not the other fails here or in ctest, never in a user's sidebar. The first seven cases are the pre-drag ordering contract (#360); the rest are shapes drag-and-drop actually produces (#364/#426), each derived from app/src/modules/collections/reorder-math.ts and the renumber in engine/src/http/routes/reorder.cpp rather than invented - a drop leaves the block dense with distinct orders, so what these pin is that the two former tiebreaks stop mattering on exactly the rows a drag rewrote. A case holds one block: subfolders and requests are separately ordered, so the `order` column never interleaves them.",
 	"cases": [
 		{
 			"name": "explicit order wins over both creation time and id",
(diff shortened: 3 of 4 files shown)


athrvk:
Claiming this - master has moved 4 commits ahead (`25f7fd6` -> `0f31fe8`) since this branch was cut, so the PR is `behind`. One of those commits (`32f85c3`, "one definition of the scratch-database file set") touches `engine/tests/`, which this PR also touches, so I'm merging master in, resolving anything that conflicts, and re-running the suites before pushing.

---

athrvk:
Updated - `origin/master` merged in at `ee1ff83`. The branch is no longer behind.

**The merge was clean, including the one file both sides touched.** `32f85c3` replaced this test's hand-rolled scratch-file cleanup with `vayu::tests::remove_database_files` from the new `temp_database.hpp`; my change to the same file is three integer literals (the `5u` -> `12u` fixture guards). Git took both, and I checked the result rather than trusting it: the merged `tree_order_test.cpp` includes `temp_database.hpp`, `cleanup()` delegates to the shared helper, and all three `EXPECT_GE (..., 12u)` guards survive. The PR's own contribution is unchanged - still 169/5 across the same 4 files.

**Re-verified on the merged tree, not the pre-merge one:**

| Check | Result |
|---|---|
| `python build.py -e -t` | clean (6m47s; picks up master's new `graphql_body.cpp`) |
| `ctest --preset linux-dev` | **1137/1137 passed** (was 1124 pre-merge - master added 13) |
| `cd app && pnpm test` | **3441 passed / 352 files** |
| `cd app && pnpm type-check` | clean |

Nothing new was fixed or hidden; this is a base update only. The body's engine count of 1124/1124 refers to the pre-merge run and is now 1137/1137 - noting it here rather than editing the body, since the earlier number was accurate for the tree it was run on.

---

PR #432, opened 2026-08-09
fix(engine/app): run a folder in the order the sidebar shows it

### LINKED ISSUE #431: engine/app(collection-tree): a recursive &#34;run this folder&#34; executes in an order the sidebar never shows - the walk is requests-first, the tree renders folders-first
Surfaced while verifying #364's fourth tracking criterion for #426 (verified at master `25f7fd6`). #426 extends the conformance fixture with post-drag shapes and found no SQL-vs-comparator disagreement *within* a block. This is the disagreement *between* the two blocks, and it is drag-independent - true with zero drags - so it was out of #426's scope and is filed here instead.

This is the #360 defect class one level up: "run this folder" executes in an order the user has never seen. #360 fixed it for rows inside one block; the ordering *between* a folder's own requests and its subfolders' was never pinned by anything.

## Problem

The two consumers disagree on where a collection's own requests sit relative to its subfolders':

- **The engine's recursive scenario walk is pre-order, own-requests-first.** `collect_requests` (`engine/src/core/scenario_plan.cpp:89-101`) pushes `get_requests_in_collection(id)` for the current node, *then* descends into its children.
- **The sidebar renders folders-first.** `CollectionItem.tsx:400-425` renders `childCollections.map(...)` and only then `requests.map(...)`, so every subfolder's subtree appears **above** the folder's own requests.

Using the tree already in `scenario_plan_test.cpp:159` (`RecursiveDescentIsDepthFirstByCollectionOrder`):

```
root            [root_a, root_b]
  child_a (order 0)  [a_1]
    grandchild       [g_1]
  child_b (order 1)  [b_1]
```

| | Order |
|---|---|
| Sidebar, top to bottom (expanded) | `a_1, g_1, b_1, root_a, root_b` |
| Recursive run (pinned by that test) | `root_a, root_b, a_1, g_1, b_1` |

Both are internally consistent and each has a test asserting it, which is exactly why neither side noticed - the same shape #360 described. Severity is moderate: it only bites a collection that holds *both* its own requests and subfolders, and only on a recursive run.

**Not affected, verified**: MCP `run_collection_smoke` runs a collection's DIRECT requests only and discloses how many sub-collections it left out (`app/electron/mcp/tools.ts:1552`), so it has no cross-block order to disagree about. Non-recursive scenario runs are single-block and already agree.

## Fix / solution pathway - needs a decision

The two sides are each defensible and the fix is one line either way; the question is which order is *right*, not which code is wrong.

- **A - descend before running own requests** (change `collect_requests` to push children's requests first, matching the sidebar). The run then matches what the user sees top-to-bottom, which is the property #360 established. Cost: reverses the run order of every existing mixed collection, and `RecursiveDescentIsDepthFirstByCollectionOrder` is rewritten - a behaviour change to shipped run semantics.
- **B - render requests above subfolders** (change `CollectionItem`). Same agreement, reached from the other end. Cost: every file tree users know (VS Code, Finder, Postman) is folders-first, so this trades a correct run order for a sidebar that reads as wrong.
- **C - state that they differ** and leave both: document the pre-order rule in `docs/engine/scripting.md`/the run UI so a recursive run says it runs own-requests-first. Cheapest, and honest, but it keeps a user-visible mismatch that #360 decided was not acceptable one level down.

A is the one to beat: it makes the run order derivable from the tree the user is looking at, which is the invariant #364's fourth criterion actually asks for, and the sidebar convention in B is the one users have the strongest prior about. The cost of A is real though - it changes the order existing scenario runs execute in, so it wants a release note.

## Engine changes

`collect_requests` in `engine/src/core/scenario_plan.cpp` for A (the descent already reverses children onto the stack; the own-requests push moves after it). App: none for A, `CollectionItem.tsx` for B.

## Tests

- `scenario_plan_test.cpp`: `RecursiveDescentIsDepthFirstByCollectionOrder` restated for the chosen order, plus a case where a collection holds own requests *and* subfolders at colliding `order` values (the post-densify shape - both blocks are renumbered `0..n-1` independently, so the collision is the normal case after a drag, not an edge one).
- The cross-consumer half is what has no home today: a test that derives the sidebar's displayed request sequence and the plan's step sequence from one tree and asserts they match. `CollectionTree.folders-first.test.tsx` (added in #426) pins the render side; the run side is `step_ids`.
- Commands: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`; `cd app && pnpm test`.

## Docs (same commit)

`docs/engine/api-reference.md` if the scenario/run ordering is described there, and a release note for A since it changes the order a stored scenario executes in. Verify whether `docs/app/state-management.md`'s sorting paragraph should gain the cross-block sentence - it currently describes only the within-block rule.

## Sequencing

Parallel-safe against everything open. #426 lands the fixture extension and the folders-first render test first (it touches `CollectionItem`-adjacent tests); this issue should land after it so the render-side guard already exists. **#364 cannot close until this is resolved** - its fourth criterion says the sidebar, MCP smoke and ScenarioPlan orders agree, and #426 satisfies that only within a block.

## Acceptance criteria

- [ ] The recursive run order and the sidebar's displayed order agree for a collection holding both its own requests and subfolders, pinned by a test that derives both from one tree.
- [ ] The chosen option is stated in the PR with the rejected alternatives and the behaviour-change cost.
- [ ] Colliding `order` values across the two blocks (the post-densify shape) are covered.
- [ ] Engine and app suites green; docs and a release note updated same-commit if the run order changes.

---

athrvk:
## Description

Option **A** from the issue: the engine's recursive walk now matches the sidebar, and the mismatch is pinned across both consumers by one fixture.

**Root cause.** `collect_requests` (`engine/src/core/scenario_plan.cpp`) was pre-order: it pushed `get_requests_in_collection(id)` for the current node, *then* descended. `CollectionItem.tsx` renders `childCollections.map(...)` before `requests.map(...)` at every depth. Both were internally consistent and each had a test asserting its own order, so a folder holding both its own requests and subfolders ran in a sequence that appeared nowhere in the UI.

**Approach.** The walk keeps one stack and gains an explicit `Emit` marker, pushed *underneath* a node's children - so each sub-collection's whole subtree is emitted before the parent's own requests, sub-collections still by `collections.order`. The `visited` guard against a corrupt `parent_id` is unchanged, and because an `Emit` is only queued past that guard, a cycle still cannot double-emit. The cross-consumer half had no home, which is why neither side noticed: a new fixture, `engine/tests/fixtures/recursive-run-order-conformance.json`, is read by the engine (`scenario_plan_test.cpp` seeds the tree and reads the plan's `step_ids`) and by the renderer (`CollectionTree.run-order.conformance.test.tsx` renders the same tree fully expanded and reads the request rows out of the DOM). The render itself is the assertion subject - a comparator copy in the app test would be the second source of truth the fixture exists to prevent.

**Rejected.** **B** (render requests above subfolders) reaches the same agreement from the other end, but every file tree users know is folders-first, and `drop-position.ts` plus `reorder-math.ts` already plan drags against the folders-first two-block model - it would trade a correct run order for a sidebar that reads as wrong and a DnD model to rework. **C** (document the divergence) is cheapest and keeps a user-visible mismatch #360 already decided was unacceptable one level down.

**Cost, stated plainly.** This changes the order existing collections execute in. A recursive run of a folder with both sub-folders and requests visits its steps in a different sequence than before, so a scenario whose steps depend on each other through variables or cookies is worth re-checking. Folders holding only requests, only sub-folders, or run without "Include sub-folders" are unaffected.

**Two notes on the issue text.**
- The issue's sidebar column reads `a_1, g_1, b_1, root_a, root_b`, which is off by one level: folders-first applies at *every* depth, so `child_a`'s subfolder renders above `child_a`'s own request and the displayed order is `g_1, a_1, b_1, root_a, root_b`. The fix produces that (deepest-subtree-first), not the issue's sequence - verified by rendering the tree, not by reading the code.
- `MCP run_collection_smoke` was re-verified as direct-requests-only, so it has no cross-block order to disagree about, and the app has no second recursive walk to update.

**Deliberate deviation from the issue spec.** Its last acceptance criterion asks for a release note alongside the docs. One was written and then removed at the reviewer's request (`.github/release-notes/vX.Y.Z.md` is authored when a release is planned, from `git log`, so a file started mid-cycle is not the repo's flow). The behaviour change is recorded in the commit history and here instead, for whoever writes the 0.15.0 notes.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Breaking change (behaviour change: the sequence a recursive collection run executes in)
- [x] Documentation update

## Testing

All green on this branch, run from a clean tree:

- `python build.py -e -t` then `ctest --preset linux-dev` - **1139/1139 passed**. Three of those are new/changed: `RecursiveDescentRunsEachSubtreeBeforeTheFoldersOwnRequests` (restated for the new order), `RunOrderConformanceFixtureIsNonEmpty`, `RecursiveRunFollowsTheSidebarOrderInEveryFixtureCase` (5 fixture cases), plus `ParentIdCycleTerminatesAndVisitsEachCollectionOnce` whose degenerate sequence follows from the new rule.
- `cd app && pnpm test` - **3448/3448 passed across 353 files**, including the new `CollectionTree.run-order.conformance.test.tsx` (7 tests).
- `mkdocs build --strict` - clean (docs changed).
- No pre-existing failures reproduced on the base commit.

**Mutation-checked both halves.** Restoring the pre-order push in `collect_requests` and rebuilding reddens exactly the three ordering tests and nothing else (1136/1139 pass), so no other engine test was silently relying on the old order. Flipping `CollectionItem`'s two render blocks reddens 6 of the 7 new app tests plus the `folders-first` pair - the app test reads the real render, not a copy of the rule.

Fixture cases: a three-level tree (deepest subtree first), two subfolders plus own requests, colliding `order` values across the two blocks (the post-densify shape, both blocks dense `0..n-1`), all-zero orders falling to `createdAt` block by block, and an empty subfolder contributing nothing.

## Checklist

Docs updated in the same commit: `docs/engine/api-reference.md` (the scenario block's ordering paragraph, plus a two-block note in **Ordering**), `docs/plans/collection-runner-design.md`, `docs/app/state-management.md` (the sorting paragraph gains the cross-block sentence the issue asked about), the `recursive` doc comments in `scenario_plan.hpp` / `app/src/types/api.ts`, and the Run dialog's module comment plus its user-facing "Include sub-folders" hint.

## Related Issues
Closes #431

### DIFF
--- a/.github/release-notes/v0.15.0.md
+++ b/.github/release-notes/v0.15.0.md
@@ -1,9 +1,0 @@
-## [0.15.0] - 2026-08-09
-
-In-progress notes for the next release. Entries are added as the changes land and the lead paragraph is written when the version is cut.
-
-### Changed
-
-- **A recursive collection run executes in the order the sidebar shows.** "Include sub-folders" walked a folder's *own* requests first and descended afterwards, while the tree renders every sub-folder above every request - so a folder holding both ran in a sequence that appeared nowhere in the UI. The run now follows the tree: each sub-folder's whole subtree, then the folder's own requests, sub-folders by their own order and requests by theirs. **This changes the order existing collections execute in** - a run of a folder with both sub-folders and requests will visit its steps in a different sequence than before this release, so a scenario whose steps depend on each other through variables or cookies is worth re-checking. Folders holding only requests, only sub-folders, or run without "Include sub-folders" are unaffected. The two consumers can no longer drift: one fixture now pins the rendered order and the engine's plan against each other.
-
-[0.15.0]: https://github.com/athrvk/vayu/compare/v0.14.0...v0.15.0
--- a/.github/release-notes/v0.15.0.md
+++ b/.github/release-notes/v0.15.0.md
@@ -1,0 +1,9 @@
+## [0.15.0] - 2026-08-09
+
+In-progress notes for the next release. Entries are added as the changes land and the lead paragraph is written when the version is cut.
+
+### Changed
+
+- **A recursive collection run executes in the order the sidebar shows.** "Include sub-folders" walked a folder's *own* requests first and descended afterwards, while the tree renders every sub-folder above every request - so a folder holding both ran in a sequence that appeared nowhere in the UI. The run now follows the tree: each sub-folder's whole subtree, then the folder's own requests, sub-folders by their own order and requests by theirs. **This changes the order existing collections execute in** - a run of a folder with both sub-folders and requests will visit its steps in a different sequence than before this release, so a scenario whose steps depend on each other through variables or cookies is worth re-checking. Folders holding only requests, only sub-folders, or run without "Include sub-folders" are unaffected. The two consumers can no longer drift: one fixture now pins the rendered order and the engine's plan against each other.
+
+[0.15.0]: https://github.com/athrvk/vayu/compare/v0.14.0...v0.15.0
--- a/app/src/modules/collections/CollectionTree.run-order.conformance.test.tsx
+++ b/app/src/modules/collections/CollectionTree.run-order.conformance.test.tsx
@@ -1,0 +1,221 @@
+/**
+ * @vitest-environment jsdom
+ */
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * Cross-consumer conformance: the request sequence the sidebar *displays* for a
+ * recursive run against the sequence the engine's plan *executes* (issue #431).
+ *
+ * `CollectionTree.folders-first.test.tsx` pins the relationship between one
+ * folder's two blocks, and `tree-order-conformance.json` pins the order within a
+ * block. Neither could notice the case this file exists for: the engine's walk
+ * was pre-order (a folder's own requests, then its subfolders'), the tree renders
+ * subfolders above requests at every depth, and each side had a test asserting
+ * its own order. "Run this folder" therefore executed in a sequence the user had
+ * never seen - the #360 defect one level up.
+ *
+ * So the fixture is read by both sides:
+ * `engine/tests/scenario_plan_test.cpp` seeds each case and reads
+ * `resolve_scenario`'s step ids; this file renders the same tree fully expanded
+ * and reads the request rows top to bottom out of the DOM. The render is the
+ * assertion subject, not a re-derivation of it - a comparator copy here would be
+ * the second source of truth the fixture exists to prevent.
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render } from "@testing-library/react";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { TooltipProvider } from "@/components/ui";
+import { useTabsStore } from "@/stores";
+import { compareTreeOrder } from "@/types";
+import { useCollectionsStore } from "./collections-store";
+import CollectionTree from "./CollectionTree";
+
+const here = dirname(fileURLToPath(import.meta.url));
+const fixturePath = join(
+	here,
+	"..",
+	"..",
+	"..",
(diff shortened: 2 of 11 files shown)


athrvk:
remove the release notes .md file, it will be generated when a release is planned

PR #434, opened 2026-08-09
fix(engine/scripting): offer pm.iterationData, and close the callable-implies-offered hole

### LINKED ISSUE #418: engine(scripting): pm.iterationData has no editor completion - 126 entries, zero for the one API a data-driven run is written against - and the completions guard still cannot catch the next omission
**Owner-reported 2026-08-09, verified at master `25f7fd6`.** The script editor's completion surface (`engine/src/http/routes/scripting.cpp`) has 126 entries and none for `pm.iterationData`, although the sandbox implements it (`engine/src/runtime/script_engine.cpp:4694-4720`, shipped by #356/PR #403) and `docs/engine/scripting.md:782-790` documents it. The one API a data-driven run is written against is the one the editor refuses to suggest - the repo's written-but-never-read shape, this time written-but-never-offered.

This omission was *predicted*: the #337 closure verification recorded that "the existing completions guard only checks offered-implies-callable, not callable-implies-offered, so forgetting the entries fails nothing" - and #337's own fix asserted its five new entries by name without closing the general hole. The very next pm.* surface fell straight through it.

## Problem

- Zero `iterationData` hits in `scripting.cpp` (verified; 126 completion entries present).
- The guard direction is still one-way: `engine/tests/script_completions_test.cpp` asserts offered entries are callable (plus the #337-era by-name assertions), but nothing derives "what the sandbox binds" and requires an entry for it - so every future `pm.*` addition can silently repeat this.

## Fix / solution pathway

1. Add the `pm.iterationData` completion entries: the object itself plus `get`, `has`, `toObject` (match the accessor surface `script_engine.cpp:4694+` actually binds - read it, do not guess; the owner verified `get()` at `:4744`), with descriptions matching `scripting.md:782-790`'s wording (row binding, scenario-runs-only).
2. Close the guard direction: a callable-implies-offered assertion. The sustainable form: one authoritative list of bound `pm.*` names (either exported from the runtime as a static table the binding code and the completions route both consume - the single-source fix - or a test-side extraction that walks `setup_pm_*`'s registrations) asserted equal to the completions surface, with an allowlist for deliberately-uncompletable internals if any exist (each with a stated reason). The by-name spot checks stay; this makes the *class* impossible, not the instance.
3. Note in the entry description that `pm.iterationData` is `undefined` outside scenario runs (the #300/#356 boundary) so the completion does not overpromise in plain request scripts.

## App changes

None - the renderer consumes `GET /script-completions` as data (verified by the existing flow); no MCP surface.

## Tests

- `script_completions_test.cpp`: the new entries asserted by name with the non-empty-scan guard (the established pattern), AND the new inverse guard - mutation check: remove any single completion entry for a bound API and the suite reddens; add a fake binding without an entry and it reddens.
- Commands: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`.

## Docs

None - `scripting.md` already documents the API; completions now match it. Verify no doc lists the completion count and state so.

## Sequencing

Parallel-safe against everything open. Touches `scripting.cpp`'s completion table only - no open issue edits it (#415's data-token warning is `scenario_plan`/composer territory).

## Acceptance criteria

- [ ] Typing `pm.it` in the script editor offers `iterationData` with `get`/`has`/`toObject`, descriptions matching the doc.
- [ ] The inverse guard exists: an unbound-entry or unoffered-binding mutation reddens the suite - the next pm.* surface cannot repeat this silently.
- [ ] Engine suite green.

---

athrvk:
## Description

**Plan.** The root cause is not the missing `pm.iterationData` entries - it is that the completions guards only ever ran *offered-implies-callable*, plus by-name spot checks for three surfaces, so an entire `pm.*` surface could be bound and listed nowhere and nothing reddened. #356 shipped `pm.iterationData`, `scripting.md` documented it, and all 126 completion entries missed it. So this adds the entries **and** closes the direction: a new guard asks the *runtime* what it bound and requires a completion for each. I rejected the issue's other suggested shape - exporting a static table of bound `pm.*` names from the runtime for the binding code and the completions route to share - because that table is itself a second copy that can drift from the actual `JS_SetPropertyStr` calls, and maintaining it is the same act of remembering that failed here. Enumerating the live `pm` object cannot drift from what the runtime bound, because it *is* what the runtime bound.

Verified at `afbfaba`: zero `iterationData` hits in `scripting.cpp`, 126 entries. Blast radius traced - the completion table has two consumers, `GET /scripting/completions` (rendered by `useScriptCompletionProvider`) and `GET /scripting/types`, whose `.d.ts` is *generated from the same table* by `script_types.cpp`. No renderer or MCP code hardcodes the list, and no doc states the entry count, so both claims in the issue hold.

**What the guard found, beyond the reported instance.** Rather than weaken it, each was fixed:

- `pm.info.iteration` / `pm.info.iterationCount` - bound by the collection runner since #356, never offered. `pm.info`'s own completion text also still claimed they are "deliberately absent", which stopped being true when the runner started setting them.
- `pm.response.errorCode` / `pm.response.errorMessage` - bound on a transport failure, never offered, and undocumented in `scripting.md`.
- `pm.response.to` - the assertion chain root, allowlisted with its reason (it is offered as the terminal `.to.be.*` labels, and reading it asserts nothing).

**Deliberate deviations from the issue spec**, both stated because the issue asked me to read rather than guess:

1. The issue expected `get` / `has` / `toObject`. The runtime binds **`get` and `toObject`** only (`script_engine.cpp:4820+`) - there is no `has`. Offering one would be the #182 defect, so a test asserts `pm.iterationData.has` is *not* offered.
2. `set` / `unset` / `clear` are bound but throw by design, so they are allowlisted as deliberately-uncompletable alongside the existing `pm.variables.set`, not offered.

**One fix outside the issue's stated files, caused by it.** Adding the `pm.info` entries exposed the same drift one layer down: `field_type` in `script_types.cpp` matched a hand-listed set that contained `string | undefined` but not `number | undefined`, so `pm.info.iteration` was declared **`void`** in the generated `.d.ts` and `pm.info.iteration + 1` was an error in the editor for a script the runtime runs happily. Shipping the completions without this would have shipped that defect, so it is fixed here by splitting the `| undefined` suffix off and re-attaching it rather than enumerating the pairs.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`

**1142/1142 passed** (1139 on the base commit, plus the 3 tests added here). No pre-existing failures on the base commit.

Three tests added:

- `ScriptCompletions.EveryPmMemberTheRuntimeBindsIsOffered` - the general guard. Walks `pm` two levels deep against a context with **every conditional surface populated** (a data row, a scenario iteration, a transport error), because a guard run against a bare context would silently not check the members that only exist in those cases. Asserts it inspected more than 40 members, so it cannot pass by scanning nothing.
- `ScriptCompletions.TheIterationDataAccessorsAreOffered` - the named half, pinning the three entries and asserting each says it is `undefined` outside a data-driven run, plus that `has` is not offered.
- `ScriptTypesTest.OptionalFieldsKeepTheirUnderlyingType` - pins `number | undefined` and `string | undefined` through the generator, and that `void | undefined` is never emitted.

**Mutation check, both directions** (the acceptance criterion), applied together and confirmed red, then reverted:

- Removed the `pm.iterationData.get` entry → `the runtime binds these but the completion list never offers them: pm.iterationData.get, ...`
- Added a `pm.mutationFakeSurface` binding with no entry → same test named it. The next `pm.*` surface cannot repeat this silently.

Generated `.d.ts` verified against a running engine (`GET /scripting/types`) before and after: `iteration: number | undefined;`, `iterationCount: number | undefined;`, `errorCode: string | undefined;`, `iterationData: { get(column: string): any; ... }`.

No app changes - the renderer consumes both endpoints as data. `mkdocs build --strict` could not be run (mkdocs is not installed in this environment); the docs edit adds no link, heading or nav entry, so there is no anchor for it to break.

## Checklist

`scripting.cpp` was clang-format clean before and is formatted; `script_completions_test.cpp` was already not clang-format clean on master, so per CLAUDE.md it was left alone and the added code matches the file's prevailing style.

## Related Issues
Closes #418

---
_Generated by [Claude Code](https://claude.ai/code/session_01QoyTt72zJgqNEWhbPC8Kng)_

### DIFF
--- a/docs/engine/scripting.md
+++ b/docs/engine/scripting.md
@@ -140,7 +140,14 @@ pm.response.reason()          // Status reason phrase ('OK', 'Not Found')
 pm.response.size()            // { body, header, total } in bytes
 pm.response.cookies           // Set-Cookie, parsed - see below
                               // (the stored session is pm.cookies)
+pm.response.errorCode         // string | undefined - TIMEOUT, DNS_ERROR, ...
+pm.response.errorMessage      // string | undefined - the same failure in words
 ```
+
+`errorCode` and `errorMessage` are present **only when the send failed before a
+response arrived**, so `if (pm.response.errorCode)` is the test for a transport
+failure - the status code in that case is vayu's synthetic `0`. A response that
+reached the script from a server carries neither.
 
 `reason()` reports the reason phrase from the status line. Where none was
 received it falls back to the canonical text for the code, so a client-side
--- a/engine/src/http/routes/script_types.cpp
+++ b/engine/src/http/routes/script_types.cpp
@@ -312,12 +312,13 @@ std::string field_type (const std::string& detail) {
     // undefined` and not `number | undefined`, which silently declared
     // `pm.info.iteration` as `void` and made `pm.info.iteration + 1` an error
     // in the editor.
-    static constexpr std::string_view OPTIONAL = " | undefined";
-    bool optional                              = false;
-    if (base.size () > OPTIONAL.size () &&
-    base.compare (base.size () - OPTIONAL.size (), OPTIONAL.size (), OPTIONAL) == 0) {
+    static constexpr std::string_view OPTIONAL_SUFFIX = " | undefined";
+    bool optional                                     = false;
+    if (base.size () > OPTIONAL_SUFFIX.size () &&
+    base.compare (base.size () - OPTIONAL_SUFFIX.size (),
+    OPTIONAL_SUFFIX.size (), OPTIONAL_SUFFIX) == 0) {
         optional = true;
-        base     = trim (base.substr (0, base.size () - OPTIONAL.size ()));
+        base = trim (base.substr (0, base.size () - OPTIONAL_SUFFIX.size ()));
     }
 
     std::string resolved;
@@ -331,7 +332,7 @@ std::string field_type (const std::string& detail) {
         // `void | undefined` is not a type, so the suffix goes with it.
         return "void";
     }
-    return optional ? resolved + std::string (OPTIONAL) : resolved;
+    return optional ? resolved + std::string (OPTIONAL_SUFFIX) : resolved;
 }
 
 void append_doc (std::string& out, const TypeNode& node, const std::string& indent) {
--- a/engine/src/http/routes/script_types.cpp
+++ b/engine/src/http/routes/script_types.cpp
@@ -36,6 +36,7 @@
 #include <cstddef>
 #include <map>
 #include <string>
+#include <string_view>
 #include <vector>
 
 #include "vayu/http/routes.hpp"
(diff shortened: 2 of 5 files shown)


athrvk:
Claiming this - CI is red (Engine on windows-latest, plus the Conventional title check). Diagnosing and pushing a fix to this branch this run.

---

athrvk:
file up a new issue for the deviations

athrvk:
Filed as #435, sequenced after this PR.

It covers three things rather than two, because writing them up surfaced one I had only stated implicitly:

1. **`pm.iterationData.has`** - the runtime binds `get`/`toObject` only, so this PR offers those two. But every sibling scope has a `has()`, and `get('col') !== undefined` cannot tell an absent column from one whose value is JSON `null` - so the issue leans toward binding it rather than documenting the gap.
2. **The guard's depth-2 limit** - it walks `pm.<key>.<member>` and skips the header/cookie maps, because those objects also carry one property per wire header and descending would demand a completion for `Authorization`. So `pm.response.headers.get` and the `pm.cookies.jar()` members are still only covered in the offered-implies-callable direction. A narrowed hole, not a closed one.
3. **Optionality is not in the generated `.d.ts`** - `pm.iterationData` is `undefined` outside a data-driven run and the completion text says so, but the generated type declares it always-present, so the editor does not flag `pm.iterationData.get(...)` in a plain request script.

`set`/`unset`/`clear` are recorded there as settled, not as open work - they are bound-to-throw by design and allowlisted alongside `pm.variables.set`.

Also on this PR since you last looked: the first push failed the Windows engine build. `OPTIONAL` is an empty parameter-annotation macro in the Windows SDK's `winnt.h`, so my `static constexpr std::string_view OPTIONAL` expanded to a declaration with no variable name. Renamed to `OPTIONAL_SUFFIX` in 48800af. Ubuntu and macOS engine jobs are green on that commit; Windows is still building.

---

PR #442, opened 2026-08-09
feat(engine/scripting): bind pm.iterationData.has, widen the completions guard to depth 3, and declare an optional surface optional

### LINKED ISSUE #435: engine(scripting): the three things #418's guard deliberately does not cover - pm.iterationData.has, depth-3 members, and optionality in the generated .d.ts
Deferred out of #418 / PR #434, which added the `pm.iterationData` completion entries and closed the callable-implies-offered direction. Filed at the owner's request so the deviations that PR stated are tracked rather than living in its description.

Three separate things, in the order they are worth doing. Each is independent; none blocks the others.

## 1. `pm.iterationData.has` does not exist - decide parity or divergence

**Problem.** #418 asked for `get` / `has` / `toObject` entries. The runtime binds **`get` and `toObject` only** (`setup_pm_iteration_data`, `engine/src/runtime/script_engine.cpp:4807+`), so PR #434 offered those two and added a test asserting `pm.iterationData.has` is *not* offered - offering an unbound member is the #182 defect.

But Postman's `pm.iterationData` is a `VariableScope`, and `has()` is part of it. Every *other* vayu scope has one: `pm.environment.has`, `pm.globals.has`, `pm.collectionVariables.has`, `pm.variables.has`, `pm.cookies.has`. `pm.iterationData` is the only reader in the `pm` surface that answers `get` and not `has`, which reads as an oversight rather than a decision - and today the workaround is `pm.iterationData.get('col') !== undefined`, which cannot tell an absent column from a column whose value is JSON `null`.

**Fix / solution pathway - needs a decision.**

- **A - bind `has(key)`.** Matches Postman, matches every sibling scope, and gives the one answer `get` cannot: absent vs. present-and-null. Costs a C function beside `js_iteration_data_get`, the same `iteration_data_context` guard, one completion entry, and a line in `scripting.md`. **This is the one to beat.**
- **B - leave it and document the divergence.** Cheaper by one function, but `pm-api-compatibility.md` then has to carry a row explaining why one scope reader is missing the method the others have, which is more words than the function.

`toObject()` already exposes the whole row, so nothing is *unreachable* today - this is about the ergonomic and the null case, which is why it is an enhancement and not a bug.

**Out of scope:** `set` / `unset` / `clear`. They are bound and throw on purpose (the rows are a run input, not a scope - see #356), and PR #434 allowlisted them as deliberately-uncompletable alongside `pm.variables.set`. That is settled; do not revisit it here.

## 2. The new guard walks two levels, so depth-3 members can still go unoffered silently

**Problem.** `ScriptCompletions.EveryPmMemberTheRuntimeBindsIsOffered` (added in #434) enumerates `pm.<key>` and `pm.<key>.<member>` and requires a completion for each. That is the shape of the surfaces it was written for, but the API is three deep in four places:

- `pm.response.headers.get` / `.has`
- `pm.request.headers.get` / `.has` / and its mutators
- `pm.response.cookies.*`
- `pm.cookies.jar().*` - behind a *call*, so it is not reachable by property walking at all

The first three are excluded from the walk by `is_data_map` because those objects also carry one property per header/cookie on the wire, so descending would demand a completion entry for `Authorization`. All four are still covered in the *offered-implies-callable* direction by the older guards, and the jar's write half is pinned by name (#337) - so this is a narrowed hole, not an open one. But a `get()` added to `pm.response.headers` tomorrow, or a fifth method on the jar, is exactly the omission #418 set out to make impossible, and it would still fail nothing.

**Fix / solution pathway.** Separate the API members from the data entries at the boundary rather than skipping the object wholesale. The runtime already knows which is which: `install_header_methods` binds the accessors and `define_header_entry` defines the wire entries, so the accessors could be marked (non-enumerable, or recorded on a side table the test reads) and the walk could then descend into a header map and require exactly the marked members. For `pm.cookies.jar()`, the walk needs a call rather than a property read - a small named exception in the test, since there is exactly one such factory.

## 3. `pm.iterationData` is typed as always-present in the generated `.d.ts`

**Problem.** `pm.iterationData` is **`undefined`** outside a data-driven collection run - that is the documented, deliberate design (`ScriptContext::iteration_data`), and the completion documentation added in #434 says so in words. The generated declarations do not: `script_types.cpp` builds a nested object type from the dotted labels, so the editor sees

```typescript
iterationData: { get(column: string): any; toObject(): { [key: string]: any } };
```

and a plain request script writing `pm.iterationData.get('x')` gets no squiggle for something that throws at run time. `pm.execution` has the mirror-image version of this: it is always bound but its methods throw outside a collection run, which no type can express either.

**Fix / solution pathway.** Let an entry declare its member optional - a flag on the completion item (or a convention in `detail`) that `script_types.cpp` turns into `iterationData?: {...}`. With `strictNullChecks` the editor would then require the `pm.iterationData ? ... : ...` guard the docs already recommend. Worth checking first whether the app's TypeScript worker runs with `strictNullChecks` on; if it does not, the optional marker buys nothing and this reduces to a docs note.

## Engine changes

`engine/src/runtime/script_engine.cpp` and `engine/src/http/routes/scripting.cpp` for 1; `engine/tests/script_completions_test.cpp` plus whatever marking 2 needs in `script_engine.cpp`; `engine/src/http/routes/script_types.cpp` for 3. No app change for 1 or 2. For 3, the renderer sets the compiler options (`useScriptTypeDefinitions.ts`) so a `strictNullChecks` check belongs there.

## Tests

- 1: `script_engine_test.cpp` - `has` answers true for a present column, false for an absent one, **true for a column whose value is `null`** (the case `get` cannot express), and throws outside a data-driven run the way `get` does. Plus the completion entry, which the #434 guard then requires automatically. Mutation-checked.
- 2: the widened walk must redden when a bound accessor is unoffered, and must **not** demand an entry for a wire header name - assert both, or the widening trades one silent hole for a noisy false positive.
- 3: `script_types_test.cpp` - the optional member emits `iterationData?:` and a non-optional one does not.
- Commands: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`.

## Docs (same commit)

`docs/engine/scripting.md` for 1 (the `pm.iterationData` section's code block and the read-only note); `docs/app/pm-api-compatibility.md` for 1 either way - it gains a method or it gains a divergence row. Nothing for 2. For 3, only if the outcome is a documented limitation rather than a type.

## Sequencing

**After PR #434** (#418), which adds the guard, the entries and the `field_type` optional-suffix fix this builds on - none of it is on master until that merges. The three items are independent of each other and parallel-safe against everything else open; #411 also touches script-surface territory but a different file (`script_body_view`), so there is no overlap. Land whenever after #434.

## Acceptance criteria

- [ ] Item 1's decision is stated in the PR with its rejected alternative, and if `has` is bound it answers the present-but-null case distinguishably from absent.
- [ ] Item 2: a bound accessor on a header/cookie/jar object that is not offered reddens the suite, and a wire header name still demands no entry.
- [ ] Item 3: either the generated declarations express optionality, or the PR states why they cannot and the limitation is documented.
- [ ] Engine suite green, mutation-checked; docs updated same-commit.

athrvk:
## Description

**Plan.** The three items #418 deliberately left uncovered, all engine-side and independent of each other. The root cause common to 1 and 3 is that `pm.iterationData` was shipped as a partial copy of a shape the rest of the surface already has - it reads like the other scopes but answers one fewer question, and it is the only surface that can be *absent* while being declared as always present. Item 2 is the same class of defect one layer up: the guard added in PR #434 closed the callable-implies-offered direction only as deep as the surfaces it was written for.

For each I took the option that removes a special case rather than adding one, and the alternative I rejected is stated inline below.

**Verified against the code at `cc8b015` before editing.** `setup_pm_iteration_data` binds `get` / `toObject` / the three throwing writers and no `has`; the walk in `EveryPmMemberTheRuntimeBindsIsOffered` skips `is_data_map` objects wholesale; `render_member` emits `iterationData: {` unconditionally. Blast radius traced: the completion table has exactly two consumers, `GET /scripting/completions` (rendered by `useScriptCompletionProvider`) and `GET /scripting/types` (whose `.d.ts` is generated from the same table by `script_types.cpp`), plus the MCP `vayu://scripting/completions` resource which re-serves the endpoint. No renderer or MCP code hardcodes the list, so adding an entry reaches all three.

**One correction to the issue's premise, stated because it changes the argument and not the outcome.** The issue says the `get(col) !== undefined` workaround "cannot tell an absent column from a column whose value is JSON `null`". In vayu it can: `js_from_json` round-trips through `JS_ParseJSON`, so a null column arrives as JS `null` and an absent one as `undefined`. So the workaround is correct here - it is just non-obvious, and it requires the reader to know that fact. The case for `has` is therefore parity (five sibling scopes and Postman all have one) and directness, not reachability. A test now pins both halves of that null behaviour so the claim stops being folklore.

### 1. `pm.iterationData.has(column)` - option A, bind it

Chosen over **B, leave it and document the divergence**: `pm-api-compatibility.md` would have to carry a row explaining why one scope reader lacks the method the other five have, which is more words than the function, and every reader of that row would still want the method. `has` answers `true` for a present-but-null column - the row carries it - which is the fact `get` can only state if you already know how a null column comes back. No argument is `false`, matching `pm.variables.has`. Outside a data-driven run it throws through the same `iteration_data_context` guard `get` uses, naming itself in the message.

`set` / `unset` / `clear` are untouched and stay allowlisted as deliberately-uncompletable, as the issue instructed.

### 2. The guard now walks three levels

Rather than skipping the data maps, the walk descends into them and asks the runtime which members are API. **It does not need a new marking to do that** - the distinction is already in the object model, and load-bearing for a different reason: `install_header_methods` defines the accessors with `HEADER_METHOD_FLAGS` (configurable + writable, *not* enumerable) while `define_header_entry` defines wire entries with `JS_PROP_C_W_E`. So a non-enumerable own property that is a function is API and must be offered; everything else - a header, a cookie, an array index, `length` - is data and must not be. A wire header literally named `get` overwrites the method as an enumerable string and is classified as data, which is exactly what those flags exist to produce.

I rejected the issue's suggested side table of marked accessors for the reason PR #434 rejected the static name table: a second record of what the runtime bound drifts from the binding calls, and keeping it in step is the same act of remembering that failed in the first place.

`pm.cookies.jar()` is behind a call and unreachable by property reads, so it is named explicitly - there is exactly one such factory, and a general mechanism for discovering factories would be more machinery than the one case is worth.

**The widening found no new omissions** - all fourteen depth-3 accessors are already offered. That is the expected result: this closes the direction, it does not fix an instance.

### 3. Optionality in the generated `.d.ts`

`pm.iterationData` is `undefined` outside a data-driven collection run, and the declarations said otherwise. The table already had a way to say "may be absent" - a `detail` ending in ` | undefined`, which `field_type` reads for leaves - so this extends that one convention instead of adding a second: the suffix detection moves into `split_optional_suffix`, used by both readers, and a node that has children renders `iterationData?: { ... }` because an object type has no union to carry the suffix in.

**The `strictNullChecks` check the issue asked for, and its outcome.** `useScriptTypeDefinitions` spreads Monaco's defaults and sets `target` / `lib` / `allowJs` / `checkJs` / `noEmit`; neither `strict` nor `strictNullChecks` is set, and Monaco does not default them on. So the marker shows in hover but produces no squiggle today. I did **not** turn `strictNullChecks` on in this PR: it would apply to every `T | undefined` in the surface at once - `pm.environment.get`, `pm.response.errorCode`, `pm.info.iteration` - so `pm.environment.get('token').trim()` would newly error. That is a decision about how noisy this editor should be for all script authors, and the file it lives in explicitly warns against trading false positives for coverage. It is filed separately as #443 rather than folded in here. The limitation is documented alongside the marker.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`

**1162/1162 passed**, including the 4 tests added here. No failures anywhere in the suite.

Tests added:

- `ScriptEngineTest.IterationDataHasAnswersPresenceIncludingANullColumn` - true for a present column, false for an absent one, **true for a column whose value is `null`**, and `get` on those same two columns answers `null` vs `undefined` so the two methods are pinned as agreeing. No argument is `false`.
- `ScriptEngineTest.IterationDataHasThrowsOutsideADataDrivenRun` - a stashed `pm.iterationData` called from a later script in the pooled context refuses, and the message names `pm.iterationData.has`.
- `ScriptCompletions.TheWalkSeparatesDepthThreeAccessorsFromWireEntries` - both halves of the widening, since it trades a silent hole for a noisy false positive if either fails: all fourteen depth-3 accessors are *reached and required*, and five wire entries the fixture puts on the wire are classified as data and demanded of nobody.
- `ScriptTypesTest.AnOptionalSurfaceIsDeclaredOptional` - `iterationData?: {` is emitted, and six always-bound surfaces keep their required declarations so the marker is not sprayed over everything.

`ScriptCompletions.TheIterationDataAccessorsAreOffered` flips: it asserted `pm.iterationData.has` is *not* offered (correct while the runtime lacked it) and now requires it alongside the others.

**Mutation checks, all four applied, confirmed red, reverted:**

| Mutation | Result |
|---|---|
| Unbind `has`, keep the completion entry | both `IterationDataHas*` tests fail |
| Stop the walk at two levels | `TheWalkSeparatesDepthThreeAccessorsFromWireEntries` fails - "nothing was classified as wire data, so the split below proves nothing" |
| Descend into data maps indiscriminately | that test fails *and* `EveryPmMemberTheRuntimeBindsIsOffered` fails naming `pm.request.headers.Authorization`, `pm.response.headers.content-type`, `pm.response.cookies.length` - the false positive the classification exists to prevent |
| Drop the `?` from the optional render | `AnOptionalSurfaceIsDeclaredOptional` fails |

Generated declarations verified against a running engine (`GET /scripting/types`): `iterationData?: {` is the only property in the file that gained a `?`, and it contains `get(column: string): any;`, `has(column: string): boolean;`, `toObject(): object;`.

No app changes. Item 3's app-side work was the `strictNullChecks` *check* the issue asked for, which is a read of `useScriptTypeDefinitions.ts`; its outcome is documented rather than coded. `mkdocs build --strict` could not be run (mkdocs is not installed in this environment) - the docs edits add no link and no nav entry; the one new heading has no inbound link.

## Checklist

`script_types.cpp` was clang-format clean before and is formatted. `script_engine.cpp`, `script_completions_test.cpp` and `script_engine_test.cpp` were already not clang-format clean on master, so per CLAUDE.md they were left alone and the added code matches each file's prevailing style. `scripting.cpp` and `script_types_test.cpp` were clean and remain clean.

Docs updated same-commit: `docs/engine/scripting.md` (the `pm.iterationData` code block and the `has` semantics including the null column), `docs/app/pm-api-compatibility.md` (the summary-table row, the code block, and a new subsection under Type declarations covering the optional-member convention and the `strictNullChecks` limitation).

## Related Issues
Closes #435

Follow-up filed: #443 (whether the script editor should run with `strictNullChecks`, which is what would make item 3's marker produce a diagnostic rather than only hover text).

### DIFF
--- a/docs/app/pm-api-compatibility.md
+++ b/docs/app/pm-api-compatibility.md
@@ -32,7 +32,7 @@ intent is that the most common Postman scripts paste in and run unchanged.
 | Crypto              | `pm.crypto.sha256(data, encoding?)`, `.hmacSha256(key, data, encoding?)` - synchronous, see below |
 | Send from script    | `pm.sendRequest(urlOrOptions, callback)` - synchronous, callback only, refused for agent-started runs, see below |
 | Flow control        | `pm.execution.setNextRequest(name \| null)`, `.skipRequest()` - collection runs only, see below |
-| Data rows           | `pm.iterationData.get(name)`, `.toObject()` - read-only, data-driven collection runs only, see below |
+| Data rows           | `pm.iterationData.get(name)`, `.has(name)`, `.toObject()` - read-only, data-driven collection runs only, see below |
 | Base64              | `btoa(binaryString)`, `atob(base64)` - globals, standard web semantics           |
 | Console             | `console.log/info/warn/error`                                                    |
 
@@ -398,6 +398,7 @@ Details, including every case that throws, are in
 
 ```javascript
 pm.iterationData.get('username');  // this iteration's value for that column
+pm.iterationData.has('coupon');    // whether the row carries that column
 pm.iterationData.toObject();       // the whole row
 ```
 
@@ -614,6 +615,28 @@ entry - it caught `queueMicrotask`, which quickjs-ng does provide, on its first 
 
 Narrow that suppression list rather than widening it: each code on it is a real mistake
 going unreported in exchange for not crying wolf on correct code.
+
+#### Optional members, and the one thing the editor cannot yet tell you
+
+A completion entry says a member may be absent by ending its `detail` in
+` | undefined`. The generator reads that one convention in two places: a leaf keeps the
+union (`iteration: number | undefined`), and a surface that has members of its own has no
+union to carry it, so the optionality moves onto the property name -
+`pm.iterationData` is declared **`iterationData?: { ... }`**, because it is `undefined`
+outside a data-driven collection run.
+
+That marker shows in hover today, but it does **not** produce a squiggle: the worker runs
+with `strictNullChecks` off, and without it an optional property is indistinguishable from
+a required one at a use site. So a plain request script calling
+`pm.iterationData.get('x')` still gets no editor error for something that throws at run
+time - the guard the docs recommend (`pm.iterationData ? ... : ...`) is advice, not a rule
+the type system enforces. Turning `strictNullChecks` on would change that for this member
+*and* for every `T | undefined` in the surface at once (`pm.environment.get` among them),
+which is a decision about how noisy this editor should be rather than a detail of the
+declarations - so it is tracked separately, not folded in here.
+
+`pm.execution` has the mirror-image limitation and no fix at all: it is always bound, and
+its methods throw outside a collection run, which no type can express.
 
 ### The second consumer: MCP agents
 
--- a/docs/engine/scripting.md
+++ b/docs/engine/scripting.md
@@ -803,6 +803,7 @@ reads:
 
 ```javascript
 pm.iterationData.get('username');  // this iteration's value for that column
+pm.iterationData.has('coupon');    // whether the row carries that column
 pm.iterationData.toObject();       // the whole row as a plain object
 ```
 
@@ -810,6 +811,11 @@ pm.iterationData.toObject();       // the whole row as a plain object
 `pm` scope reader does. Values keep their JSON type: a JSON file's `3` arrives
 as a number, and a CSV column arrives as a string, because that is what the
 file said.
+
+`has` answers presence, the same way `pm.environment.has` and its siblings do.
+A column whose value is `null` is `true` - the row carries it - which is the
+fact `get` alone cannot state without the reader knowing that an absent column
+comes back as `undefined` while a null one comes back as `null`.
 
 **`pm.iterationData` is `undefined` where there is no row** - a single Send, a
 load run's deferred `tests` script, and a collection run started without a data
--- a/engine/src/http/routes/script_types.cpp
+++ b/engine/src/http/routes/script_types.cpp
@@ -37,6 +37,7 @@
 #include <map>
 #include <string>
 #include <string_view>
+#include <utility>
 #include <vector>
 
 #include "vayu/http/routes.hpp"
@@ -144,6 +145,9 @@ struct TypeNode {
     std::string documentation;
     int kind    = 0;
     bool listed = false; // false for an interior node the labels only imply
+    // The member may be absent at run time - see split_optional_suffix. Only
+    // read for a node that has children; a leaf carries it in its own type.
+    bool optional = false;
     // Set only where the table cannot say it: `pm.expect` is the entry point
     // into the chain, and its `detail` documents no return type because the
     // completion popup has nothing useful to show there.
@@ -172,6 +176,33 @@ std::string trim (const std::string& s) {
     }
     const auto last = s.find_last_not_of (" \t\n");
     return s.substr (first, last - first + 1);
+}
+
+constexpr std::string_view OPTIONAL_SUFFIX = " | undefined";
+
+/**
+ * @brief The table's one way of saying "this member may not be there".
+ *
+ * A `detail` ending in ` | undefined` marks the member optional, with the same
+ * spelling for a leaf (`pm.info.iteration` is `number | undefined`) and for a
+ * whole surface (`pm.iterationData` is `undefined` outside a data-driven
+ * collection run). Only the rendering differs: a leaf keeps the union, and a
+ * surface that has members has no union to put it in, so it becomes a `?` on
+ * the property name.
+ *
+ * One reader of the convention rather than two, so `field_type` and
+ * `render_member` cannot come to disagree about what the suffix means.
+ *
+ * @return the detail with the suffix removed, and whether it was there.
+ */
+std::pair<std::string, bool> split_optional_suffix (const std::string& detail) {
+    const std::string base = trim (detail);
+    if (base.size () > OPTIONAL_SUFFIX.size () &&
+    base.compare (base.size () - OPTIONAL_SUFFIX.size (),
+    OPTIONAL_SUFFIX.size (), OPTIONAL_SUFFIX) == 0) {
+        return { trim (base.substr (0, base.size () - OPTIONAL_SUFFIX.size ())), true };
+    }
+    return { base, false };
 }
 
 /**
@@ -312,14 +343,8 @@ std::string field_type (const std::string& detail) {
     // undefined` and not `number | undefined`, which silently declared
     // `pm.info.iteration` as `void` and made `pm.info.iteration + 1` an error
     // in the editor.
-    static constexpr std::string_view OPTIONAL_SUFFIX = " | undefined";
-    bool optional                                     = false;
-    if (base.size () > OPTIONAL_SUFFIX.size () &&
-    base.compare (base.size () - OPTIONAL_SUFFIX.size (),
-    OPTIONAL_SUFFIX.size (), OPTIONAL_SUFFIX) == 0) {
-        optional = true;
-        base = trim (base.substr (0, base.size () - OPTIONAL_SUFFIX.size ()));
-    }
+    const auto [stripped, optional] = split_optional_suffix (base);
+    base                            = stripped;
 
     std::string resolved;
     if (base == "number" || base == "string" || base == "boolean" || base == "any") {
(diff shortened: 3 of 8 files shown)


athrvk:
Claiming this - master moved to `8bbc4ac` (#440, #441 merged) and this branch is behind, so `mergeable_state` is `behind`. Merging master in and re-running the engine suite this run.

---

athrvk:
Done - master merged in (`a04a096`), no conflicts and no semantic overlap: the two new commits (#440, #441) touch `db/database`, `http/routes/runs`, `engine/vcpkg.json` and app code, and nothing under `engine/src/scripting` or `script_completions` / `script_types`.

Rebuilt and re-ran against the merged tree rather than trusting the pre-merge run, since master changed `vcpkg.json` and `CMakeLists.txt`:

- `python build.py -e -t` - clean
- `ctest --preset linux-dev` - **1169/1169 passed, 0 failed** (up from 1162 as master added 7 `runs_route_test` cases)

CI is green on the merge commit across all three platforms - Engine on ubuntu/macos/windows plus C++ Tests on each, and the CI gate. `mergeable_state` moved from `behind` to `blocked`, which is now just the review requirement.

---

PR #448, opened 2026-08-09
feat(app/graphql): schema explorer beside the editor - browse, search and insert typed operation skeletons

### LINKED ISSUE #387: app(graphql): schema explorer - browse the endpoint's queries, mutations and types beside the editor, insert-at-cursor with argument and variable skeletons
**Owner direction, 2026-08-08**: show the queries/mutations/types the GraphQL API exposes, insertable into the body editor - "similar to how Postman shows it, but in a more improved manner, more user friendly, maybe context bar could be useful here? or rethink the UI/UX". Design grounded in the 2026-08-08 GraphQL sweep (4 readers + verification + critic at `93a2f93`); every fork below is decided with rationale.

## The UX decision: beside the editor, not in the context bar

**Primary surface: a toggleable explorer pane docked inside the GraphQL body editor area** - a split beside the query pane, GraphiQL-style adjacency. Rationale: the context bar is 220-480px wide (the layout-store clamps) - a schema tree plus docs plus search in that width is exactly Postman's cramped-sidebar failure mode, and insert-at-cursor wants to live next to the cursor. What Postman gets wrong and this design fixes: their explorer is a flat list far from the editor, insert ignores the cursor and existing selections, arguments arrive without types, and there is no path from a type you are reading to a runnable operation. Vayu's explorer is a tree with docs inline, cursor-aware insertion, and typed skeletons.

**Secondary surface (phase 2): a compact "GraphQL" context-bar section** on #344's shipped registry - schema status, freshness, refresh, endpoint, and an outline of the *current document's* operations (click scrolls the editor). At-a-glance status belongs in the bar; browsing does not.

## The explorer (phase 1)

- **Tree**: Query / Mutation / Subscription / Types branches, fields with argument signatures and return types, deprecations struck-through with the reason, descriptions inline (plain text in v1 unless a markdown renderer already exists in the dep tree - GraphQL descriptions are spec-markdown; state the choice in the PR). Search filters across fields and types with match highlighting; the tree is virtualized (GitHub-scale schemas are the sizing case).
- **Insert semantics (the decided contract)**:
  - A root field inserts a full operation skeleton: operation keyword, required arguments as `$variables` with their types declared in the signature, nested required-argument fields stubbed, scalar leaves selected, object leaves left as `{ }` with the cursor inside.
  - A nested field inserts into the current selection set when the cursor sits inside a compatible parent type; otherwise it builds the enclosing path from the nearest valid ancestor. Never silently produces an invalid document - if no valid insertion point exists, append a new operation.
  - A type inserts a fragment definition on that type.
  - **Variables pane merge policy**: argument defaults are written into the variables pane only when its text parses as strict JSON (merge, never replace); otherwise only the `$var` declarations go into the operation and the pane gets a badge ("2 variables need values") - never touch unparseable draft text (the draft-preservation rule #384 items 3/5 establish).
- **Subscriptions are shown but not insertable**, grouped under "Subscription (not executable)" with a tooltip stating the engine's one-shot HTTP transport - honesty over hiding; transport is out of scope.
- **Introspection surface unchanged**: the explorer renders whatever the schema cache holds and adds NO new introspection triggers beyond the existing body-tab lifecycle plus its own explicit refresh button (#382 makes those side-effect-free; the consent budget stays where it is).
- **Freshness**: the explorer header reuses #383's fetchedAt affordance and keep-last-good behavior (a stale-but-present schema browses with a "schema from Xm ago" notice, not a blank pane).
- **State lifetime**: search text, expansion, and scroll live in a store keyed by schema identity - not component state - so a Radix tab glance does not reset them (the body-drafts lesson, verified as a real trap by the critic).
- **Keyboard/AT**: the tree follows the WAI-ARIA tree pattern (the roving-focus + typeahead precedents from #362 apply; reuse the pattern, not necessarily the collections hook), `/` focuses search, Enter inserts at cursor, insertions announced via a polite live region.

## Implementation shape

Pure modules first (all node-env testable against the shared schema fixture #384 introduces): `schema-tree.ts` (introspection -> tree model + search index), `insert-skeleton.ts` (field/type + cursor context -> document edit: text, range, cursor). The pane component consumes both plus the schema cache; Monaco insertion via `executeEdits` with cursor placement. No engine work anywhere in this issue.

## App changes

New `app/src/modules/request-builder/components/RequestTabs/panels/graphql-explorer/` (or beside GraphQLBody - follow `docs/app/COMPONENTS.md` and update it), the two pure modules under `app/src/lib/graphql/`, a small store slice. Phase 2 adds the context-bar section on the shipped registry. MCP: none.

## Tests

- `schema-tree.test.ts` / `insert-skeleton.test.ts` (node env, fixture-driven): tree shape incl. deprecations and all type kinds; skeleton generation for every insert case above (root/nested/type/subscription-refused), argument and variable emission, invalid-cursor fallback; search index.
- jsdom: insert round-trip into a Monaco stub (edit + cursor asserted), variables-pane merge policy (strict-JSON merge vs badge, mutation-checked - the never-touch-drafts case especially), state survives a tab glance, virtualization renders a bounded row count, keyboard flow, live-region announcement.

## Docs (same commit)

`docs/app/COMPONENTS.md` (the new module), the app's GraphQL user-facing doc if one exists under `docs/` (verify; add a short page via the docs-site rules if the feature warrants it - it does, this is a headline feature), `docs/app/architecture.md` one paragraph (explorer consumes the schema cache read-only).

## Sequencing

All prerequisites landed: #382 (df0aad3), #383 (0f7d897), #384 (PR #407, including the shared fixture at app/src/test/graphql-schema-fixture.ts), and PR #376 with the phase-2 registry sections (1905170) - both phases are unblocked. Parallel-safe against #381, #385, #386, the tree track (#366/#367), the scenario phases, and the perf pair.

Out of scope: subscription transport; introspection consent changes; markdown renderer adoption as its own decision; MCP exposure of the schema.

## Acceptance criteria

- [ ] For a fixture schema with all type kinds: every branch browses with docs and deprecations; search filters; the tree virtualizes.
- [ ] Every insert case produces a valid document with typed variables, pinned by pure-module tests; the variables pane is never corrupted (merge-or-badge, never replace drafts).
- [ ] Subscriptions visible, refused with the stated reason.
- [ ] Explorer state survives tab glances; keyboard-only flow works end to end with announcements.
- [ ] Phase 2's context-bar section ships on the #344 registry (merged) with status + outline only.
- [ ] All tests green, mutation-checked; docs same-commit.

---

athrvk:
## Description

The GraphQL schema explorer: a WAI-ARIA tree over the endpoint's Query / Mutation / Subscription / Types, docked beside the query editor, with cursor-aware insertion. Plus phase 2's context-bar section.

**The plan.** The root cause is that Vayu introspects a schema and then shows it to nobody - the schema reaches Monaco's completion popup and hover, and there is no way to *read* the API. The approach is two pure modules under `lib/graphql/` plus a pane that consumes them, because both halves have decisions worth pinning without a DOM: `schema-tree.ts` computes children on demand per expanded row, and `insert-skeleton.ts` turns a row into a document edit. **The alternative I rejected was materialising the tree up front.** A schema is a graph, not a tree - `User.posts` leads to `Post`, whose `author` leads back to `User` - so eager expansion either recurses forever or needs a visited-set whose pruning decides for the user which branch they may open. Asking for one row's children when it expands has neither problem and keeps a GitHub-scale schema off the render path. The same reasoning drives the second decision: a field row carries the `rootPath` it was reached by (`[Query.user, User.posts, Post.title]`) rather than the insertion module searching the schema graph for a route, because the graph has many answers or none and the user already picked one by expanding rows.

**The contract everything else follows from: the document still parses afterwards.**

- A field lands in the cursor's selection set only when that set's type *owns* it - no inline-fragment guessing on an interface or union, since "probably compatible" is how an invalid document gets written on the user's behalf. Failing that, an enclosing set; failing that, a new named operation.
- Required arguments become `$variables` **and are declared on the operation that gained them**, including promoting a shorthand `{ … }` operation to `query ( … )` - which otherwise cannot carry variables at all. A `$id` nobody declared is a document the server rejects, produced by a click rather than typed.
- A row with nowhere valid to go (a `Types`-branch field with no compatible cursor, a subscription, an enum value) is **refused out loud** through a live region, never silently dropped.

**The Variables pane is merged into, never replaced,** and only when its text is strict JSON. A pane holding `{"id": {{userId}}}` is a working template that `JSON.parse` rejects; rewriting it would delete the draft to make room for a placeholder. Existing keys keep their values. What could not be written comes back as `pending` and becomes a badge naming it.

Explorer state (search, expansion, scroll) lives in a store keyed by schema identity, not component state - Radix unmounts the whole Body tab on every glance at Headers, the same trap the body drafts exist for.

### Deviations from the issue spec

Two, both forced by the parse contract, plus one deferral:

1. **Object leaves get their scalar fields, not the empty `{ }` the issue sketched.** An empty selection set is a syntax error, so the sketch and "never silently produces an invalid document" disagreed; the contract wins. The cursor still lands inside the braces where the sketch wanted it. Deprecated fields are browsable but never selected for you.
2. **Descriptions render as plain text, not markdown.** The issue said "plain text in v1 unless a markdown renderer already exists in the dep tree" - one does (`MarkdownView`), but its allow-list is *block* elements and a tree row is one line high, so it cannot render into one. The full text is in the row's tooltip.
3. **Click-to-scroll from the context-bar outline is deferred.** It needs an editor-focus channel across the module boundary that does not exist; the acceptance criterion for phase 2 is "status + outline only", which ships. Happy to file it as a follow-up if you want it.

Two smaller notes: `useRovingTreeFocus` is imported from `modules/collections/` rather than copied, since it is generic over `[role="treeitem"]` and its data attributes - a hand-rolled copy would not receive the primitive's fixes. And the bar's `graphql` section applies to every request tab and says "does not send a GraphQL body" for the rest, because `appliesTo` sees only the tab, which carries no body mode - the same shape `cookies` already has for a host with no cookies.

## Type of Change
- [x] New feature
- [x] Documentation update

## Testing

All app-side; no engine change, so the engine suite was not run (nothing it covers can change colour).

- `cd app && pnpm test` - **3589 passed, 361 files, 0 failed**. 99 of those are new across 6 files: `insert-skeleton.test.ts` (25), `schema-tree.test.ts` (24), `SchemaExplorer.test.tsx` (19), `GraphQLBody.explorer.test.tsx` (17), `GraphQLSection.test.tsx` (12), `explorer-store.test.ts` (4).
- `pnpm type-check`, `pnpm lint`, `pnpm format:check` - all clean.
- `mkdocs build --strict -f .github/mkdocs.yml` - clean, with the new page in the nav.

No pre-existing failures on the base commit.

What the tests actually pin, mutation-check style:

- Every insertion is run through `graphql.validate` against the fixture schema, not just `parse` - which is what catches an undeclared `$variable` rather than a syntax slip.
- The never-touch-drafts case: revert the strict-JSON guard in `mergeVariables` and the templated pane is replaced by `{"term": ""}`, and the test reddens.
- The caret round trip goes through a Monaco stub whose model computes offsets against *whatever the component last rendered*, so an offset into the new document only lands if the component waited for it - inline placement fails it.
- The bounded row count stubs `IntersectionObserver`, because jsdom has none and `useGrowingWindow` honestly degrades to rendering everything - without the stub the test would measure the fallback instead of the window.
- Keyboard flow (arrows, Right to expand, Enter to insert, `/` to search) and the live-region announcement are asserted on the rendered tree.

## Checklist

## Related Issues
Closes #387

---
_Generated by [Claude Code](https://claude.ai/code/session_01E94MRXaS7ibv7qTYQUUDaK)_


athrvk:
fix merge conflict

athrvk:
Conflicts resolved, branch is current with `master` (`6a963b4`).

Two merges were needed - `#442`/`#444`/`#445`/`#447` landed while this was open, then `#451`. Only the first had conflicts, both in files `#444` also touched:

- **`GraphQLBody.tsx` imports** - additive on both sides, kept both. `#444`'s `attachVariablesDiagnostics` (the Variables pane's masked-twin markers) and this branch's explorer imports are independent, and the merged mount handler keeps `#444`'s twin lifecycle alongside this branch's query-editor ref. Nothing here was a semantic collision: `#444` owns the *variables* pane's markers, this owns the *query* pane's insertion.
- **`docs/app/COMPONENTS.md`'s `lib/graphql` table** - kept this branch's three new rows (`schema-tree.ts`, `insert-skeleton.ts`, `explorer-store.ts`) plus `#444`'s rewritten `variables-schema.ts` row and its new `variables-diagnostics.ts` row. This branch's copy of the `variables-schema.ts` row was the pre-`#444` text and was dropped in favour of theirs.

One real fix fell out of the merge rather than the textual conflict: mounting the variables editor now creates a masked twin model, and this PR's explorer test was the only place that mounts it with a stub Monaco. The stub gained the `Uri` and `editor` model registry `attachVariablesDiagnostics` needs - without it all 17 cases in that file failed on a merge that `git` called clean, which is exactly the kind of conflict a diff cannot show.

Re-verified on the merged tree, not on the pre-merge run:

- `pnpm test` - **3692 passed, 373 files, 0 failed**
- `mkdocs build --strict` - clean

CI is green across ubuntu / macOS / windows on the merge commit. The engine changes now in the branch came from `master` and are unmodified here, so the engine suite was not re-run.

---

Issue #455, opened 2026-08-10
app(graphql): body panel layout rethink - schema controls crowd the Query header, refresh exists twice, and the Variables pane idles at 35%


athrvk:
**Owner direction, 2026-08-10**: "this graphql body section can be rethought from a UI/UX perspective and its functionality, current UI has slightly non intuitive button placements, there is empty space which could be useful". Grounded in the shipped #387/#408 surface, verified at master `054059a` with a live screenshot review. This is a layout/interaction pass over an already-working feature - nothing here is broken, and the editors' pane structure (the Monaco-identity and min-h-0 lessons recorded in `GraphQLBody.tsx`) must survive the rethink.

## Problem - the specific non-intuitive placements, verified

1. **The Query pane's header hosts four controls, and only one is about the query.** `GraphQLBody.tsx:504-578` puts the operation Select, the `SchemaStatusBadge`, the schema Refresh button, and the explorer toggle all in the Query `PaneHeader`. Three of the four describe the *schema*, whose pane is a sibling. The code's own comment (`:552-555`, "the toggle sits beside the schema badge because it is the same subject") is internally consistent - but the conclusion should have been that the *cluster* is on the wrong pane, not that the toggle should join it.
2. **Refresh exists twice, visible simultaneously.** The explorer header has its own refresh (`graphql-explorer/SchemaExplorer.tsx:163`, beside the search box) and the Query header has another (`GraphQLBody.tsx:535-540`); both call the same `refresh` (`:438-440` / the `onRefresh` prop). With the explorer open - the state the screenshot shows - two identical 12px spin-icons on one screen do the same thing from different corners, and nothing says whether they differ.
3. **Schema status is stated twice with different vocabularies.** The Query header shows the `SchemaStatusBadge` (`:120`), while the explorer header carries the #383 freshness affordance (`fetchedAt`). Same subject, two surfaces, split across panes.
4. **The Variables pane reserves 35% for what is usually a few lines.** `GraphQLBody.tsx:605` gives it `defaultSize="35%" minSize="15%"` with no collapse; a `{ "code": "IN" }` document leaves the region overwhelmingly empty - the owner's "empty space which could be useful". The pane header holds only the two badges (`:608-612`).
[...]

PR #458, opened 2026-08-10
feat(app/graphql): explorer search marks its matches, spans descriptions, and shows them on demand

### LINKED ISSUE #452: app(graphql): explorer search computes matchStart "for highlighting" and the pane drops it - search results render unhighlighted
From the 2026-08-10 closure verification of #387 (verified at master `054059a`). The #387 spec said "Search filters across fields and types with match highlighting"; filtering shipped, highlighting did not - and the halfway state is the repo's signature defect class, written-but-never-read: the search function computes and documents where the match sits, and the one consumer throws it away. #387's acceptance checkbox only promised "search filters", so its closure stands; this is the follow-up. Cheap polish, not urgent.

## Problem

- `searchSchema` returns `SchemaSearchMatch` whose `matchStart` field is documented "Where the term matched in `node.name`, for highlighting. -1 for no match" (`app/src/lib/graphql/schema-tree.ts:333-336`), computed at `:426-428` (name matches carry the real index, signature-only matches carry -1 and rank below - that split already works).
- The sole consumer maps it away: `SchemaExplorer.tsx:96` does `searchSchema(searchIndex, search).map((m) => ({ node: m.node, depth: 0 }))` (`app/src/modules/request-builder/components/RequestTabs/panels/body/graphql-explorer/SchemaExplorer.tsx`). Verified: zero other readers of `matchStart` (`rg matchStart app/src` hits only `schema-tree.ts` and its test).
- `ExplorerRow` renders `node.name` as plain text; no highlight styling exists anywhere in the explorer directory.

So a search for `post` lists `Post`, `Query.post` and every field returning one, with nothing marking why each row matched - exactly the flat-list failure #387's UX section was written against.

## Fix / solution pathway

1. Thread `matchStart` (and the needle length, i.e. the search term) from `SchemaExplorer.tsx:96` into the row model instead of dropping it - the search-mode rows already diverge from tree rows (`depth: 0`), so widening that shape is local.
2. In `ExplorerRow`, when a row carries `matchStart >= 0`, split `node.name` into before/match/after and wrap the match in the design system's existing emphasis treatment - check for an existing search-highlight primitive first per the repo rule ("a hand-rolled copy of a primitive does not receive the primitive's fixes"): `rg -i "highlight" app/src/components` before styling by hand. A `matchStart` of -1 (signature-only match) renders unchanged - that asymmetry is informative, not a bug.
3. Keep the pure module pure: the split arithmetic belongs in `schema-tree.ts` (a small `splitAtMatch(name, matchStart, length)` beside `searchSchema`) so it is node-env testable; the component only renders the three segments.

## App changes

`app/src/lib/graphql/schema-tree.ts`, `graphql-explorer/SchemaExplorer.tsx` + `ExplorerRow`. Engine: none. MCP: none, verified - the explorer is renderer-only.

## Tests

- `schema-tree.test.ts` (node env): the split helper on start/middle/end matches, unicode names, and the -1 passthrough - mutation-check by returning the name unsplit.
- The explorer jsdom suite (`GraphQLBody.explorer.test.tsx` or the SchemaExplorer test file): a search renders the matched segment inside the emphasis element and a signature-only match renders plain; assert on the accessible text staying identical (highlighting must not break the row's name for AT).

## Docs (same commit)

None, verified - `docs/app/COMPONENTS.md` describes the explorer's search as filtering and does not claim highlighting; add the word once it exists if that paragraph is touched anyway.

## Sequencing

Parallel-safe against everything open (#134, #197, #287, #348, #443, #446, #449, #450, and the sibling follow-ups filed today - the outline click-to-scroll issue touches `GraphQLSection.tsx`/`GraphQLBody.tsx`, disjoint from these two files).

## Acceptance criteria

- [ ] A name-matched search row visually marks the matched substring; signature-only matches render unchanged.
- [ ] `matchStart` has a real reader (verified by the new tests, mutation-checked); the split logic lives in the pure module.
- [ ] Accessible row text is unchanged by the highlight; app suite, type-check, lint, format green.

---

athrvk:
## Description

Started as #452 - `searchSchema` computed `matchStart`, documented it "for highlighting", and its one consumer mapped it straight off the row, so search results said nothing about why each row was in the list. Extended on your review to cover descriptions and an on-demand show/hide.

**Plan.** Three commits' worth of one idea: *the pane should say why a row matched, and let you read the thing that matched.*

1. **Highlighting (the original #452 fix).** Root cause was a dropped field, not a missing computation - the search already knew the offset, the row model never carried it. `matchStart` is threaded into the row model, and the index arithmetic goes into `splitAtMatch(name, matchStart, length)` beside `searchSchema`, so an off-by-one is testable without a DOM and the row only renders three segments.
2. **Descriptions are searched.** The index was `name + signature`, so a word the row was *displaying* returned "Nothing matches" - the pane showing text and then denying it exists. Now indexed.
3. **Descriptions can be read.** They were clipped to one line with the full text only in the native `title` tooltip, which is no use to anyone reading rather than pointing. A show/hide in the pane header lifts the clipping.

**The ranking is the substantive design call, not the concatenation.** Descriptions are prose - whole sentences on a documented schema - so a common word (`the`, `id`, `user`) matches most of the schema through them. Folded into the existing signature tier that flood drowns the type matches that tier exists to surface (the reason the name/signature split was there in the first place). So descriptions are a **third tier**: name, then signature, then description. The existing "ranks a name match above a signature-only match" test passes unchanged, and `IndexEntry` now keeps the three fields lowercased separately rather than as one haystack, since the ranking turns on *which* one matched - which also stops a keystroke re-lowercasing the whole schema.

**Two rejected alternatives.** A shared search-highlight primitive: I checked first per the repo rule (`rg` over `app/src/components`) - there is none, and `.variable-highlight` is the `{{var}}` colour, a different subject. With one consumer that would be speculative abstraction, so the mark is the sanctioned accent tint (`bg-primary/20`) inline, with `text-inherit` so a field name stays `text-primary` and a type name `text-foreground` through the tint (a bare `<mark>` also arrives with the UA's yellow-on-black). And a **per-row** description disclosure instead of a pane-level toggle: a third target inside a 24px row is exactly the composite-row hit-area trap `drawer-row-hit-area` was written against, and it would take its width from the activator - so the toggle is one control in the header, remembered per schema in `explorer-store` alongside search and expansion.

**One behaviour that is deliberately not the toggle's to decide:** a row matched *through its description* is always drawn in full. Clipped to one line the description is usually cut off before the word that put the row in the results, which leaves a row the user cannot connect to what they typed - the exact failure the mark exists to prevent.

Verified before writing: the #452 problem still existed at `5e6141f`, and `matchStart` had no other reader in `app/src`. Engine: none. MCP: none - the explorer is renderer-only. The new header control is consistent with #455's direction (schema affordances consolidating into the explorer header) and does not collide with it.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`cd app && pnpm test && pnpm type-check && pnpm lint` - all green, plus `prettier --write` limited to the six touched source files (which were clean before, per the repo's no-repo-wide-formatting rule).

- App suite: **376 files, 3744 tests passed**; 25 of those are new here (3719 before).
- `schema-tree.test.ts` (node env): `splitAtMatch` at start/middle/end, the -1 passthrough, empty term, a term running past the end of the name, an offset past the end, astral/accented names split on code units, and a loop asserting the segments always reassemble into the original name. Then description search - a word only in a description finds its row, both offsets reported when name and description each match, -1 for a node with no description, and a purpose-built schema where `zebra` reaches each tier exactly once so the order proves the tiers rather than the type map's iteration order.
- `explorer-store.test.ts`: the new field defaults off, toggles both ways, stays independent of search/expansion/scroll, and is kept apart per schema.
- `SchemaExplorer.test.tsx` (jsdom): name-match marking with the accessible row text unchanged; a signature-only match unmarked in the same render pass where the name-matched type is marked; a description-only match found and marked in the description with the name left alone; a description-matched row rendered unclipped; the toggle clipping and unclipping, reporting `aria-pressed`, adding nothing to an undocumented row, and remembered across an unmount and per schema.
- Mutation-checked, each restored and re-run green afterwards: dropping the description from the index (4 tests red), folding the description tier into the signature tier (the ordering test red), removing the forced-full on a description match (2 red), and forcing `matchStart={-1}` at the call site, i.e. the original #452 bug (2 red).

No pre-existing failures on the base commit. CI was green on the first commit before this one was pushed.

## Checklist

Docs in the same commits: `docs/app/COMPONENTS.md` (the explorer paragraph now states the three tiers, the forced-full behaviour and the header show/hide; the `schema-tree.ts` and `explorer-store.ts` rows follow), `docs/app/state-management.md` (the store's shape and method list), `docs/app/architecture.md` (the one-line store summary). No new design-system token, so `docs/design-system.md` is untouched.

## Related Issues

Closes #452. Leaves #460 open and narrowed to its remaining item - enum values and input-object fields are browsable in the tree but still absent from the search index, which is a separate gap from descriptions.


athrvk:
- does the search also searches in description ? 
- are we showing the description to the user (with hide/show capability)?

athrvk:
Checked both against the code rather than answering from memory. Short answers: **no** to the first, **yes but with no toggle** to the second. Neither is something this PR changed - both predate it - so I have filed #460 rather than widening this diff, and left this one to the one thing #452 asked for.

**1. No, search does not cover descriptions.** `entry()` builds the haystack as `` `${node.name} ${node.signature}` `` lowercased (`schema-tree.ts:384`), and `searchSchema` matches that and nothing else. So on the fixture's `search` field the row displays "Search across users and posts.", and typing `across` returns **Nothing matches** - the pane shows the text and then denies it exists.

While confirming that I found a second coverage gap in the same function: `buildSearchIndex` indexes types plus the fields of object/interface types only, so **enum values and input-object fields are browsable in the tree but unsearchable** - which is squarely against that function's own stated reason for existing ("a search has to see rows nobody expanded").

**2. Yes, descriptions are shown - inline, truncated, and with no hide/show.** `ExplorerRow` renders `- {node.description}` after the name and signature, `truncate`d to the row's single line in `text-muted-foreground font-sans`. The full untruncated text reaches the user only through the row's native `title` tooltip, joined there with the deprecation reason and the subscription note. There is no toggle, no density setting and no expand-in-place.

Those two interact, which is the part worth deciding before anyone writes the code, and it is why #460 is an issue rather than a one-line follow-up commit:

- **Ranking.** `searchSchema` has two tiers today (name, then everything else) and that split exists deliberately - without it, searching `Post` buries the `Post` type under every field that returns one. Descriptions are prose, so folding them into the existing second tier lets a common word pull in most of the schema and drown the signature matches that tier was built for. A third tier (name > signature > description) is the conservative answer.
- **Visibility.** This PR marks the matched run of the *name* and leaves `matchStart: -1` rows unmarked on purpose, so "unmarked" reads as "the name is not what matched". A description match would land in that same bucket **and** would often be truncated before the matched word - a row the user cannot connect to their term at all. The recommendation in #460 is to reuse `splitAtMatch` on the description span, which is exactly what this PR makes available.

If you would rather have the description search in here instead of in #460, say so and I will extend this branch - it is a bigger change than it sounds because of the ranking, but it is not a large one.

CI is green on this PR: App on ubuntu, macos and windows all passed, CI gate success.

---

athrvk:
add the search capability and allow the use to see the full desc on demand show/hide

athrvk:
On it - extending this branch with description search and an on-demand show/hide for the full description. Will report back when it is pushed and green.

---

athrvk:
Both landed in `f231993`. PR body and title updated to cover the widened scope.

**Search now spans descriptions**, ranked as a **third tier** rather than folded into the existing haystack - name, then signature, then description. That is the one real design call here: descriptions are prose, so a common word (`the`, `id`, `user`) matches most of the schema through them, and folded in with signatures that flood drowns the type matches the signature tier exists to surface. `IndexEntry` now keeps the three fields lowercased separately, since the ranking turns on which one matched - which also stops every keystroke re-lowercasing the whole schema. The existing name-above-signature test passes unchanged.

A match reports its offset in the description as well as the name, and both get marked, so a row that matches in both marks both.

**Descriptions can now be read in full**, via a show/hide in the pane header (the `Text` icon beside Refresh), remembered per schema alongside search and expansion. One pane-level control rather than a per-row disclosure: a third target inside a 24px row is the composite-row hit-area trap `drawer-row-hit-area` was written against, and it would take its width from the activator. It is also consistent with #455's direction of consolidating schema affordances into the explorer header.

One thing I decided rather than asked, flagging it because it is not what the toggle strictly implies: **a row matched through its description is always shown in full, whatever the toggle says.** Clipped to one line, the description is usually cut off *before* the word that put the row in the results - so the row arrives with a mark the user cannot see and no way to tell why it matched, which is the failure the marking exists to prevent. Say the word if you would rather the toggle be absolute and I will make it so.

Verification: **3744 tests green** (25 new), type-check, lint, `prettier --write` limited to the six files I touched. Mutation-checked each new behaviour and restored it: dropping descriptions from the index turns 4 red, folding the description tier into the signature tier turns the ordering test red, removing the forced-full on a description match turns 2 red.

#460 is retitled and narrowed to what is genuinely left over - enum values and input-object fields are still browsable in the tree but absent from the index, which is a separate gap from descriptions and untouched by this branch.

---

PR #482, opened 2026-08-10
feat(engine/app): per-run performance thresholds with a pass/fail verdict

### LINKED ISSUE #470: engine/app(loadtest): per-run performance thresholds - declare latency/error-rate/throughput budgets on POST /runs and report a pass/fail verdict beside testValidation
First of a three-part arc (thresholds → capacity discovery → baselines/diffing) toward making Vayu judge a load run, not just measure it. This one is the foundation the other two reuse.

## Problem

A load run measures everything and judges nothing. The report carries `latency{p50..p999}`, `summary.errorRate`, `summary.throughput` - but there is no way to declare "this run passes if p99 < 50ms and error rate < 0.1%" and get a verdict. The two pass/fail-shaped things that exist today both miss:

- **`testValidation`** (`validate_scripts`, `engine/src/core/run_manager.cpp:56-212`) replays the `pm.*` test script against sampled *individual responses*. A script sees one response at a time; it structurally cannot assert an aggregate like p99 or error rate.
- **The app's SLO** is a single client-only setting: `sloThresholdMs` in `app/src/stores/client-settings-store.ts:69` (default `DEFAULT_SLO_THRESHOLD_MS = 200`, `app/src/constants/client-settings.ts:59`), consumed by `computeBreakpoint` (`app/src/modules/dashboard/utils/computeBreakpoint.ts:34`). It never reaches the engine, is global rather than per-run, is not persisted with the run, and produces a chart annotation, not a verdict.

So "did this run meet its budget" is answered by a human reading numbers - which also means it cannot be answered by CI, by the MCP `get_run_report` consumer, or by the run-diff work planned next.

## Fix / solution pathway

Mirror the existing four-hop `testValidation` pattern exactly - payload field → evaluated post-run → optional summary section → optional report section. Every hop has a worked example in the tree.

**1. Payload.** Optional `thresholds` object on `POST /runs`:

```json
"thresholds": {
  "latencyP50Ms": 20,
  "latencyP95Ms": 40,
  "latencyP99Ms": 50,
  "maxErrorRatePct": 0.1,
  "minThroughputRps": 10000
}
```

All keys optional; at least one required for the object to be meaningful (empty object = reject). Validate in `validate_run_config` (`engine/src/http/routes/execution.cpp:434-532`, called at :802 before `create_run` - a rejected request must leave no trace). The `NumericRunField` table (:311-316, :486-524) is flat-key only, so `thresholds` needs its own small loop beside it: must be an object, keys drawn from the known set (unknown key = reject, name it), values finite and > 0 (`maxErrorRatePct` in [0, 100]). Reuse the existing message style and the `invalid_run_config` error code. The config snapshot already persists it for free (`sanitize_config_snapshot`, :895-898).

**2. Evaluate at end of run.** In `execute_load_test` (`engine/src/core/run_manager.cpp`), right beside the `validate_scripts` call at :784-790 and before the summary write. Add a pure function (own file pair, e.g. `engine/include/vayu/core/threshold_eval.hpp` + tests, following `load_pacing.hpp`'s pure-core precedent):

```cpp
struct ThresholdCheck { std::string metric; double limit; double actual; bool passed; };
struct ThresholdOutcome { std::vector<ThresholdCheck> checks; size_t passed; size_t failed; };
std::optional<ThresholdOutcome> evaluate_thresholds (const nlohmann::json& config,
                                                     const RunSummaryInputs& inputs);
```

Returns `nullopt` when the config carries no `thresholds` - that is what keeps the section absent rather than zeroed, same convention `validate_scripts` uses (:58-69). Evaluate off the same `RunSummaryInputs` the summary is built from (`engine/include/vayu/core/run_manager.hpp:362-390` - `latency` percentiles, `throughput`, and error rate derived from `total_requests`/errors) so the verdict and the reported numbers cannot disagree. A stopped-early run still evaluates against what it measured; note that in the report consumer docs rather than special-casing.

**3. Persist.** `RunSummaryInputs` gains `std::optional<ThresholdOutcome> thresholds` beside `std::optional<ScriptValidationTotals> tests` (run_manager.hpp:383). `build_run_summary_payload` (`run_manager.cpp:958-1013`) writes an optional snake_case `"thresholds"` section beside `"tests"` (:1002-1005): `{ "checks": [{"metric","limit","actual","passed"}], "passed": N, "failed": M }`. Run status stays `completed` - the verdict lives in the report, it does not fail the run.

**4. Report.** `apply_run_summary` (`engine/src/http/routes/runs.cpp:203-306`) + `ReportExtras` (:119-180) gain the read side (null-tolerant, like the existing `read_number` handling - rows written before this change have no section). `run_report_response` emits camelCase beside `testValidation` (:775-784):

```json
"thresholdValidation": {
  "checks": [{ "metric": "latencyP99Ms", "limit": 50, "actual": 47.2, "passed": true }],
  "passed": 4, "failed": 1, "verdict": "failed"
}
```

`verdict` is `"passed"` iff `failed == 0`. Section absent when no thresholds were declared.

## App changes

- **Types:** `LoadTestConfig` (`app/src/types/domain.ts:691-725`), `StartLoadTestRequest` (`app/src/types/api.ts:344-394`), `RunReport` (`domain.ts:869-1096`, new optional `thresholdValidation` beside `testValidation` at :976-981).
- **Dialog:** a new collapsible "Pass/fail budgets" section in `LoadTestConfigDialog` beside "Recording & limits" (`app/src/modules/request-builder/components/LoadTestConfigDialog/index.tsx:488-604`), using the local `NumberField` helper (:101), persisted in `SavedLoadTestConfig` (:61-73). Prefill the p99 field from `sloThresholdMs` so the existing client setting becomes the default budget instead of a parallel concept. Forward through `handleConfirmLoadTest` (`app/src/modules/request-builder/index.tsx:395-523`).
- **Rendering:** a verdict banner + per-check rows in `RequestResponseView` beside the "Test Validation" card (`app/src/modules/dashboard/components/RequestResponseView.tsx:221-257`), and in `LoadTestDetail`'s Overview tab (`app/src/modules/history/main/LoadTestDetail.tsx` - note it renders no `testValidation` today; add only `thresholdValidation` here, don't grow scope). Use `--status-*` three-token colors per the enforced token tests; no raw palette; reports without the section render exactly as today.
- **MCP:** `start_load_run` input schema (`app/electron/mcp/tools.ts:1738-1823`) gains `thresholds`; `get_run_report` returns the verdict for free. Update the tool table in `docs/engine/mcp.md:143-163` in the same commit.

## Tests

- Engine: `evaluate_thresholds` unit tests (each metric passes and fails; empty/absent config → `nullopt`; error-rate math matches `MetricsCollector::error_rate()` percent semantics). Validation tests in `engine/tests/run_config_validation_test.cpp` (drives `validate_run_config` directly). Round-trip in `engine/tests/runs_route_test.cpp` following `ReportReadsTheStoredSummary` (:642) and an absence twin following `ReportOmitsTestValidationWhenNoScriptRan` (:736). Mutation-check: flip a comparison in `evaluate_thresholds`, the pass/fail attribution test must fail, restore.
- App: dialog builds a payload carrying `thresholds`; verdict renders from a report with the section and is absent without it (mutation-check the preference); `pnpm type-check` clean.

## Docs (same commit)

`docs/engine/api-reference.md` - `POST /runs` payload (:1637-1690) and the report shape (:2596-2740). `docs/engine/db-schema.md` - the `summary` column semantics (:230-264). `docs/engine/mcp.md` tool table. App-side: `docs/app/api-integration.md` if the request/report types are documented there.

## Sequencing

Independent of everything open. #450 works in the same `execute_load_test` region (the `validate_scripts` call site) - whichever lands second rebases trivially, the two sections are siblings. The capacity-discovery mode (filed next) reuses this issue's SLO field naming, and the CI gate issue consumes `thresholdValidation.verdict` - land this first.

## Acceptance criteria

- [ ] `POST /runs` accepts and validates an optional `thresholds` object; a bad key or value is a 400 `invalid_run_config` naming the field, with no run row created.
- [ ] A completed run's report carries `thresholdValidation` with per-check `metric/limit/actual/passed` and a `verdict`, evaluated from the same numbers the summary reports.
- [ ] A run without thresholds omits the section entirely - no zeros - and pre-existing runs still render everywhere.
- [ ] The load-test dialog can declare budgets, and both the live dashboard and history detail show the verdict.
- [ ] MCP `start_load_run` forwards thresholds; suites green; docs updated same commit.

athrvk:
## Description

**Root cause.** A load run measures everything and judges nothing. The two pass/fail-shaped things that exist both miss: `testValidation` replays a `pm.*` script against individual sampled responses, so it structurally cannot assert an aggregate like p99 or error rate; the app's `sloThresholdMs` is a global client-only setting that never reaches the engine, is not persisted with the run, and produces a chart annotation rather than a verdict. Verified still live at `21ba739` - the report carries `latency{p50..p999}`, `summary.errorRate` and `summary.throughput` with nothing that compares them to anything.

**Approach.** Mirror the four-hop `testValidation` shape exactly: payload field → evaluated post-run → optional summary section → optional report section. The one design decision worth naming is that `validate_thresholds` (the route's gate) and `evaluate_thresholds` (the run's verdict) live in one file pair and read the **same metric table**. Two lists would drift, and the way that failure shows up is silent - a key the route accepts and the evaluator never checks, so the user declares a budget the run is never judged against. The verdict is computed from the filled `RunSummaryInputs` immediately before the summary write rather than from the collector, so the verdict and the numbers printed beside it cannot disagree.

**Rejected:** deriving the error rate from `MetricsCollector::error_rate()`, which is what "error rate" means everywhere else in the engine. It counts transport failures only, so a run of nothing but HTTP 500s scores a clean 0% - and it would contradict the `summary.errorRate` printed in the same report. The verdict follows the report's rule (every status outside 2xx/3xx, including the 0 a transport failure is recorded under). That needed a third copy of the 2xx/3xx test, so the rule is now `vayu::is_success_status` and the two existing copies call it.

**Deliberate deviation, stated.** The issue's payload spec says values must be finite and `> 0`, with `maxErrorRatePct` in `[0, 100]`. Implemented as written, and the asymmetry is load-bearing: `maxErrorRatePct: 0` ("no request may fail") is the strictest budget a user can declare, where a latency ceiling or a throughput floor of 0 is a typo nothing could meet. A `null` value reads as absent, following the config-wide rule the flat numeric fields already use - so an object of only nulls is rejected as declaring nothing, rather than accepted as a budget the report will never mention.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- `python build.py -e -t` then `ctest --preset linux-dev --output-on-failure`: **1225/1225 passed**.
- `cd app && pnpm test`: **3802 passed** across 380 files. `pnpm type-check` and `pnpm lint` clean; `prettier --check` clean on the touched files.
- `mkdocs build --strict -f .github/mkdocs.yml`: clean.
- No pre-existing failures encountered on the base commit.

**24 new engine tests** (`threshold_eval_test.cpp` ×17, `runs_route_test.cpp` ×3, `run_config_validation_test.cpp` ×4):

- a config with no `thresholds`, an explicit `null`, an object of only nulls, and a non-object value each produce no section and (where present) a rejection;
- an empty object is rejected rather than stored - a caller that sent it believes the run will be judged;
- an unknown key is rejected **and named**; a non-numeric or out-of-range value likewise;
- `maxErrorRatePct: 0` is accepted where `latencyP50Ms: 0` and `minThroughputRps: 0` are not;
- each percentile reads its own measurement (catches a copy-paste in the metric table - three budgets all judging p99 would look right until a p50 breached);
- the error rate counts 4xx/5xx **and** transport failures: 8 server errors + 2 code-0 out of 1000 is 1%, not the 0.2% the collector would say;
- the throughput floor passes above its limit, where every other budget passes below;
- a run that sent nothing is judged on what it measured (0/0 is 0%, not a division that takes the summary write down);
- a stored budget of the wrong type is skipped rather than guessed at, so a hand-edited or older snapshot cannot take the whole section down;
- the report round-trip asserts the check's `actual` **equals** `summary.errorRate` - the one assertion that pins the verdict to the numbers beside it;
- `verdict` is `"passed"` only when `failed == 0`, and the section is absent for a run that declared none.

**Mutation-checked**: flipping the `AtMost`/`AtLeast` comparison in `evaluate_thresholds` fails 6 tests (4 unit, 2 route round-trip); restored, all green.

**27 new app tests** (`budgets.test.ts` ×11, `ThresholdVerdict.test.tsx` ×8, `LoadTestConfigDialog.test.tsx` ×5, `electron/mcp/tools.test.ts` ×3):

- the dialog seeds p99 from `sloThresholdMs` and sends it (this is the reader that stops the setting being written-and-never-read);
- a cleared budget stays cleared across dialog opens - re-seeding every time would silently undo the user's decision to run without a latency budget;
- an out-of-range budget blocks Start with a named message instead of being dropped from the payload;
- `thresholds` is absent, never `{}`, when nothing was declared;
- every field in the table keys to a real `RunThresholds` member, so none can be typed into a control and sent nowhere;
- the verdict renders nothing for a run that declared no budgets, prints the throughput floor as `≥` and the ceilings as `≤`, and shows a metric it has no label for rather than dropping a check whose count is still in the total;
- MCP forwards the budgets under the engine's own keys, rejects an empty object by name, and rejects a value the engine would reject.

## Checklist

App changes were required and are included: `LoadTestConfig` / `StartLoadTestRequest` / `RunReport` types, a **Pass/fail budgets** disclosure in `LoadTestConfigDialog` (its field table and rules extracted to `budgets.ts`, component-free and unit-tested), forwarding through `handleConfirmLoadTest`, one shared `ThresholdVerdict` rendered by both the dashboard report and the history Overview (a second copy would not receive the first's fixes), and `thresholds` on MCP `start_load_run`.

Docs updated in the same commit: `docs/engine/api-reference.md` (the `POST /runs` payload, a new `thresholds` block section, the report shape), `docs/engine/db-schema.md` (the `summary` column), `docs/engine/mcp.md` (the tool table), `docs/app/api-integration.md` and `docs/app/COMPONENTS.md`.

## Related Issues
Closes #470

### DIFF
--- a/app/electron/mcp/tools.test.ts
+++ b/app/electron/mcp/tools.test.ts
@@ -1442,6 +1442,53 @@ describe("dispatchTool", () => {
 		expect(payload.postRequestScript).toBeUndefined();
 	});
 
+	test("start_load_run forwards thresholds under the engine's own metric keys", async () => {
+		// The keys travel verbatim to POST /runs and come back in the report's
+		// thresholdValidation, so a rename anywhere on this path is a budget
+		// the engine rejects - or worse, one it accepts and never judges.
+		const client = fakeClient();
+		const res = await dispatchTool(
+			"start_load_run",
+			parseArgs("start_load_run", {
+				url: "https://api.example.com",
+				confirmed: true,
+				thresholds: { latencyP99Ms: 50, maxErrorRatePct: 0.1 },
+			}),
+			ctxWith(client, { allowlist: ["api.example.com"] })
+		);
+
+		expect(res.isError).toBeFalsy();
+		const payload = (client.startRun as ReturnType<typeof vi.fn>).mock.calls[0][0];
+		expect(payload.thresholds).toEqual({ latencyP99Ms: 50, maxErrorRatePct: 0.1 });
+	});
+
+	test("start_load_run sends no thresholds key when none were declared", () => {
+		// An empty object is a 400 from POST /runs, so "no budgets" has to be
+		// an absent key rather than an empty one.
+		const parsed = parseArgs("start_load_run", { url: "https://api.example.com" });
+		expect(parsed.thresholds).toBeUndefined();
+		expect(() =>
+			parseArgs("start_load_run", { url: "https://api.example.com", thresholds: {} })
+		).toThrow(/at least one budget/i);
+	});
+
+	test("start_load_run rejects a budget the engine would reject", () => {
+		// The schema's bounds mirror the engine's so the agent is told which
+		// field is wrong, rather than meeting an HTTP 400 with no field named.
+		expect(() =>
+			parseArgs("start_load_run", {
+				url: "https://api.example.com",
+				thresholds: { maxErrorRatePct: 150 },
+			})
+		).toThrow();
+		expect(() =>
+			parseArgs("start_load_run", {
+				url: "https://api.example.com",
+				thresholds: { latencyP99Ms: 0 },
+			})
+		).toThrow();
+	});
+
 	test("start_load_run still accepts the engine's own `tests` spelling", async () => {
 		const client = fakeClient();
 		const res = await dispatchTool(
--- a/app/electron/mcp/tools.ts
+++ b/app/electron/mcp/tools.ts
@@ -1811,6 +1811,25 @@ export const TOOLS: McpTool[] = [
 				.describe(
 					`In-flight cap (constant_rps only), 1-${MAX_IN_FLIGHT_BOUND}. Default max(targetRps * 10, 1000).`
 				),
+			// Pass/fail budgets for the whole run. The bounds mirror the
+			// engine's, so a value this schema accepts is one POST /runs
+			// accepts; an empty object is rejected there rather than starting a
+			// run nothing will judge, so it is rejected here too, by name.
+			thresholds: z
+				.object({
+					latencyP50Ms: z.number().positive().max(86_400_000).optional(),
+					latencyP95Ms: z.number().positive().max(86_400_000).optional(),
+					latencyP99Ms: z.number().positive().max(86_400_000).optional(),
+					maxErrorRatePct: z.number().min(0).max(100).optional(),
+					minThroughputRps: z.number().positive().max(1_000_000_000).optional(),
+				})
+				.refine((t) => Object.keys(t).length > 0, {
+					message: "Declare at least one budget, or omit `thresholds` entirely.",
+				})
+				.optional()
+				.describe(
+					"Pass/fail budgets for this run. The report comes back with `thresholdValidation`: one check per budget plus a verdict of passed/failed. Omit for a run that is measured but not judged."
+				),
 			requestId: z
 				.string()
 				.optional()
@@ -1873,6 +1892,12 @@ export const TOOLS: McpTool[] = [
 			for (const key of ["duration", "rampUpDuration"]) {
 				const v = str(args, key);
 				if (v !== undefined) payload[key] = v;
+			}
+			// Forwarded verbatim - the keys are the engine's own metric names,
+			// and they come back unchanged in `get_run_report`'s
+			// `thresholdValidation`. Zod has already bounded every value.
+			if (args.thresholds && typeof args.thresholds === "object") {
+				payload.thresholds = args.thresholds;
 			}
 			// An omitted duration is 60s engine-side, not "unbounded" and not
 			// "capped" - so a cap under 60s has to be sent as an explicit field
--- a/app/src/components/shared/ThresholdVerdict.test.tsx
+++ b/app/src/components/shared/ThresholdVerdict.test.tsx
@@ -1,0 +1,127 @@
+/**
+ * @vitest-environment jsdom
+ */
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * The verdict's job is to be unambiguous about a run that failed its budget, so
+ * the cases that matter are the ways it could mislead: showing nothing for a
+ * run that was judged, showing a section for one that was not, printing a floor
+ * as if it were a ceiling, or silently dropping a check a newer engine sent.
+ */
+
+import { describe, it, expect } from "vitest";
+import { render, screen, cleanup } from "@testing-library/react";
+import { ThresholdVerdict } from "./ThresholdVerdict";
+import type { RunReport } from "@/types/domain";
+
+type Verdict = NonNullable<RunReport["thresholdValidation"]>;
+
+const verdict = (over: Partial<Verdict> = {}): Verdict => ({
+	checks: [{ metric: "latencyP99Ms", limit: 50, actual: 47.2, passed: true }],
+	passed: 1,
+	failed: 0,
+	verdict: "passed",
+	...over,
+});
+
+describe("ThresholdVerdict", () => {
+	it("renders nothing for a run that declared no budgets", () => {
+		// Every run recorded before budgets existed is this run. "Not judged"
+		// has to read as absence, not as a card full of zeros.
+		const { container } = render(<ThresholdVerdict verdict={undefined} />);
+		expect(container).toBeEmptyDOMElement();
+	});
+
+	it("renders nothing for a section carrying no checks", () => {
+		const { container } = render(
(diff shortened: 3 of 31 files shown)


athrvk:
Claiming this - master has moved 4 commits ahead of this branch (`c21f562`), so I'm merging master in and re-running the suites against the updated base this run.

---

athrvk:
Withdrawing the claim above - this PR had already merged at 14:44 UTC, a minute before the run picked it up, so the "behind master" state I saw was stale. No action needed here; the branch I briefly recreated has been deleted again.

---

PR #488, opened 2026-08-10
feat(engine/app): capacity-discovery run mode

### LINKED ISSUE #471: engine/app(loadtest): capacity-discovery run mode - an adaptive step-ramp that stops at the latency knee and reports what the service can sustain
Second of the thresholds → capacity → baselines arc (#470 is first). The headline capability: "find my service's capacity" as a one-click run mode, instead of a human bisecting concurrency by hand.

## Problem

Finding capacity with Vayu today is manual bisection: guess a `concurrency`, run `ramp_up`, read the dashboard, adjust, repeat. The pieces of an answer all exist and none of them close the loop:

- The engine already produces the exact per-tick signal a controller needs - `MetricTickSample` (`engine/include/vayu/core/run_manager.hpp:290-309`) carries `latency_p50_ms/p95_ms/p99_ms` (windowed), `current_rps`, `current_concurrency`. The comment at `engine/src/core/run_manager.cpp:1390` even says these power the "capacity breakpoint / saturation" derivations.
- The app already *detects* the knee, but only post-hoc and client-side: `computeBreakpoint` (`app/src/modules/dashboard/utils/computeBreakpoint.ts:34-50`) scans ticks for the first `latency_p99_ms > sloThresholdMs` and `SaturationCard` / `BreakpointStat` / the scatter's SLO marker render it. It observes whatever ramp the user happened to configure; it does not drive one.
- No strategy has a feedback path. Every `target_fn` passed to `maintain_concurrency` (`engine/src/core/load_strategy.cpp:258-309`; signature `engine/include/vayu/core/load_strategy.hpp:42-46`) is a pure function of `elapsed_ms` and constants - `RampUpLoadStrategy` (:503-547) interpolates `startConcurrency → concurrency` over `rampUpDuration` and is blind to what latency is doing.

So the tool that can outrun the server cannot tell you where the server gives out.

## Fix / solution pathway

**1. New mode.** Add `Capacity` to `enum class LoadTestType` (`engine/include/vayu/types.hpp:741`), `to_string` (:743-751, string `"capacity"`) and `parse_load_test_type` (:753-765). Register in `LoadStrategy::create` (`engine/src/core/load_strategy.cpp:553-574`). `execute_scenario_load` must reject it the way it rejects `constant_rps` (`engine/src/core/scenario_load.cpp:222-227`) - capacity discovery over a multi-step scenario is its own issue, not a side effect.

**2. Pure controller core first.** Following the `load_pacing.hpp` precedent (pure functions, unit-tested without a server), add `engine/include/vayu/core/capacity_controller.hpp`:

```cpp
struct CapacityWindow { double p99_ms; double rps; size_t concurrency; };
enum class CapacityAction { Hold, StepUp, Stop };
struct CapacityDecision { CapacityAction action; size_t next_concurrency; const char* stop_reason; };
CapacityDecision decide_next_level (const CapacityConfig&, const std::vector& history);
```

Policy (all tunable via config, defaults in `constants.hpp` beside `run_config`): hold each concurrency level for `stepDuration` (default 5s), then evaluate that window. **Step up** (e.g. +25%, min +1) while healthy. **Stop** when any of: windowed p99 > `sloMs` for 2 consecutive windows (`slo_exceeded`); rps gained < 5% across the last two step-ups while concurrency grew (`plateau`); `concurrency` cap reached (`cap_reached`); `duration` deadline (`deadline`). The knee is the last level *before* the stop condition.

**3. Feedback plumbing - the one sharp edge, decided here.** The strategy must NOT call `MetricsCollector::sample_window_percentiles()` - it is documented single-reader/mutating ("call once per metrics tick from the producer thread only", `engine/include/vayu/core/metrics_collector.hpp:523-539`) and the metrics thread already consumes it in `emit_live_tick` (`run_manager.cpp:1212`). Instead, publish what that thread already computes: give `RunContext` a small mutex-guarded `std::optional latest_tick` written by `collect_metrics` each tick and copied out by the strategy. One writer, cheap reads, no second consumer of the interval recorder, and the controller sees exactly the numbers the dashboard shows. `CapacityLoadStrategy` then drives `maintain_concurrency` with a `target_fn` that reads the controller's current level and a `should_continue` that goes false on `Stop` - both closures already capture `context`, exactly as `IterationsLoadStrategy` does (`load_strategy.cpp:478-484`).

**4. Config fields.** `sloMs` (default 200, matching the app's `DEFAULT_SLO_MS`), `stepDuration` (same string format as `duration`), reuse `startConcurrency` (start of the search, default 1), `concurrency` (the cap), `duration` (the deadline). Validate the numerics in `validate_run_config`'s `NumericRunField` table (`engine/src/http/routes/execution.cpp:486-524`); `sloMs` bounds can mirror the app's clamp (1..60000, `app/src/constants/client-settings.ts:57-58`).

**5. Report.** Same four-hop optional-section pattern as `testValidation` and #470's `thresholdValidation`: `RunSummaryInputs` optional → `build_run_summary_payload` (`run_manager.cpp:1110`) → `apply_run_summary`/`ReportExtras` (`engine/src/http/routes/runs.cpp:213`, :119) → `run_report_response`:

```json
"capacity": {
  "sloMs": 200, "stopReason": "slo_exceeded",
  "maxHealthyConcurrency": 48, "maxHealthyRps": 23400, "p99AtMaxHealthyMs": 41.2,
  "kneeConcurrency": 64, "kneeP99Ms": 312.0,
  "levels": [{ "concurrency": 1, "rps": 980, "p99Ms": 1.4 }]
}
```

`levels[]` is the per-step audit trail (bounded - one entry per level, a capacity run has tens of levels, not thousands). Absent for every other mode.

## App changes

- `LoadTestMode` union (`app/src/types/domain.ts:689`) gains `"capacity"`; MCP `KNOWN_LOAD_MODES` (`app/electron/mcp/safety.ts:153`) likewise, and `checkLoadCaps` (`safety.ts:203`) must treat the `concurrency` cap as the bounding value.
- `LoadTestConfigDialog`: a Capacity entry in `ProfilePicker` (`app/src/modules/request-builder/components/LoadTestConfigDialog/ProfilePicker.tsx`) with fields sloMs (prefilled from `sloThresholdMs`), step duration, start/max concurrency, deadline; wire through `handleConfirmLoadTest` (`app/src/modules/request-builder/index.tsx:395-523`) and `StartLoadTestRequest` (`app/src/types/api.ts:344-394`). Mode metadata lives in `app/src/constants/load-test-modes.ts`.
- Dashboard: `HeroRow.renderModeCards` (`app/src/modules/dashboard/components/hero/HeroRow.tsx:39`) and `ModeStatsRow.renderModeStats` (`.../stats/ModeStatsRow.tsx:50`) gain a capacity arm - `SaturationCard`, `BreakpointStat` and `P99AtPeakStat` already exist and are the natural cards; live view works unchanged since SSE ticks carry the needed fields.
- History: `LoadTestDetail` (`app/src/modules/history/main/LoadTestDetail.tsx`) renders the `capacity` section (headline sentence + levels table or a small uPlot wrapper; `UPlotChart` is the one charting primitive, semantic wrappers only). `RunReport` type gains the section (`domain.ts:904`).
- Docs table in `docs/engine/mcp.md` for the `start_load_run` mode enum, same commit.

## Tests

- Engine, pure: `capacity_controller` unit tests need no server - step-up while healthy, stop on 2-window SLO breach (1 window must NOT stop - mutation-check by dropping the consecutive requirement), plateau detection, cap and deadline, knee attribution to the last healthy level.
- Engine, integration (`engine/tests/load_strategy_test.cpp` fixture, `SlowMockServer`): against `/fast` a low deadline stops with `deadline`/`cap_reached`; against `/slow` (500ms) with `sloMs: 100` it stops `slo_exceeded` at low concurrency and the report's `kneeP99Ms` exceeds the SLO. Validation tests in `run_config_validation_test.cpp`; summary→report round-trip + absence twin in `runs_route_test.cpp` (pattern: `ReportReadsTheStoredSummary` :642, `ReportOmitsTestValidationWhenNoScriptRan` :736).
- App: mode plumbs through payload building; capacity cards render from a capacity report and are absent otherwise; `KNOWN_LOAD_MODES` accepts it (existing safety tests).

## Docs (same commit)

`docs/engine/api-reference.md` - `POST /runs` modes and the report's new section. `docs/engine/architecture.md` - strategies section gains the capacity mode and the `latest_tick` feedback path (it changes the "strategies are pure functions of elapsed time" invariant, which must be restated). `docs/engine/mcp.md`. `docs/app/COMPONENTS.md` if the new cards are catalogued there.

## Sequencing

#470 merged 2026-08-10 (e1402eb, PR #482) - the dependency is satisfied; align the SLO field naming with the shipped threshold keys (engine/src/core/threshold_eval.cpp:76).

After #470, to share the SLO naming (`sloMs` here, `latencyP99Ms` budget there) rather than invent two spellings. Not blocked by #197 (high-RPS default re-tune), but a capacity run leans on those defaults harder than any fixed run - whoever implements should re-read #197's findings on `workers`/`maxInFlight` and say in the PR whether defaults were touched. Parallel-safe (#450 closed 2026-08-10 via PR #474). Conflict surface with siblings: #475/#476/#478 extend the same RunSummaryInputs -> build_run_summary_payload -> apply_run_summary -> run_report_response four-hop in run_manager.cpp/runs.cpp - coordinate whichever lands second; trivial rebases.

## Acceptance criteria

- [ ] `mode: "capacity"` runs an adaptive step-ramp that needs only an SLO and a cap - no hand-tuned target - and stops itself with a named `stopReason`.
- [ ] The report answers the headline question in numbers: max healthy concurrency, max healthy rps, p99 at that level, and the knee, with a per-level audit trail.
- [ ] The strategy reads metrics only through the published tick - `sample_window_percentiles()` still has exactly one caller (mutation-check or a static assertion in review).
- [ ] Scenario load runs reject the mode with a clear 400.
- [ ] Dialog, live dashboard, and history all speak the new mode; suites green; docs updated same commit.

---

athrvk:
## Description

`mode: "capacity"` - an adaptive step-ramp that finds the latency knee and reports what the service can sustain, instead of a human bisecting concurrency by hand.

**Plan.** The root cause is that no strategy had a feedback path: every `target_fn` passed to `maintain_concurrency` is a pure function of `elapsed_ms` and constants, so the tool that can outrun the server could not tell you where the server gives out. The pieces of an answer all existed (`MetricTickSample` carries the windowed percentiles, `computeBreakpoint` detects the knee client-side and post-hoc) and none of them closed the loop. The approach is the one the issue lays out: a pure controller core (`decide_next_level` over a level history, no clock and no collector, unit-tested without a server) driving a strategy that reads metrics through a **published** tick. The alternative I rejected is the obvious one - having the strategy call `MetricsCollector::sample_window_percentiles()` directly. That call is documented single-reader and *resets* the rolling window on read, and the metrics thread already consumes it once per tick; a second consumer would silently halve both readers' sample counts, and neither the live chart nor the search would look wrong. So the metrics thread publishes what it already computed (`RunContext::publish_live_tick()`, one writer) and the strategy copies it out - which also means the controller steers by exactly the numbers the dashboard is drawing.

I verified the problem still exists as described; the anchors in the issue were accurate. Three things are worth calling out as deviations, all noted inline in the code:

1. **`Percentiles` gained a `count` field.** Not in the issue's plan, but required for correctness: an empty window returns all zeros, so averaging a level's per-tick p99s pulled a 500ms endpoint's measured p99 well under a 100ms budget. The first integration test caught it - the search climbed 4 → 64 against `/slow` and stopped `cap_reached`. The controller now ignores windows in which nothing completed.
2. **A fifth stop reason, `stopped`.** The issue names four. An operator-cancelled search has not run out of clock, and reporting `deadline` for it would claim a measurement never taken.
3. **The `deadline` stop lives in the strategy, not `decide_next_level`.** It is the one condition that is about the clock rather than about what the service did, and the pure core holds no clock. `summarize_capacity` takes the reason from the caller for the same reason.

The `target_fn` invariant this breaks is restated explicitly in `docs/engine/architecture.md` rather than left to be rediscovered.

**Engine**
- `core/capacity_controller.{hpp,cpp}` - step up 25% (min +1) while healthy; a *single* breaching window `Hold`s and re-measures the same level; stop on two consecutive breaches (`slo_exceeded`), a sub-5% throughput gain across two step-ups (`plateau`), the ceiling (`cap_reached`), the deadline, or an operator stop. Growth/plateau/breach-count are constants in `constants::capacity`, not new wire fields.
- Report gains an optional `capacity` section. `maxHealthy*` is omitted when the first level already breached and `knee*` unless latency ended the search - a search that ran out of room never watched the target give out.
- `sloMs` (1-60000, the app's own clamp) and `stepDuration` validated in the run route; `duration` and `stepDuration` now share one parser rather than a copy. `capacity` on a scenario run is a 400.

**App**
- Capacity Discovery profile in the load dialog: latency budget (prefilled from `sloThresholdMs`), step duration, start/ceiling, deadline. Blocked when the start is already at the ceiling - the engine would run it, but a one-level search is never what the profile was picked for.
- `HeroRow` / `ModeStatsRow` capacity arms; `CurrentConcurrencyCard` is deliberately *not* reused (its subtitle states a ramp duration and lag, and a search has neither).
- `CapacitySummary` card in the history Overview: the finding as a sentence, over the levels table it came from.
- MCP `start_load_run` accepts the mode and forwards `sloMs` / `stepDuration`; the existing concurrency cap already bounds both `concurrency` (the ceiling) and `startConcurrency`, which is exactly what stops an adaptive run from outgrowing it.

**#197 (high-RPS default re-tune), as the issue asks:** no defaults were touched. `workers` / `maxInFlight` are untouched by this PR - a capacity run leans on them harder than a fixed run does, but re-tuning them is #197's scope and mixing it in would make both changes harder to attribute.

## Type of Change
- [x] New feature
- [x] Documentation update

## Testing

All green on this branch; no pre-existing failures observed.

- **Engine:** `python build.py -e -t && ctest --preset linux-dev` - **1282/1282 passed** (128.9s). 34 of those are new or capacity-related:
  - `capacity_controller_test.cpp` (20) - step-up, growth floor, cap clamp, the consecutive-breach requirement (mutation check: drop it and `OneBreachingWindowHoldsRatherThanStopping` reports `Stop`), plateau detection, a hold not counting as a step-up, knee attribution, payload omissions, config clamping, plus the source scan pinning `sample_window_percentiles()`'s single production caller (the scan asserts it read >10 files and >1000 bytes first).
  - `load_strategy_test.cpp` (4) - end-to-end against `SlowMockServer` with a real metrics thread: `/slow` at `sloMs: 100` stops `slo_exceeded` with `kneeP99Ms` over the SLO and no `maxHealthy`; `/fast` stops `cap_reached`; a short deadline stops `deadline` with the partial window absent from the trail.
  - `run_config_validation_test.cpp` (5) and `runs_route_test.cpp` (3) - field validation and the summary → report round-trip plus its two absence twins.
- **App:** `pnpm test` - **3823/3823 passed** (381 files); `pnpm type-check`, `pnpm lint`, `pnpm format:check` clean. New: `CapacitySummary.test.tsx` (7, incl. a mutation check on the over-budget row colour), dialog field/payload tests, `validateCapacityRange`, summary prose, and MCP safety caps. The `load-test-modes` vocabulary guard caught the new mode and was updated.
- **Docs:** `mkdocs build --strict` clean.

## Checklist

## Related Issues
Closes #471

---
_Generated by [Claude Code](https://claude.ai/code/session_01VMERxmB31k3ih8XYgVztHg)_

### DIFF
--- a/app/electron/mcp/safety.test.ts
+++ b/app/electron/mcp/safety.test.ts
@@ -248,3 +248,34 @@ describe("defaultDurationUnderCap", () => {
 		expect(defaultDurationUnderCap({ mode: "iterations", iterations: 10 }, config)).toBeNull();
 	});
 });
+
+describe("capacity discovery", () => {
+	const config = resolveSafetyConfig({ maxConcurrency: 500, maxDurationSeconds: 300 });
+
+	test("caps the ceiling the search climbs toward, not just a fixed target", () => {
+		// `concurrency` is the ceiling in this mode rather than a level held, so
+		// it is exactly the field that decides how large the run can get.
+		expect(checkLoadCaps({ mode: "capacity", concurrency: 5000 }, config).ok).toBe(false);
+		expect(checkLoadCaps({ mode: "capacity", startConcurrency: 5000 }, config).ok).toBe(false);
+		expect(
+			checkLoadCaps({ mode: "capacity", startConcurrency: 4, concurrency: 256 }, config).ok
+		).toBe(true);
+	});
+
+	test("rejects a step duration the engine cannot read", () => {
+		// It goes through the same string parser `duration` does, so a value the
+		// engine would throw on has to fail here rather than at run time.
+		expect(checkLoadCaps({ mode: "capacity", stepDuration: "a bit" }, config).ok).toBe(false);
+		expect(checkLoadCaps({ mode: "capacity", stepDuration: "5s" }, config).ok).toBe(true);
+	});
+
+	test("is a duration mode, so the duration cap still reaches it", () => {
+		// The guard mirrors `LoadStrategy::create`: an unrecognised mode with no
+		// `iterations` falls through to a duration strategy. `capacity` is a
+		// duration mode either way - the check is that adding it to the known
+		// set did not accidentally route it to the iterations branch, which
+		// takes no duration cap at all.
+		const tight = resolveSafetyConfig({ maxDurationSeconds: 30 });
+		expect(defaultDurationUnderCap({ mode: "capacity" }, tight)).toBe("30s");
+	});
+});
--- a/app/electron/mcp/safety.test.ts
+++ b/app/electron/mcp/safety.test.ts
@@ -5,6 +5,9 @@
  * LICENSE file in the "app" directory of this source tree.
  */
 
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
 import { describe, expect, test } from "vitest";
 import {
 	extractHost,
@@ -277,5 +280,48 @@ describe("capacity discovery", () => {
 		// takes no duration cap at all.
 		const tight = resolveSafetyConfig({ maxDurationSeconds: 30 });
 		expect(defaultDurationUnderCap({ mode: "capacity" }, tight)).toBe("30s");
+	});
+
+	test("caps an omitted duration against this mode's own engine default", () => {
+		// The regression: `capacity` falls back to 300s engine-side, not the 60s
+		// every other mode uses. A 120s cap is *above* 60 and *below* 300, so a
+		// guard keyed on one default injected nothing and let the search run
+		// 2.5x over the ceiling. `checkLoadCaps` cannot catch it either - with
+		// `duration` omitted there is nothing for it to check.
+		const cap = resolveSafetyConfig({ maxDurationSeconds: 120 });
+		expect(defaultDurationUnderCap({ mode: "capacity" }, cap)).toBe("120s");
+		// The same cap over a 60s-default mode still needs no field.
+		expect(defaultDurationUnderCap({ mode: "constant_concurrency" }, cap)).toBeNull();
+	});
+
+	test("supplies nothing once the cap clears the capacity default", () => {
+		const roomy = resolveSafetyConfig({ maxDurationSeconds: 600 });
+		expect(defaultDurationUnderCap({ mode: "capacity" }, roomy)).toBeNull();
+	});
+
+	test("mirrors the engine's own capacity deadline", () => {
+		// A drift guard, not a restatement: the number above is only a cap if it
+		// is the number the engine actually falls back to, and nothing else
+		// connects the two files. Same pattern as
+		// `src/constants/load-test.engine-parity.test.ts`.
+		const repoRoot = join(dirname(fileURLToPath(import.meta.url)), "..", "..", "..");
+		const constantsHpp = readFileSync(
+			join(repoRoot, "engine", "include", "vayu", "core", "constants.hpp"),
+			"utf8"
+		);
+		// CLAUDE.md's documented failure mode: a source scan reading "" passes.
+		expect(constantsHpp.length).toBeGreaterThan(0);
+
+		const match = constantsHpp.match(/constexpr\s+int64_t\s+DEADLINE_MS\s*=\s*(\d+);/);
+		expect(match, "capacity::DEADLINE_MS not found in constants.hpp").not.toBeNull();
+		const engineSeconds = Number(match![1]) / 1000;
+
+		// Read through the exported behaviour rather than the private table: a
+		// cap one second under the engine's default must produce a field, and a
+		// cap exactly at it must not.
+		const under = resolveSafetyConfig({ maxDurationSeconds: engineSeconds - 1 });
+		const at = resolveSafetyConfig({ maxDurationSeconds: engineSeconds });
+		expect(defaultDurationUnderCap({ mode: "capacity" }, under)).toBe(`${engineSeconds - 1}s`);
+		expect(defaultDurationUnderCap({ mode: "capacity" }, at)).toBeNull();
 	});
 });
--- a/app/electron/mcp/safety.ts
+++ b/app/electron/mcp/safety.ts
@@ -142,6 +142,7 @@ export interface LoadRunParams {
 	startConcurrency?: number;
 	duration?: string | number;
 	rampUpDuration?: string | number;
+	stepDuration?: string | number;
 	iterations?: number;
 }
 
@@ -150,7 +151,13 @@ export interface LoadRunParams {
  * Anything else falls through `LoadStrategy::create`, so the guard mirrors that
  * fallback rather than trusting the string it was given.
  */
-const KNOWN_LOAD_MODES = new Set(["constant_rps", "constant_concurrency", "ramp_up", "iterations"]);
+const KNOWN_LOAD_MODES = new Set([
+	"constant_rps",
+	"constant_concurrency",
+	"ramp_up",
+	"iterations",
+	"capacity",
+]);
 
 /**
  * What the engine runs when `duration` is absent - `duration_field_ms(config,
@@ -209,7 +216,10 @@ export function checkLoadCaps(params: LoadRunParams, config: McpSafetyConfig): G
 	}
 	// `startConcurrency` rides the same ceiling as `concurrency`: `ramp_up` seeds
 	// the run with it (`RampUpLoadStrategy`, `target_fn(0) = startConcurrency`),
-	// so an uncapped start is an uncapped run however low the target is.
+	// so an uncapped start is an uncapped run however low the target is. For
+	// `capacity` the same pair is what bounds the search - `concurrency` is the
+	// ceiling it climbs toward rather than a target it holds - so capping both
+	// is exactly what stops an adaptive run from outgrowing the cap.
 	for (const field of ["concurrency", "startConcurrency"] as const) {
 		const value = params[field];
 		if (typeof value === "number" && value > config.maxConcurrency) {
@@ -221,7 +231,7 @@ export function checkLoadCaps(params: LoadRunParams, config: McpSafetyConfig): G
 	}
 	// A duration the engine cannot read now fails the run rather than quietly
 	// becoming 60s, so say so here instead of starting a run that dies.
-	for (const field of ["duration", "rampUpDuration"] as const) {
+	for (const field of ["duration", "rampUpDuration", "stepDuration"] as const) {
 		const value = params[field];
 		if (value !== undefined && parseDurationGrammar(value) === null) {
 			return {
--- a/app/electron/mcp/safety.ts
+++ b/app/electron/mcp/safety.ts
@@ -160,11 +160,32 @@ const KNOWN_LOAD_MODES = new Set([
 ]);
 
 /**
- * What the engine runs when `duration` is absent - `duration_field_ms(config,
- * "duration", 60000)` in `load_strategy.cpp`. An omitted duration is therefore
- * not "no duration", and a cap below this one only binds if the field is sent.
+ * What the engine runs when `duration` is absent, **per mode**.
+ *
+ * An omitted duration is not "no duration": each strategy passes its own
+ * fallback to `duration_field_ms`, so a cap only binds when the field is sent
+ * *or* when the cap is under the default the engine would otherwise use.
+ *
+ * This was a single number, on the assumption that every mode falls back to
+ * 60s. `capacity` does not - it walks a level every `stepDuration`, so its
+ * `constants::capacity::DEADLINE_MS` is 300s - and the assumption became a
+ * hole: with the cap set to 120s and an agent omitting `duration`,
+ * `checkLoadCaps` had nothing to check and this function returned null because
+ * 120 >= 60, so the search ran for 300s against a 120s cap. Keyed by mode, the
+ * guard now models what the engine actually does rather than one mode's version
+ * of it, and `safety.test.ts` reads `constants.hpp` to keep the numbers in step.
  */
-const ENGINE_DEFAULT_DURATION_SECONDS = 60;
+const ENGINE_DEFAULT_DURATION_SECONDS: Readonly<Record<string, number>> = {
+	capacity: 300,
+};
+
+/** The fallback for every mode that has no entry of its own. */
+const ENGINE_FALLBACK_DURATION_SECONDS = 60;
+
+/** What the engine would run for, in seconds, if `duration` were omitted. */
+function engineDefaultDurationSeconds(mode: string): number {
+	return ENGINE_DEFAULT_DURATION_SECONDS[mode] ?? ENGINE_FALLBACK_DURATION_SECONDS;
+}
 
 /** What the engine runs when `iterations` is absent (`IterationsLoadStrategy`). */
 const ENGINE_DEFAULT_ITERATIONS = 1000;
@@ -196,7 +217,11 @@ export function defaultDurationUnderCap(
 ): string | null {
 	if (params.duration !== undefined) return null;
 	if (isIterationsRun(params)) return null;
-	if (config.maxDurationSeconds >= ENGINE_DEFAULT_DURATION_SECONDS) return null;
+	// The engine's own default for an absent `mode` is a duration mode, and the
+	// same string `isIterationsRun` reads - the two must agree about which run
+	// this is or the cap lands on the wrong strategy.
+	const mode = params.mode ?? "constant_rps";
+	if (config.maxDurationSeconds >= engineDefaultDurationSeconds(mode)) return null;
 	return `${config.maxDurationSeconds}s`;
 }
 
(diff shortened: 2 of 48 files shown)


athrvk:
## Code review - capacity-discovery run mode (head `ddf4ac3`)

Reviewed the full diff. The core engine controller (`decide_next_level`, hold/plateau/knee logic, the published-tick feedback path) held up well under scrutiny - the findings are in the wiring around it.

### Findings

**1. `KNOWN_MODES` missing `"capacity"` makes the new dashboard arms unreachable** - `app/src/modules/dashboard/hooks/useMode.ts:22`

`resolveMode`/`useMode` normalize the unknown string `"capacity"` to `"constant_rps"`, so the new `case "capacity"` arms in `HeroRow.tsx` and `ModeStatsRow.tsx` are dead code. Run any capacity test and open the live dashboard or history Overview tab: the user sees the open-loop constant_rps hero cards (Target-RPS-based) instead of the ConcurrencyUtil/Saturation cards this PR added specifically for the mode - the repo's own "written but never read" defect class.

**2. MCP duration cap bypassed by the capacity default deadline** - `app/electron/mcp/safety.ts:199`

`defaultDurationUnderCap` assumes the engine's absent-duration default is 60s for every mode, but `CapacityLoadStrategy` defaults its deadline to 300000ms (5 min). If the user lowers the MCP duration cap to e.g. 120s and an agent calls `start_load_run` with `mode:"capacity"` and no `duration`: `checkLoadCaps` passes (nothing to check), `defaultDurationUnderCap` returns null because 120 >= 60, so no capped duration is injected - and the engine runs the search for 300s, 2.5x over the cap the guard exists to enforce. The new safety test only covers caps below 60s, so it pins the wrong assumption.

**3. False "already over budget at the lowest level" claim when zero levels were judged** - `app/src/components/shared/CapacitySummary.tsx:125`

Start a capacity run with `stepDuration` 10s and `duration` 5s (both accepted by the dialog and the engine). No level is judged; `summarize_capacity` stores `levels: []` with stopReason `"deadline"`, and the section is still reported because `levels` is a (empty) array. `CapacitySummary`'s `!sustained` branch then prints that the target breached the budget at the lowest concurrency the search tried - a measurement that was never taken, directly contradicting the "Ran out of time" badge next to it. The engine deliberately omits `maxHealthy*` to avoid exactly this false claim; the sentence needs a `levels.length` check.

**4. UB float-to-size_t cast on hand-edited config snapshots** - `engine/src/core/capacity_controller.cpp:195`

`capacity_config_from` casts arbitrarily large finite doubles straight to `size_t` (`static_cast<size_t>(start)` / `(cap)`), which is undefined behavior for values above `SIZE_MAX` - despite the function's documented contract of being total over arbitrary JSON from stored config snapshots. A snapshot containing `{"startConcurrency": 1e30}` passes the finite/positive checks, then the cast yields 0 or an arbitrary huge value depending on platform, so the search may run at a garbage concurrency level. The lambda clamps for finiteness and positivity but not against a usable upper bound.

**5. Stop-reason chip hand-rolls the Badge primitive** - `app/src/components/shared/CapacitySummary.tsx:94`

The chip is a bare `<span>` painting `bg-warning/10` / `bg-status-success/10` with `px-2 py-0.5` and no radius class - exactly what `Badge variant="chip"` exists for. Per `app/CLAUDE.md`, a hand-rolled copy of a primitive does not receive the primitive's fixes, and the missing radius class pins the box square for users who chose the Rounded setting.

### Checked and deliberately not flagged

`Percentiles` gaining `count` as its first member has no positional aggregate-init call sites to break; the single-reader rule for `sample_window_percentiles()` is preserved (and pinned by a source-scan test); `maintain_concurrency`'s 50ms wake cadence keeps `CapacitySearch::advance` polling faster than the 100ms tick, so no ticks are structurally lost; the hold/plateau/`per_level` collapse logic and knee/maxHealthy omission rules are internally consistent with their tests; the MCP omitted-`concurrency` case is safe under the default cap; no em-dashes were introduced and all governing docs were updated in-commit.

---

athrvk:
claimed

athrvk:
All five findings confirmed against the code and fixed in `9f7984a`. Verified each rather than taking it on trust; the reproduction paths were accurate as written.

**1. `KNOWN_MODES` - dead dashboard arms.** Correct, and the worst of the five: `resolveMode` normalised `capacity` to `constant_rps`, so a closed-loop adaptive search rendered open-loop Target-RPS hero cards. Worth naming *why* it shipped green - a hand-written `new Set<LoadMode>([...])` is only type-checked for *spurious* members, never missing ones, so the third copy of the vocabulary could silently fall behind the union and `LOAD_TEST_MODES`.

Rather than add the string, I derived the set from `LOAD_TEST_MODES`, which `load-test-modes.test.ts` already pins to the union. That makes the whole chain one guard instead of three lists, and the invariant becomes structural: a mode a user can *start* is a mode the dashboard can *render*. `useMode.test.ts` walks the shipped vocabulary rather than a list written in the test, so the next mode fails here without anyone thinking to add a case.

**2. MCP duration cap bypass.** Correct. The guard's model was wrong, not just its number - it assumed one engine default for all modes, and `checkLoadCaps` genuinely cannot help when the field is absent. Fixed by keying the default on mode. I kept the 300s engine-side default rather than flattening it to 60s: at the default step this mode walks a level every 5s, so a minute is ~12 levels and a search that almost always ends `deadline` having found nothing. The number is now `constants::capacity::DEADLINE_MS`, and `safety.test.ts` reads `constants.hpp` for it (same drift-guard pattern as `load-test.engine-parity.test.ts`, applied from the electron side since `electron/` production code doesn't import `@/`). The new test uses a 120s cap - above 60, below 300 - which is exactly the window that was open.

**3. False "already over budget" claim.** Correct. Empty `levels` and absent `maxHealthy*` are two different states and I had collapsed them. I kept the engine reporting the section with `levels: []` - "the search judged nothing" is itself the finding that tells you to lengthen the run or shorten the step, which an omitted section couldn't say - and gave the card a third branch that says that instead of inventing a measurement.

**4. Float-to-`size_t` UB.** Correct - finite-and-positive isn't sufficient, and the function's contract is explicitly total over stored snapshots no route ever saw. Clamped to `run_config::MAX_CONCURRENCY` before the cast, with a test on `1e30` asserting a usable level rather than a zero the search would spin at forever.

**5. Hand-rolled chip.** Correct on both counts, including the one that isn't visible in the diff: no radius class at all pins the box square for a user on the Rounded setting. Now `Badge variant="chip"`.

One thing I did **not** touch, flagging rather than folding in: `ThresholdVerdict` (from #470) has the identical hand-rolled chip - same bare `<span>`, same `px-2 py-0.5`, same missing radius. It's the same defect in the sibling component I modelled this one on, but it's #470's code and outside this diff; happy to take it in a follow-up if you want it.

CI was green on `ddf4ac3` before these changes. Re-verified locally on `9f7984a`: engine **1283/1283**, app **3830/3830** (382 files), `type-check` / `lint` / `format:check` / `mkdocs --strict` all clean. Docs updated in the same commit for the per-mode duration default, since the MCP cap now depends on it.

---

athrvk:
take ThresholdVerdict in this PR itself

athrvk:
Done in `d3585ef` - `ThresholdVerdict`'s chip is now `Badge variant="chip"`, same as `CapacitySummary`'s.

Two things came out of doing it that weren't in the original finding:

**A guard, not just a fix.** Both chips get their colour from a binding (`failed`, `stop.degraded`), so a source scan can't see the class list - and the missing-radius half is the escape hatch `app/CLAUDE.md` says no scan can flag at all. So each component's test now renders and asserts on `element.className`. Mutation-checked both ways: reverting the `<Badge>` back to the `<span>` fails the radius assertion, restoring it passes.

One correction worth noting since it's in the test: my first version of the token assertion was `not.toMatch(/\btext-destructive\b/)`, which fails against the *correct* token - `-` is a word boundary, so it matches the prefix of `text-destructive-text`. It's a lookahead now (`/\btext-destructive(?!-text)\b/`), which is what actually distinguishes the bare indicator fill from the `-text` variant.

**A docs miss from my first commit.** `ThresholdVerdict` is catalogued in `docs/app/COMPONENTS.md`; `CapacitySummary` wasn't, and the same-commit rule says it should have been. Added, and the entry records the three-state rule explicitly (sustained / no level held the budget / judged nothing at all), since finding #3 in your review was precisely those last two states being collapsed - that's the kind of thing the catalogue exists to keep straight.

App-only TSX and Markdown, so no engine rebuild: **3832/3832** app tests (up 2, the new radius guards), `type-check` / `lint` / `format:check` / `mkdocs --strict` clean.

---

PR #489, opened 2026-08-10
feat(engine/app): refresh oauth2 tokens during long load runs

### LINKED ISSUE #478: engine/app(auth): oauth2 token refresh during long load runs - today an expiring token silently turns the run into a 401 storm the report never explains
Token-expiry-mid-test is a repeatedly reported pain across API tooling, and Vayu currently handles it by warning up front and then doing nothing. The app half of the story already exists; this closes the engine half.

## Problem

A load run resolves auth exactly once: `execute_load_test` (`engine/src/core/run_manager.cpp:755`) calls `build_request` before the strategy starts (:824; the comment at :813-815 states the event loop attaches `request.headers` to every transfer, "so resolving auth here covers the whole run"). When an OAuth2 token expires mid-run:

- Nothing engine-side notices. There is no per-request or periodic `is_expired` re-check (`is_expired`, `engine/src/http/oauth_client.cpp:194-199`, 45s skew), no refresh thread, and `collect_metrics` never touches `oauth_tokens`.
- 401s from the target are recorded as ordinary status-code results (`handle_result` treats any HTTP response as a completion), so the run "succeeds" with a status chart that flips to 401 partway through and a report that never says why.
- The app knows and can only warn: `OAuth2LoadTestGuard` (`app/src/modules/request-builder/components/OAuth2LoadTestGuard.tsx`; its header at L8-19 documents that the engine acquires once and does not refresh mid-run) computes coverage vs run duration and gates the Start button (`gated`, L76) - with a "Start anyway" switch as the only path for a run longer than the token's life.

Everything needed to fix it exists: `acquire_token` already implements the full refresh path (refresh grant with rotation handling, `oauth_client.cpp:232-261`, honoring `autoRefreshToken`), tokens are cached in the `oauth_tokens` table (`database.cpp:308-317`), and the per-run auxiliary-thread pattern is established (`context->metrics_thread`, `run_manager.cpp:733-734`, joined at :832-833, :855, :893-894 and :1016 - four join sites; the watchdog must be joined at each).

## Fix / solution pathway

**1. Capture what's needed on `RunContext`.** When the run's auth parses to `OAuth2Auth` (`parse_auth`, `engine/src/http/auth_resolver.cpp:115-149`), stash the resolved oauth2 config JSON and the token's `expiresAt` on the context at build time. Non-OAuth2 runs: feature entirely inert.

**2. A refresh watchdog thread** - `context->auth_refresh_thread`, spawned beside `metrics_thread`, joined the same places. Loop: sleep until `expires_at - lead` (lead = 60s, constant beside the 45s skew in `oauth_client.hpp:60`; wake early on `is_running == false`), call `acquire_token(db, config, /*force_refresh=*/true, nullopt)` - the blocking `vayu::http::Client` inside it is fine on this dedicated thread, never on the tick thread - then swap the credential (step 3) and re-arm from the new token's expiry. Tokens with `expires_in <= 0` (never expire, :198) or `autoRefreshToken == false`: thread exits immediately after logging once at verbose. An `authorization_code` config with no refresh token cannot refresh non-interactively - record the failure (step 4) and stop trying; never pop interactive auth mid-run.

**3. Swap the header safely.** The transfer path copies `request.headers` per submission; a naked mutation races. Hold the `Authorization` value as `std::shared_ptr` on `RunContext` (atomic load per transfer setup, atomic store on refresh) and have the event-loop's header-attach site read through it for the auth header only - the implementer must locate the exact copy site in `engine/src/http/event_loop/` and state it in the PR. Respect placement: `tokenPlacement == "query"` (`auth_resolver.cpp:92-109`) cannot be swapped per-transfer this way - v1 refreshes header-placement tokens only, and the guard copy says so; query-placement runs keep today's behavior, documented.

**4. Make it visible.** Count refreshes and failures on the context; report an optional `auth` section via the standard four-hop (`RunSummaryInputs` - constructed at two sites, :937 single-request and :1029 scenario; the auth section must be wired into both → `build_run_summary_payload` `run_manager.cpp:1110` → `apply_run_summary` `runs.cpp:213` → report): `{"refreshes": [{"atSeconds": 3620.4}], "refreshFailures": 1, "lastError": "oauth2_provider_error: ..."}` - absent for non-OAuth2 runs. A refresh failure does NOT stop the run; the target's 401s plus this section tell the story honestly.

## App changes

- `OAuth2LoadTestGuard` copy: the `refresh`/`too-long` blocking branch (L147-180) becomes accurate - a refreshable token over a long run is now `covered` (adjust `coverageState` in `oauth2-load-test-coverage.ts`; keep blocking for non-refreshable/query-placement cases). Its header doc (L8-19) rewritten to the new contract.
- `RunReport` type gains the optional `auth` section (`app/src/types/domain.ts:904`; it sits beside the `thresholdValidation?` field at :1030 that #470 added); `LoadTestDetail` renders a one-line note when refreshes/failures are present ("token refreshed at 1h00m; 1 refresh failure at 1h32m").
- `TokenStatusRow` (`app/src/components/shared/OAuth2Form/TokenStatusRow.tsx`) is already the expiry surface - no change required; do not grow scope here.

## Tests

- Engine: the watchdog against a stub token endpoint on `SlowMockServer` - a short-`expires_in` token gets refreshed before expiry and subsequent transfers carry the new value (assert via the echo of received `Authorization` headers, `EchoServer` pattern); refresh failure records the section and the run completes; `expires_in <= 0` spawns no work; non-OAuth2 run untouched (mutation-check: disable the swap, the new-header assertion must fail). Timing: drive with a 2-3s `expires_in` so the test stays inside the 60s ctest timeout - the lead must be configurable for tests (constant injectable or config key).
- App: `coverageState` matrix updated (refreshable long run → covered; query-placement long run → still gated) with both branches asserted; report note renders with and without the section.

## Docs (same commit)

`docs/engine/db-schema.md:339` - the oauth_tokens section currently states "There is no mid-run refresh" verbatim (:358); rewrite it. `docs/engine/api-reference.md` report shape. `docs/engine/architecture.md` per-run thread inventory (same table #475 touches - coordinate). App: none beyond code.

## Sequencing

Independent of #470-#473 and #475-#477; touches `run_manager.cpp` thread setup like #475 does - trivial rebase for whichever lands second. The mock-issuer issue (filed next) makes this issue's tests nicer but is NOT a dependency - the stub token endpoint above suffices. #470 merged 2026-08-10 (PR #482); the summary-payload anchors above were re-verified against it.

## Acceptance criteria

- [ ] A load run outliving its token keeps sending fresh `Authorization` headers, with each refresh timestamped in the report's `auth` section.
- [ ] Refresh failure degrades gracefully: run completes, section says what happened, no interactive prompt ever fires mid-run.
- [ ] Non-OAuth2 and never-expiring-token runs are provably untouched (no thread, no section).
- [ ] The pre-run guard's copy matches the new reality, both coverage branches tested.
- [ ] Suites green; the db-schema doc's "no mid-run refresh" sentence is gone; docs same commit.

---

athrvk:
## Description

A load run resolved auth exactly once, before the strategy started (`execute_load_test` → `build_request`), so a run outliving its OAuth 2.0 access token turned into a 401 storm the report never explained. The app could only warn up front and gate the Start button; the engine did nothing.

**Plan.** Root cause: the credential is resolved into a single `vayu::Request` that every transfer copies, and nothing re-checks its expiry. The fix arms a per-run refresh watchdog (beside the existing `metrics_thread`) that re-acquires the token `oauth2RefreshLeadMs` (default 60s) before expiry and publishes the new `Authorization` value on `RunContext::auth_refresh`; the **strategy thread** - the event loop's sole producer - copies it onto its own `Request` when the cell's generation moves, immediately before `EventLoop::submit`, which copies the request wholesale into the transfer. So a swap reaches every later transfer, none already queued, and costs one relaxed atomic load per submission with no lock on the hot path.

The alternative I rejected is the one the issue sketched: reading a `shared_ptr<const std::string>` through the event loop's own header-attach site (`curl_utils.cpp:418`). That site works on the transfer's *copy* of the request, and `EventLoop` is deliberately run-agnostic - wiring a `RunContext` (or a run-specific field on `vayu::Request`, which every non-load caller would then carry) into it puts run policy inside the generic transport for no gain. `SubmissionRequest` in `load_strategy.cpp` keeps the swap where the copy is made. The exact copy site is `EventLoopImpl::submit` (`event_loop_impl.cpp:106`, `data->request = request`), on the calling thread.

Deliberate scope notes, all documented:
- Only header-placed credentials are renewed. `plan_auth_refresh` returns nothing - and the run behaves exactly as before - for `tokenPlacement: "query"`, `autoRefreshToken: false`, an `authorization_code` grant with no refresh token, a non-expiring token, a request whose `Authorization` header the user supplied, and scenario load runs (each step resolved its own auth at plan time, so there is no single credential).
- `auth_refresh_delay_ms` is floored at 1s: a token whose whole lifetime is shorter than the lead is always inside its refresh window, and an unfloored schedule would hammer the token endpoint.
- A failed refresh is recorded, retried with a 5s→60s backoff, and **never** fails the run; an `authorization_code` refusal never pops interactive auth.
- The four worker exit paths now join through `RunContext::join_aux_threads` rather than four copies of a per-thread join.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

All green locally on this branch (base `ec194d1`).

- `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` - **1262/1262 passed**. New `auth_refresh_test.cpp` (12 tests) covers the schedule, the swap, the eligibility matrix, and three whole runs against an in-process mock IdP minting 3s tokens: a run that outlives its token sends `Bearer AT1` then `Bearer AT2` (mutation check: drop the `sync_auth_header` call and the target only ever sees one credential), a refused refresh is recorded while the run still completes, and a non-expiring token spawns no watchdog and reports no section. `runs_route_test.cpp` gains the `auth` summary round-trip plus its absence twin.
- `cd app && pnpm test` - **382 files / 3822 tests passed**, including the reworked `coverageState` matrix (refreshable long run → covered; query-placed / opted-out / interactive → still gated) and `LoadTestDetail.auth.test.tsx`, which renders the note rather than scanning source for it.
- `pnpm type-check`, `pnpm lint`, `prettier --check` on the touched files - clean. `clang-tidy -p build` on the new/changed engine sources - only the pre-existing `missing-field-initializers` warning at `auth_resolver.cpp:84`, which this change does not touch.

## Checklist

Docs in the same commit: `docs/engine/db-schema.md` (the "There is **no** mid-run refresh" sentence is gone, replaced by the eligibility list), `docs/engine/api-reference.md` (report `auth` section + the conditional-sections line), `docs/engine/architecture.md` (per-run thread inventory, the swap's threading contract, `plan_auth_refresh` in the auth-resolver bullet), `docs/engine/benchmarks.md` (the `oauth2RefreshLeadMs` key), `docs/app/api-integration.md` (`report.auth` and the guard's mirrored rule).

## Related Issues
Closes #478

---
_Generated by [Claude Code](https://claude.ai/code/session_01GrsuygKbeyS7AzFfApz8SF)_

### DIFF
--- a/app/src/modules/history/main/LoadTestDetail.auth.test.tsx
+++ b/app/src/modules/history/main/LoadTestDetail.auth.test.tsx
@@ -1,0 +1,118 @@
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * @vitest-environment jsdom
+ */
+
+/**
+ * The mid-run OAuth 2.0 refresh note in LoadTestDetail's header (#478).
+ *
+ * The wording is unit-tested in auth-refresh-note.test.ts; what this file adds
+ * is that the pane actually draws it - a note computed and never rendered is
+ * this codebase's most repeated defect, and no source scan can see it.
+ */
+import { describe, it, expect } from "vitest";
+import { render, screen } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { TooltipProvider } from "@/components/ui";
+import type { ReactElement } from "react";
+import LoadTestDetail from "./LoadTestDetail";
+import type { RunReport } from "@/types";
+
+function renderWithClient(ui: ReactElement) {
+	const qc = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+	return render(
+		<QueryClientProvider client={qc}>
+			<TooltipProvider>{ui}</TooltipProvider>
+		</QueryClientProvider>
+	);
+}
+
+function report(auth?: RunReport["auth"]): RunReport {
+	return {
+		metadata: {
+			runId: "r",
+			runType: "load",
+			status: "completed",
+			startTime: 0,
+			endTime: 1000,
+			requestUrl: "http://127.0.0.1:8080/x",
+			requestMethod: "GET",
+			configuration: { mode: "constant_rps", duration: "2h" },
+		},
+		summary: {
+			totalRequests: 600,
+			successfulRequests: 600,
+			failedRequests: 0,
+			errorRate: 0,
+			totalDurationSeconds: 7200,
+			avgRps: 0.08,
+		},
+		latency: {
+			min: 100,
+			max: 130,
+			avg: 101,
+			median: 101,
+			p50: 101,
+			p75: 101,
+			p90: 102,
+			p95: 103,
+			p99: 108,
+			p999: 120,
+		},
+		statusCodes: { "200": 600 },
+		errors: { total: 0, withDetails: 0, types: {} },
+		auth,
+	};
+}
+
+describe("LoadTestDetail auth-refresh note", () => {
+	it("says when the run's token was refreshed", () => {
+		renderWithClient(
+			<LoadTestDetail
+				report={report({ refreshes: [{ atSeconds: 3600 }], refreshFailures: 0 })}
+				runId="r"
+			/>
+		);
+
+		expect(screen.getByText("Access token refreshed at 1h 0m")).toBeInTheDocument();
+	});
+
+	// The case that explains a 401 storm: the run kept going on a credential it
+	// could not renew, so the reason has to reach the reader.
+	it("names the reason a refresh failed", () => {
+		renderWithClient(
+			<LoadTestDetail
+				report={report({
+					refreshes: [],
+					refreshFailures: 1,
+					lastError: "oauth2_provider_error: invalid_grant",
+				})}
+				runId="r"
+			/>
+		);
+
+		expect(
+			screen.getByText("1 refresh failure - oauth2_provider_error: invalid_grant")
+		).toBeInTheDocument();
+	});
+
+	it("stays quiet for a run that could not refresh at all", () => {
+		renderWithClient(<LoadTestDetail report={report(undefined)} runId="r" />);
+
+		expect(screen.queryByText(/refresh/i)).not.toBeInTheDocument();
+	});
+
+	it("stays quiet for a watched run that never needed a refresh", () => {
+		renderWithClient(
+			<LoadTestDetail report={report({ refreshes: [], refreshFailures: 0 })} runId="r" />
+		);
+
+		expect(screen.queryByText(/refresh/i)).not.toBeInTheDocument();
+	});
+});
--- a/app/src/modules/history/main/LoadTestDetail.tsx
+++ b/app/src/modules/history/main/LoadTestDetail.tsx
@@ -20,6 +20,7 @@ import {
 	Settings2,
 	AlertTriangle,
 	ListOrdered,
+	KeyRound,
 } from "lucide-react";
 import {
 	Badge,
@@ -39,6 +40,7 @@ import { computeBreakpoint } from "@/modules/dashboard/utils/computeBreakpoint";
 import { useRunTimeSeriesQuery } from "@/queries/runs";
 import { useClientSettingsStore } from "@/stores";
 import { OverviewTab, PerformanceTab, SamplesTab, ScenarioStepsTab } from "./components";
+import { authRefreshNote } from "./auth-refresh-note";
 import type { LoadTestDetailProps, TimeSeriesResponse } from "../types";
 
 export default function LoadTestDetail({ report, runId }: LoadTestDetailProps) {
@@ -119,6 +121,10 @@ export default function LoadTestDetail({ report, runId }: LoadTestDetailProps) {
 		if (timeSeries.length < 2) return base;
 		return { ...base, breakpoint: computeBreakpoint(timeSeries, sloThresholdMs) };
 	}, [report, timeSeries, sloThresholdMs]);
+
+	// One line on whether the run's OAuth 2.0 credential was kept current - the
+	// answer to 401s that appear partway through an otherwise healthy run.
+	const authNote = useMemo(() => authRefreshNote(report.auth), [report.auth]);
 
 	const successRate =
 		report.summary.totalRequests > 0
@@ -238,6 +244,24 @@ export default function LoadTestDetail({ report, runId }: LoadTestDetailProps) {
 								<span className="text-foreground/90 italic">{config.comment}</span>
 							</div>
 						)}
+					</div>
+				)}
+
+				{/* What kept the run authorized. Drawn only when the engine was able to
+				    refresh at all - a run without the section reports nothing here,
+				    exactly as every run did before mid-run refresh existed. */}
+				{authNote && (
+					<div
+						className={`flex items-center gap-2 text-sm mb-3 p-3 border rounded-md bg-background/50 ${
+							authNote.warning ? "text-status-warning-text" : "text-muted-foreground"
+						}`}
+					>
+						{authNote.warning ? (
+							<AlertTriangle className="w-4 h-4 shrink-0" />
+						) : (
+							<KeyRound className="w-4 h-4 shrink-0" />
+						)}
+						<span>{authNote.text}</span>
 					</div>
 				)}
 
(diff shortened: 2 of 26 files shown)


athrvk:
make the new keys user configurable - constants etc

athrvk:
Done in 9132df3 - all four numbers the watchdog runs on are settings now, not constants.

Seeded under **Network & Connectivity**, each with a label, description and a range `POST /config` enforces:

|---|---:|---|---|
| `oauth2RefreshLeadMs` | 60000 | 1000 - 3600000 | How far ahead of expiry a run renews |
| `oauth2RefreshMinIntervalMs` | 1000 | 100 - 3600000 | Floor between two renewals |
| `oauth2RefreshRetryMs` | 5000 | 250 - 600000 | First wait after a refused renewal |
| `oauth2RefreshRetryMaxMs` | 60000 | 1000 - 3600000 | Ceiling on that backoff |

The constants stay as the seeded defaults, so there is one set of numbers rather than two, and `AuthRefreshTuning` / `read_auth_refresh_tuning` carry them into both the watchdog and the schedule. They are read once when a run arms its watchdog - the same "applies to the next run started, no restart" behaviour as `workers` and `dnsCacheTimeout` - so a run's cadence cannot shift under it half way through. No app change was needed: Settings renders engine entries dynamically from `GET /config`, and `network_performance` is an existing category.

One number stayed a constant on purpose: the 100 ms slice the watchdog sleeps in. It only bounds how promptly a finished run joins the thread - nothing about a run's behaviour changes with it - so it would be a knob with no meaning to a user. Say the word if you want it exposed anyway.

Also hardened while in there: a non-positive stored value can only come from a hand-edited row (the route rejects one against the seeded minimum), and a zero floor would turn the schedule into a tight loop against the token endpoint, so the reader falls back to the default rather than trusting it - the same treatment `live_ring_size` gives its inputs.

Tests: four new cases - every key is seeded with a label/description/range, the stored values are what a run reads (mutation-checked per key), a non-positive value falls back, and the schedule's floor follows `min_interval_ms` rather than the constant. `ctest -R "AuthRefresh|SyncAuth|PlanAuth|MidRunRefresh|ConfigRoute|RunsRoute"` is 66/66 green locally; docs (`db-schema.md`, `benchmarks.md`, `architecture.md`) updated in the same commit.

---

athrvk:
Added the watchdog's own knob too, in 10a787a - `oauth2RefreshPollIntervalMs` (default 100ms, range 10ms - 5s), so all five numbers the watchdog runs on are settings and none are constants.

It controls how often the watchdog wakes while waiting, to notice the run has ended. That is not invisible after all: the worker joins that thread before it writes the report, so this is what bounds how long a finished run's last moments take. It is carried on `AuthRefreshTuning` and honoured at **both** wait sites - the expiry schedule and the retry backoff - so it cannot half-apply. The seeded-metadata, stored-value and non-positive-fallback tests all cover it alongside the other four.

Docs updated in the same commit (`benchmarks.md` table, `db-schema.md`, `architecture.md`).

Note on verification for this last commit: the local engine rebuild is still running - touching `constants.hpp` invalidates most translation units and this container has 4 cores - so CI is the first full check on it. I'm watching the run and will push a fix if anything reddens.

---

athrvk:
fix merge conflicts

athrvk:
Conflicts resolved and pushed (722653a) - merged `master` after #488 landed.

Five files conflicted, all of them "keep both": #488's capacity work and this branch add sibling report sections, never competing ones.

| File | Resolution |
|---|---|
| `run_manager.hpp` | Both includes (`auth_refresh.hpp` + `capacity_controller.hpp`) |
| `routes/runs.cpp` | Both `ReportExtras` blocks, and both summary readers as separate `if`s |
| `runs_route_test.cpp` | Both round-trip tests kept - the conflict boundary had split the auth-omission test mid-body, so its closing brace was restored |
| `types/domain.ts` | Both `RunReport` sections: `auth`, then `capacity` |
| `api-reference.md` | The conditional-sections line now names both; capacity's own section restored with the `auth` paragraph after it |

Nothing about either feature changed in the merge - `auth` and `capacity` are independent optional sections on the same report, and both round-trip tests now sit side by side.

Verified so far: `pnpm type-check` clean, `prettier --check` clean on the merged type file. The engine rebuild is still running locally (the merge brought in the capacity sources); CI on the merge is the fuller check and I'm watching it.

---

PR #490, opened 2026-08-10
feat(engine/app): server-vitals overlay for load runs

### LINKED ISSUE #475: engine/app(loadtest): server-vitals overlay - scrape a user-supplied metrics endpoint during a load run and chart it on the same timeline as p99 and rps
First of an observability trio (this, per-phase histograms, anomaly annotations) attacking the most-reported load-testing pain: "the tool showed response times degrading and I had no idea why." The accepted community answer is "set up Grafana + Prometheus and re-run" - a whole second stack to see the server's CPU next to the client's p99. Vayu can put both on one screen, locally.

## Problem

A Vayu run knows everything about the client side and nothing about the target. When p99 climbs, the user cannot see whether server CPU, memory, or connection count climbed with it - the engine's per-tick series (`MetricTickSample`, `engine/include/vayu/core/run_manager.hpp:290-309`) has no channel for external data, and the dashboard has nothing to draw that it doesn't produce itself. Most services already expose the needed numbers on a Prometheus `/metrics` or a JSON status endpoint; the gap is purely that nothing samples it during the run and stores it beside the ticks.

## Fix / solution pathway

**1. Payload.** Optional `monitor` block on `POST /runs`:

```json
"monitor": {
  "url": "http://localhost:9100/metrics",
  "intervalMs": 1000,
  "format": "prometheus",
  "series": ["node_cpu_seconds_total", "process_resident_memory_bytes"]
}
```

`format` is `"prometheus"` (text exposition; a minimal parser for `name{labels} value` lines, no dependency needed) or `"json"` (flat object of numbers; `series` lists keys). Validate in `validate_run_config` (`engine/src/http/routes/execution.cpp:434-532`): http(s) URL, `intervalMs` in [250, 60000], `series` non-empty, max 8 entries. Loopback/private targets are allowed - this is a local tool scraping the user's own infra; no allowlist.

**2. Scrape thread - NOT the tick thread.** The metrics thread must not make blocking calls: it is a fixed-cadence sampler (`sleep_for(tick_interval_ms)` with no deadline compensation, `run_manager.cpp:1295`), its whole body sits in one `try` whose catch ends live metrics for the run (:1263-1267), and the only HTTP client available is the synchronous `vayu::http::Client`. Follow the existing per-run auxiliary-thread precedent instead - `context->metrics_thread` spawned in `RunManager::start_run` (`run_manager.cpp:733`) and joined by the worker (:893): add a `monitor_thread` on `RunContext` that loops on its own `intervalMs`, GETs the URL with a timeout well under the interval, parses the named series, and appends samples to a mutex-guarded buffer on `RunContext`. A failed scrape records a gap, never throws outward, and after 5 consecutive failures logs once and keeps trying at a backed-off interval.

**3. Persist + serve.** New table `monitor_samples (id, run_id, timestamp, payload)` shaped exactly like `metric_ticks` (`engine/src/db/database.cpp:255-259`), written at the scrape cadence, deleted by the run cascade (`remove_run_cascade_locked`, `engine/include/vayu/db/database.hpp:298-305`). Do NOT extend `metric_ticks`' payload - its key set is the `GET /runs/:id/metrics` contract pinned by `stats_route_test.cpp` (`run_manager.hpp:311-320` says so). Serve history as `GET /runs/:id/monitor` with the same `{data[], pagination{}}` envelope as `run_time_series_response` (`engine/src/http/routes/metrics.cpp:92-103`). For live: emit `event: monitor` frames through the same `RunContext` tick ring the SSE endpoint drains (`ticks_since`/`TickBatch`, `run_manager.hpp:195-228`) - frames are pre-framed strings sharing one id counter, so `Last-Event-ID` replay keeps working; verify that assumption against `live_metrics` (`metrics.cpp:345-404`) and say so in the PR.

**4. Report.** Optional `monitor` summary section (per-series min/max/avg + sample count) via the standard four-hop pattern (`RunSummaryInputs` → `build_run_summary_payload` `run_manager.cpp:1110` → `apply_run_summary`/`ReportExtras` `runs.cpp:203-306`/:119-180 → `run_report_response`). Absent when no monitor was configured.

## App changes

- Config: a "Server monitoring" collapsible in `LoadTestConfigDialog` (beside "Recording & limits", `.../LoadTestConfigDialog/index.tsx:547`) - URL, interval, series names; persisted in `SavedLoadTestConfig` (:68); through `StartLoadTestRequest` (`app/src/types/api.ts:344-394`). Note: #470's new "Pass/fail budgets" card sits in the same dialog (~:660).
- Live: `sse-client.ts` gains a `monitor` event handler beside `metrics`/`step` (`SSEClient.connect`, `app/src/services/sse-client.ts:139`); a monitor slice on `useDashboardStore`.
- Charts: `UPlotChart` already supports a second scale - `y2Format` (`UPlotChart.tsx:78`), `hasY2` (:150), exercised by `RequestRateChart`. The constraint is `uPlot.AlignedData`'s shared x column: resample monitor samples onto the tick timeline (nearest-tick join in a pure util, unit-tested). Add a "Server vitals" chart row in `MetricsView` (chart rows at `MetricsView.tsx:179-483`) as a semantic wrapper - consumers never touch `UPlotChart` directly - joined to the `CHART_SYNC` live/history group so cursor sync lines up CPU with p99.
- History: `LoadTestDetail` fetches `GET /runs/:id/monitor` (new query beside `useRunTimeSeriesQuery`, `app/src/queries/runs.ts:207`) and renders the same row.

## Tests

- Engine: prometheus-line parser unit tests (typical exposition text, labels, NaN, comments; mutation-check a parse rule). Scrape loop against a `SlowMockServer` route serving canned metrics - samples land in `monitor_samples`, a hanging endpoint (`/hang`) does not stall the run or the tick cadence (assert tick count unaffected). Round-trip + absence twin in `runs_route_test.cpp`. Validation tests in `run_config_validation_test.cpp`.
- App: resample util (uneven cadences join correctly - mutation-check), monitor event → store, chart wrapper renders with y2 series, history tab fetch.

## Docs (same commit)

`docs/engine/api-reference.md` (`POST /runs` payload, new `GET /runs/:id/monitor`, live SSE event types, report section), `docs/engine/db-schema.md` (new table + cascade), `docs/engine/architecture.md` (the per-run thread inventory), `docs/engine/mcp.md` if `start_load_run` grows the block.

## Sequencing

Independent of #470-#473 and of everything open. The anomaly-annotation issue (filed after this) can optionally mark monitor-series spikes too - coordinate naming, nothing more. Parallel-safe. Conflict surface, not dependency: #478 (run_manager.cpp thread setup + runs.cpp report section), #471 (RunContext + the report four-hop), #477 (MetricsView.tsx, LoadTestDetail.tsx) - trivial rebases, coordinate whichever lands second.

## Acceptance criteria

- [ ] A run configured with a `monitor` block records external samples for its whole duration, on its own thread, without perturbing tick cadence even against a hanging endpoint.
- [ ] Live dashboard and history detail both chart the scraped series on the shared timeline with cursor sync against the run's own charts.
- [ ] Samples are pruned with the run; `metric_ticks`' contract is untouched (its pinning test unmodified).
- [ ] A failing scrape degrades to gaps and a single log line, never a failed run.
- [ ] Report carries the summary section only when configured; suites green; docs same commit.

---

athrvk:
## Description

A run knew everything about the client side and nothing about the target: when p99 climbed, nothing said whether the server's CPU climbed with it. A run may now declare a `monitor` block naming a metrics endpoint, which the engine scrapes for the life of the run and both dashboards chart on the same timeline as p99 and throughput.

**Plan.** The root cause is that the per-tick series (`MetricTickSample`) has no channel for external data and the dashboard can only draw what it produces itself. The approach follows the existing per-run auxiliary-thread precedent: a `monitor_thread` on `RunContext` beside the metrics thread, its own `monitor_samples` table shaped exactly like `metric_ticks`, its own `GET /runs/:id/monitor` endpoint in the shared envelope, live `monitor` frames on the ring the SSE endpoint already drains, and a `monitor` report section through the standard four-hop. **The alternative I rejected was widening `metric_ticks`** - either as extra payload keys or by scraping on the metrics thread. Both fail for the same reason from two directions: that payload's key set *is* the `GET /runs/:id/metrics` contract (pinned by `stats_route_test.cpp`), and the metrics thread is a fixed-cadence sampler with no deadline compensation, so a blocking scrape inside it delays every later tick and a hanging endpoint would end live metrics for the whole run.

Two design points worth a reviewer's eye:

- **Frames are framed under the ring's lock.** `RunContext::append_event` assigns the SSE id inside `tick_mtx`, because a monitored run has *two* producers on that ring and the previous `published_count.load()`-then-`append_tick` pattern would hand both the same id, breaking `Last-Event-ID` resume. I verified the assumption the issue asked about: `live_metrics` (`metrics.cpp`) replays pre-framed strings and terminates on `closed && offset >= published_count`, so interleaved `monitor` frames sharing that id space replay correctly. `build_tick_payload` / `build_step_payload` now delegate to the same framing helper, so there is one copy of the wire shape.
- **Samples are joined by wall clock, not elapsed seconds.** A tick's `elapsed_seconds` starts at the first *persisted* tick while the scrape starts with the run, so joining on it would offset the two series by ~1s. `joinMonitorToTimeline` joins on `timestamp` and holds each reading until the next scrape, going null once it is older than the observed cadence allows - which is what makes a failed scrape a gap rather than a plateau over the outage.

### Settings (review follow-up)

Two of the five constants this feature introduced are things a user would reach for, and are seeded as engine settings the way #478's `oauth2Refresh*` keys are, so the Settings panel renders them with no app change:

|---|---|---|---|
| `monitorIntervalMs` | 1000 | 250–60000 | Cadence a run scrapes at when its `monitor` block names none |
| `monitorMaxSeries` | 8 | 1–64 | How many metrics one run may chart |

Both are read per run - a change applies to the next run started. The other three stay constants deliberately: the interval *bounds* stop a cadence that measures the scraper rather than the target, and the backoff threshold is how politely the loop gives up on a dead endpoint. `read_monitor_limits(db)` is the single reader, and the decision functions take a resolved `MonitorLimits` rather than a `Database` - the split `resolve_request_timeout_ms` already draws in the execution route. The app reads the same two entries through `useMonitorSettings`, because this dialog always sends an explicit `intervalMs`, so a renderer-local default would leave the engine's copy with no reader on the UI path.

### Deliberate deviations

Both tracked in #491 rather than left silent:

- **`start_load_run` (MCP) does not grow the block**, which the issue listed as conditional ("if"). Keeping it out holds this diff to the app's own path - and the monitor URL is a second host an ad-hoc agent run would contact, which is an allowlist decision that deserves its own review rather than a footnote here.
- **The app's `RunReport` type carries the report's `monitor` section but nothing renders it as a panel.** It has a reader - the history view gates its `GET /runs/:id/monitor` fetch on `report.monitor.samples` - so it is not the repo's "written but never read" defect; a per-series summary panel beside the chart is deferred.

Also: three categorical chart roles were added for a plot whose series count the *user* decides. `--chart-4` (amber) is deliberately excluded - it measures 2.38 against a light card, under the 3.0 a series must clear (`status-token-contrast.test.ts`, which derives its list from `ROLE_TOKEN` and caught this).

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Testing

All green locally on the merged branch (base `9880a02`, which includes #488 and #478):

- **Engine** - `python build.py -e -t && cd engine && ctest --preset linux-dev`: **1330/1330 passed** (159s). New: `engine/tests/monitor_test.cpp` (26 tests - validation gate and run-reader parity, Prometheus parsing incl. labelled-family summing, `NaN`/`+Inf`, CRLF, a `}` inside a quoted label value, JSON parsing, `MonitorTotals`, the four config-limit tests, plus five whole-run tests driving `RunManager::start_run` against a canned `/vitals` endpoint). The acceptance test is `AHangingEndpointStallsNeitherTheRunNorTheTicks`: the run points its monitor at `/hang` (30s hold), still finishes in time, still persists its per-second ticks, stores no sample, and reports the gaps. Also added: 2 report round-trip/absence tests (`runs_route_test.cpp`), 3 series-endpoint tests (`stats_route_test.cpp`), 4 route-validation tests (`run_config_validation_test.cpp`), 1 cascade test.
- **App** - `pnpm test`: **3889/3889 passed** (386 files). New: `monitorSeries.test.ts` (10), `monitor.test.ts` (12), plus cases in the SSE, store, chart-smoke and dialog suites. `pnpm type-check`, `pnpm lint` and `pnpm format:check` clean.
- **Docs** - `mkdocs build --strict` clean.

Two config-backed behaviours are mutation-checked rather than merely asserted: reverting the configured-cadence line turns `TheConfiguredIntervalIsWhatABlockWithoutOneScrapesAt` red, and the cap test binds in both directions (9 names rejected at 8, accepted at 12).

One pre-existing flake, unrelated and not touched: `SentRequestHeadersTest.DefaultUserAgentReachesTheTestScript` timed out once under a loaded parallel run and passes in isolation and on re-run.

## Checklist

## Related Issues
Closes #475
Follow-ups deferred here: #491

### DIFF
--- a/app/src/config/api-endpoints.ts
+++ b/app/src/config/api-endpoints.ts
@@ -108,6 +108,12 @@ export const API_ENDPOINTS = {
 	STATS_TIME_SERIES: (runId: string, limit = STATS_PAGE_LIMIT, offset = 0) =>
 		`/runs/${runId}/metrics?limit=${limit}&offset=${offset}`,
 
+	// Server vitals scraped during the run (JSON, paginated, same envelope).
+	// Its own endpoint rather than extra keys on the tick objects: those keys
+	// are the /metrics contract, and scrapes land on the user's own cadence.
+	RUN_MONITOR: (runId: string, limit = STATS_PAGE_LIMIT, offset = 0) =>
+		`/runs/${runId}/monitor?limit=${limit}&offset=${offset}`,
+
 	// Import. FETCH proxies a remote collection past CORS; APPLY persists the
 	// whole parsed tree in one atomic call and returns the temp-id -> real-id map
 	// (the import path no longer creates items one POST at a time).
--- a/app/src/constants/monitor.ts
+++ b/app/src/constants/monitor.ts
@@ -1,0 +1,20 @@
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * Server-monitoring bounds, mirroring `constants::monitor` engine-side.
+ *
+ * The interval **bounds** are fixed on both sides - they exist to stop a
+ * cadence that measures the scraper rather than the target. `DEFAULT` and
+ * `MONITOR_MAX_SERIES` are only the pre-config seeds: the live values come from
+ * the `monitorIntervalMs` / `monitorMaxSeries` engine settings via
+ * {@link useMonitorSettings}, and these stand in until that query resolves.
+ */
+export const MONITOR_INTERVAL_MS = { MIN: 250, MAX: 60_000, DEFAULT: 1000 } as const;
+
+/** Seed for `monitorMaxSeries`; the engine rejects a longer list with a 400. */
+export const MONITOR_MAX_SERIES = 8;
--- a/app/src/hooks/useMonitorSettings.ts
+++ b/app/src/hooks/useMonitorSettings.ts
@@ -1,0 +1,60 @@
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * useMonitorSettings
+ *
+ * The two server-monitoring limits a user can move: the default scrape cadence
+ * and how many metrics one run may chart.
+ *
+ * They live **engine-side** (`monitorIntervalMs` / `monitorMaxSeries`), like
+ * the live-chart settings beside them, and for the same reason: the engine
+ * applies the cadence to any run that names none and rejects a series list over
+ * the cap with a 400. A renderer-local copy could only drift - and it would
+ * drift in the two directions that hurt most, a dialog that always sends an
+ * explicit interval (so the setting would never apply to a run started here)
+ * and one that refuses a metric list the engine would have accepted.
+ *
+ * Read-only: both are edited from the engine settings list, and this hook is
+ * how the load-test dialog seeds and bounds its own fields from them. Until the
+ * config query resolves, the module defaults stand - the same numbers the
+ * engine seeds, so the gap changes nothing.
+ */
+
+import { useConfigQuery } from "@/queries";
+import { MONITOR_INTERVAL_MS, MONITOR_MAX_SERIES } from "@/constants/monitor";
+
+export interface MonitorSettings {
+	/** Cadence to seed a fresh draft with, in ms. */
+	defaultIntervalMs: number;
+	/** How many metric names the dialog accepts before it blocks Start. */
+	maxSeries: number;
+}
+
+export function useMonitorSettings(): MonitorSettings {
+	const { data: config } = useConfigQuery();
+
+	const entryValue = (key: string): string | undefined =>
+		config?.entries?.find((e) => e.key === key)?.value;
+
+	// Entry values are strings. An absent key - config still loading, or an
+	// engine older than these settings - leaves the default rather than parsing
+	// `undefined` to NaN.
+	const rawInterval = Number(entryValue("monitorIntervalMs"));
+	const defaultIntervalMs =
+		Number.isFinite(rawInterval) &&
+		rawInterval >= MONITOR_INTERVAL_MS.MIN &&
+		rawInterval <= MONITOR_INTERVAL_MS.MAX
+			? rawInterval
+			: MONITOR_INTERVAL_MS.DEFAULT;
+
+	const rawMaxSeries = Number(entryValue("monitorMaxSeries"));
+	const maxSeries =
+		Number.isFinite(rawMaxSeries) && rawMaxSeries > 0 ? rawMaxSeries : MONITOR_MAX_SERIES;
+
+	return { defaultIntervalMs, maxSeries };
+}
--- a/app/src/modules/dashboard/components/MetricsView.tsx
+++ b/app/src/modules/dashboard/components/MetricsView.tsx
@@ -48,6 +48,7 @@ import {
 	StatusCodesOverTimeChart,
 	ResponseTimeVsConcurrencyChart,
 	HdrPercentileChart,
+	ServerVitalsChart,
 	CHART_SYNC,
 } from "./charts/uplot";
 import { SkeletonHdrPlot } from "./charts/HdrPercentilePlot";
@@ -93,6 +94,9 @@ function MetricsView({
 	// of an O(n) scan over the full historicalMetrics buffer.
 	const peakConcurrency = useDashboardStore((s) => s.peakConcurrency);
 	const breakpoint = useDashboardStore((s) => s.breakpoint);
+	// Empty for a run that configured no monitor - the row below reads that
+	// rather than the config, so it is right for a replayed run too.
+	const monitorSamples = useDashboardStore((s) => s.monitorSamples);
 
 	// Only the latest tick's elapsed_seconds is consumed below - depending on the
 	// whole historicalMetrics array would rebuild `derived` every tick (10 Hz) and
@@ -419,6 +423,28 @@ function MetricsView({
 					</div>
 					<StatusCodesOverTimeChart
 						history={chartWindow}
+						isCompleted={isCompleted}
+						syncKey={CHART_SYNC.live}
+					/>
+				</div>
+			)}
+
+			{/* Server vitals - only for a run that scraped an endpoint, so a run
+			    without a monitor block keeps the dashboard it always had. */}
+			{monitorSamples.length > 0 && chartWindow.length > 1 && (
+				<div className="bg-card border border-border rounded-md p-3.5">
+					<div className="flex items-baseline justify-between mb-3">
+						<h3 className="text-xs font-semibold text-foreground">
+							Server vitals
+							<InfoChip tip={TOOLTIPS.serverVitals} />
+						</h3>
+						<span className="text-[10px] font-mono text-muted-foreground">
+							scraped from the target
+						</span>
+					</div>
+					<ServerVitalsChart
+						history={chartWindow}
+						samples={monitorSamples}
 						isCompleted={isCompleted}
 						syncKey={CHART_SYNC.live}
 					/>
--- a/app/src/modules/dashboard/components/charts/uplot/TimeSeriesCharts.tsx
+++ b/app/src/modules/dashboard/components/charts/uplot/TimeSeriesCharts.tsx
@@ -16,13 +16,14 @@
 
 import { useMemo } from "react";
 import type uPlot from "uplot";
-import type { LoadTestMetrics } from "@/types";
+import type { LoadTestMetrics, MonitorSample } from "@/types";
 import {
 	buildLatencyChartData,
 	buildPercentileChartData,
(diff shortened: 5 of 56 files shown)


athrvk:
- fix merge conflicts
- are the new constants/configs user configurable? If it makes sense to have them user configurable make them.

athrvk:
Claiming this - picking up your feedback on this PR now (merge conflicts + the user-configurability question on the new constants). Will push to this branch.

---

athrvk:
Both done - pushed as two commits.

## Merge conflicts

Master had gained #488 (capacity) and #478 (oauth2 refresh), which collided in three real places rather than textually. Each is resolved toward master's shape rather than beside it:

- **#478 added `RunContext::join_aux_threads()`** - the same idea as this branch's free `join_run_threads` helper. The free one is deleted and the scrape loop joins inside the member, so all three auxiliary threads (tick, refresh, scrape) have one join list instead of two that could disagree.
- **The tick emitter** keeps #488's `publish_live_tick` feedback call *and* frames through `append_event`, so the capacity search still gets its per-tick numbers and the SSE id is still assigned under the ring's lock.
- **The summary payload, the run-config validator and the load-test dialog** each carry all three features' sections side by side; the dialog's blocking-error chain reads ramp → start → capacity → budgets → monitoring.

Verified after merging, not just after resolving: engine **1330/1330**, app **3889/3889**, type-check + lint + format + `mkdocs --strict` clean.

## Configurable constants

Five constants came in with this feature. Two of them are things a user would actually reach for, and they are now settings - seeded through `seed_default_config` exactly like your `oauth2Refresh*` keys, so the Settings panel renders them with no app change:

|---|---|---|---|
| `monitorIntervalMs` | 1000 | 250–60000 | Cadence a run scrapes at when its `monitor` block names none |
| `monitorMaxSeries` | 8 | 1–64 | How many metrics one run may chart; the rejection message now names the setting |

Both are read per run, so a change applies to the next run started - no restart.

**The other three stay constants, deliberately.** The interval *bounds* (250/60000) exist to stop a cadence that measures the scraper rather than the target, and `FAILURES_BEFORE_BACKOFF` is how politely the loop gives up on a dead endpoint. Those are rails and internal manners, not preferences - the same reason `threshold_eval`'s budget ranges are constants while its budgets are per-run. Happy to expose any of them if you disagree; it's a one-line seed each. All five now live in `constants::monitor` beside every other tunable's seed.

Two things worth flagging in how it's wired:

- `read_monitor_limits(db)` is the single reader for both keys, and the decision functions take a resolved `MonitorLimits` rather than a `Database` - the same split `resolve_request_timeout_ms` already draws in the execution route, so the core stays total over arbitrary JSON and testable without a database.
- **The app reads the same two entries** (`useMonitorSettings`, mirroring `useLiveChartSettings`). Without that, `monitorIntervalMs` would have been a setting with no reader on the path the app takes - this dialog always sends an explicit `intervalMs`, so the engine's default would never have applied to a run started from the UI - and a raised cap would still have been refused client-side.

New tests: four on the limits reader (seeded defaults; the configured cadence reaching a block that names none, **mutation-checked**; the cap binding in both directions; a hand-edited out-of-range value falling back to the seed), and three on the dialog seeding and bounding itself from the settings.

---

athrvk:
Standing down on my claim above - a concurrent routine run pushed the same two fixes (`bc1d524` merge, `90a6bcf` settings) while I was working, and CI is green across all three platforms on that head. I've discarded my duplicate rather than force-pushing over it.

One substantive difference worth your call, since we landed on opposite sides of it. What shipped exposes `monitorIntervalMs` and `monitorMaxSeries` (and wires the dialog to read them, which is the right half to have). I had instead exposed the scrape loop's three operational knobs, which that commit deliberately keeps as constants:

- **`monitorScrapeTimeoutMs`** - one scrape's timeout. Today it is always three quarters of the interval; an exposition slow enough to exceed that reads as a permanent gap with no way to fix it short of slowing the whole cadence. This is the one I'd argue is a real gap rather than a preference.
- **`monitorPollIntervalMs`** - the 50ms slice the loop sleeps in. Exactly the knob `oauth2RefreshPollIntervalMs` already is, for the same reason (it bounds how long a finished run waits on the join).
- **`monitorFailuresBeforeBackoff`** - I'd agree this one is closer to a rail than a setting; the "internal manners" framing holds.

Nothing here blocks the PR. If any of it is worth having I'll open a follow-up issue rather than reopening this branch - say the word.

---

PR #493, opened 2026-08-10
feat(app/cli): headless load-run gate for CI

### LINKED ISSUE #473: engine(cli): extend vayu-cli into the headless CI gate - wait for the run, read the threshold verdict, exit 0/1/2 (PARKED)
> **Parked by owner request (2026-08-11). Do not pick this up until explicitly asked.** Re-scoped from the Node/`dist-electron` home to the C++ `vayu-cli` after PR #493 was closed unmerged - the placement decision and its rationale are recorded in that PR's closing comment.

## Problem

#470 gave the load-run report a machine verdict (`thresholdValidation.verdict`) and nothing headless can act on it. The existing C++ `vayu-cli` (`engine/src/cli.cpp`, `docs/engine/cli.md`) starts a load run from a JSON config against the running daemon, prints a monitor URL, and exits - no wait, no verdict, no gating exit code - so a CI job built on it goes green regardless of latency. The MCP path reads the verdict but requires the app installation; a CI runner has only the engine binary.

**Why the C++ CLI is the home (decided)**: the gate assumes a running engine, so CI already ships the engine binary - a gate inside `vayu-cli` adds zero dependencies, where a Node gate would add Node plus the built `dist-electron` tree. The engine itself validates thresholds (`validate_run_config`) and writes the report the verdict lives in; no TS types are needed to read one field.

## Fix / solution pathway

Extend `vayu-cli` (same binary, new flags), porting the decided semantics from closed PR #493 - its PR body and `gate.test.ts` list are the behavioural spec:

1. **Flags**: budget flags mapping to the engine's threshold keys (`--p50/--p95/--p99` ms, `--error-rate` pct, `--min-rps`), at least one required to gate; `--wait` (or gating implied by any budget flag - decide and state); `--timeout` for the wait deadline (default: run duration + grace); `--json` to emit the raw report; `--engine-url`. Load shape flags as `vayu-cli` has today, restricted to `constant_concurrency` and `constant_rps` (the duration-bounded modes; `capacity`/`ramp_up`/`iterations` refused by name with a pointer to the app) with the `--target-rps`-vs-`--concurrency` pairing rules from the PR (each refused in the other's mode).
2. **Behaviour**: build the `POST /runs` payload carrying `thresholds` (the engine validates ranges - do not duplicate bounds client-side beyond flag parsing); poll `GET /runs/:id` through `pending`/`running` to a terminal status with the deadline; fetch the report; map to exit codes.
3. **Exit codes (the contract)**: `0` every budget met; `1` a budget missed; `2` no verdict possible - usage error, engine unreachable, run `failed`/`stopped`, wait timeout, or a report with no `thresholdValidation` (absence is "broken", never "pass"). Human output names every failed check with limit and actual; `--json` emits the report verbatim.
4. Keep the logic in a testable core (a function taking parsed args + an injected HTTP client + clock), with `main` as the thin shell - the same discipline PR #493 used TS-side.

## App changes

None, verified - the app and MCP layers are untouched; `buildLoadRunPayload`/`start_load_run` stay as they are.

## Tests

Engine gtest over the extracted core: flag grammar (unknown/duplicate/misparied flags exit 2; bounds refused by the engine's own 400 surfaced verbatim), poll loop (terminal statuses, deadline, injected clock), exit-code mapping including the absent-verdict case (mutation-check: swap 0/1, treat absence as pass - both must redden), `--json` passthrough. One live smoke against `scripts/test/mock-server.go` mirroring PR #493's table. Commands: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`.

## Docs (same commit)

`docs/engine/cli.md` - the gate flags, exit-code contract, and a CI snippet; cross-link from `docs/engine/mcp.md`'s `start_load_run` row (the agent path vs the CI path).

## Sequencing

Parked - excluded from the wave map until unparked. When picked up: parallel-safe against everything currently open; touches `engine/src/cli.cpp` and `cli.md`, which nothing open edits. #470's thresholds and the report shape are merged and stable.

## Acceptance criteria

- [ ] `vayu-cli` can start a duration-bounded load run with budgets, wait for it, and exit 0/1/2 per the contract above - single static binary, no Node.
- [ ] Absent verdict, timeout, and non-terminal failures all exit 2 with distinct messages; failed checks are named with limit and actual.
- [ ] Engine suite green incl. the mutation checks; docs same-commit.

---

athrvk:
## Description

A pipeline can now fail a build on a performance budget:

```bash
node dist-electron/mcp/gate-cli.js \
  --url https://api.example.com/health \
  --concurrency 32 --duration 60s \
  --p99 50 --error-rate 0.5 --min-rps 1000
```

It starts one load run against a **running** engine, waits for it, prints per-budget pass/fail, and exits `0` (every budget met), `1` (a budget missed) or `2` (no verdict possible - bad flags, engine unreachable, run failed/stopped, wait timed out). `1` and `2` are separate so a pipeline can tell "too slow" from "broken".

**Plan.** Root cause: #470 gave the report a machine verdict (`thresholdValidation`) and nothing outside the app or an MCP client can read it - `vayu-cli` starts a run and exits before there is one, so a job built on it goes green whatever the latency was. Approach: a sibling Node entry beside the stdio MCP server (`gate-cli.ts`), built by the same `tsconfig.node.json` sweep of `electron/` and packaged by the same `dist-electron/**/*` glob, with all logic in a testable `gate.ts` core so argv, the clock, the streams and `process.exit` are the only untested lines. Rather than re-derive the run payload, `start_load_run`'s compose-and-assemble step is extracted as `buildLoadRunPayload` and shared. **Rejected: extending the C++ `vayu-cli`** - the gate needs the compose/threshold payload building and the TS report types that already exist in the main-process MCP layer, so a C++ version would mean a second copy of both, in the language with no report types at all.

**Verified against reality first.** Everything the issue anchors to still exists as described: `engine-client.ts` `startRun` / `getRunReport`, `cli.ts` as the standalone-entry pattern, `start_load_run`'s payload building at `tools.ts`, and #470's `verdict` in both `runs.cpp` and `domain.ts`. One thing the issue assumed that was not there: `EngineClient` had no `GET /runs/:id`, only `listRuns` and the report - so `getRun` is added, with the gate as its reader.

### Deliberate scope decisions

- **The gate assumes a running engine and never manages one.** `--engine-url` is the only coupling; lifecycle stays with the app and the CI step that launches `vayu-engine`. As specified in the issue.
- **The MCP allowlist and load caps do not apply**, and the docs say so. This is a human putting a target in their own pipeline, exactly like a human pointing the app at one; `safety.ts` exists to bound an autonomous agent, and there is no agent here. Pinned by a test that runs a non-allowlisted host at 5000 concurrency for 30m - both guards would refuse it.
- **`--mode` accepts `constant_concurrency` and `constant_rps` only.** Those two are duration-bounded and fully described by the flags, so the same command twice is the same run twice and the wait has a deadline known before it starts. `capacity`, `ramp_up` and `iterations` are named non-goals in the issue and are refused by name, with a message saying where to start them instead - they carry parameters this CLI does not expose, and an iterations run stops on a request count so nothing bounds the wait.
- **One addition beyond the issue's flag list: `--target-rps`.** `--mode` is in the spec and `constant_rps` is unusable without a rate. It is refused in `constant_concurrency` mode, and `--concurrency` is refused in `constant_rps` mode - the engine's `LoadStrategy::create` reads only one of the two per strategy, so accepting the other would be a flag the run never reads.
- **A usage error exits `2`**, not a fourth code: a mistyped flag is "broken", not "too slow".

### Notes on the shared code

- `buildLoadRunPayload` builds the payload; the allowlist, the caps and the capped-duration default stay in `start_load_run`. The tool's behaviour is unchanged - the existing `tools.test.ts` coverage is the regression guard, and it stays green.
- `engineDefaultDurationSeconds` is now exported instead of copied: an omitted `--duration` is not an unbounded run, and the wait deadline needs the same "60, except capacity" the caps use.
- The client parameter on the compose path is narrowed to `Pick<EngineClient, "composeRequest">`, so the gate reuses it without standing up a client it does not otherwise need.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`gate.test.ts` - **39 new tests**, all in the `node` environment:

- **Flag parsing**: each budget flag lands on its own engine metric key; no budget at all is a usage error; bounds mirror the engine's (`--error-rate 101`, `--p99 0`, `--p99 abc` all refused, `--error-rate 0` accepted); unknown flag refused by name; a flag given twice refused; `--flag=value` equals `--flag value`; both mode pairings refused in both directions; the three exploratory modes refused.
- **Poll loop**: waits through `pending`/`running` and answers with the terminal status; times out on the budget; a row with no status is an error, not an assumed finish. The clock and the sleep are injected rather than faked, so the timeout branch is driven directly instead of raced.
- **Exit codes**: `verdictExitCode` passed→0 / failed→1; run `failed` or `stopped` → 2; report with no `thresholdValidation` → 2 naming #470's absence semantics rather than crashing or reading it as a pass; engine unreachable → 2; `EngineRequestError` surfaces its status and body; accepted run with no id → 2.
- **End to end** through a fake client: budgets and load shape reach `POST /runs` unchanged; `--json` emits the raw report; the human output names every failed check with limit and actual; dropped pre-request scripts are reported.

**Mutation-checked** (reverted, confirmed red, restored): swapping the 0/1 exit codes fails 6 tests; returning the running status instead of throwing at the deadline fails the timeout test; treating an absent `thresholdValidation` as present fails 2 tests.

**Suites** (`cd app`): `pnpm test` **385 files / 3890 tests passed**; `pnpm type-check` clean; `pnpm lint` clean; `prettier --check` clean on every file touched (the three docs files were already unformatted at HEAD and were deliberately left alone per the repo's no-repo-wide-formatting rule). `mkdocs build --strict` clean. No engine code changed, so no engine suite run - per the issue's own test section. No pre-existing failures observed.

**Also driven live**, against a locally built engine and `scripts/test/mock-server.go`:

| Invocation | Result |
|---|---|
| `--url .../health --concurrency 8 --duration 5s --p99 5000 --error-rate 5 --min-rps 10` | 3/3 budgets met, **exit 0** (108,598 requests @ 21,716 rps) |
| `--url .../slow/50 --p99 1 --min-rps 1000000` | 2/2 budgets missed, **exit 1** (p99 52.031 vs limit 1) |
| `--mode constant_rps --target-rps 200 --json` | raw report on stdout, **exit 0** |
| `--request-id <saved>` | composed engine-side, ran, **exit 0** |
| `--request-id req_nope` | `No saved request with id "req_nope"`, **exit 2** |
| engine down | `fetch failed`, **exit 2** |
| no budget flags | usage message, **exit 2** |
| `--help` | usage, **exit 0** |

## Checklist

## Related Issues
Closes #473

---
_Generated by [Claude Code](https://claude.ai/code/session_01HVYhoHAzBjDCm5LdNSPvQ2)_


athrvk:
Branch refreshed onto master - it was 10 commits behind (`9880a02` → `af95ada`, picking up #490, #492, #496 and #497), which is what had it showing as `behind`.

The merge was clean with no conflicts: nothing that landed touches `app/electron/mcp/`, `docs/engine/mcp.md`, `docs/engine/cli.md` or the backlog file, so this diff is unchanged against the new base - still the same 9 files.

Re-verified on the merged tree rather than only on the old base: app suite **3959 tests across 388 files** green (up from 3890/385 - the difference is what master brought in), `pnpm type-check`, `pnpm lint` and `mkdocs build --strict` all clean, and `gate.test.ts` + `tools.test.ts` green together (156 tests) since #490's monitor work and this PR both touch the load-run payload path.

**Left as a draft on purpose** - you converted it, so I have not marked it ready. If the draft was for something specific in the diff, say what and I will fix it here; the two calls most worth a second opinion are the ones in the description: keeping the MCP allowlist and caps off this CLI, and restricting `--mode` to the two duration-bounded modes. Otherwise mark it ready whenever it suits you.

---

athrvk:
Claiming this - master has moved 18 commits ahead (`af95ada` → `3a7c937`) and the PR is showing as conflicting. Refreshing the branch and re-verifying this run.

---

athrvk:
Merged master (`af95ada` → `3a7c937`, 18 commits) and resolved two real conflicts - both cases of master and this branch adding to the same place, not textual noise:

**1. `EngineClient.getRun`** - added by both sides. This branch added it for the gate's terminal-status poll; master added it for `compare_runs` finding a target's baseline when the caller named none. Kept one method with one doc comment naming both readers, and kept master's neighbouring `listBaselineRuns` untouched.

**2. `start_load_run`'s payload build** - master grew a `monitor` block (#490) in the handler while this branch extracted the whole payload build into `buildLoadRunPayload`. Split it along the same line the extraction already draws: the monitor block is part of the *run description*, so it is forwarded from the shared builder right beside `thresholds`; the second-host allowlist check (`checkMonitorHost`) is a *guard*, so it stays in the tool handler next to `checkAllowlist`. Taking master's version verbatim would have left the payload built in two places again, which is the drift the extraction exists to prevent.

The gate CLI passes no monitor today, so its behaviour is unchanged - but if a `--monitor` flag is ever added it gets the same payload shape for free, rather than a second copy of the block.

**Verified on the merged tree:** `pnpm test` **4050 tests across 397 files** green (up from 3959/388 - the delta is what master brought in), `pnpm type-check` clean, `pnpm lint` clean, `prettier --check` clean on both touched files. Mutation-checked the relocated line: deleting `payload.monitor = args.monitor` from the builder turns "start_load_run forwards the monitor block to /runs unchanged" red, so the move is pinned by the existing test rather than merely compiling.

Still 9 files in the diff and still a draft - I have not changed either. The two calls flagged earlier (MCP allowlist/caps not applying to this operator-invoked CLI, and `--mode` limited to the two duration-bounded modes) are the ones worth a second opinion before it goes ready.

---

athrvk:
CI note on `503b1d1`, so it is on the record rather than only in the Actions UI: **App on ubuntu-latest and macos-latest both passed** (06:53 and 06:54), and every other check is green or skipped. **App on windows-latest has not reported** - it entered "Run frontend tests" at 06:51:22 and the API still shows it `in_progress` well past the job's own `timeout-minutes: 20`, which should have reaped it at 07:10. Its baseline for this step is ~6 min.

The two readings I cannot separate from here are a stale API view (the same endpoint reported ubuntu and macOS as `in_progress` for ~20 min after they had actually finished) and a wedged runner. Nothing points at the diff: the tree is identical on all three platforms, the other two passed it, and the full suite is green locally (4050 tests). If it is still showing stuck when you look, a re-run of that job is the cheapest way to tell the two apart - if it wedges a second time at the same step, that is a real Windows-only problem and worth its own issue rather than a fix squeezed into this PR.

I did not cancel or re-run anything, and I have not touched the draft status.

---

athrvk:
Correction to the note above: **Windows was never stuck, and the whole run is green.** `App on windows-latest` completed at **06:57:30** with its test step taking 6:03 (06:51:22 → 06:57:25) - dead on its ~6 min baseline - and the **CI gate passed at 06:57:35**. Every check on `503b1d1` is now success or skipped.

The "stuck job" reading was an artifact of the API: it served a view of the run that was roughly 50 minutes behind reality, reporting the job `in_progress` long after it had finished, exactly as it had earlier done for the Ubuntu and macOS jobs. Nothing to re-run and no Windows-only problem to file - disregard that suggestion.

So: master merged, conflicts resolved, and green on all three platforms. Still a draft, still yours to mark ready.

---

athrvk:
Closing unmerged by owner decision (2026-08-11) - a placement call, not a quality one.

The gate's own constraint decided this: it "assumes a running engine", so CI already has the engine binary present - a gate built into the existing C++ `vayu-cli` costs CI zero additional dependencies, while this Node entry requires Node plus the built `dist-electron` tree in a CI runner. The TS-reuse rationale also thins on inspection: the engine's `validate_run_config` owns the threshold ranges (as this PR itself argues for the monitor block), and the verdict is one field in a report the engine writes.

Nothing here is wasted: the design decisions this PR settled - the 0/1/2 exit-code split with usage-errors-as-2, the duration-bounded mode restriction with `--target-rps`/`--concurrency` pairing rules, absent `thresholdValidation` read as "broken" rather than "pass", the poll-with-deadline shape, `--json` passthrough - are now the contract recorded in #473, which has been re-scoped to the C++ `vayu-cli` and **parked by owner request** (see the issue). Whoever eventually picks it up should treat this PR's test list as the behavioural spec to port.

---

---

PR #494, opened 2026-08-10
feat(app/dashboard): anomaly annotations on the run timeline

### LINKED ISSUE #477: app(dashboard): anomaly annotations on the run timeline - auto-mark degradation windows and error onset so a 3-second spike cannot hide in the averages
Third of the observability trio (#475, #476). The reported pain: cumulative averages mask point-in-time degradation - a spike or an error burst that lived in one window of the test disappears into the summary, and users of other tools call out that nothing natively splits windowed from cumulative views. Vayu already has the windowed data (per-tick `latency_p99_ms` was built for exactly this); what's missing is the tool *pointing at* the bad windows instead of leaving the user to squint at charts.

## Problem

- The per-tick series (`LoadTestMetrics`, `app/src/types/domain.ts:831-856`: `latency_p99_ms`, `current_rps`, `requests_failed`, `status_codes?`, `elapsed_seconds`) contains every anomaly a run had, but no code looks for them. The only derived event today is the single SLO breakpoint - `computeBreakpoint` (`app/src/modules/dashboard/utils/computeBreakpoint.ts:34-50`) finds the *first* `p99 > threshold` crossing and stops.
- The chart layer can only draw whole-chart reference lines: `Marker` (`.../charts/uplot/plugins.ts:28-38`) is `{orient, value, scale?, label, role, align?}` and `markersPlugin` (:43-86) strokes one dashed line + one label per marker - no spans, no hit-testing, no per-point attachment.
- Post-run, a user reading `LoadTestDetail` sees summary numbers plus charts; "p99 spiked 4x for 3 seconds at t=41 coinciding with the first 5xx burst" is knowledge the data holds and the UI never states.

App-only: everything needed is already persisted per tick; no engine change.

## Fix / solution pathway

**1. Pure detection util** - `app/src/modules/dashboard/utils/detectAnomalies.ts`, shaped like `computeBreakpoint` (pure function over `LoadTestMetrics[]`, unit-testable with synthetic arrays):

```ts
export interface Anomaly {
  kind: "latency_spike" | "error_burst" | "throughput_drop" | "first_5xx";
  startSeconds: number; endSeconds: number;
  magnitude: number;        // e.g. 4.2 for "4.2x baseline"
  label: string;            // "p99 4.2x baseline for 3s"
}
export function detectAnomalies(history: LoadTestMetrics[]): Anomaly[]
```

Detection rules, deliberately simple and deterministic (no tunable ML, no flakiness):
- **latency_spike**: `latency_p99_ms > 3x` the trailing-median p99 (median over the previous 15 ticks, requiring >= 15 ticks of warmup) for >= 2 consecutive ticks; window extends until it drops back under 1.5x.
- **error_burst**: per-tick failure delta (`requests_failed` is cumulative - diff it) exceeding 1% of that tick's completions, >= 2 consecutive ticks.
- **throughput_drop**: `current_rps < 60%` of trailing median while concurrency did not drop, >= 2 consecutive ticks.
- **first_5xx**: the first tick whose `status_codes` (when present) contains a 5xx key - a point, `endSeconds == startSeconds`.
Merge overlapping windows of the same kind; cap output at 20 anomalies (keep the largest by magnitude) and note the cap in the last label. Constants live in the file, exported for tests.

**2. Chart rendering - a spans plugin.** Extend the uPlot plugin layer with `annotationsPlugin(getAnnotations, theme)` beside `markersPlugin` (`plugins.ts:43`): translucent vertical bands from `startSeconds` to `endSeconds` (full chart height, like markers use `top→top+height`, :61-62), color by kind via the existing `ColorRole` mapping (error → status-error, latency → warning), point anomalies drawn as the existing dashed-line style. Add an `annotations?: Anomaly[]` prop to `UPlotChartProps` (`UPlotChart.tsx:57-98`) wired the same way `markers` is (`markersRef` pattern, :132-143 - redraw on change, not in `shapeSig`). Hover detail can ride the existing `tooltipPlugin` label row or a title attribute pass - keep interaction minimal in v1; the bands + labels are the value.

**3. Consumers.**
- **Live**: derive in `MetricsView`'s memoized bundle (`DashboardDerived` memo, `MetricsView.tsx:106-167`) from the same history array the charts consume - per the repo rule that no card re-derives from raw metrics - and pass to `LatencyPercentilesChart` / `ErrorRateChart` / `RequestRateChart` (`TimeSeriesCharts.tsx`).
- **History**: `LoadTestDetail` already recomputes breakpoints from the fetched series (`LoadTestDetail.tsx:113-120` calls `computeBreakpoint(timeSeries, sloThresholdMs)`); call `detectAnomalies(timeSeries)` in the same memo, pass to the historical charts (`HistoricalChartsSection.tsx`), and render a compact "Events" list (kind, time range, label) in the Overview tab so the finding is stated in words, not only paint.
- The existing breakpoint marker stays - it is SLO-relative; anomalies are baseline-relative. Same chart, both layers.

## Tests

All pure-logic first, jsdom second:
- `detectAnomalies` unit tests with synthetic tick arrays: a clean run yields `[]` (mutation-check: loosen the 3x factor to 1x, this must fail); a 2-tick 4x spike is one window with correct bounds and magnitude; a 1-tick spike is NOT flagged (consecutive requirement, mutation-check by dropping it); cumulative `requests_failed` diffing (a constant high cumulative value with zero delta must not flag); the 20-cap keeps largest.
- Chart layer: `UPlotChart` accepts `annotations` and the plugin registers (className/spy level, jsdom has no canvas layout - assert the plugin list, not pixels, and assert the test scanned a non-empty annotation set per the repo's source-scanning rule).
- `LoadTestDetail` renders the Events list from a series containing a spike, and no list for a clean series.

## Docs (same commit)

`docs/app/COMPONENTS.md` if chart plugins are catalogued there; `docs/design-system.md` only if a new color use needs recording (prefer existing status tokens so it doesn't).

## Sequencing

Independent; app-only. If #475 lands, its monitor series can feed the same detector later - design `detectAnomalies` over `LoadTestMetrics[]` now, don't generalize speculatively. If #471 lands, capacity runs get annotations for free since they stream the same ticks. Parallel-safe; note #475 edits the same MetricsView.tsx / LoadTestDetail.tsx and rides the same uplot chart layer this issue extends (UPlotChart.tsx props, plugins.ts) - coordinate whichever lands second.

## Acceptance criteria

- [ ] A run with a mid-run latency spike or error burst shows shaded windows on the live and historical charts and a worded Events list in history - with zero configuration.
- [ ] A clean run shows nothing new anywhere.
- [ ] Detection is deterministic pure logic with constants exported and mutation-checked; no engine or storage change.
- [ ] App suite, type-check, lint, format all green; docs updated if touched surfaces are documented.

---

athrvk:
## Description

**The plan.** Root cause: the per-tick series already contains every anomaly a run had (`latency_p99_ms`, the cumulative failure/completion counters, `current_rps`, the per-tick status map), and no code looks for them - `computeBreakpoint` is the only derived event, and it is SLO-relative and stops at the first crossing. So a 3-second collapse that recovered leaves the summary numbers almost unchanged and shows up only as a bump the reader has to spot. The approach: one pure detector over `LoadTestMetrics[]` (`detectAnomalies`), derived once per view and read twice - the charts shade the windows, the history Overview states them in words. **Rejected:** deriving inside `MetricsView`'s `DashboardDerived` bundle, which the issue suggests. That bundle deliberately depends on the latest tick and not on the history array (its comment says so), so folding a history-wide scan into it would rebuild it at 10 Hz and defeat the `React.memo` on `HeroRow` / `ModeStatsRow`. It is derived beside the other series transforms instead, which keeps the "no card re-derives from raw metrics" rule intact - the charts are the consumers and take it directly. **Also rejected:** teaching the chart layer about anomaly kinds. `plugins.ts` gets an `Annotation` (span + role + label) the way it already has a `Marker` (value + role + label), and `TimeSeriesCharts` maps the domain shape onto it beside the existing `breakpointMarker` helper.

Verified against the code before writing: every anchor in the issue still holds (`computeBreakpoint.ts:34-50`, `Marker`/`markersPlugin` at `plugins.ts:28-86`, `UPlotChartProps` markers wiring, `MetricsView`'s derived memo, `LoadTestDetail.tsx` recomputing the breakpoint from the fetched series).

**What lands:**

- **`detectAnomalies.ts`** - four rules, deliberately dumb and deterministic: latency spike (p99 past `3x` a trailing median of 15 ticks for `>= 2` consecutive ticks; the window stays open until p99 falls back under `1.5x`, with the baseline **frozen at the opening tick** so a long degradation cannot raise its own bar and report itself recovered), error burst (per-tick failure delta past 1% of that tick's completions - both counters are cumulative and are diffed), throughput drop (under 60% of the trailing median **while concurrency held**, so a ramp winding down is not an anomaly), and the first 5xx as a point. Windows are disjoint by construction, so there is no merge pass to go stale. Over 20 findings the list is triaged rather than truncated - the largest survive and the last one says how many did not. Every constant is exported and pinned by a test.
- **`annotationsPlugin`** (`plugins.ts`) - translucent bands from `startSeconds` to `endSeconds`, an instant drawn as a dashed rule, both clipped to the plot box so a zoomed-out window cannot paint over the axis. Registered before `markersPlugin` so the breakpoint's dashed rule stays legible over a band. `UPlotChart` takes `annotations` through the same ref-and-redraw path as `markers` (re-creating the plot would drop the user's zoom mid-run).
- **Live** (`MetricsView`): windows shade Throughput, Latency, Percentiles and Server Vitals over time.
- **History**: `LoadTestDetail` derives once and hands the result to both tabs - Performance shades the same charts, Overview renders a new **`RunEvents`** card (kind, the finding in words, the time range). Silent for a clean run: an "Events (0)" card puts the eye on the one thing with nothing to say.

No engine, storage or API change; a run with no findings and a run with no stored series both render exactly as before.

**Merged with master** after #490 and #492 landed - see [the resolution comment](https://github.com/athrvk/vayu/pull/494#issuecomment-5249283293). The notable half: #490's Server Vitals row now shades the same windows through the shared `BaseProps`, which is where the shading earns the most - "p99 spiked 4.2x at t=41" is answered right there by what the server was doing then.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check` - **3911 passed / 389 files** (22 new), type-check, ESLint and Prettier all clean, re-run after the master merge rather than only before it. No engine code is touched, so per root `CLAUDE.md` ("if no test could possibly go from green to red, do not run tests") the C++ suite was not run. No pre-existing failures observed.

Every rule is mutation-checked by actually reverting it and confirming red, not by inspection:

| Mutation | Result |
|---|---|
| `LATENCY_SPIKE_FACTOR` 3 → 1 | 6 failed |
| `MIN_CONSECUTIVE_TICKS` 2 → 1 | 2 failed |
| read `requests_failed` undiffed | 2 failed |
| drop the concurrency guard | 1 failed |
| cap keeps insertion order instead of the largest | 1 failed |
| annotation band not clipped to the plot box | 1 failed |
| always draw a band, never a rule | 1 failed |
| draw offscreen windows | 1 failed |

The clean-run test is the one that could pass for the wrong reason, so its fixture drifts rather than sitting flat, and a companion test asserts the drift really does put consecutive ticks above the trailing median - i.e. the run is quiet *because of* the 3x factor, not because nothing in it could ever trip a rule. The plugin is driven directly against a recording 2D context rather than asserted through a rendered chart: jsdom has no layout and a stubbed context, so a mounted chart would pass just as happily with a plugin that drew nothing.

## Checklist

## Related Issues
Closes #477


athrvk:
fix merge conflict

athrvk:
Claiming this - merging master (#490 and #492 landed) and resolving the conflicts on this branch now.

---

athrvk:
Merged - master had gained #490 (server-vitals overlay) and #492 (AssertionError). Three conflicts, all in the places #490 also touched, and each resolved toward master's shape rather than beside it:

- **`TimeSeriesCharts.tsx`** - both sides added imports and a chart concept. #490's **Server Vitals row now shades the same windows as every other time chart**: it takes `anomalies` through the shared `BaseProps` and maps it with the same `anomalyAnnotations` helper, so there is one mapping rather than a second copy. This is the row where the shading earns the most - "p99 spiked 4.2x at t=41" is answered right there by what the server was doing then - and both consumers (live `MetricsView`, history `HistoricalChartsSection`) pass it.
- **`charts.smoke.test.tsx`** - both sides appended a case; kept both, and the anomaly case now mounts `ServerVitalsChart` with windows attached as well.
- **`LoadTestDetail.tsx`** - both sides added an import and a derivation. The pane runs the monitor-series query and `detectAnomalies` side by side; `LoadTestDetail.anomalies.test.tsx` gained the `useRunMonitorSeriesQuery` mock the pane's other tests already carry (its absence was the one real breakage the merge produced, and it failed loudly rather than passing quietly).

Everything else auto-merged - `types.ts` carries `monitorSamples` and `anomalies` side by side, and the engine half of both PRs came across untouched.

One thing swept up while resolving, flagged rather than hidden: `docs/app/COMPONENTS.md` had no `ServerVitalsChart` in its chart catalogue. Since I was editing that exact sentence, it is now listed - two words, no behaviour.

Verified after merging, not just after resolving: app suite **3911 passed / 389 files**, type-check, lint and format all clean. No engine file was touched by the resolution.

---

PR #495, opened 2026-08-11
feat(engine/app): pin a run as baseline and diff any run against it

### LINKED ISSUE #472: engine/app(history): pin a run as baseline and diff any run against it - the comparison engine already exists in the MCP layer and the renderer never sees it
Third of the thresholds (#470) → capacity (#471) → baselines arc. This is the "did my change make the API slower?" answer: pin a known-good run, and every later run of the same endpoint shows its deltas.

## Problem

Run-to-run comparison is already implemented in Vayu - and only an MCP client can use it:

- `app/electron/mcp/compare.ts` has the whole diff engine: `MetricDelta` (:28), `RunComparison` (:46), `compareReports(baseRunId, targetRunId, base, target)` (:56) covering latency `p50/p90/p95/p99/avg/max`, `summary.avgRps`/`throughput`, `summary.errorRate`/`totalRequests`, and per-status-code counts. It is pure, main-process-only, consumed by the `compare_runs` tool (`app/electron/mcp/tools.ts:1982`) and the `compare_runs` prompt (`app/electron/mcp/prompts.ts:71`). No renderer code imports any of it.
- The history UI cannot even show two runs side by side: `HistoryDetail` (`app/src/modules/history/main/HistoryDetail.tsx`) routes exactly one run, and `LoadTestDetail` renders one report. A human answers "slower than last week?" by opening two reports and eyeballing.
- There is no baseline concept anywhere: the `runs` table (`engine/src/db/database.cpp:240-252`) has no such column, `RunFilter` (`engine/include/vayu/db/database.hpp:37-43`) cannot filter on one, and a repo-wide grep for `baseline` finds only the app's unrelated CSS hits and backlog D9's prose.

Leaving `compareReports` where it is and hand-rolling a renderer copy would recreate this repo's most-documented defect class ("a hand-rolled copy of a primitive does not receive the primitive's fixes", root `CLAUDE.md`), so the move to a shared module is part of the work, not an aside.

## Fix / solution pathway

**1. Engine: a `baseline` flag on runs.**
- Column `baseline` INTEGER, `default_value(0)`, on the `runs` table (`database.cpp:240-252`; sqlite_orm `sync_schema` adds defaulted columns to existing DBs - verify on an existing dev DB and state the result in the PR).
- `RunFilter` gains `std::optional baseline` (`database.hpp:37-43`), honored by `get_runs_paginated`/`count_runs`.
- `GET /runs` accepts `baseline=true|false` on the envelope path (`engine/src/http/routes/runs.cpp:854-940`), and `serialize_run_row` (:335) emits `"baseline"`.
- Set/unset endpoint: per the repo's API rule (#95: POST is create-only, PUT updates), add `PUT /runs/:id/baseline` with body `{"baseline": true|false}` → 200 with the updated row, 404 on a missing run, 400 on a non-boolean. Multiple baselines are allowed (one per endpoint/request is the expected use); the *selection* of which baseline applies to a run is client-side (below), so the engine stays policy-free.
- Retention: `prune_runs` (`database.cpp:1040-1108`) must never delete a baseline run - that is the whole point of pinning. Add the WHERE clause and a test whose mutation-check is removing the clause.

**2. Shared comparison module.** Move `compareReports` + `MetricDelta` + `RunComparison` to a pure module importable by both the renderer and the MCP main process, behavior-identical. Candidate home: `app/src/lib/run-compare.ts` with `app/electron/mcp/compare.ts` reduced to a re-export - but the implementer must verify the electron-main build (its tsconfig/bundler) can import from `app/src/lib` and state what they found; if the boundary forbids it, invert the direction (shared file under a path both configs include) rather than duplicating. `MetricDelta` already carries direction (which way is an improvement) - preserve it; the renderer colors from it.

## App changes

**3. Renderer UI.**
- Pin/unpin action on `RunItem` (`app/src/modules/history/sidebar/RunItem.tsx`) with a baseline chip (`Badge variant="chip"` per the enforced badge test), plus a baseline filter in the history store's `FilterType` if it falls out naturally (`app/src/modules/history/history-store.ts`).
- A mutation hook beside the existing ones in `app/src/queries/runs.ts` (`useDeleteRunMutation` is the pattern: invalidate `queryKeys.runs.lists()` on success).
- `LoadTestDetail` (`app/src/modules/history/main/LoadTestDetail.tsx`) gains a "vs baseline" strip: find the most recent baseline run for the same `requestId` (fall back to same method+URL for unsaved requests; `GET /runs?baseline=true&requestId=...` provides it), fetch its report with the existing `useRunReportQuery` (`app/src/queries/runs.ts:169`), run the shared `compareReports`, and render `MetricDelta` rows - direction-aware coloring via `--status-*` tokens (latency down is good, throughput up is good), no raw palette. When no baseline exists or the baseline IS the open run, the strip is absent - not an empty shell.

**4. MCP.** `compare_runs` (`tools.ts:1982`) gains an optional convenience: omitting `baseRunId` resolves the target's baseline the same way the renderer does. `list_runs` output including `baseline` comes free from `serialize_run_row`. Update `docs/engine/mcp.md` tool table (:143-163) same commit.

## Tests

- Engine (`runs_route_test.cpp` + a db test): PUT toggles and persists; `GET /runs?baseline=true` filters; the row carries the flag; prune spares baselines (mutation-check: drop the WHERE clause, the test must fail); older DBs open cleanly with the defaulted column.
- App: pin action fires the mutation and the row shows the chip; the vs-baseline strip renders deltas from two canned reports with correct improvement direction (mutation-check: invert a direction, the color assertion must fail); no strip without a baseline; the moved module's existing MCP behavior pinned by whatever `compare` tests exist today staying green from the new import path.

## Docs (same commit)

`docs/engine/api-reference.md` - `GET /runs` params, the new `PUT /runs/:id/baseline`, the run-row shape. `docs/engine/db-schema.md` - `runs` table (:204-330) gains the column and the prune exemption. `docs/engine/mcp.md`. `docs/request-storage-design.md` only if it describes run rows.

## Sequencing

Independent of #470 and #471 - no shared code. #470 merged 2026-08-10 (e1402eb, PR #482) - thresholdValidation.verdict is live (runs.cpp:813-816, domain.ts:1030-1039), so the 'regressed vs baseline AND broke its budget' pairing is available immediately. Nothing else open touches the runs schema. The shared-module move (step 2) is the only part with a repo-structure unknown; it is deliberately scoped so a wrong first guess costs a file move, not a rewrite.

## Acceptance criteria

- [ ] A run can be pinned/unpinned as baseline from the history sidebar, survives restarts, and is never pruned.
- [ ] Opening a load run with a matching baseline shows direction-aware deltas for latency percentiles, throughput, error rate, and status codes - computed by the same single implementation the MCP `compare_runs` tool uses.
- [ ] No baseline → no strip; pre-existing runs and older databases keep working.
- [ ] `GET /runs` can filter to baselines; MCP `compare_runs` can default to the baseline.
- [ ] Suites green; docs updated same commit.

---

athrvk:
## Description

**Root cause.** Run-to-run comparison is already implemented - `compareReports` in `electron/mcp/compare.ts` diffs latency percentiles, throughput, error rate and the status mix - and only an MCP client can reach it. There is no baseline concept anywhere (no column, no filter, no endpoint), so a human answering "did my change make this slower?" opens two reports and eyeballs them.

**Approach.** A `baseline` flag on the run row, a `PUT /runs/:id/baseline` to set it, a `baseline` filter on `GET /runs` so a client can resolve "the pin for this request" in one query, and a retention exemption so the pin outlives the cap. On top of that: a pin action on the history row and a vs-baseline strip in the load report, plus `compare_runs` defaulting its base to the target's pin so an agent and the UI answer the question about the same pair of runs.

**Alternative rejected.** Adding the flag as a general field on a run update: a run has no general update endpoint, and everything else on the row is written by the engine as the run executes. A dedicated `PUT` keeps the one client-settable property from implying the rest are settable too.

### Verified against the code, two drifts from the issue

- **`MetricDelta` does not carry direction today** (the issue says it does, and says to preserve it). It carries `metric/base/target/delta/pctChange` only. Added as a `direction` field - `lower-is-better` / `higher-is-better` / `neutral` - computed in the shared core and surfaced in the MCP output schema too, since an agent reading `delta: -12` otherwise has no way to know whether that is good. `neutral` is `summary.totalRequests`, which moves with how long a run was told to run; calling either direction an improvement would paint a shorter run as a regression.
- **The shared-module move is not possible in this repo, in either direction** - the unknown the issue flagged. Demonstrated, not assumed:
  - main → `src/`: `tsconfig.node.json` emits with `rootDir: "electron"`, so a file from `src/` is **TS6059**. The rationale is documented in that file (giving it the `@/*` mapping would drag renderer sources under the main process's `outDir`).
  - renderer → `electron/`: `electron/` is a **referenced composite project**, so importing one of its files from `src/` is **TS6305** ("output not built"). Excluding the file from that project to dodge it is **TS6307** ("must list all files").
  - Widening `rootDir` fixes both and changes the emit layout to `dist-electron/electron/…`, which breaks the published `node dist-electron/mcp/cli.js` path users already have in their MCP client configs. Not worth it for one file.

  So the two copies stay, pinned the way this repo pins its other cross-boundary mirrors (`lib/dynamic-variables.ts`, `McpDataEntity`, and the C++/TS `variable-resolution-conformance.json`): **`electron/mcp/compare.conformance.test.ts` runs both implementations over the same five fixtures and fails on any divergence**, in either direction, plus a case set for `deltaVerdict`. That does not meet the issue's "same single implementation" criterion literally; it is the strongest available substitute, and the header of each file points at the other.

### Deliberate omissions

- **No baseline entry in the history `FilterType`.** The issue said "if it falls out naturally" - it does not: `FilterType` is `"all" | Run["type"]` and `filterRuns` compares it to `run.type` directly, so a pin (an orthogonal dimension) would need a second filter control, not a value. `GET /runs?baseline=` exists and is used by the lookups; the sidebar control is a separate, small piece of work if wanted.
- **`compare_runs`'s status-code merge mishandles a raw engine report** where `statusCodes` arrives as an array of tuples (the renderer's transformer converts it; the MCP path does not). Pre-existing, untouched here, and worth its own issue.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Testing

All from a clean build at `9880a02`, all green, no pre-existing failures encountered.

| Suite | Result |
|---|---|
| `python build.py -e -t` + `ctest --preset linux-dev` | **1312 / 1312 passed** (176s) |
| `cd app && pnpm test` | **3873 / 3873 passed**, 387 files (257s) |
| `pnpm type-check` (all three tsconfigs) | clean |
| `pnpm lint` (0 warnings) / `pnpm format:check` | clean |
| `mkdocs build --strict` | built |

New tests - 12 engine, 17 app:

- **Engine, retention (`db_test.cpp`)**: a pinned run survives the count cap and the age cap, with its ticks and results; the cap still keeps its quota of unpinned runs (the pin is skipped, not merely spared). **Mutation-checked for real**: disabling the exemption and rebuilding fails both, and they pass again on restore.
- **Engine, storage**: pin toggles and survives a reopen; a missing run stores nothing; the filter selects in both directions and stays a wildcard when unset; a database whose `baseline` column is dropped (the pre-column simulation `MigratesHttpVersionColumnOntoAPreExistingRequestsTable` uses, with the same "guard the guard" check that the drop happened) reopens and reads as unpinned.
- **Engine, route (`runs_route_test.cpp`)**: PUT pins/unpins and answers the list row; 404 on a missing run; 400 for seven unusable bodies (`"true"`, `1`, `null`, `{}`, wrong key, non-JSON, array) with nothing written; the flag on a list row, on the filtered list, and on the single-run payload.
- **App**: the pin action sends the *negation* of the current state (a button that always pins cannot unpin), the chip renders only when pinned and carries no `hover:bg-*` (rendered `className`, since a source scan cannot see it), no action on design/scenario rows; the strip colours by direction - a rise in latency red and a rise in throughput green *from the same sign* - reads an improvement as one, draws nothing with no pin / when the run is its own pin (and fetches no report in either case), and matches an unsaved request by url **and** method exactly against the engine's over-matching `q`; the endpoint builder emits `baseline=false` rather than dropping it.
- **MCP**: an explicit base skips the lookup; an omitted base resolves through `getRun` → `listBaselineRuns` → report; the three refusals (no saved request, no pin, target *is* the pin) each report an argument error and compare nothing; every metric carries its `direction`.

## Checklist
- [x] I have added tests

Docs updated in the same commit: `docs/engine/api-reference.md` (the new endpoint, the `baseline` param and row field, the retention note), `docs/engine/db-schema.md` (column + prune exemption), `docs/engine/mcp.md` (the tool's optional base and `direction`, the mirrored-file note), `docs/app/state-management.md` (the `runs.baseline` key family and why it is invalidated rather than patched), `docs/app/COMPONENTS.md` (the row action and the strip).

## Related Issues
Closes #472

---
_Generated by [Claude Code](https://claude.ai/code/session_01PxjgmowZmjeJhw6RSmVj1A)_


athrvk:
fix merge conflicts

athrvk:
Claiming this - merging master into the branch now and re-verifying both suites before I push.

---

athrvk:
Merged - master had gained 17 commits, most relevantly #475 (server vitals) and #477 (anomaly annotations), which land on the same three surfaces this branch touches.

Two textual conflicts, both resolved by taking **both** sides rather than either:

- **`prune_runs`' doc paragraph** (`docs/engine/db-schema.md`) - master added `monitor_samples` to the cascade sentence while this branch added the baseline exemption to the same sentence. The merged line names both.
- **`LoadTestDetail.scenario.test.tsx`'s query mock** - master added `useRunMonitorSeriesQuery`, this branch added the three the vs-baseline strip reads. Both are in the mock now.

One thing the merge did *not* flag, which is the part worth knowing about: **`LoadTestDetail.anomalies.test.tsx` arrived from master and broke**, because it mocks `@/queries/runs` wholesale and the strip now renders inside the same header - so its three hooks came back undefined. Auto-merge could not see that; only running the suite did. Stubbed to "no data", which is the state those cases already render in (no pin, no strip), so the Events assertions are untouched.

Checked by hand rather than trusted to the merge:

- `remove_run_cascade_locked` keeps master's `monitor_samples` line **and** the baseline skip is still in `prune_runs` - the two edits were adjacent and are independent.
- `RunFilter`, `serialize_run_row` and the report four-hop in `runs.cpp` carry all three features' additions side by side.

Re-verified on the merge result, not just on the resolution: engine **1386/1386**, app **3964/3964** (393 files), `pnpm type-check` across all three tsconfigs, `pnpm lint`, `pnpm format:check`, `mkdocs build --strict` - all clean.

---

Issue #481, opened 2026-08-10
engine/app(mock): local mock server from a collection's examples - which first requires an examples model, because imports currently drop every saved response on the floor


athrvk:
"I just want a fake server so the frontend can build against the spec" is one of the most repeated workflow asks, and the incumbent gates mocks behind cloud/paid plans. Vayu has the importers, the storage, and a C++ HTTP server - what it lacks is anywhere to keep an example response, and that gap is the real first half of this issue.

## Problem

Two stacked gaps:

**Examples do not survive import, and cannot be stored.**
- The Postman parser never reads `item.response[]` (saved examples): `pmRequest` (`app/src/services/importers/postman.ts:235-246`) reads `item.request`, `item.name`, `item.event` and `protocolProfileBehavior`; examples are dropped without even a `SkippedItem` tally.
- The OpenAPI parsers sample **request** bodies only: `buildBody` (`app/src/services/importers/openapi-v3.ts:213-241`) collapses `example ?? sampleSchema(schema)` into a request-body string, and `op.responses` is never visited by any code path (`schema-sampler.ts` reads `const`/`example`/`examples[0]` - for requests).
- There is nowhere to put them anyway: `RequestDraft` (`importers/types.ts:63-85`) has no response field, the `Request` type (`app/src/types/domain.ts:233-262`) has none, and the engine's full table list (`engine/src/db/database.cpp:157-318`) contains no examples table. The only stored responses are run-scoped (`results.trace_data`, `body_blobs`/`result_bodies`) and pruned with their run.

**No mock server.** The only mock in the repo is a Go benchmark fixture (`scripts/test/mock-server.go`). The engine binds exactly one loopback server (`server.cpp:53`), though the second-listener pattern is proven (`OAuth2AuthorizeManager::start`, `engine/src/http/routes/oauth_authorize.cpp:168-280`).

### Phase 1 - the examples model

- **Engine**: new table `request_examples (id, request_id, name, status, headers TEXT, body TEXT, content_type, created_at, updated_at)`, cascade-deleted with its request, indexed on `request_id` (index block: `database.cpp:169-183`). CRUD nested under requests: `GET /requests/:id/examples`, `POST` (create-only, 409 on existing id), `PUT /requests/:id/examples/:exampleId`, `DELETE` - per the repo's API rule (#95). New registrar in `server.cpp:121-135`. Body size cap with a named constant.
[...]

Issue #479, opened 2026-08-10
engine(auth): built-in oauth2 mock issuer - mint tokens with chosen claims, expiry and failure modes so auth flows are testable without a real identity provider


athrvk:
Companion to #478. The reported pain: developing against real IdPs means 2FA prompts, captchas, provider rate limits and "suspicious login" emails - so teams bolt on external oauth2 mock-server packages just to develop. Vayu's whole posture is "works offline, air-gapped"; auth is currently the one part of a workspace that cannot run offline.

## Problem

Every OAuth2-authed request in Vayu ultimately needs a live token endpoint: `acquire_token` (`engine/src/http/oauth_client.cpp:201-293`) POSTs to the config's `accessTokenUrl`, and there is nothing local to point it at. Consequences: collections with OAuth2 auth are dead offline; testing edge behavior (expiry, slow issuer, provider 500s, refresh rotation) requires a cooperative real provider or hand-rolled infrastructure; and engine tests for auth paths (#478's watchdog) each build ad-hoc stubs.

The engine already contains every ingredient:
- **Second-listener precedent**: the interactive-auth loopback callback already constructs an independent `httplib::Server`, binds `bind_to_any_port("127.0.0.1")` and runs it on its own thread with TTL reaping and teardown (`OAuth2AuthorizeManager::start`, `engine/src/http/routes/oauth_authorize.cpp:168-280`; `Attempt` struct :109-124; `kAttemptTtlMs` :34).
- **Crypto for JWTs**: `hmac_sha256` (`engine/include/vayu/utils/sha256.hpp:56-61`) + `base64url_encode` (`engine/include/vayu/utils/encoding.hpp:131`) are exactly HS256 JWT material; the sha256 header's own comment warns MAC *verification* needs `sodium_memcmp` - the issuer only signs, but note it for the token-introspection route if added.

## Fix / solution pathway

**1. Lifecycle API** on the main engine server (new `register_mock_issuer_routes` in the 15-registrar list, `engine/src/http/server.cpp:121-135`, `RouteContext` in `engine/include/vayu/http/routes.hpp:479-490`):

[...]

PR #499, opened 2026-08-11
feat(engine/app): local webhook inbox - record inbound requests with no tunnel

### LINKED ISSUE #480: engine/app(webhooks): local webhook inbox - an engine-hosted listener that records incoming requests so webhook development needs no cloud tunnel
Testing the *receiving* side of webhooks today means webhook.site or an ngrok tunnel - routing your payloads through someone else's cloud, which is precisely what Vayu's air-gapped/corporate users cannot do. For the dominant dev-loop case (a local or LAN service emitting webhooks at a local consumer-to-be), no tunnel is needed at all - just something local that listens, records, and shows.

## Problem

Vayu can send in every direction and receive in none. There is no listener concept: the engine binds exactly one loopback server (`server_.listen("127.0.0.1", port_)`, `engine/src/http/server.cpp:53`), and the app has no surface for an inbound exchange - grep for `webhook` across `app/` and `docs/app/` returns nothing. A developer building a webhook consumer bounces between Vayu (to trigger the event) and a browser tab on webhook.site (to see what arrived), leaking payloads to a third party on the way.

The building blocks all exist:
- **Second listener**: proven by the OAuth loopback callback - independent `httplib::Server`, `bind_to_any_port("127.0.0.1")`, own thread, TTL reaping (`OAuth2AuthorizeManager::start`, `engine/src/http/routes/oauth_authorize.cpp:168-280`).
- **Rendering an exchange**: the shared viewer's `RequestData` is a transport-agnostic `{method?, url?, headers?, body?}` bag (`app/src/components/shared/response-viewer/types.ts:42-47`), and `UnifiedResponseViewer` already renders a request with no response (empty-state guard, `UnifiedResponseViewer.tsx:46-56`); `buildRawRequest` (`app/src/components/shared/response-viewer/utils.ts:273`, covered by `build-raw-request.test.ts`) builds the raw-HTTP text view.
- **Live streaming to the UI**: the SSE pattern (`sse-client.ts`, engine `set_content_provider` loops) is established.

## Fix / solution pathway

**1. Engine: inbox lifecycle** (new `register_inbox_routes` in the registrar list, `server.cpp:121-135`):
- `POST /inbox/start` body `{"port"?: number, "bind"?: "127.0.0.1", "response"?: {"status": 200, "body"?: "", "headers"?: {}, "delayMs"?: 0}}` → `{"inboxId", "url": "http://127.0.0.1:/"}`. The listener accepts **any** method on **any** path and records it; the configured `response` is what callers receive (default 200 empty) - enough to test retry logic by setting 500 + delay.
- `POST /inbox/:id/stop`, `GET /inbox` (list), `PUT /inbox/:id` (update the canned response live).
- **Bind address is a real trust boundary, decided here**: default `127.0.0.1`. `"bind": "0.0.0.0"` is allowed - LAN webhook sources are the point for some users - but the engine's posture (no route auth, CORS `*`, `server.cpp:105-107`) makes any wider bind on the *management* API unacceptable; only the inbox listener itself may bind wide, it must serve nothing but capture + canned response (no engine routes), and the response body for `GET /inbox` must state the bind so the UI can badge it. Reject `0.0.0.0` unless the request carries `"confirmNonLoopback": true`.
- Owned by a manager member on `Server` per the member-order rule (`server.hpp:46-54`); all inboxes stop on shutdown.

**2. Engine: capture storage.** New table `inbox_requests (id, inbox_id, received_at, method, path, query, headers TEXT, body TEXT, body_truncated, remote_addr)` - body capped (reuse the sample-body cap constants pattern, `constants.hpp` `run_config` block) and the table bounded per inbox (e.g. keep last 500, delete-oldest on insert; constant, documented). `GET /inbox/:id/requests` with the standard `{data[], pagination{}}` envelope; `DELETE /inbox/:id/requests` clears. Live: `GET /inbox/:id/live` SSE emitting one event per capture, following `live_metrics`' framing (`engine/src/http/routes/metrics.cpp:345-404`) - note each SSE handler occupies a cpp-httplib pool thread (default pool, no `new_task_queue` override exists), so document that N concurrently-watched inboxes cost N pool threads and cap watchable inboxes accordingly (reject a second concurrent `/live` per inbox).

**3. App: an Inbox surface.** New tab type: add `"inbox"` to `TabType` (`app/src/stores/tabs-store.ts:14-21`), the `isTabDirty` switch (:81-114 - a missing case reads as clean and gets LRU-evicted; inbox is never dirty, add the explicit case with a comment), and the `Shell.tsx` content switch (:25-37). Surface: start/stop + URL-copy header, live-updating capture list (SSE), detail pane mapping a capture onto `RequestData` + `UnifiedResponseViewer` and the raw view via `buildRawRequest` (`app/src/components/shared/response-viewer/utils.ts:273`). Entry point: a small action in the Dock/collections area is an owner call - propose it in the PR, don't bikeshed here. Badge non-loopback inboxes with a warning chip (`--status-warning` tokens).

**4. Deliberately out of scope** (named so they arrive on purpose): script assertions on captures (`pm.*` over received payloads - powerful, own issue), tunnels of any kind, request replay/forwarding, signature verification helpers (HMAC verify would need `sodium_memcmp` per `sha256.hpp:12-19`'s warning - good follow-up).

## Tests

- Engine (`inbox_test.cpp` in the `vayu_tests` target): start → POST a payload at the inbox with the engine's own `Client` → capture row has method/path/headers/body; canned 500+delay is served; body over cap truncates with the flag set; ring bound evicts oldest (mutation-check the bound); non-loopback without `confirmNonLoopback` is 400; stop refuses further connections; clean shutdown with a live inbox (pattern: `run_shutdown_test.cpp`).
- App: tab type registered in all three switches (a test that walks `TabType` values through `Shell`'s switch guards the isTabDirty gap - assert it scanned a non-empty type list); capture renders through `UnifiedResponseViewer` from a canned capture; live-append updates the list.

## Docs (same commit)

`docs/engine/api-reference.md` (route family + SSE), `docs/engine/db-schema.md` (new table + bounds), `docs/engine/architecture.md` (listener inventory + the bind-address trust decision - this is the doc that must carry it), `docs/app/COMPONENTS.md` for the new surface.

## Sequencing

Independent of everything open. Shares the "manager-owned extra listener" shape with #479 - if both land, extract the common lifecycle (TTL, teardown, thread ownership) only when the second one lands, not speculatively. Same note for #481 phase 2 - three listeners would then share the `server.cpp:121-135` registrar list, the `server.hpp:46-54` member block, and the `docs/engine/architecture.md` listener inventory; rebase-level, whoever lands later coordinates. Parallel-safe.

## Acceptance criteria

- [ ] Start an inbox, point any webhook source at its URL, and watch captures appear live in a Vayu tab - no tunnel, no third party.
- [ ] Captures render with the same viewer quality as outbound exchanges, including raw HTTP text.
- [ ] Canned response (status/headers/body/delay) is configurable live; retry-logic testing works.
- [ ] Loopback by default; wider bind requires explicit confirmation and is visibly badged; storage bounded and truncation flagged.
- [ ] Suites green; docs (including the trust-boundary write-up) same commit.

---

athrvk:
## Description

**Plan.** Root cause: Vayu can send in every direction and receive in none - the engine binds exactly one listener (`server.cpp:53`) and `rg webhook app/` returns nothing - so testing the *receiving* side of a webhook means webhook.site or an ngrok tunnel, which is precisely what air-gapped users cannot use. Verified still true at `9880a02`. The approach follows the second-listener precedent the engine already owns (`OAuth2AuthorizeManager`): an `InboxManager` member on `Server`, declared **before** `server_` so its dtor stops and joins the listener threads after the route lambdas that reference it are gone, with `db_` (external) still alive - the reverse-member-order rule `server.hpp:46-54` documents. Each inbox is an independent `httplib::Server` accepting any method on any path.

**The alternative I rejected:** capturing on the *main* engine listener under an `/inbox/:id/capture` prefix. It needs no second listener, but it makes the capture URL indistinguishable from the management API, so the LAN-source case (#480's stated reason for allowing a wider bind at all) would mean publishing the whole management surface with it - and the inbox could not be stopped independently. A separate listener is what lets the bind decision be made per listener rather than per process.

**A note on `set_pre_routing_handler`, since it looks like the obvious way to catch every method:** cpp-httplib invokes it *before* reading the body (`routing()`, httplib.h:11858), so `req.body` is empty there. The inbox registers catch-all handlers per method instead; `Get(".*")` also serves HEAD, and CONNECT/TRACE/PRI are the remainder cpp-httplib will not route (documented - a webhook is never one).

### Engine

- **Routes**: `POST /inbox/start`, `POST /inbox/:id/stop`, `GET /inbox`, `PUT /inbox/:id`, `GET /inbox/:id/requests`, `DELETE /inbox/:id/requests`, `GET /inbox/:id/live`.
- **The bind is a trust boundary, decided in the doc as the issue asked.** Default `127.0.0.1`; anything outside 127.0.0.0/8 and `::1` is a `400` (`inbox_non_loopback_bind`) unless the caller sends `"confirmNonLoopback": true`, and every read then reports `loopback: false` so the UI badges it. `docs/engine/architecture.md` carries **why only the inbox may bind wide**: the management API has no route auth and answers CORS `*`, so its blast radius is every stored request and credential; an inbox serves capture-and-echo and nothing else.
- **Validation is loud, never a fallback.** A status outside 100-599, a negative or over-cap `delayMs`, a non-object `headers`, an out-of-range `port` and an empty `bind` are each a 400 naming the field. A listener quietly answering something other than what it was asked to is invisible on both sides of the wire.
- **`PUT /inbox/:id` is a live merge-patch** - the next caller gets the new response, no restart, no captures lost - and accepts either the response object or `{"response": {...}}`, so a client can send back what `start` handed it.
- **SSE**: one event per capture, SSE `id:` = the capture's storage id, so a reconnect with `Last-Event-ID` resumes exactly. **One stream per inbox** (`409`, `inbox_live_in_use`): each SSE handler occupies a cpp-httplib pool thread for its whole life, and the server overrides no `new_task_queue`.
- **Storage**: new `inbox_requests` table. `add_inbox_request` inserts and trims in one transaction, keeping the newest 500 per inbox and evicting the oldest **by `id`**, not `received_at` - two captures can share a millisecond, and a timestamp cutoff would evict both or neither. Bodies store 64 KiB with `body_bytes`/`body_truncated` stating what was cut; a request over 8 MiB is refused at the transport. `init()` clears the table wholesale, for the same reason it reconciles orphaned runs: no inbox survives its process, so every surviving row belongs to an inbox nothing can list.
- **A store failure answers 500**, not the canned response - a dropped capture that looked accepted is invisible to both sides.

### App changes

- `"inbox"` added to `TabType` and answered explicitly in all three switches that dispatch on one - `Shell.renderTabContent`, `TabStrip`'s descriptor/icon, and `isTabDirty`. The last is stated rather than left to `default` because a missing branch reads as *clean* and is LRU-evictable; `tab-type-coverage.test.ts` now walks the union through all three and fails on a gap (mutation-checked: dropping the `inbox` case fails it).
- `modules/inbox/` - start/stop, URL with copy, running/live badge, non-loopback warning chip, canned-response controls, live capture list, and a detail pane that renders a capture through `UnifiedResponseViewer` + `buildRawRequest`. A capture is an exchange with no response, which that viewer already handles: **a request you received should read like one you sent.**
- Streamed captures are merged into the same query cache the first fetch fills, keyed on capture id, rather than kept in a second list beside it - two lists would need reconciling on every clear, and whichever the detail pane read would decide which was true.

### Deviations from the issue spec, both flagged rather than silent

- **Entry point** (the issue left it as an owner call): a fifth tile on the welcome Launcher, `Inbox`, and the row goes 4 → 5 columns. Not the Dock - its left group is a `<nav>` of *drawer-view switchers*, and an inbox has no drawer view to switch to. Say the word and I will move it.
- **`GET /inbox` lists stopped inboxes too**, and stopping frees the listener without deleting the record. The issue's route list has no `DELETE /inbox/:id`, so a stop that dropped the record would strand its captures behind an id nothing could name - the "written but never read" defect, one route over.

## Type of Change
- [x] New feature
- [x] Documentation update

## Testing

- **Engine `ctest --preset linux-dev`: 1317/1317** (17 new in `inbox_test.cpp`) - payload validation incl. the non-loopback refusal and the merge-patch keeping absent fields; a live listener capturing method/path/query/headers/body/remote-addr on POST *and* DELETE; the canned 500 with its delay asserted as a lower bound; an over-cap body stored truncated with the flag; the ring bound evicting the oldest and only within its inbox (mutation-check: remove the trim and the count is 5, not 3); `get_inbox_requests_since` feeding the live poll; per-inbox clear plus the startup clear; the list core's 404 and both pagination pages; one-live-stream-per-inbox; stop refusing further connections while keeping the history; and teardown with a listener still running.
- **App `pnpm test`: 3861/3861** (12 new) - `CaptureDetail` rendering a capture through the shared viewer and the raw HTTP text, the truncation notice, the merge being idempotent on capture id, malformed events dropped rather than defaulted, and the tab-type coverage guard (which asserts it parsed a **non-empty** union first, per the CSS-import guard that spent weeks reading `""`).
- `pnpm type-check`, `pnpm lint`, `pnpm format:check` and `mkdocs build --strict` all clean. No pre-existing failures observed on this base.

Routes are covered through their extracted cores rather than an in-process `vayu::http::Server`, matching the suite's convention - no existing test starts the full server.

## Checklist

## Related Issues
Closes #480

Coordination note for #479 (PR #498) and #481 phase 2, as that issue's Sequencing anticipates: this is the second manager-owned listener, so `server.cpp`'s registrar list, `server.hpp`'s member block and the new listener inventory in `docs/engine/architecture.md` are the three places that will conflict at merge time. Whichever lands second rebases; the common lifecycle is worth extracting once the third one exists, not before.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DBo5r3gwqfgoTeoccLDimc)_


athrvk:
are the inbox constants/settings user configurable

athrvk:
Claiming this - working through which of the inbox constants should become settings now, and pushing to this branch.

---

athrvk:
Three of the five now are - pushed as `af92a0f`.

Five constants came in with this feature. Two are things a user would actually reach for and a third is the same knob `oauth2RefreshPollIntervalMs` already is, so all three are settings, seeded through `seed_default_config` exactly like your `oauth2Refresh*` keys - so the Settings panel renders them with no app change:

|---|---|---|---|
| `inboxMaxBodyBytes` | 65536 | 256-8388608 | How much of an inbound body a capture stores. Past it the body is kept as a prefix and flagged; `bodyBytes` always reports the size as received |
| `inboxMaxCaptures` | 500 | 1-10000 | Captures retained per inbox, and the ceiling on one page of the capture list |
| `inboxLivePollIntervalMs` | 250 | 25-5000 | Delay between a webhook landing and its row appearing |

All three land in **Observability & Data**, since they are about captured data and its live stream rather than the listener's networking.

**The other two stay constants, deliberately** - and unlike the monitor case, the argument here is not "internal manners":

- **The 8 MB transport limit** bounds what an *unauthenticated remote caller* can make the engine buffer. Every other limit in this feature spends the local user's disk on the local user's behalf; this one spends the engine's memory on a stranger's behalf, and an inbox is the one listener that may be bound beyond loopback. That is a rail, not a preference.
- **The 30s ceiling on the canned `delayMs`** is how long a *stop* can be made to take: the delay holds a listener thread and teardown joins it. Raising it would be the user configuring their own hang.

`DEFAULT_PAGE_LIMIT` (50) is the third non-setting and the weakest case of the three - it is only the fallback for a client that names no `limit`, and the app always names one, so exposing it would be a setting with no reader on the path the app takes. Happy to expose any of these if you disagree; each is one seed entry.

Two things worth flagging in how it is wired:

- **`read_inbox_limits(db)` is the single reader**, and it resolves **once when an inbox starts** - so a change applies to the next inbox started, and a running listener keeps what it was started with. That is deliberate rather than incidental: resolving per capture would make one inbox's list a mix of rows truncated at 64 KiB and rows truncated at 256, with nothing on the row saying which rule it met. `InboxManager::limits(id)` is how the routes read them back, so the page cap and the live cadence describe the inbox they belong to.
- **The capture-list page cap now follows `inboxMaxCaptures`**, not the constant. Without that, raising retention to 2000 would have left the engine refusing to hand back more than 500 per page - a setting that half-applied, which is the "written but never read" shape one level over.

New tests, 22 in the file now: the seeded defaults and that each is a real config row with min/max (a constant with no entry is invisible in Settings); the configured body limit truncating and the configured retention binding the ring, both **mutation-checked** - pointed back at the constants, those two fail; a running inbox keeping its limits while a newly started one takes the change; and out-of-range stored values falling back to their seeds.

Engine **1322/1322**, prettier and `mkdocs --strict` clean. `app/src/config/network.ts` picked up a comment fix in the same commit - it claimed the engine "retains 500 captures per inbox", which is now a setting rather than a fact.

---

athrvk:
fix merge conflicts

athrvk:
Merged, pushed as `5ba6dff`. Master had gained 17 commits, including #498 (mock issuer) - the collision this PR's body predicted, in exactly the three places it named plus two more.

Five conflicts, each resolved toward master's shape rather than beside it:

- **`server.hpp` / `server.cpp` / `routes.hpp` / `CMakeLists.txt`** - both sides added a manager-owned listener. `MockIssuerManager` and `InboxManager` are now both members declared before `server_`, both in `RouteContext`, both registered; the inbox's comment says "and once more, for the third listener family" rather than restating the rule a third time.
- **`database.cpp`** - #475's `add_monitor_sample` family and this branch's inbox-capture family landed in the same spot after `count_metric_ticks`. Both blocks kept, in master's order.

**Two documentation inventories disagreed after the merge, which is the part worth reading.** #498 added a "Listener inventory" paragraph under Auth Resolution ending *"None of them may bind wider than `127.0.0.1`"*; this branch added a "Listeners" section under HTTP Server whose table did not know about mock issuers. Left alone the merge would have produced a doc that is wrong in one place and incomplete in the other - the exact stale-line failure CLAUDE.md warns about. So:

- The **Listeners** section is now the single inventory: four rows, including the mock issuer, with the bind rule stated per listener. The management API may not bind wide because it is unauthenticated with CORS `*`; the callback and the mock issuer may not because one carries an authorization code and the other hands out bearer tokens; the inbox may, because it serves none of that.
- The auth section's paragraph keeps the two auth-specific facts (TTL, at most 8) and points at that inventory instead of restating a rule that is no longer true engine-wide.

Verified after merging rather than after resolving: engine **1396/1396**, app **3952/3952**, type-check, lint, format and `mkdocs --strict` all clean.

---

PR #501, opened 2026-08-11
feat(app/mcp): monitor block on start_load_run, and a panel for the report's vitals summary

### LINKED ISSUE #491: engine/app(monitor): the two pieces #475 deferred - `start_load_run` cannot ask for server vitals, and the report's monitor summary has no panel
Deferred out of #475 (PR #490), which shipped the scrape, the storage, the endpoint, the live frames and both chart rows. Filed so the recorded deferrals have a tracker rather than living in a PR body. Two small, independent pieces; neither blocks anything.

## Problem

**1. An MCP agent cannot ask for server vitals.** #475's "Docs (same commit)" line made the MCP change conditional ("`docs/engine/mcp.md` if `start_load_run` grows the block"), and PR #490 kept it out to hold the diff to the app's own path. So a `monitor` block is reachable from the load-test dialog and from a hand-written `POST /runs`, but not from `start_load_run` (`app/electron/mcp/`) - the one caller that has no dialog behind it. An agent asked to "load-test this and tell me why it slowed down" has the engine-side capability and no way to name it.

**2. The report's `monitor` section is fetched but never displayed.** The engine writes a per-series summary (`min`/`max`/`avg`/`count` plus `samples` and `failures`) and `RunReport.monitor` types it app-side, but the only reader is the gate on the history view's `GET /runs/:id/monitor` fetch (`LoadTestDetail.tsx`). The numbers themselves are on screen nowhere: a reader who wants "what was peak CPU during this run" has to hover the chart. **Verify before building** - if a panel has since appeared, this half is done.

Neither is a defect in the "written but never read" sense (both have a reader today); both are capability the engine already has and the surfaces do not.

## Fix / solution pathway

**1. MCP.** Add `monitor` to `start_load_run`'s schema, shaped exactly as the engine's block (`url`, `intervalMs`, `format`, `series`) - passed through verbatim, not re-validated, since `validate_run_config` already gates it with the messages the tool should surface. Two things to check against the existing safety layer rather than assumed:
- `app/electron/mcp/safety.ts` gates ad-hoc runs on the composed URL allowlist. The monitor URL is a **second** host this run will contact, and it is not the request's - decide explicitly whether it is allowlisted too, and say which in the PR. Scraping the user's own `localhost:9100` is the whole point of the feature, so a blanket allowlist check would break it; a loopback/private exemption is the likely answer, and it needs to be a decision rather than an oversight.
- The scrape outlives nothing the duration cap does not already bound, so the duration caps need no change - confirm that against `checkLoadCaps`.

**2. Panel.** Render `report.monitor` beside the Server Vitals chart in the history Performance tab (`PerformanceTab.tsx`), as a small per-series min/max/avg row. Absent section = no panel, as everywhere else. `failures > 0` is the one figure worth calling out in words: a run with `samples: 0` and a non-zero `failures` scraped an endpoint that was never reachable, and today that reads as an empty chart with no explanation.

## Tests

- MCP: schema round-trip through `start_load_run` (the block reaches the engine payload unchanged); the allowlist decision, whichever way it goes, asserted with a loopback monitor URL against a non-loopback request URL - mutation-check it.
- App: panel renders with the section, absent without it; the all-failures case says so rather than showing zeros.
- Commands: `cd app && pnpm test && pnpm type-check && pnpm lint`; engine suites only if the MCP change touches engine code (it should not).

## Docs (same commit)

`docs/engine/mcp.md` (the `start_load_run` schema and the allowlist decision) for part 1; nothing for part 2 unless the panel changes a documented shape.

## Sequencing

**Blocked on #475 / PR #490 landing** - both halves edit surfaces that PR creates. Parallel-safe against everything else open; the two halves are independent of each other and can land separately. Conflict surface, not dependency: #476 and #477 also touch `PerformanceTab.tsx`.

## Acceptance criteria

- [ ] `start_load_run` accepts a `monitor` block and the run scrapes it; the allowlist decision for the monitor host is stated in the PR and pinned by a test.
- [ ] The history Performance tab shows the per-series summary when the run recorded one, and says so when every scrape failed.
- [ ] Absent-section behaviour unchanged everywhere; suites green; docs same commit.

athrvk:
## Description

The two pieces #475 deferred. Both halves verified against the code first, and both were still exactly as the issue describes at `6e17ce6`: `start_load_run`'s schema has no `monitor` key, and `report.monitor`'s only reader is `LoadTestDetail`'s gate on the `GET /runs/:id/monitor` fetch - the numbers themselves are on screen nowhere. No panel has appeared since the issue was filed.

**Plan.** Root cause is the same on both sides and is not the repo's "written but never read" defect: the engine has the capability (`monitor` on `POST /runs`, the per-series summary in the report) and two surfaces never grew the way to reach it. Approach: forward the block verbatim from the MCP tool the way `thresholds` is forwarded, and render the summary as its own card in the Performance tab the way `PhasePercentiles` is. **The alternative I rejected** was re-validating the block MCP-side to mirror `thresholds`' bounds: `monitor.series`' ceiling is the `monitorMaxSeries` **setting**, not a compile-time constant, so a second copy here would start refusing blocks the engine accepts the moment a user raises it - and the MCP layer cannot import `app/src/constants/monitor.ts` anyway (`tsconfig.node.json` pins `rootDir: "electron"`; TS6059). This layer types the shape, the engine owns the ranges, and `validate_run_config` already answers each one by field name.

### The allowlist decision, stated

**A loopback or private-network `monitor.url` is exempt from the allowlist; a public one is checked exactly as the target URL is.** The allowlist exists to stop an agent generating traffic against third parties it was never pointed at, and a private address is by definition the user's own network - which is also the feature's own case, since the endpoint a load run wants beside it is the target's own `localhost:9100`. A blanket check would make the capability unreachable in the case it was built for. The engine draws the same line for the same reason (`read_monitor_block`, `engine/src/core/monitor.cpp`), and a public monitor host is one GET per scrape at some third party - exactly what the allowlist is for.

Exempt: `localhost`, `127.0.0.0/8`, `10/8`, `172.16/12`, `192.168/16`, `169.254/16`, `::1`, `fc00::/7`, `fe80::/10`, and `0.0.0.0`. The test is **textual**, like the allowlist itself: a DNS name that *resolves* to a private address still needs an entry, because resolving it here would make the guard's answer depend on the network it was asked on. `URL.hostname` normalises the alternate IPv4 spellings, so `http://2130706433/` is read as `127.0.0.1` rather than as an unrecognised string.

**Caps: confirmed unchanged, as the issue asked.** `checkLoadCaps` needs nothing new - the monitor thread is joined when the run ends, so whatever bounds the run bounds the scrape, and the cadence bounds and the series ceiling are the engine's own.

### The panel

`report.monitor` renders as a per-series min/avg/max card under the Performance tab's chart, through the chart's **own** `fmtVitals` (now exported from the uplot module rather than copied), so a peak read in the panel and the same peak read off the line are spelled identically. It states the two outcomes a chart structurally cannot draw, because both look like an empty panel:

- **every scrape failed** (`samples: 0`, `failures > 0`) - a section, a count, and no line above it, which until now read as an unexplained empty chart;
- **scrapes succeeded but carried none of the requested metric names** (`samples > 0`, no series) - a typo in a metric name, not an outage, and identical to one on the chart.

A series' own reading count is printed only when it lags the run's sample count, since that is what explains a sparse line; an equal count says nothing. Absent section = no card, as everywhere else.

### Conflict note

`electron/mcp/tools.ts`'s `start_load_run` is also touched by PR #493, which extracts `buildLoadRunPayload` out of the same handler. Whichever lands second rebases; the two changes are adjacent lines in the same function, not competing designs.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

All from `6e17ce6`. No engine code changed, so no engine suite run - per the issue's own test section.

| Suite | Result |
|---|---|
| `cd app && pnpm test` | **3973 / 3973 passed**, 391 files (182s) |
| `pnpm type-check` (all three tsconfigs) | clean |
| `pnpm lint` (0 warnings) | clean |
| `prettier --check` on every file touched | clean (all were clean at HEAD, so formatting them is in scope) |
| `mkdocs build --strict` | built |

New tests - 12 MCP, 6 app:

- **`safety.test.ts`**: 12 private/loopback spellings exempt from an *empty* allowlist (incl. `[::1]`, `fd00::`, `fe80::`, and the decimal-IPv4 form); 7 public ones refused, with the RFC1918 edges pinned in both directions (`172.15` and `172.32` are not private, `172.16`/`172.31` are); a public host that *is* allowlisted; an unresolvable `{{variable}}` monitor URL; `allowAll` covering it too.
- **`tools.test.ts`**: the block reaches `POST /runs` unchanged; no `monitor` key when none was asked for; **a loopback monitor beside a non-loopback target** (the shape that motivates the decision) starts the run; a non-allowlisted public monitor is refused **and no run starts**; the ranges are left to the engine while a wrong *type* is still caught.
- **`MonitorSummary.test.tsx`**: rendered **through `PerformanceTab`**, because the half that was missing was the reading - a panel test that never goes through the tab would pass with the tab still ignoring the section. Covers the min/avg/max values, no panel for an absent section, the all-failures wording, the partial-failure count, the succeeded-but-no-metrics case, and the per-series count appearing only when it lags.

**Mutation-checked** (reverted, confirmed red, restored):

- dropping the private-host exemption → **14 tests red** across both MCP files;
- dropping the monitor gate from the handler → the public-refusal test red;
- renaming the forwarded payload key → the round-trip test red;
- disabling the `report.monitor` render in `PerformanceTab` → **5 of the 6 panel tests red** (the "absent section draws nothing" case correctly stays green).

## Checklist

Docs updated in the same commit: `docs/engine/mcp.md` (the tool table row, the `monitor` block and why its bounds stay engine-side, and the allowlist decision under Safety model), `docs/app/COMPONENTS.md` (the new component and what it says that a chart cannot).

## Related Issues
Closes #491

---
_Generated by [Claude Code](https://claude.ai/code/session_01SGUQEQToLAdxpnsQsn7Rqo)_


athrvk:
fix merge conflict

athrvk:
Conflict resolved and pushed - the branch was 8 commits behind (`6e17ce6` → `fe6a7af`, picking up #495 and #499).

**One conflict, in `docs/engine/mcp.md`**: this PR and #495 each appended a bullet to the same `Notes:` list under the tool table - my `start_load_run`'s `monitor` block, and #495's `compare_runs` baseline resolution. Both kept, mine first so it sits beside the `start_load_run` bullet it extends. No text from either side was dropped, and nothing about the two claims interacts.

Everything else auto-merged. Worth noting since the overlap was real rather than textual: `electron/mcp/tools.ts` took both #495's `compare_runs` tool and this PR's `monitor` block on `start_load_run` cleanly - verified after the merge that the schema field, the second allowlist gate and the payload forwarding are all still in the handler, not just that git reported no conflict.

Re-verified on the merged tree rather than only on the old base:

| Check | Result |
|---|---|
| `pnpm test` | **4009 / 4009 passed**, 396 files (240s) - up from 3973/391, the difference being what master brought in |
| `pnpm type-check` (all three tsconfigs) | clean |
| `pnpm lint` (0 warnings) | clean |
| `mkdocs build --strict` | built |

No engine suite: this branch still changes no engine code, and the engine work the merge brought in (#499's inbox) is already green on master.

---

Issue #523, opened 2026-08-11
app(settings): the shell pass - search, one NumberSettingRow, one card rule, one save-model story, and the end of window.alert


athrvk:
Sub-issue 5 of the settings overhaul #518 (verified at master `298e174`). The structural pass, deliberately last in the wave - it rebuilds the rows and cards the wave-1 copy lands in.

## Problem / Fix, by finding (each anchored in the #518 sweep)

1. **No search (high).** ~45 engine entries plus 7 app panels, findable only by guessing the category. Add a search/filter input at the top of the settings sidebar matching key, label, and description (all already in the payloads); results navigate to the owning panel/category with the entry highlighted.
2. **Hand-rolled numeric row, four divergent copies (high).** The "number input + unit suffix + range hint + default line" row exists in four places with divergent markup AND behaviour (engine rows validated+explicit-save; DashboardPanel immediate-commit; LoadTestingPanel clamp-on-blur; MCP debounced). Add `NumberSettingRow` to `SettingControls.tsx` (label, description, unit suffix, min/max hint, validation + `aria-invalid`, optional Default line, pluggable commit strategy) and migrate all four - the repo's hand-rolled-primitive rule applied to its own settings.
3. **Save-model story (high).** Three models coexist (engine explicit-save, app autosave, MCP debounced) and nothing on screen says which applies. `ClientSettingsPanel` owns every app-panel header - one line ("Changes are saved automatically") covers seven panels; the engine view states its explicit model beside the Save bar; MCP states debounce.
4. **Silent flush (medium).** Leaving an engine category silently saves pending edits and silently DISCARDS invalid ones. Route the flush through the save-store (Dock shows Saving/Saved) and toast discarded edits ("1 invalid change was discarded").
5. **Card granularity rule (medium).** One rule: a Card is a topic; settings are rows inside it (the Editor/Interface pattern). Refactor Dashboard and Notifications (currently one card per setting) onto it.
6. **ToggleRow migration (medium).** McpSettingsPanel hand-rolls the ToggleRow shape; extend `ToggleRow` (disabled prop, right-slot) and migrate - after #522's copy lands.
7. **Category metadata registry (medium).** `ENGINE_CATEGORY_META` (tree) and `CATEGORY_TITLES` (main) are two hand-maintained maps; extract one engine-categories registry mirroring `app-panels.ts`; sidebar order stops depending on map iteration.
[...]

Issue #521, opened 2026-08-11
app(settings): the default font contradicts the caption and the design-system doc, the reset dialog undersells what it wipes, and three descriptions need the reword pass


athrvk:
Sub-issue 3 of the settings overhaul #518 (verified at master `298e174`). App-side copy and one hard defect; the client-settings inventory found zero written-never-read keys - this surface is healthy, these are the residue.

## Problem

1. **Default-font contradiction (the one hard defect).** `DEFAULT_UI_FONT = "inter"` (`app/src/constants/appearance.ts:55`) while the Space Grotesk picker option is captioned "Default - geometric sans" (`appearance.ts:30`) and `docs/design-system.md:818,827` names Space Grotesk as the UI/body default. A fresh install renders Inter; the UI and the doc both claim otherwise. **Owner decision: which is the intended default?** Then align all three (code constant, option caption, doc) - and if the constant changes, note the visual diff for existing users whose stored value is absent.
2. **Reset scope undersells.** The "Reset app settings" card and its confirm dialog say "Appearance, editor, and dashboard preferences" (`GeneralPanel.tsx:250-252,272`) but `resetAll` clears `STORAGE_KEYS.CLIENT_SETTINGS` wholesale - notifications, auto-save, and the load-test ceilings go too. Fix the dialog text to enumerate truthfully (or scope the reset; text fix recommended - a partial reset is its own surprise).
3. **Rewords** (drafts per #518 voice conventions, refine in PR):
   - `themeSource`: card description restates the three option labels beneath it. Draft: "Sets the app's light or dark palette. System follows your operating system and switches automatically."
   - `sloThresholdMs`: text predates #470 - it should say what the setting now actually does (drives the dashboard breakpoint/saturation markers AND prefills the capacity run's SLO in the load dialog, `LoadTestConfigDialog/index.tsx:272-291`).
   - `lineNumbers` and the Auto-save toggle rows omit descriptions their siblings carry - add one-liners.
4. **Recorded decision, owner's call**: `LOAD_TEST_DEFAULTS` (`load-test.ts:32-51` - the dialog's starting values: 100 rps, 60s, 10 connections, 10% sample rate) are fixed while the panel only offers *ceilings*. Either add editable defaults to the Load testing panel (small, the FIELDS array pattern is there) or record "starting values are deliberate, ceilings are the policy surface" here and close that gap as a non-goal.

## App changes

[...]

Issue #519, opened 2026-08-11
engine(config): 12 dead knobs - wire or retire the settings that do nothing, and fix the two descriptions that contradict their neighbours


athrvk:
Sub-issue 1 of the settings overhaul #518 (verified at master `298e174`). The engine seeds 51 config entries; these 12 change nothing when a user edits them. The retire mechanism already exists: `seed_default_config` deletes retired keys at `engine/src/db/database.cpp:1604-1616` (it removed `requestBatchSize`/`contextPoolSize` "so the Settings UI stops offering a dead knob") - most of this issue is adding to that list.

## Problem - each verified, reader search stated

**Logged but never applied (wire-or-retire, decide per entry)**: `dbCacheSize`, `dbTempStore`, `dbMmapSize`, `dbWalAutocheckpoint` - each is read into a startup log line (`database.cpp:583-595`) while the actual PRAGMAs in `on_open` apply compile-time constants (`database.cpp:383-385,400`). The user changes the setting, restarts as the label demands, and the constant applies anyway.

**Written-never-read (retire)**: `maxConnections` (no reader anywhere; the `server::MAX_CONNECTIONS` constant it mirrors is itself barely referenced); `tcpKeepAliveIdle` + `tcpKeepAliveInterval` (`curl_utils.cpp:505-506` sets the CURL options from constants, never the config); `statsInterval` (superseded by `liveTickIntervalMs` - same 100ms default, same purpose, renders beside it, only the latter is read at `run_manager.cpp:1380`); `maxJsonFieldSize` (`json.cpp:822` uses the constant; the description's "increase only if..." escape hatch is fiction); `sseConnectTimeout`, `sseMaxRetry`, `sseSendLastEventId` (the `sse::*` constants appear only in the seeds; no engine or app reader - the dashboard's reconnect logic never consults them).

**Copy defects in live entries (same commit)**: `liveReplayWindowMs`'s closing claim "Memory is bounded at 20,000 ticks per run" contradicts the next entry - the ceiling is `liveMaxRetainedTicks`, default 50,000, configurable; reword to defer to it. `dbBusyTimeout` description has a trailing space. `eventLoopMaxPerHost` label breaks Title Case. `dbTempStore`/`dbSynchronous` encode enums as "0 = X, 1 = Y" prose over integer inputs - if either survives the wire-or-retire call, convert per #518 convention 5.

## Fix / solution pathway

[...]

Issue #518, opened 2026-08-11
Settings overhaul (2026-08-11 sweep): a third of the engine catalogue is dead or internal, copy contradicts mechanisms and docs, and the shell needs search, one numeric row, and one save-model story - tracking issue


athrvk:
Owner direction, 2026-08-11: "review the settings page, all the settings - simplify the UI/UX, reword the descs, and triage what should and should not be user-configurable." Full sweep at master `298e174`: 26 app client settings, 51 engine config entries, 11 MCP panel controls, structural UX review, and a voice audit of every description string. Every claim in the sub-issues carries verified anchors.

## Headline findings

- **12 of 51 engine entries are dead** (written-never-read or logged-never-applied) and 5 more are internals with no user story - a third of the catalogue offers knobs that do nothing or should not be front-row. Sub-issue 1.
- **The restart-required channel is a string convention** - the label substring `(Requires Restart)` is parsed by two consumers, stated three redundant ways per card, and stale on `workers` (#197 owns that one label). Sub-issue 2.
- **App copy has one hard defect** - the default UI font contradicts the picker caption and `docs/design-system.md` - plus a reset dialog that undersells what it wipes. Sub-issue 3.
- **Two MCP safety-cap descriptions are wrong about their own mechanism** (`maxRps` is inert for closed-loop modes; `maxConcurrency` does not govern in-flight), on the one panel where wrong text is a safety problem. Sub-issue 4.
- **The shell**: no search over ~45 engine entries, key-sorted ordering, four divergent hand-rolled numeric rows, three coexisting save models with no on-screen statement, a silent save-flush that discards invalid edits, `window.alert` in General, and duplicated category-metadata maps. Sub-issue 5.

## Voice conventions (all sub-issues write copy against these)

1. Descriptions are declarative sentences with periods; labels are Title Case nouns, no trailing punctuation.
2. Say what changes and when it takes effect, in user language; never restate the label; one to three sentences (the six-sentence engine essays get trimmed, detail moves to docs).
3. Units live once, as an input suffix - not in labels, not in parens.
4. "Max"/"Min" abbreviated in labels everywhere (matching the app side); numbers with units in descriptions spell the unit.
5. Enumerable options render as selects/option buttons, never as "0 = X, 1 = Y" prose over an integer input.
6. Restart-required is stated once, by the structured chip - not in labels, not in description tails.

## Sequencing - the wave map

[...]

PR #528, opened 2026-08-11
feat(config): make restart-required and advanced typed fields on a config entry

### LINKED ISSUE #520: engine/app(config): one structured restart-required channel, an Advanced group for the five internal knobs, and a pending-restart signal in the Dock
Sub-issue 2 of the settings overhaul #518 (verified at master `298e174`).

## Problem

1. **Restart-required is a string convention stated three redundant ways.** The only signal the engine sends is the literal substring `(Requires Restart)` in the label, substring-parsed by BOTH consumers - `SettingsMain.tsx:64` (`isRestartRequired`) and `app/electron/mcp/tools.ts:818` (`restartRequiredAmong`, feeding `update_config`'s result at `:1238`). One card then says it three times: label suffix, rendered "Restart Required" chip, and a closing description sentence. A stale label (the `workers` case, #197) lies through all three at once, and nothing anywhere tells the user a restart is currently *pending*.
2. **Five live entries are internals no user story reaches**, rendering as first-class rows beside daily-use settings: `dbBusyTimeout` (SQLite lock pragma), `oauth2RefreshRetryMs` / `oauth2RefreshRetryMaxMs` / `oauth2RefreshPollIntervalMs` (watchdog backoff internals - the ceiling's description even depends on reading its sibling), `inboxLivePollIntervalMs` (a 250ms poll cadence below anyone's tuning threshold). Membership is the owner's call; the mechanism is this issue's.

## Fix / solution pathway

1. **Structured flag**: `ConfigEntry` gains (or starts honouring - a typed `requiresRestart` field partially exists; verify and complete) a boolean serialized on `GET /config`; both consumers read the field, the substring parsing dies, labels drop the suffix, descriptions drop the tail sentence - the chip becomes the single statement per #518 convention 6. Seeds updated accordingly.
2. **Advanced group**: entries gain an `advanced` boolean (same seed mechanism); the app renders advanced entries in a collapsed "Advanced" section at the bottom of their category (collapsed state not persisted - rediscoverable). Initial membership: the five above; owner may amend on this issue before implementation.
3. **Pending-restart signal**: when a restart-required entry's saved value differs from the engine's running value (the engine can report its effective value, or the app tracks saves since connect - choose the honest one and state it), the Dock's engine-status area shows an amber "restart pending" state with the existing restart action; the affected entry's card shows a dot. This is the "what confirms my change took effect" answer for the one class where autos-ave cannot.

## App changes

`SettingsMain.tsx` (flag read, Advanced section), `app/electron/mcp/tools.ts:818` (field read - `update_config`'s `restartRequired` result shape unchanged for agents), `Dock.tsx` (pending state beside `EngineStatus`, mirroring its guard-test pattern). Engine: `ConfigEntry` serialization + seeds.

## Tests

Engine: the flag serializes; seeds carry no `(Requires Restart)` substrings (a grep-shaped test so the convention cannot creep back - assert it scanned a non-empty catalogue per the repo rule). App: advanced section collapses/expands and contains exactly the flagged entries; Dock pending state appears when a restart-required save lands and clears after restart (mutation-check: render it unconditionally, the absent-case fails); `restartRequiredAmong` behaviour pinned unchanged through the field. Commands: both suites, full commands per #518.

## Docs (same commit)

`docs/engine/api-reference.md` config-entry shape (the new fields); `docs/engine/mcp.md` `update_config` row only if its text names the label convention (verify).

## Sequencing

Wave 1 of #518 - parallel-safe against sub-issues 1/3/4 (different files; sub-issue 1 edits seed *values*, this edits seed *shape* - coordinate the one-line overlap in `database.cpp` seeds, trivial rebase). Land before sub-issue 5, which consumes the structured flag for its restart-communication work. #197's `workers` label fix becomes simpler after this (the suffix is gone catalogue-wide) - note added there is unnecessary; #197 already owns only the stale *description sentence*.

## Acceptance criteria

- [ ] Restart-required flows as a typed field end to end; the substring convention is gone and grep-guarded against return.
- [ ] The five internals (or the owner-amended list) render collapsed under Advanced; the list is recorded here.
- [ ] A pending restart is visible in the Dock and on the entry; suites green, mutation-checked.

---

athrvk:
## Description

The only signal the engine sent for "this needs a restart" was the literal substring `(Requires Restart)` in an entry's **label**, regex-matched by *both* consumers - `SettingsMain.tsx:65` and `tools.ts:818` (feeding `update_config`'s `restartRequired` result) - so one stale label misinformed the settings screen and an agent at the same time, and a single card stated the fact three times over (label suffix, chip, closing description sentence). Separately, five live entries are internals no user story reaches, rendering as first-class rows beside daily-use settings because key order is the only ordering the panel has.

**Plan.** Root cause: the restart requirement was carried as prose, so it had no single owner - the copy could not be reworded without minding two parsers, and the two parsers could not be trusted without minding the copy. Approach: `ConfigEntry` gains `requires_restart` and `advanced` as real columns, serialized on every entry by the existing `config_entry_json` (the one shape both `GET /config` and `POST /config` answer with), and both consumers read the field. The alternative rejected: deriving the flag app-side from a key list, which is the "config one branch defines and another re-derives" defect `CLAUDE.md` calls out - the engine knows which values it reads at startup, and a list in the renderer would drift the moment a read site moved.

### Decisions worth reviewing

1. **`workers` keeps `requiresRestart: true`.** That is a faithful translation of the label it carried, not a new claim. Whether it is *true of the mechanism* is #197's question (`benchmarks.md` records that it is applied per run) - that doc line now points at the field instead of the label, so the discrepancy stays visible rather than being quietly resolved by this PR.
2. **Advanced membership, as recorded here:** `dbBusyTimeout`, `oauth2RefreshRetryMs`, `oauth2RefreshRetryMaxMs`, `oauth2RefreshPollIntervalMs`, `inboxLivePollIntervalMs` - the issue's list unchanged. They are hidden, never removed: each is still live and still settable. Pinned by key in a test, so amending the group has to say so.
3. **The pending signal tracks saves, and says so.** The issue asked to choose the honest source. The engine does not report its *effective* values, so a comparison against them would be inferred rather than known - the Dock therefore reports restart-required saves made since this renderer connected (the existing `engine-store` latch) and the copy says "saved", not "in effect". Consequence, stated in the code and in `COMPONENTS.md`: it does not survive a renderer reload.
4. **`window.alert` is gone from the restart path.** The restart moved into a shared `useEngineRestart` hook (banner + Dock), so the wait-then-invalidate-then-clear sequence exists once; a modal alert raised from the Dock would block the whole app over a failure every other surface reports as a toast. A failed restart now leaves the pending signal standing - it is the user's evidence the saved value is not live yet. (The `window.alert` calls in `GeneralPanel` are #523's, untouched.)
5. **Seeds carry the flags through two named markers** - `restart_required (ConfigEntry{…})` / `advanced (…)` - rather than trailing positional booleans after `now`, which would have been two unlabelled `true`s at the end of a 10-field aggregate.
6. **Migration.** Both columns are NOT NULL with a `default_value`, so `sync_schema` adds them with `ALTER TABLE ADD COLUMN` (the pattern `requests.follow_redirects` already uses) rather than a copy-and-recreate. Verified by hand, not assumed: a database written in the pre-change shape with a user-modified `workers = 7` was opened with the new binary - the value survived, the columns were added, the label lost its suffix and `GET /config` reported `requiresRestart: true`.

### App / MCP, traced

- **App (required by the issue, done):** `SettingsMain.tsx` reads the field, renders the Advanced group, and no longer strips a substring out of two labels; `Dock.tsx` gains the pending signal; `domain.ts` types both fields as required booleans (the app and engine ship together, and an optional field would have re-introduced "maybe it just was not sent").
- **MCP:** `restartRequiredAmong` reads `requiresRestart === true`; `update_config`'s result shape is unchanged, so nothing an agent consumes moves. `docs/engine/mcp.md` did name the label convention ("the engine marks these in each entry's label") and is corrected.
- **Other consumers, checked:** nothing else in the app or the CLI enumerates config keys or parses labels (`useMonitorSettings`/`useLiveChartSettings` read named keys), and `apply_config_update` copies the stored row, so a value write cannot flatten the metadata - which now has its own test.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`python build.py -e -t` then `cd engine && ctest --preset linux-dev --output-on-failure`: **1431/1431 passed** (178s). `cd app && pnpm test`: **4047/4047 passed** (401 files, 167s). `pnpm type-check`, `pnpm lint`, `pnpm format:check` all clean. `mkdocs build --strict -f .github/mkdocs.yml` passes (2.59s). No pre-existing failures on either side.

Engine - five new tests in `config_route_test.cpp`:

- `RestartRequiredSerializesAsATypedFlagBothWays` - asserts both values, so a serializer that always sent `true` cannot pass.
- `NoSeededLabelOrDescriptionSpellsOutTheRestartRequirement` - the grep-shaped guard the issue asked for, over the whole catalogue, asserting it scanned something non-empty (per the repo rule) *and* that at least one entry still carries the flag - dropping the suffix without setting the field would otherwise satisfy it while telling the user nothing.
- `AdvancedFlagsExactlyTheRecordedInternals` - set equality by key, not a count.
- `AdvancedSerializesOnTheWire`
- `UpdatingAValueKeepsItsMetadataFlags` - the written-never-read guard on the update path: if `apply_config_update` ever rebuilt the entry instead of copying it, every restart chip would vanish after one save.

App - two new files plus the existing banner test extended:

- `SettingsMain.advanced-section.test.tsx` (4) - collapsed by default with the flagged entry *not rendered* (not merely last, which is what "collapsed" has to mean for a screen reader), expands and collapses, counts, and is absent entirely from a category with no internals. Fixtures are keyed so key-sort alone would interleave them.
- `Dock.pending-restart.test.tsx` (4) - absent when nothing is pending, present after a save, restarts and lowers itself, and - on a failed restart - keeps standing and reports the reason as an `error` toast.
- `SettingsMain.restart-banner.test.tsx` gains the twin case: an entry whose label still reads `(Requires Restart)` while the flag says `false` must raise **nothing**.

**Mutation checks (run, not assumed):**

- `isRestartRequired` reverted to `entry.label.includes("(Requires Restart)")` → 4 of the 6 restart-banner tests **FAILED**, including both new ones; restored, all pass.
- `primaryEntries` reverted to the unfiltered `categoryEntries` → 2 of the 4 advanced-section tests **FAILED**; restored, all pass.

clang-format was applied to the changed line ranges only - the touched engine files are not clang-format clean at HEAD and `CLAUDE.md` forbids reformatting the rest. Prettier ran only on files that were already clean.

## Checklist

Same commit: `docs/engine/api-reference.md` (both fields in the `GET /config` shape, with what each means and why consumers must not parse the label), `docs/engine/db-schema.md` (the two columns and the ALTER-vs-recreate reason), `docs/engine/mcp.md` (the corrected `update_config` sentence), `docs/engine/benchmarks.md` (two lines that referred to the now-deleted label suffix - they would have been stale on merge), `docs/app/COMPONENTS.md` (the Dock's pending signal and how the engine categories render from metadata).

## Related Issues

Closes #520

Sub-issue 2 of #518. Wave 1, parallel-safe against #521/#522; #519 is gated on PR #517 merging (per the owner's coordination note there) and was skipped this run. The one-line overlap with #519 in the `database.cpp` seed block is the trivial rebase the issue anticipated - this edits seed *shape*, #519 edits seed *values*. #523 (wave 2) consumes this structured flag.

---
_Generated by [Claude Code](https://claude.ai/code/session_016corypqqbyAKiAw9eL9bzw)_

### DIFF
--- a/app/electron/mcp/tools.test.ts
+++ b/app/electron/mcp/tools.test.ts
@@ -188,8 +188,18 @@ describe("engine config tools", () => {
 		const client = fakeClient({
 			getConfig: vi.fn().mockResolvedValue({
 				entries: [
-					{ key: "workers", value: "16", label: "Worker threads (Requires Restart)" },
-					{ key: "timeoutMs", value: "5000", label: "Request timeout" },
+					{
+						key: "workers",
+						value: "16",
+						label: "Worker Threads",
+						requiresRestart: true,
+					},
+					{
+						key: "timeoutMs",
+						value: "5000",
+						label: "Request timeout",
+						requiresRestart: false,
+					},
 				],
 			}),
 		});
--- a/app/electron/mcp/tools.ts
+++ b/app/electron/mcp/tools.ts
@@ -807,13 +807,19 @@ const configUpdateSchema = z
 /**
  * Of the changed config keys, which require an engine restart to take effect.
  *
- * The label is the whole signal: the engine writes "… (Requires Restart)" into
- * each entry's `label` (`engine/src/db/database.cpp`) and emits no boolean
- * beside it. A `requiresRestart === true` branch used to run ahead of this
- * regex on a field nothing has ever written - a check that could not fire,
- * mirrored in `SettingsMain.tsx` and typed in `domain.ts`, which read as if the
- * engine had a second, more reliable channel. If one is ever added, this is the
- * one place that decides.
+ * The engine says so in a typed `requiresRestart` field on each entry
+ * (`engine/src/http/routes/config.cpp`), read here and in `SettingsMain.tsx`.
+ * It replaced a "… (Requires Restart)" substring in the entry's `label` that
+ * both consumers regex-matched out of the prose - so a label that fell out of
+ * step with the mechanism (the `workers` case, #197) misinformed an agent and
+ * the settings screen at once, and no rewording of the copy could be made
+ * without minding this parser.
+ *
+ * Strictly `=== true`: this decides whether the tool result tells an agent its
+ * change is live, and a payload that omits the field (an engine older than the
+ * app it is paired with) has not said that it does not need one.
+ *
+ * `update_config`'s `restartRequired` result shape is unchanged.
  */
 function restartRequiredAmong(configResponse: unknown, changedKeys: string[]): string[] {
 	const raw = Array.isArray(configResponse)
@@ -823,11 +829,7 @@ function restartRequiredAmong(configResponse: unknown, changedKeys: string[]): s
 			: [];
 	const entries = Array.isArray(raw) ? (raw as Array<Record<string, unknown>>) : [];
 	const byKey = new Map(entries.map((e) => [String(e.key), e]));
-	return changedKeys.filter((key) => {
-		const e = byKey.get(key);
-		if (!e) return false;
-		return /requires restart/i.test(String(e.label ?? ""));
-	});
+	return changedKeys.filter((key) => byKey.get(key)?.requiresRestart === true);
 }
 
 /**
--- a/app/src/components/layout/Dock.pending-restart.test.tsx
+++ b/app/src/components/layout/Dock.pending-restart.test.tsx
@@ -1,0 +1,108 @@
+/**
+ * @vitest-environment jsdom
+ */
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * A restart the engine is waiting for has to be visible outside Settings.
+ *
+ * `pendingRestart` was raised by the settings save path and read by exactly one
+ * screen - the one the user has already left by the time it matters. Every
+ * other setting confirms itself; the restart-required ones cannot, so the Dock
+ * says so and offers the restart from wherever the user is.
+ *
+ * Rendered, not source-scanned: the signal arrives through a store binding, and
+ * both halves are asserted - present when pending, absent when not. A component
+ * that rendered the line unconditionally would pass the first case alone.
+ */
+
+import { describe, it, expect, beforeEach, vi } from "vitest";
+import { render, screen, cleanup, fireEvent, waitFor } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { TooltipProvider } from "@/components/ui";
+import { Dock } from "./Dock";
+import { useEngineStore, useToastStore } from "@/stores";
+import { TIMING } from "@/config/timing";
+
+vi.stubGlobal("__VAYU_VERSION__", "0.0.0-test");
+
+const restartEngine = vi.fn();
+
+const renderDock = () =>
+	render(
+		<QueryClientProvider
+			client={new QueryClient({ defaultOptions: { queries: { retry: false } } })}
+		>
+			<TooltipProvider>
+				<Dock />
+			</TooltipProvider>
+		</QueryClientProvider>
+	);
+
+const signal = () => screen.queryByRole("button", { name: /Restart pending/i });
+
+beforeEach(() => {
+	cleanup();
+	vi.clearAllMocks();
+	useEngineStore.setState({
+		isEngineConnected: true,
+		engineError: null,
+		pendingRestart: false,
+		restartRequiredKeys: [],
+	});
+	useToastStore.setState({ toasts: [] });
+	restartEngine.mockResolvedValue({ success: true });
+	vi.stubGlobal("electronAPI", { restartEngine });
+	// The renderer reads it off `window`, which is what the component sees.
+	(window as unknown as { electronAPI: unknown }).electronAPI = { restartEngine };
+});
+
+describe("the Dock's pending-restart signal", () => {
+	it("stays away while nothing is pending", () => {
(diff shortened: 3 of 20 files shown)


athrvk:
Claiming this - master moved (#517 merged) and this branch is behind it. Merging master in, resolving the seed-block overlap, and re-running the suites before pushing.

---

athrvk:
Updated against master (`58bcf4e` merges `63175c0`, which brought in #517).

The overlap the issue anticipated turned out to need no edit: #517's new `monitorScrapeTimeoutMs` seed uses positional aggregate init ending at `now`, and this PR appends `requires_restart` / `advanced` at the end of `ConfigEntry` with `= false` defaults, so the new entry compiles unchanged and carries `requiresRestart: false` - which is correct for it, since the monitor settings are read per run. Its label and description also clear the catalogue-wide guard (`NoSeededLabelOrDescriptionSpellsOutTheRestartRequirement`), and `AdvancedFlagsExactlyTheRecordedInternals` still matches the recorded five, so no test needed amending.

Verified on the merged tree: `python build.py -e -t` clean, `ctest --preset linux-dev` **1437/1437 passed** (210s) - 1431 before, plus the six monitor tests #517 added. The merge touched no app files, so the app suite (4047/4047 on the previous head) was not re-run.

---

Issue #505, opened 2026-08-11
engine(refactor): extract the manager-owned listener lifecycle - the "third listener" trigger PR #499 named has fired


athrvk:
From the 2026-08-11 deferral audit (verified at master `3a7c937`). PR #499's coordination note set the trigger explicitly: "the common lifecycle is worth extracting once the third one exists, not before." It exists. `engine/include/vayu/http/server.hpp` now carries three sibling manager members - `OAuth2AuthorizeManager` (`:52`), `MockIssuerManager` (`:60`), `InboxManager` (`:64`) - each hand-implementing the same pattern: declared before `server_` so destruction stops and joins listener threads after the route lambdas die (the reverse-member-order rule the header documents), each with its own start/stop/join, ephemeral-port bind, and thread ownership. #481 phase 2 would add a fourth. Verified: no shared base or helper exists (`grep` for a common ListenerManager abstraction returns nothing).

## Problem

Three hand-rolled copies of a lifecycle whose correctness is subtle (member order, stop-before-join, bind_to_any_port error paths, shutdown while a handler runs) is the repo's most-documented defect class - "a hand-rolled copy of a primitive does not receive the primitive's fixes." The inbox's teardown-with-listener-running test exists; the issuer's equivalents exist; a fix discovered in one (say, a shutdown race) now needs manual porting to the other two, and #481 would make it three.

## Fix / solution pathway

Extract the shared shape - suggested: a `ManagedListener` (or CRTP/composition helper) owning the `httplib::Server`, the thread, `start(bind, port)` returning the bound port, `stop()` with join, and the destructor discipline - and rebase the three managers onto it, keeping their per-service state (attempts/issuers/inboxes) and route registration as the variable part. Constraints:

- **Behaviour-preserving**: all three managers' existing tests must pass unchanged - they are the spec. No route, payload, or lifecycle semantics move.
- The `server.hpp` member-order comment updates to describe the rule once, on the helper.
- Do not force-fit: if one manager's lifecycle genuinely diverges (e.g. the authorize manager's TTL sweep), keep that outside the helper rather than widening the helper's contract - state what stayed out and why in the PR.

## App changes

None, verified - pure engine refactor; no route or shape changes.

## Tests

[...]

PR #557, opened 2026-08-13
feat(engine/app): an examples model, so imports stop dropping every saved response

athrvk:
## Description

**Phase 1 of #481 only** - the examples model. Phase 2 (the mock server) stays open on that issue behind #505, per its 2026-08-11 sequencing update.

**Plan.** Root cause, verified at `f88adfb`: the issue's premises still hold exactly as filed. `pmRequest` (`postman.ts`) read `item.request` and never `item.response[]`; `op.responses` was visited by no code path in either OpenAPI parser; `RequestDraft`, `ImportApplyRequestItem` and the engine's table list had nowhere to put one. So a collection whose value *was* its documented responses imported with none, and the loss was not even counted as a skip. Approach: build the storage first and wire the parsers to it - a new `request_examples` table owned by its request, CRUD nested under `/requests/:id/examples`, examples riding nested on the request item of `POST /import/apply` so the whole tree still lands in one transaction, and a read-only Examples tab so the user can see they survived. **Rejected: a top-level `examples` section in the import payload, beside collections/requests/environments.** It would need temp ids in the shared namespace and would put entries in the response's `idMap` that no client asked for - nothing references an example, so nesting it on its owner is both smaller and truer.

**One deliberate addition to the issue's schema: an `order` column.** The sketch lists `(id, request_id, name, status, headers, body, content_type, created_at, updated_at)`, and that set cannot keep the order Phase 2 depends on. A bulk import writes every example of one request in the same millisecond, so `created_at` ties for all of them and the `id` tiebreak returns the author's list shuffled - caught by a test that failed on exactly that. Since "the first example on the matched request" is what a mock server serves, position has to be stored, the same way `requests.order` stores it within a collection. A create that states no `order` appends; a bulk import numbers by payload index.

### Engine

- **`request_examples`** (`id`, `request_id`, `name`, `status`, `headers`, `body`, `content_type`, `order`, timestamps), indexed on `request_id` - which backs every access the table has. Cascade-deleted with the request, and through the collection cascade one request at a time: every read here is by request id, so a row left behind would be unreachable rather than merely stale.
- **`headers` is a `KeyValueEntry[]`**, the shape a request's headers use - not the JSON object `result_bodies` / `inbox_requests` store. A stored example is re-served rather than only displayed, so repeated names (`Set-Cookie`) and the author's order both have to survive. `apply_key_value_field` moved from `requests.cpp` into `routes.hpp` so both resources validate that shape through one copy.
- **CRUD** `GET` / `POST` / `PUT` / `DELETE` under `/requests/:id/examples`, testable cores per the repo's route convention. The owner is checked before the example on every path, so one reached through the wrong request is a `404`, not a cross-request write. `status` outside 100-599 and a body over 1 MiB are **400s, not coercions** - a clamped status or a half body would be served as if it were what the caller stored.
- **`POST /import/apply`** writes nested examples through `apply_request_example_fields`, the same applier the single-item route uses, in the same transaction. They carry no `tempId` and so are absent from the `idMap`; they do count toward the 10,000-item cap, because they are rows the call allocates and writes.

### App

- **Parsers**: Postman `item.response[]` (name, `code`, `header[]`, `body`; a malformed entry counts as `malformed_item`) and both OpenAPI parsers' `op.responses` (`example` → first `examples[*].value` → `sampleSchema`, the precedence request bodies already use) map to an `ExampleDraft`. `_postman_previewlanguage` is deliberately not read into `contentType` - it is an editor mode, not a media type.
- A response keyed `default` or `2XX` is **counted as a new `example_no_status` skip kind**, not guessed at: it documents a real response, but an example is served under one status line.
- `ImportMeta.exampleCount`, read off the finished drafts by `countExamples` (the approach `unattachedFileParts` uses, so the number and the drafts cannot disagree), shown in the import preview.
- **An Examples tab** listing what an import stored - status chip, name, expanding to the recorded headers and a `ResponseBody` view. Read-only: examples arrive by import in this phase, so there is no create/edit hook and no endpoint constant beyond the list path - a constant with no reader is the defect this repo keeps finding.

## Type of Change
- [x] New feature
- [x] Documentation update

## Testing

All green on this branch, from a clean `master` base at `f88adfb`:

- `python build.py -e -t && cd engine && ctest --preset linux-dev` - **1463 passed, 0 failed** (up from 1449: 14 new in `examples_route_test.cpp`).
- `cd app && pnpm test` - **4273 passed, 420 files** (up from 4252: 15 new in `examples-import.test.ts`, 6 in `examples-tab.test.tsx`).
- `cd app && pnpm type-check`, `pnpm lint` (zero warnings), `pnpm format:check` - clean.

No pre-existing failures encountered on either suite.

Mutation-checked:

- Revert `order` to a constant → `ListIsInCreationOrderAndScopedToTheRequest` and `ImportApplyWritesNestedExamples` fail. This is how the ordering gap above was found rather than reasoned about.
- Drop the cascade from `delete_request` → `DeletingTheRequestDeletesItsExamples` fails; drop it from `delete_collection` → the descendant test fails.
- Delete the `item.response` read in `pmExamples`, or the `buildExamples` call in `buildOperation` → the first test of that parser's block fails.
- Collapse the Examples tab's error branch into the empty branch → "says the read failed rather than claiming there are none" fails.

C++ formatting: only `routes.hpp` and the two new files were clang-formatted - the other engine files this touches were already unformatted at `HEAD`, and reformatting them would bury the diff.

## Checklist
- [x] I have added tests

Docs in the same commit: `docs/engine/db-schema.md` (the table, its ordering rule, the cascade, the caps, the index), `docs/engine/api-reference.md` (a Request examples section and the `examples` field on `/import/apply`), `docs/request-storage-design.md` (why an example is not a `results` row), `engine/CLAUDE.md` (endpoint table + the nesting rule), `docs/app/api-integration.md`, `docs/app/COMPONENTS.md`, `docs/app/state-management.md`, and the import-collections pages for Postman, OpenAPI 3 and Swagger 2.0 plus their shared README.

## Related Issues

Part of #481 - **phase 1 only**, so this does not close it. Phase 2 (`POST /mock/start` and the listener) remains, blocked on #505 landing first.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SPsceSLgkJxFdShcm96Zey)_


athrvk:
CI red on `Engine on ubuntu-latest`, and it is **not this diff** - filed as #559.

The failure is `ReorderRouteTest.NormalizesTheRootCollectionsWhenParentIdIsNull`, in a file this branch does not touch (`git diff origin/master...HEAD -- engine/tests/reorder_route_test.cpp engine/src/http/routes/reorder.cpp engine/src/http/routes/collections.cpp` is empty), and **the identical suite passed on macos-latest in the same run**.

The test creates two root collections, deliberately ties their `order` to reproduce the pre-#360 shape, and then asserts which of the two normalization renumbers first. `normalize_scope` walks `get_collections()`, which orders by `order`, then `created_at`, then `id` - so once `order` is tied by construction, a runner fast enough to create both rows inside one millisecond ties `created_at` too and the decision falls to the **random UUID** in `id`. The reported value (`b->order` is `0`, expected `1`) is exactly that coin landing the other way. It passed 25/25 locally, which is what this defect looks like on a slower host rather than evidence against it.

Left alone rather than fixed here: it is a different issue's scope, and #559 carries the diagnosis and three options. Re-running the failed job once the in-progress Windows engine job finishes - I'll report where it lands.

Worth noting the connection, since it is the same root cause this PR's `order` column exists to avoid: `created_at` does not break a tie among rows written in one millisecond, and an id tiebreak past it is a lottery. That is why examples carry an explicit position rather than relying on insertion time.

---

athrvk:
**CI green.** Re-ran the failed job with no code change and `Engine on ubuntu-latest` passed - 1463/1463, same commit (`dc1b627`). `CI gate` is success, and every other job (engine and app on all three platforms) passed first time.

That is the confirmation the diagnosis wanted: nothing changed between the red and the green run, so the failure was the UUID tiebreak in #559 landing the other way, not this diff.

---

Issue #555, opened 2026-08-13
app(services): drawer correctness batch - a stale green indicator on a dead engine, toggle-close on the indicator click, anonymous unordered rows, silent create, and rows invisible to screen readers


athrvk:
From the 2026-08-13 services-drawer review (verified at master `c76f874`). Seven verified defects in the just-shipped drawer, batched by surface - none blocks daily use, together they are the polish the surface needs. #554 (addressing) and #553 (delete) land first; this batch follows.

## Problems, each anchored

1. **(medium) The Dock indicator lies when the engine dies**: it renders "N services" in green from TanStack's last-good cache while `EngineStatus` beside it says Disconnected - services are engine-process state, so none can be running. Gate the indicator on `isEngineConnected` (the store is already in the Dock).
2. **(medium) Clicking the indicator with the drawer already on services CLOSES it**: `activateDrawerView` toggles-when-current. An ambient status chip should reveal, never hide - call the non-toggling open (or guard on current view).
3. **(medium) Creating an inbox gives zero success feedback and the new row lands anywhere**: the start mutation has `onError` only, and the engine lists inboxes in map order, so the row appears at an arbitrary position unhighlighted. Sort rows by a stable key (creation time if the record carries it - verify; else port) and select-or-flash the new row on success.
4. **(medium) Rows are anonymous**: an inbox's only text is its full URL - three inboxes are three near-identical monospace strings. Label rows "Port NNNN" with the URL demoted to the copy affordance's tooltip/subtitle. **Decision point, not assumed**: a user-editable inbox *name* would need an engine field (`PUT /inbox/:id` merge-patch could carry it) - say yes/no here; the port label ships regardless.
5. **(medium) `aria-label` on the row activator hides everything from screen readers**: the label replaces the content, so the URL, "Stopped", the bind badge and failure-mode badge are inaudible. Compose the accessible name from the content (drop the label or use `aria-labelledby`).
6. **(low) Copy affordances report success unconditionally**: `useCopy` fires `clipboard.writeText` with `void` and toasts "copied" even on rejection - every other copy site in the app awaits. Align; also give a stopped row's "Copy inbox URL" a cue that the listener is gone (the copied URL refuses connections far from the cause).
7. **(low) Two error disciplines, one lifecycle**: the drawer toasts stop/clear failures; the inbox tab's same mutations have no `onError` at all. Unify on the toast (fix the tab side here or in #553's tab pass - state where).

[...]

Issue #554, opened 2026-08-13
app(inbox): the drawer row cannot open the inbox it names - the singleton tab has no selector and always falls back to the first inbox


athrvk:
From the 2026-08-13 services-drawer review (verified at master `c76f874`). Two independent reviewers hit the same root: **inbox addressing does not exist**.

## Problem

1. **The drawer row lies.** `InboxRow`'s activator is labelled "Open inbox on port N" but calls `openTab({type:"inbox", entityId:null})` (`app/src/modules/services/ServicesPanel.tsx:164`) - the singleton tab with no address.
2. **The tab cannot show a chosen inbox.** `InboxView` renders `inboxes.find(id === selectedInboxId) ?? inboxes[0]`, and `selectedInboxId` is set ONLY by the tab's own start mutation. So with inboxes A and B, clicking B's drawer row shows A. There is no selector anywhere in the tab.
3. **External stops reach the tab 10 seconds late.** An inbox stopped from the drawer (or curl) ends its live stream immediately, but the tab's state updates only on the `SERVICES_POLL_INTERVAL_MS` (10s) poll - in the window, the tab shows a live-looking inbox whose listener is gone, and the #506 reconnect machinery burns its retries against a deliberately-stopped stream.

## Fix / solution pathway

1. **Thread the address**: `openTab({type:"inbox", entityId: inboxId})`. The inbox stays a singleton TAB (one tab, retargetable) - the tab-store already carries `entityId`; the singleton semantics just need "opening with a new entityId retargets the existing tab" (verify how `SINGLETON_TYPES` handles entityId today - likely ignores it; that is the change). `InboxView` prefers the tab's entityId, then its own last selection, then `inboxes[0]`.
2. **A selector in the tab header**: a compact inbox switcher (Select over ports/labels, ordered by creation) beside the status badge - the multi-inbox story the tab never had. Selection state per the existing `selectedInboxId` plumbing.
3. **Immediate external-stop reflection**: on the live stream's terminal close, refetch the inbox list at once instead of waiting for the poll (one invalidation in the `useInboxLive` close path), and suppress #506's retry loop when the refetched record says `running: false` - stopped-is-not-an-error. The stopped tab state then renders the honest affordances (coordinate with the delete issue #553 - it lands delete on this stable selection).

## App changes

[...]

PR #563, opened 2026-08-14
fix(app/inbox): tab honesty batch - paged captures with the cut visible, all four canned fields, and a stopped panel that stops taking edits

### LINKED ISSUE #556: app(inbox): tab honesty batch - stopped-inbox controls that lie, a written-never-read hasMore, a canned response missing half its fields, and a detail pane that rewrites under the reader
From the 2026-08-13 inbox-tab review (verified at master `c76f874`). Six verified defects in the tab, batched; #554 (addressing/selector) lands first, #553 (delete) beside it.

## Problems, each anchored

1. **(medium) A stopped inbox presents active-looking controls that are lies or no-ops.** The `CannedResponseControls` stay fully enabled - `PUT /inbox/:id` happily merge-patches a record whose listener is gone, so the user edits a response nothing will ever serve. Disable-with-explanation (or hide) the canned controls and any other active affordance on a stopped record; the stopped state's honest verbs arrive with #553 (delete) and #554 (switch away).
2. **(medium) `hasMore` is written and never read - the repo's signature defect, in the capture list.** The engine returns it, `mergeCapture` preserves it, and no UI reads it: the tab fetches exactly one page of 50 (`INBOX_CAPTURES_PAGE_LIMIT`) with no load-more, so an inbox holding up to 500 ring-retained captures silently shows the newest 50. Add load-more (or infinite scroll) driven by `hasMore`, and show "N captured" vs "showing 50" so the cut is visible.
3. **(medium) The canned response has four engine fields and the tab exposes two.** `InboxCannedResponse` carries status, body, headers, delayMs (all validated by `apply_response_fields`); the tab renders status + delayMs only - a reply *body* or *header* set an agent or curl configured is invisible and uneditable from the UI, and cannot be authored there. Expose all four (body as a small editor, headers as the key-value row primitive).
4. **(medium/low) The detail pane rewrites under the reader.** `selectedCapture` falls back to `captures[0]` and the SSE merge prepends - on a streaming inbox the pane re-renders to each arriving webhook while the user reads one. Make selection sticky once a row is clicked (fall back to newest only before any explicit selection), and keep the selected row highlighted as new rows push it down.
5. **(low) SSE/initial-fetch clobber race**: a capture merged via `setQueryData` while the initial GET is in flight is overwritten when the fetch resolves. Merge the fetch result INTO the cache (union by id) instead of replacing it.
6. **(low) Truncation is invisible in the list**: `bodyTruncated` renders only in the detail chip; rows print "{bodyBytes} bytes" with no marker. Add the marker to the row.

Polish, same commit if cheap: the warning chips' `bg-status-warning-fill text-primary-foreground` pairing diverges from every other status-fill chip (use the matching on-fill token); align the tab's "Reachable on 0.0.0.0" with the drawer's bare "0.0.0.0" chip text (one wording, both places).

## App changes

`modules/inbox/index.tsx`, `CannedResponseControls.tsx`, `CaptureDetail.tsx`, `queries/inbox.ts` (pagination + merge-into-cache), the row component. Engine: none, verified - all four canned fields and the pagination already exist server-side. MCP: none.

## Tests

Stopped record disables the canned controls (mutation-check: re-enable, fails); load-more appends and the count line matches; all four canned fields round-trip through the PUT; sticky selection survives an arriving capture (fake SSE); the clobber race pinned with a delayed fetch; the row truncation marker renders. Commands: `cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check`.

## Docs (same commit)

`docs/app/COMPONENTS.md` - the inbox tab section (pagination, canned fields, stopped-state behaviour).

## Sequencing

**After #554** (selector/addressing reshapes the tab header this batch edits); parallel-safe with #553 and #555 except one brush: #555 item 7 may fix the tab's missing `onError` - whichever lands second skips the duplicate (stated in both). Parallel-safe against everything else open.

## Acceptance criteria

- [ ] No control on a stopped inbox pretends to work; every capture is reachable through load-more with the cut visible; all four canned-response fields are editable and displayed.
- [ ] Reading a capture is stable under a streaming inbox; the fetch/SSE race is pinned; truncation shows in the list.
- [ ] App suite, type-check, lint, format green; docs same-commit.

---

athrvk:
## Description

**Plan.** The root cause across items 1-3 is the same shape: the tab showed a partial view of engine state without saying it was partial. `hasMore` was written into the cache and read nowhere (one page of 50 out of up to 500 retained, indistinguishable from an inbox that had received 50); the canned panel exposed 2 of the 4 fields `apply_response_fields` validates, so a reply body or header set configured by an MCP tool or curl was invisible and uneditable here; and `PUT /inbox/:id` happily merge-patches a stopped record, so the panel stayed live over a listener that was gone. The approach is one cache entry per inbox with **one union rule** (`mergeCaptures` in `queries/inbox.ts`) that all three writers - first fetch, load-more, live stream - go through; that same union is the fix for item 5's clobber race, since the fetch now reads the cache *after* it resolves and merges into it. The alternative I rejected was `useInfiniteQuery`: its pages are independently-cached slices, and the SSE stream writes into the same list, so every page cache would have to be reconciled with the stream separately - exactly the two-lists-one-truth problem `useInboxLive` was written to avoid.

Two deviations from the issue spec, both verified against the code rather than assumed:

- **Item 4 (detail pane rewrites under the reader) is already fixed.** `selection` carries `{inboxId, captureId}` and the `captures[0]` fallback applies only when nothing has been picked - which is what the issue asks for ("fall back to newest only before any explicit selection"). This landed with #554. I added the test that locks it (a fake SSE arrival while a row is selected) rather than a change.
- **Headers use local inputs, not `KeyValueEditor`.** The issue says "headers as the key-value row primitive", but `KeyValueRow` calls `useRequestBuilderContext()` unconditionally (`KeyValueRow.tsx:105`) and that hook *throws* outside `RequestBuilderProvider`, and the row also carries variable resolution, `ResolvedPeek` and file parts that a canned response has no use for. Reusing it would mean mounting a fake provider or refactoring the primitive - the second is worth doing but is not this issue's diff. Filed as a follow-up (see below).

Two smaller notes:

- **The chip-token polish item does not reproduce.** `bg-status-warning-fill text-primary-foreground` is what `StatusCodeBadge` - the canonical status chip - uses, and `--primary-foreground` is `0 0% 100%` in every scheme, so there is no visual divergence. The real inconsistency is `text-white` in `modules/history`, which is out of scope here. The *wording* half of that item is fixed: both surfaces now render one `NonLoopbackBadge`.
- **#555's item 7** (the tab's missing `onError` on stop/clear) is left to #555 as the issues agreed, since #555 was claimed first. The one new mutation here (`loadMore`) carries its own `onError` toast.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check` - all green.

- Full app suite: **426 files / 4333 tests passed** (180s). Inbox + services + queries subset: 18 files / 136 tests.
- `python3 -m mkdocs build --strict -f .github/mkdocs.yml` - built clean (docs change).
- Engine untouched (verified: all four canned fields and the pagination already exist server-side), so no engine build or ctest run - nothing there could change colour.

New tests, mutation-checked by actually reverting each fix and confirming the failure:

| Fix reverted | Test that went red |
|---|---|
| fetch returns the page instead of merging | `keeps a capture streamed in while the first fetch was in flight` |
| `hasMore: page.hasMore` without `&& data.length < total` | `stops offering more once the tail is loaded, whatever a first page says` |
| clear no longer empties the cache before invalidating | `does not resurrect cleared captures on the refetch that follows a clear` |

Plus: load-more appends with `offset === accumulated length` and the count line tracks it; no control offered when the list already holds everything; the tab hands `stopped` to the panel (the wiring, which is the half that goes missing); every canned field round-trips through one apply; a removed header is sent as removed rather than omitted (a merge-patch omission is how a deleted header comes back); status/delay/header validation names the field instead of letting a bare 400 arrive; a stopped panel takes no input, says why, and still reads out what the record holds; the row truncation marker renders.

## Checklist

## Related Issues
Closes #556

---
_Generated by [Claude Code](https://claude.ai/code/session_01TVGRRs9Yoq2zuV4VLXhpvL)_


athrvk:
fix the merge conflict

athrvk:
Conflict resolved and pushed (`584dfdd`). #555 landed in #562 while this branch was open, and it collided in two places:

- **`ServicesPanel.tsx`** - master now leads the inbox row with "Port N" and demotes the URL; this branch replaced the inline non-loopback chip with the shared `NonLoopbackBadge`. Kept both; they are independent edits to adjacent lines.
- **`index.tsx` + the inbox docs** - #555 took item 7, the shared brush both issues named, so every mutation the tab owns now reports through one `reportFailure` helper. The load-more mutation added here had a hand-rolled copy of those same three lines, so it goes through the helper too rather than being the one error path with its own spelling. The docs paragraph is master's, with this PR's `CannedResponseControls` entry beside it.

Re-verified on the merge commit: app suite **427 files / 4354 tests** passed, plus `type-check`, `lint`, `format:check`, and `mkdocs build --strict`.

One thing worth your eye: `PR CI - run tests` never dispatched for the first push - only `Labeler` ran, and no workflow run exists for `b1c64b5`. Per the note at the top of `pr-tests.yml`, a required check that never starts blocks the merge outright. Watching whether this push starts it.

---

Issue #553, opened 2026-08-13
engine/app(inbox): a stopped inbox is a permanent zombie - no DELETE /inbox/:id, no restart, and the drawer's one-click "+" mints more of them


athrvk:
Owner report, 2026-08-13 (verified at master `c76f874`): the services drawer creates an inbox per "+" click and offers no delete anywhere. Root cause is inherited, not #502's: the engine has **no `DELETE /inbox/:id`** - the full route set is start, `:id/stop`, list, `PUT :id`, `:id/requests` GET+DELETE, `:id/live` (`engine/src/http/routes/inbox.cpp:743-886`; the only DELETE is captures-only at `:859-873`). `InboxManager::stop` is terminal by design ("The record and its captures stay readable until the engine exits, so a stop is not a delete", `inbox.cpp:771-774` - PR #499's recorded rationale: issue #480's route list had no delete, and dropping the record on stop would strand captures). No restart route exists either. Net: every stopped inbox is an unremovable, unrestartable row until the engine process exits - and the drawer made minting them one click.

Adjacent lie, same fix: the drawer's inbox-group affordance is a Play icon labelled "Start inbox", which beside a stopped row reads as "restart it" - but it always creates a NEW inbox.

## Fix / solution pathway

1. **Engine: `DELETE /inbox/:id`.** Registered beside its siblings; running inbox: stop first then delete (one call - the user's intent is "make it gone"), or 409-require-stopped if the teardown ordering makes stop-inside-delete racy (decide, state in the PR - the `ManagedListener` lifecycle from #505 makes stop-then-delete clean). Cascades `inbox_requests` rows in one transaction (the `add_inbox_request` trim shows the shape). 404 for unknown id. This resolves #499's stranded-captures objection properly: the captures die WITH the record, by explicit user intent, instead of the record living forever to protect them.
2. **Drawer row**: a delete affordance (trash icon) on every row; confirm via the destructive-dialog pattern when the inbox has captures, naming the count ("Delete this inbox and its 37 captures?") - instant when empty. Invalidate the list; a deleted selected inbox deselects.
3. **Inbox tab**: the same delete beside Stop/Clear, same confirm; after delete the tab falls back to the remaining-inbox behaviour (coordinate with the multi-inbox addressing issue filed alongside - land that first).
[...]

PR #569, opened 2026-08-14
feat(engine/app/mock): serve a collection's saved examples on a local listener

### LINKED ISSUE #481: engine/app(mock): local mock server from a collection's examples - which first requires an examples model, because imports currently drop every saved response on the floor

athrvk:
## Description

**Plan.** Phase 1 gave examples somewhere to live; nothing read them back, which is the root cause this closes - an imported spec's documented responses sat in `request_examples` and no code path could turn them into a running service. The approach is a fourth on-demand listener whose **route table is a start-time snapshot** of the collection subtree: built once from `build_mock_routes`, const thereafter, so the listener thread reads immutable state and the manager's lock is never taken from a handler. The alternative I rejected was a mock that watches the collection and hot-reloads - it makes every read of the table need a lock on the request-serving hot path, and it makes "which version of my collection is this answering?" unanswerable mid-run, which is exactly the question a load run against the mock has to be able to answer. Restart-to-reload is documented instead.

The second decision worth stating: **path matching is per whole segment, not a regex over the path.** A template is only ever a whole segment in every format Vayu imports, and a substring matcher would have `/petsfoo` answering a route for `/pets`.

### What lands

**Engine** - `mock_server.hpp` / `routes/mock_server.cpp`, `MockServerManager` beside the other three:

- `POST /mock/start` (`collectionId` required; `port`, `latencyMs`, `errorRatePct`), `POST /mock/:id/stop`, `GET /mock`, `GET /mock/:id/routes`.
- The table walks the collection **and every collection under it** - an OpenAPI import files its requests in a folder per tag, so a table built from direct children would serve nothing at all.
- All three template spellings collapse to one wildcard segment: `{{petId}}` (Vayu/Postman), `{petId}` (OpenAPI), `:petId` (Postman's other form). The `{x}` half mirrors `normalizeVars(path, {pathTemplates: true})` and is pinned to it by `engine/tests/fixtures/path-template-conformance.json`, read by both suites - the app writes these URLs and the mock reads them back.
- **Specificity decides, not registration order**: `/pets/mine` beats `/pets/{{petId}}`. Without it, which route answers depends on which request an import happened to write first, and a literal route can be permanently shadowed.
- Loopback only, no `bind` field. Built on the `ManagedListener` #505 extracted - the first listener built on the helper rather than ported to it, which was the point of extracting it when the third one landed.

**App** - `CollectionDetail/MockServerControl` (the only surface holding a collection, so the only one that can start a mock) and a **Mock servers** group in the Services drawer that lists, expands to the route table, and stops. Plus the phase-1 leftover: `ImportMeta.exampleCount` was computed by all three parsers, declared by the type, documented as "shown in the import preview", and rendered nowhere - now the fifth count on the preview's summary line.

### Three deliberate deviations from the issue spec, each with its reason

1. **A matched route with no saved example answers `501 mock_no_example`, not a 404.** The issue names two near-misses (method mismatch, path miss); this is a third the code makes reachable, and it is the one an import that dropped its examples produces. `404` would send the user hunting for a path they can see in `GET /mock/:id/routes`; the 501 names the request instead.
2. **A stop drops the record**, matching a mock issuer rather than an inbox. An inbox's record survives a stop because its captures do; a mock holds nothing that outlives its listener, and its table is a start-time snapshot, so a stopped record would list something that answers nothing and never can again. This also sidesteps the #553 class of problem (a stopped row nothing can remove) rather than inheriting it.
3. **The Services drawer group has no create affordance.** The issue asked for "a per-collection chip as the contextual starter" plus a drawer group; a start button in the drawer would need a collection the drawer does not have. The collection header starts, the drawer lists - stated in both files and in COMPONENTS.md.

Not attempted, and filed rather than left implicit: no app surface sets `latencyMs` / `errorRatePct` yet (the engine accepts and the drawer displays both). Follow-up issue #570.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`; `cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check` - all green.

- Engine suite: **1488 / 1488 passed** (253s), 20 of them new.
- App suite: **431 files / 4392 tests passed** (226s).
- `python3 -m mkdocs build --strict -f .github/mkdocs.yml` - built clean (docs change).
- No pre-existing failures on this base (`c626455`).

New engine coverage in `mock_server_routes_test.cpp`: payload validation and every bound; the shared conformance fixture; subtree walk; first-example-in-stored-order; a request with no example staying in the table as unservable; disabled example headers dropped and repeated `Set-Cookie` surviving; content-type fallback; specificity; both near-miss bodies; trailing-slash equivalence; a live listener serving real examples over a socket; latency as a bound assert; error rates **0 and 100 only** (a probability is not an assertion); stop; teardown while a latent response is in flight; the server budget; and **a real load run against a started mock**, asserting the run's status histogram carries 200s.

Mutation-checked by actually reverting each change and confirming the failure:

| Change reverted | Test that went red |
|---|---|
| Specificity in `resolve_mock_route` (first match instead of fewest wildcards) | `ALiteralRouteBeatsATemplatedOneWhateverOrderTheyWereWritten` |
| The subtree walk in `build_mock_routes` (direct children only) | `TheTableCoversTheWholeCollectionSubtree` |
| The `:param` rule in `normalize_mock_path` | `MatchesTheSharedConformanceFixture` |
| `{meta.exampleCount} examples` in `PreviewView` | both cases in `ImportModal.example-count.test.tsx` |

The first of those is the case worth noting: the specificity test passed under mutation on the first attempt, because both fixture requests had `order` 0 and the id tiebreak happened to put the literal first. The test now pins the table order it depends on (`ASSERT_EQ(routes[0].path_template, "/pets/{{petId}}")`) before asserting the match, so it fails for the right reason.

Both conformance-fixture readers assert the fixture is non-empty before iterating it - vacuously-passing table tests are a defect this repo has shipped before.

## Checklist

## Related Issues

Closes #481.

Docs updated same-commit: `docs/engine/api-reference.md` (the Mock Server section, including the load-run-against-a-mock workflow the issue asked to state as supported), `docs/engine/architecture.md` (listener inventory is four now, plus why this is the one listener with no wide-bind gate at all), `docs/app/COMPONENTS.md` and `docs/app/api-integration.md`. `db-schema.md` is untouched - phase 2 adds no table.

### DIFF
--- a/app/src/config/api-endpoints.ts
+++ b/app/src/config/api-endpoints.ts
@@ -160,6 +160,15 @@ export const API_ENDPOINTS = {
 	MOCK_ISSUER_BY_ID: (issuerId: string) => `/mock-issuer/${issuerId}`,
 	MOCK_ISSUER_STOP: (issuerId: string) => `/mock-issuer/${issuerId}/stop`,
 
+	// Collection mock server (issue #481 phase 2). Engine-process state like the
+	// two above, so the same verb-path shape. There is no DELETE and no PUT: a
+	// mock's route table is a start-time snapshot of the collection, so changing
+	// what it serves means starting another one, and stopping is what ends it.
+	MOCK_SERVER: `/mock`,
+	MOCK_SERVER_START: `/mock/start`,
+	MOCK_SERVER_STOP: (mockId: string) => `/mock/${mockId}/stop`,
+	MOCK_SERVER_ROUTES: (mockId: string) => `/mock/${mockId}/routes`,
+
 	// OAuth 2.0
 	OAUTH2_TOKEN: `/oauth2/token`,
 	OAUTH2_AUTHORIZE_START: `/oauth2/authorize/start`,
--- a/app/src/modules/collections/CollectionDetail/CollectionDetail.error.test.tsx
+++ b/app/src/modules/collections/CollectionDetail/CollectionDetail.error.test.tsx
@@ -58,6 +58,9 @@ vi.mock("./AuthTab", () => ({ default: () => null }));
 vi.mock("./InfoTab", () => ({ default: () => null }));
 vi.mock("./ScriptTab", () => ({ default: () => null }));
 vi.mock("./VariablesTab", () => ({ default: () => null }));
+// The header's mock-server control reaches the engine and the toast store;
+// this file is about the tab shell, and its own suite covers the control.
+vi.mock("./MockServerControl", () => ({ default: () => null }));
 
 beforeEach(() => {
 	refetch.mockReset();
--- a/app/src/modules/collections/CollectionDetail/CollectionDetail.loading.test.tsx
+++ b/app/src/modules/collections/CollectionDetail/CollectionDetail.loading.test.tsx
@@ -47,6 +47,9 @@ vi.mock("./AuthTab", () => ({ default: () => null }));
 vi.mock("./InfoTab", () => ({ default: () => null }));
 vi.mock("./ScriptTab", () => ({ default: () => null }));
 vi.mock("./VariablesTab", () => ({ default: () => null }));
+// The header's mock-server control reaches the engine and the toast store;
+// this file is about the tab shell, and its own suite covers the control.
+vi.mock("./MockServerControl", () => ({ default: () => null }));
 
 beforeEach(() => {
 	state.collections = { data: [], isLoading: false };
--- a/app/src/modules/collections/CollectionDetail/MockServerControl.test.tsx
+++ b/app/src/modules/collections/CollectionDetail/MockServerControl.test.tsx
@@ -1,0 +1,168 @@
+/**
+ * @vitest-environment jsdom
+ */
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * The collection header's mock-server control (issue #481 phase 2).
+ *
+ * This is the only surface that can *start* a mock, because it is the only one
+ * that has a collection. What it has to get right is which mock it is talking
+ * about: the engine's list is every mock in the process, and a header showing a
+ * different collection's URL - or offering "Run mock server" beside one that is
+ * already running - is worse than showing nothing.
+ *
+ * The transport is mocked and the real query hooks run, so starting and
+ * stopping go through the same mutations and cache invalidation the app uses.
+ */
+
+import { describe, it, expect, beforeEach, vi } from "vitest";
+import { render, screen, cleanup, fireEvent, waitFor } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { TooltipProvider } from "@/components/ui";
+import { useToastStore } from "@/stores";
+import type { MockServer } from "@/types";
+import MockServerControl from "./MockServerControl";
+import { mockForCollection } from "./mock-server-selection";
+
+const listMockServers = vi.fn();
+const startMockServer = vi.fn();
+const stopMockServer = vi.fn();
+
+vi.mock("@/services/api", async (importOriginal) => {
+	const actual = await importOriginal<typeof import("@/services/api")>();
+	return {
+		...actual,
+		apiService: {
+			...actual.apiService,
+			listMockServers: () => listMockServers(),
+			startMockServer: (...a: unknown[]) => startMockServer(...a),
+			stopMockServer: (...a: unknown[]) => stopMockServer(...a),
+		},
+	};
+});
+
+const writeText = vi.fn();
+
+function mock(overrides: Partial<MockServer> = {}): MockServer {
+	return {
+		mockId: "mock_a",
+		collectionId: "col_1",
+		collectionName: "Pet Store",
+		url: "http://127.0.0.1:43100",
+		port: 43100,
+		latencyMs: 0,
+		errorRatePct: 0,
+		routeCount: 3,
+		routesWithoutExample: 0,
+		createdAt: 1700000000000,
+		...overrides,
+	};
+}
+
+function renderControl(collectionId = "col_1") {
+	const client = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+	return render(
+		<QueryClientProvider client={client}>
+			<TooltipProvider>
+				<MockServerControl collectionId={collectionId} />
+			</TooltipProvider>
+		</QueryClientProvider>
+	);
+}
+
+beforeEach(() => {
+	cleanup();
+	listMockServers.mockReset().mockResolvedValue([]);
+	startMockServer.mockReset().mockResolvedValue(mock());
+	stopMockServer.mockReset().mockResolvedValue({ mockId: "mock_a", stopped: true });
+	writeText.mockReset().mockResolvedValue(undefined);
+	vi.stubGlobal("navigator", { ...navigator, clipboard: { writeText } });
+	useToastStore.setState({ toasts: [] });
+});
+
+describe("picking the mock this header is about", () => {
+	it("ignores mocks of other collections", () => {
+		const mine = mock({ mockId: "mock_mine", collectionId: "col_1" });
+		const theirs = mock({ mockId: "mock_theirs", collectionId: "col_2", port: 43000 });
+		expect(mockForCollection([theirs, mine], "col_1")?.mockId).toBe("mock_mine");
+		expect(mockForCollection([theirs], "col_1")).toBeNull();
+		expect(mockForCollection([], "col_1")).toBeNull();
+	});
+
+	it("picks the lowest port when one collection has several", () => {
+		// Nothing stops a user starting two mocks of one collection - different
+		// latency, different error rate - and the engine lists them in map
+		// order, which is not stable across polls.
+		const high = mock({ mockId: "mock_high", port: 43900 });
+		const low = mock({ mockId: "mock_low", port: 43100 });
+		expect(mockForCollection([high, low], "col_1")?.mockId).toBe("mock_low");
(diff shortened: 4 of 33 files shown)


athrvk:
**CI: `Engine on windows-latest` failed on a runner network error, not on this diff.** Recording it rather than pushing anything, because there is nothing here to fix.

The job died inside `vcpkg install`, before a single translation unit was compiled:

```
error: curl operation failed with error code 35 (SSL connect error).
error: Not a transient network error, won't retry download from
       https://www.python.org/ftp/python/3.14.2/python-3.14.2-embed-amd64.zip
error: building vcpkg-tool-meson:x64-windows failed with: BUILD_FAILED
```

Everything after that is downstream of it - the CMake configure then reports no Ninja and no compiler, and the reporter's `No test results found for C++ Tests (windows-latest)` is the symptom, not the cause. Three things say it is the runner and not the change:

- `engine/vcpkg.json` is untouched by this PR - no dependency moved, so nothing new was being fetched.
- sccache recorded **0 compile requests** on the job. No source of mine reached a compiler.
- The job failed at ~3 minutes while `Engine on ubuntu-latest` and `Engine on macos-latest` were still compiling, having cleared the same `vcpkg install` step.

vcpkg's "Not a transient network error" is its classification of `curl` exit 35, not evidence about python.org - a TLS handshake failure to a CDN is exactly the transient case.

I will re-run the failed job once the rest of the run finishes (GitHub refuses `rerun-failed-jobs` while other jobs in the run are still in flight) and report back. Locally on this branch the engine suite is **1488/1488** and the app suite **4392** tests, both green.

---

athrvk:
**Re-run done: CI is fully green on the same commit (`281acd5`), no code changed.**

`Engine on windows-latest` re-ran as attempt 2 and passed - `Setup vcpkg` cleared in 26s where attempt 1 died in it, which confirms the earlier diagnosis (a `curl` exit 35 TLS handshake failure fetching the embeddable-python zip, not this diff).

| Job | Attempt 1 | Attempt 2 |
|---|---|---|
| Engine on windows-latest | failure (in `vcpkg install`, 0 compile requests) | **success** - build 22m, tests 4m45s |
| C++ Tests (windows-latest) | no results reported | **success** |
| CI gate | - | **success** |

Everything else was already green on attempt 1 and is unchanged: engine on ubuntu/macOS, app on all three, CodeQL, conventional title. `mergeable_state` is `blocked` on your review only.

---

athrvk:
**Green.** Re-ran the failed job once the rest of the run settled; `Engine on windows-latest` passed on the second attempt with no change to the branch, which is the confirmation that the failure was the runner's TLS handshake to python.org and not this diff.

| Check | Result |
|---|---|
| Engine on ubuntu-latest / `C++ Tests (ubuntu-latest)` | pass |
| Engine on macos-latest / `C++ Tests (macos-latest)` | pass |
| Engine on windows-latest / `C++ Tests (windows-latest)` | pass (re-run) |
| App on ubuntu / macos / windows | pass |
| CI gate | pass |

The first attempt died at 3 minutes; the re-run took 28, having cleared the same `vcpkg install` step normally. Ready for review.

---

Issue #618, opened 2026-08-15
engine(data): a {{data.*}} value bound into an XML body is inserted verbatim - `&amp;` or `&lt;` produces malformed XML


athrvk:
## Problem

Split out of #580, which added the `xml` body mode. `scenario_data.cpp`'s
`body_context` classifies a request body as either `Plain` or `JsonDocument`, and
that choice decides how a `{{data.*}}` value is encoded when it is bound into the
body text. `Json` and `JsonRpc` are `JsonDocument` (a value landing inside a
string literal is escaped); everything else, `Xml` included, falls through the
`default:` to `Plain` - the value is inserted byte for byte.

So a data file whose cell holds `Ben &amp; Jerry's` or `a &lt; b`, bound into

```xml
&lt;order&gt;&lt;customer&gt;{{data.name}}&lt;/customer&gt;&lt;/order&gt;
```

sends XML the server will reject as malformed - or, for a cell holding
`&lt;/customer&gt;&lt;injected/&gt;&lt;customer&gt;`, XML with a different shape than the author wrote.
Nothing warns: the bind succeeds, the row is sent, and the failure surfaces as
the origin's 400.

This was anticipated before the mode existed - the pre-#580 comment read "The XML
body mode #580 would add is the one to revisit: its quoting rules are not JSON's,
so it needs its own encoding rather than this one." #580 added the mode and left
the encoding at `Plain`, recording the gap in that comment and here rather than
guessing at the rule under a feature PR.

## Why it is not a one-liner

XML has no single escape set the way a JSON string literal does. The rule depends
on where the token sits:

- **Text content**: `&amp;`, `&lt;` must be escaped (`&gt;` only inside `]]&gt;`, but escaping it
  is harmless and conventional).
- **An attribute value**: the above plus the quote character that delimits the
  attribute - and which one that is depends on how the author wrote it
  (`name="{{data.n}}"` vs `name='{{data.n}}'`).
- **Inside a `&lt;[CDATA[…]]&gt;` section**: nothing is escaped, and escaping would
  corrupt the value; only `]]&gt;` matters.
- **Inside a comment or a processing instruction**: a bind there is almost
  certainly a mistake worth refusing rather than encoding.

`advance_json_string_state` answers the JSON question by scanning the literal
chunks for an unescaped quote. The XML analogue has to track at least
text/attribute/CDATA, and know the delimiter of the attribute it is inside.

## Fix

[...]

Issue #581, opened 2026-08-14
app/engine(body): a JSON-RPC 2.0 body mode with engine-side call enveloping - the second cheap win, on the GraphQL template


athrvk:
## Problem

JSON-RPC 2.0 is the wire protocol of every MCP server, most Ethereum/blockchain nodes, LSP-adjacent tooling, and a long tail of internal RPC APIs. Vayu itself speaks it (the MCP HTTP transport builds JSON-RPC frames in `app/electron/mcp/http.ts:60,95,125` and `app/electron/mcp/cli.ts:15`), yet a user testing a JSON-RPC endpoint today gets no help: they pick the `json` body mode and hand-write the `{"jsonrpc":"2.0","method":...,"params":...,"id":...}` envelope on every request, and nothing catches a forgotten `"jsonrpc"` member or a missing `id`.

This is the exact shape GraphQL had before its dedicated mode, and the GraphQL implementation is a proven template for fixing it: a stored content mode whose envelope is completed engine-side at the one chokepoint all execution paths share. The roadmap discussion named JSON-RPC one of the two "cheap wins that aren't protocols"; the sibling is the XML mode in #580.

Verified blank slate: the only JSON-RPC strings in the repo are the MCP transport internals above plus `docs/engine/mcp.md:94`. There is no request-side concept to collide with.

## What ships

A `"jsonrpc"` body mode, stored as `{mode: "jsonrpc", content: string}`, following the GraphQL chain link for link:

**App chain** (each anchor is the GraphQL precedent to extend):

- `app/src/types/domain.ts:21` - add `"jsonrpc"` to the `BodyMode` union; `:87-91` the `RequestBody` content-mode arm (one-token addition at `:89`); `:242-243` the `bodyType` mirror.
- `BodyPanel.tsx:64-75` - a `BODY_MODES` entry `{value: "jsonrpc", label: "JSON-RPC", contentType: "application/json"}` (the `contentType` chip at `:336-341` is display-only). Editor stays the plain code editor: `isCodeMode` at `:306` gains `jsonrpc`, and the Monaco language at `:396` is `"json"` for this mode. No structured sub-component - unlike GraphQL's query/variables split, a JSON-RPC call is one JSON text.
[...]

PR #619, opened 2026-08-15
feat(app/engine/body): an XML body mode, an XML pretty-printer, and implied Content-Types in codegen

### LINKED ISSUE #580: app/engine(body): an XML body mode and an XML pretty-printer - the SOAP/legacy-enterprise win, plus snippets finally carrying implied Content-Types
From the 2026-08-14 protocol-roadmap exploration (verified at master `9f222b8`; anchors re-checked at `0513a46` after the JSON-RPC sibling #581 merged via PR #585 - ordinals below updated accordingly). XML today is response-side only: `detectBodyType` recognises it (`response-viewer/utils.ts:52-58`), Monaco highlights it (`utils.ts:114`; the `xml` basic-language ships in the bundle) - but `formatBody` pretty-prints **JSON only** (`utils.ts:129-152`), so Pretty and Raw are byte-identical for XML, and there is no XML request mode at all. SOAP and legacy-enterprise APIs are just HTTP+XML; this issue captures most of that ask with no WSDL machinery.

## Fix / solution pathway - the add-a-BodyMode chain, every link anchored

**App request side** (the chain is small and fully mapped):
1. `BodyMode` union + `RequestBody` gain `"xml"` (`domain.ts:21`, one token at `:89` - it is a content-carrying mode like text).
2. `BODY_MODES` entry `{ value: "xml", label: "XML", contentType: "application/xml" }` (`BodyPanel.tsx:64-75`); `isCodeMode` gains xml (`:306`) with editor `language="xml"` (`:396` - highlighting is free, the language ships).
3. `REQUIRED_CONTENT_TYPE` gains `xml: "application/xml"` (`content-type.ts` - the third entry, after graphql and the jsonrpc entry #585 added at `:44`; the graphql machinery `switchContentType`/`contentTypeToAdd` handles add/remove/user-wins already).
4. `draftKey`: xml shares the `"raw"` bucket (`body-drafts.ts` - jsonrpc set the precedent for a content mode joining raw; it is one plain text pane, unlike graphql's structured envelope); state this in the PR per the file's own header rationale.
5. The mode chains: `initialRequest` ternaries (`index.tsx:139-150`), the `handleSave` cast (widen the literal union - #585 already widened it with `"jsonrpc"` in BOTH places, `request-builder/index.tsx` and `save-run-to-request.ts:206`; widen both again), the Source/Resolved preview gate (`BodyPanel.tsx:346` - xml joins `isCodeMode` so `{{variable}}` preview works; wire interpolation is already free and mode-agnostic per `request_composer.cpp:624-627`).

**Engine**:
6. `BodyMode` enum gains `Xml` (`types.hpp:124`); spelling table `"xml"` (`json.cpp:617-634` - "the one place a spelling is added"); serialize branch emits mode + raw content (the graphql shape, no re-parse).
7. `implied_content_type` gains `case BodyMode::Xml: return "application/xml";` (`form_body.cpp` - the fourth entry, after Form, GraphQL and the JsonRpc case #585 added at `:191`). `wire_body_bytes` and `script_body_view` need nothing - both default to `body.content`. No envelope, no new file.

**Response side**:
8. `formatBody` (`utils.ts:129-152`) gains an XML branch for `bodyType === "xml"`: a small pure-string indenter (tag-depth walk - no dependency; `DOMParser` is renderer-available but node-env tests would lack it, and prettier core has no XML parser, so pure-string is the test-friendly choice). Malformed input returns the original unchanged - the graphql `format.ts` null-on-failure precedent. Applies to responses AND (via the same helper) an optional Format action for the xml body editor if trivial; otherwise response-only, stated.

**Consumers** (each verified as a real touchpoint):
9. Importers: Postman raw-with-xml-language and Insomnia `application/xml`/`text/xml` mimeTypes map to the xml mode (`postman.ts` body mapping, `insomnia-v4.ts:184-196` pattern); imported explicit Content-Type still wins (`shared.ts:231`).
10. **Codegen finally carries implied Content-Types** - the rider that also fixes graphql and jsonrpc: `prepare.ts:109-142` falls everything through as raw content with NO header, so today a copied curl of a graphql or jsonrpc request sends the wrong Content-Type (#585 landed jsonrpc and explicitly deferred this rider here). Emit the mode's implied Content-Type header when the user declared none, for every mode `implied_content_type` covers (graphql and jsonrpc now, xml here). Golden-test per target.
11. MCP `bodyType` docstrings x4 (`tools.ts` - the four run/save docstring sites #585 just touched; follow its diff) + `bodyPayload` needs nothing (content fallthrough). `api-reference.md`: the body-union table, the Content-Type rules, `docs/engine/mcp.md` bodyType mentions.

## Tests

Engine: an `xml_body_test.cpp` on the `graphql_body_test.cpp` skeleton minus the envelope group - storage round-trip, wire verbatim, derived `application/xml`, explicit Content-Type kept, `OtherModesAreUntouched`. App: mode chain round-trip (save/load/initialRequest), auto-header add/remove/user-wins (the content-type.ts test pattern), the pretty-printer unit cases (nesting, attributes, self-closing, malformed-returns-original - mutation-checked), codegen goldens incl. the graphql+jsonrpc Content-Type fix. Commands: both suites full, per the repo's real commands.

## Docs (same commit)

`api-reference.md` sections above; `docs/app/COMPONENTS.md` body-panel mode list; import docs if they enumerate modes (`docs/app/import-collections/README.md` mentions bodyType - verify).

## Sequencing

The JSON-RPC sibling #581 has MERGED (PR #585), so the shared-table conflict this issue originally named is resolved - rebase onto current master and the diffs are additive. The codegen Content-Type rider (item 10) lands HERE, as #585 recorded. Parallel-safe against the SSE epic (#574-#576 remaining - disjoint files) and everything else open.

## Acceptance criteria

- [ ] An XML request authors with highlighting, auto-`Content-Type: application/xml` (user-declared wins), `{{variables}}` resolved, wire-verbatim; importers map it; MCP accepts `bodyType: "xml"`.
- [ ] An XML response pretty-prints in the Pretty toggle, malformed XML rendering unchanged; codegen snippets carry implied Content-Types for graphql, jsonrpc and xml (golden-tested).
- [ ] Both suites green, mutation-checked; docs same-commit.

athrvk:
## Description

**Plan (root cause, approach, rejected alternative).** Anchors verified against the issue, and the sequencing has come true - #581 merged via #585, so the shared tables took this diff additively. The root cause is that XML existed on only one side of the app: `detectBodyType` recognised it (`response-viewer/utils.ts:52-58`) and Monaco highlighted it, but `formatBody` pretty-printed **JSON only**, so Pretty and Raw were byte-identical for every XML response, and there was no XML request mode at all. The approach is the add-a-BodyMode chain the issue maps, with no envelope anywhere: the engine's `wire_body_bytes` gets **no** `Xml` arm, because sending the author's bytes unchanged is what the mode means, and the whole of what it adds over `text` is the derived `application/xml`. The alternative I rejected is the one the issue floats for the pretty-printer - `DOMParser`, which is available in the renderer and would be the obvious tool. It is absent from the `node` environment this repo's tests default to, so the one rule worth testing would have been the one thing untestable; prettier core ships no XML parser either. A tag-depth string walk needs neither and is exact about the only question being asked of it.

**What the printer refuses to do** is as load-bearing as what it does. Input it cannot walk - an unterminated tag, a close that names something else, an element left open - comes back byte for byte, the contract `lib/graphql/format.ts` already has, because a body mid-edit or one holding an unresolved `{{token}}` must render as what the user typed. **Mixed content is emitted verbatim** too: in `<p>hello <b>x</b> world</p>` the spaces around `<b>` are part of the text, and putting each child on its own line would be editing the document rather than formatting it. The tag scanner is quote-aware because an unescaped `>` inside an attribute value is well-formed XML, and a naive scan for the next `>` would end the tag inside it and take the rest of the document with it.

**The rider #585 deferred here.** `prepare.ts` fell every content mode through as raw content with **no** header, so a copied curl of a GraphQL or JSON-RPC request went out under libcurl's `application/x-www-form-urlencoded` default and most servers answer that with a 400 - the snippet not doing what the app had just done. Snippets now carry the mode's implied Content-Type when the request declares none, mirroring the engine's `implied_content_type` + `body_content_type_header` pair, declared-wins rule included.

## Type of Change
- [x] Bug fix (XML responses never pretty-printed; snippets sending the wrong Content-Type for graphql/jsonrpc)
- [x] New feature (the `xml` body mode)
- [ ] Breaking change
- [x] Documentation update

## Testing

All green locally, re-run on the merge commit after #613-#617 landed (base `74e6a55`).

- **Engine** - `python build.py -e -t && ctest --preset linux-dev`: **1641 tests, 100% passed, 0 failed.**
- **App** - `pnpm test`: **456 files, 4726 tests, all passing.** `pnpm type-check`, `pnpm lint` and `pnpm format:check` clean on the files this touches.
- **Docs** - `mkdocs build --strict -f .github/mkdocs.yml` clean.
- `clang-format --dry-run --Werror` clean on `engine/tests/xml_body_test.cpp` and on `scenario_data.cpp`, the one touched engine file that was clean before this. `json.cpp`, `form_body.cpp` and `types.hpp` each carry pre-existing violations in regions this does not touch, so per CLAUDE.md they are not reflowed.

An earlier pre-merge run showed `HttpClientTest.UnreachableHostsRawRequestNamesWhatWasAskedFor` timing out. It was CPU contention from running the app suite alongside ctest - it passes in the quiet run above, and it touches no body mode. Recorded rather than quietly dropped.

**Mutation-checked** (on the pre-merge tree; the merge changed no source file this covers), each reverted, rebuilt, confirmed red, then restored:

| Mutation | Result |
|---|---|
| drop the `Xml` arm of `implied_content_type` | 3 engine tests fail (wire, derived-header, load driver) |
| give `wire_body_bytes` an `Xml` arm that appends a newline | 4 engine tests fail - verbatim is asserted on the string, so *any* tidying reddens it |
| remove the implied Content-Type from `prepare.ts` | 3 codegen cases fail, one per mode |
| return the source unchanged from `formatXml` | the indenter cases fail; the refuses-to-touch cases stay green, which is the point of having both halves |
| count descendants rather than direct children in `isMixed` | 10 cases fail - this was a real bug the tests caught before review did |

The last row is worth stating plainly: the first draft of `isMixed` walked the whole subtree, so every ancestor of a leaf value was "mixed" and the printer emitted entire documents on one line. The nesting cases caught it.

New engine coverage (`xml_body_test.cpp`, 9 cases): storage round-trips the document and an unresolved `{{token}}` verbatim, empty content is not a wire body, the document reaches the wire byte for byte as `application/xml` on **both** drivers (`Client::send` and the `EventLoop` every `POST /runs` uses), an explicit `application/soap+xml` is kept, and `OtherModesAreUntouched` pins that adding an enum entry did not give `text` a Content-Type or hand a form body the wrong encoder.

## Checklist

## Deviations from the issue spec, each with its reason

1. **No Format action on the XML body editor.** The issue leaves this optional ("if trivial; otherwise response-only, stated"). Stated: response-only. The body editor offers no Format action for JSON either, so adding one for XML alone would be a new affordance on one mode rather than a feature of the panel - its own issue if wanted.
2. **The importer maps Postman's `xml` language only, and never sniffs.** An unlabelled `<`-shaped body stays `text`: without the language, it is as likely to be HTML, and guessing would hand the request a Content-Type the server may disagree with. Insomnia states its mime, so both `application/xml` and `text/xml` map.
3. **Two doc lines beyond the issue's list are corrected**, both in sentences this change makes wrong: `db-schema.md`'s stored-body union had never gained `jsonrpc` either (a #585 miss, fixed in the same line rather than left half-stale), and `scripting.md`'s list of engine-derived Content-Types named only `graphql`.
4. **An existing Insomnia test changed rather than being added to.** `keeps an unlisted text body (XML)` asserted the behaviour this PR deliberately changes. Its intent - an unlisted body keeps its text - is preserved by repointing it at `text/yaml`, which is still genuinely unlisted, and its variable-normalization assertion moved into the new XML block so nothing it covered is now uncovered.

## Known gap, recorded not guessed at: `{{data.*}}` into an XML body

`scenario_data.cpp`'s `body_context` classifies a body as `Plain` or `JsonDocument`, and `Xml` falls to `Plain` - so a bound value is inserted **verbatim**, and a cell holding `&` or `<` produces a document the server rejects. The pre-existing comment anticipated exactly this ("The XML body mode #580 would add is the one to revisit"). I did not guess at the encoding under a feature PR: XML has no single escape set, and which one is right depends on whether the token sits in element text, a double- or single-quoted attribute value, or a CDATA section. That is a scanner, not a one-liner. It is now stated in the code comment, in `api-reference.md` where the escaping rules are documented for users, and filed as **#618** with the rule and its test shapes worked out.

## Notes on scope

- No new config knob, no wire-shape change beyond the mode's own spelling, and no renderer state added.
- The codegen rider is header-only by design. A `jsonrpc` snippet still prints the bare call rather than the frame the engine completes, because reproducing that client-side would be a hand-rolled copy of `jsonrpc_wire_body` that stops receiving its fixes - the defect CLAUDE.md names. GraphQL is unaffected either way: the app's own panes write the envelope, so its `content` already is the JSON the header now declares.
- Nothing adjacent was refactored.
- Merged `master` in at `93dddbf` after #613-#617 landed. One textual conflict, in `codegen.test.ts`, where #612's streaming block and this PR's Content-Type block were both appended at the end of the file; both kept, neither edited. Details in the thread.

## Related Issues

Closes #580. Carries the codegen Content-Type rider #585 recorded as landing here. Opens #618 for the XML data-binding encoding.


athrvk:
fix merge conflict

athrvk:
Conflict resolved and pushed (`93dddbf`). #613-#617 all merged while this was in CI, moving the base from `7c2dda0` to `74e6a55`.

**One textual conflict**, in `services/codegen/codegen.test.ts`: #612's streaming block and this PR's Content-Type block were both appended at the end of the file. Both kept, neither edited - they test disjoint rules on the same generators.

Everything else auto-merged, including the two files where a semantic clash was plausible, so I checked both rather than trusting the merge:

- **`codegen/prepare.ts`** - #612 added the `stream` field beside the `headers` mapping this PR changed. They compose: the implied Content-Type is computed from the body mode, `stream` is passed through, and curl's implied `Accept: text/event-stream` is a different header under a different rule. No stream case reaches the new table.
- **`core/scenario_data.cpp`** - #616 added the header-collision refusal to `walk_bindable_fields` while this PR rewrote the `body_context` comment above it. Different functions; the merged file reads correctly and its suite is green.

**Re-verified on the merge commit, not just on my pre-merge tree:**

- Engine: **1641 tests, 100% passed, 0 failed.** This also settles the one failure noted in the PR body - `HttpClientTest.UnreachableHostsRawRequestNamesWhatWasAskedFor` timed out in the earlier run only because I had the app suite running concurrently; it passes here in a quiet run, so it was contention, not this diff.
- App: **456 files, 4726 tests, all passing.** `pnpm type-check`, `pnpm lint` clean; `prettier --check` clean on the resolved file.

The Testing section of the description now carries these post-merge numbers rather than the pre-merge ones.

---

Issue #590, opened 2026-08-14
app(importers): imported query params never reach the wire - the importer splits them out of the URL and nothing ever joins them back


athrvk:
## Problem

Import a Postman collection whose request is `{{baseUrl}}/users?page=1`, then send that request - by design Send, by collection scenario run, or by load run. The wire request is `{{baseUrl}}/users`. **The query is gone in every execution path**, and stays gone until the user happens to edit the Params table once, which silently repairs the URL.

The chain, verified end to end at `0513a46` (wire-proven against a live engine with an inbox capture):

1. `pmUrl` (`app/src/services/importers/postman.ts:185-204`) deliberately **splits**: stored `url` is the base before `?`, and the query becomes `params[]` entries. The importer's own test pins the split (`postman.test.ts:36-40`: `req.url` is `"{{baseUrl}}/users"`, `req.params` carries `page`/`trace`). Insomnia (`insomnia-v4.ts:282`) and the OpenAPI importers produce the same shape.
2. The orchestrator forwards `url` and `params` verbatim (`orchestrator.ts:157-158`); the engine applier stores them as-is; the engine treats `params` as opaque builder state (`engine/include/vayu/types.hpp:829`) and **no engine path ever reads it** - `request_composer.cpp` has zero `params` references.
3. The app's execute path sends `url: request.url` verbatim (`app/src/modules/request-builder/index.tsx:223, :367`; load at `:425`), and the builder loads `url: fetchedRequest.url` verbatim (`:164`). Nothing merges.
4. The only code that joins params into the URL is `buildUrlWithParams` in `ParamsPanel.tsx:29-47`, and it runs **only inside the panel's onChange handler** (`:63`) - i.e. only when the user edits the table.

App-native requests are safe by invariant: the ParamsPanel writes enabled params into the URL on every edit, so a request built in the app carries its query inside `url`. The importers violate that invariant on write, and the violation is invisible - the Params table *shows* the entries, so everything looks right while every request goes out without them.

## Fix pathway

**Restore the app's own invariant at import time** (recommended): after mapping, rebuild the stored `url` from base + *enabled* params using the same logic as `buildUrlWithParams` (extract it from `ParamsPanel.tsx` to a shared util rather than copying - the hand-rolled-primitive rule). Disabled params stay only in `params[]`, exactly the state the ParamsPanel itself produces. One call site in the orchestrator's request finalize covers all importers, including future ones.

[...]

PR #633, opened 2026-08-15
fix(app/importers): import optional value-less OpenAPI query params disabled

### LINKED ISSUE #622: app(importers): OpenAPI value-less query-param stubs now import as bare keys in the URL - decide enabled-with-no-value vs imported-disabled
## Problem

The decision PR #606 (closed #590) explicitly punted, now with a tracker. The OpenAPI importers emit a params row for every declared query parameter, including ones with no default value - a stub row with an empty value. Post-#590, enabled params join into the stored URL, so an imported operation declaring `verbose` (no default) now stores `{{baseUrl}}/pets/{{petId}}?verbose` and **sends the bare key on every request**. Before #590 the stub silently never reached the wire (the general bug); now it reaches the wire as `?verbose`, which some APIs treat as `verbose=true` - a behaviour change for imported collections that nobody chose.

PR #606's body flagged it: "If those stub rows should import *disabled* instead, that is a separate call about the importer's enabled choice" - and documented the current behaviour in both OpenAPI mapping docs. This issue is that separate call.

## Decision - recommendation first

- **(a) Recommended: import value-less stubs disabled.** The stub is documentation ("this endpoint accepts `verbose`"), not intent ("send `verbose` always"). Disabled rows keep the discoverability (the row is visible in the Params table, one click to enable) without changing the wire. This also matches how the table round-trips: a disabled row stays out of the URL by the #590 invariant.
- (b) Keep enabled-with-empty-value and keep the docs line. Defensible for APIs where a declared flag param is conventionally always sent - but that is the rarer reading of an OpenAPI parameter declaration, and (a) costs those users one click with the row already in front of them.

Scope note: **required** query params (`required: true` in the spec) should stay enabled regardless of value - a spec that says required means send it; only optional value-less stubs import disabled. State this split in the implementation.

## Fix pathway

The enabled flag at the OpenAPI mapping sites (`openapi-v3.ts` / `openapi-v2.ts` query-parameter mapping - the sites PR #606's docs edits point at), keyed on `required` and presence of a default/example value. The #590 join logic then does the right thing unchanged.

## Tests

Importer round-trip: optional value-less param imports disabled and stays out of the URL; required param imports enabled; param with a default imports enabled with the value. Mutation-check the required/optional split.

## Docs, same commit

The two OpenAPI mapping docs PR #606 already touched (`docs/app/import-collections/`) - replace the current-behaviour line with the decided rule.

## Sequencing

Parallel-safe against everything open. Touches the two OpenAPI importers, their tests, and the mapping docs.

## Acceptance criteria

- [ ] The decision (a or b) is recorded here before the PR closes this.
- [ ] (a) An imported optional value-less query param is present, disabled, and off the wire; required params still send; defaults still send.
- [ ] Mapping docs state the rule.

athrvk:
## Description

Takes option **(a)** from #622: an OpenAPI/Swagger query parameter that is optional and declares no value imports **disabled**.

**Plan (root cause, approach, rejected alternative).** The root cause is a mismatch of meaning: a spec's `parameters` list declares what an operation *accepts*, while `params[]` since #590 declares what a request *sends* - every enabled row joins the stored `url`. The importers emitted every declared parameter as `enabled: true` with an empty value, so an operation declaring `verbose` stored `{{baseUrl}}/pets/{{petId}}?verbose` and put a bare key on the wire that some APIs read as `verbose=true`. The approach keys `enabled` on the parameter itself - `required: true` or a declared value - and, while reading the declaration anyway, carries a scalar `example`/`default` into the row's value so "defaults still send" is true on the wire rather than nominally (an enabled row with an empty value would still have sent `?limit`, which is not what `default: 25` says). The rule lives once in `queryParamRow` (`openapi-shared.ts`), not once per parser. The rejected alternative is option (b), keeping enabled-with-empty-value: it is defensible only for APIs where a declared flag is conventionally always sent, which is the rarer reading of an OpenAPI parameter declaration, and it leaves imported collections quietly sending keys nobody chose - (a) costs those users one click with the row already in front of them.

The rule, applied by `queryParamRow`:

| Parameter declares | `value` | `enabled` |
|---|---|---|
| a scalar `example` / `examples` / `schema.example` / `schema.default` (v3), `default` (v2) | that value as text | `true` |
| `required: true`, no value | `""` | `true` - joins as a bare key, the cue to fill it in |
| nothing, and not required | `""` | **`false`** |

Value precedence in v3 is `example` → first `examples` entry → `schema.example` → `schema.default`, the same "concrete example beats generated stub" order `buildBody` already uses; a `default` describes what the server assumes when the parameter is *absent*, so it ranks last. The schema is followed one `$ref` hop. Swagger 2.0 reads `default` only - it has no `example` keyword for a non-body parameter.

Two deliberate boundaries, both documented and tested:

- **Non-scalar values are not carried.** An array or object is serialized by `style`/`explode` (v3) or `collectionFormat` (v2), neither of which these parsers read; a row holds one string, so joining on a guessed separator would send what the spec did not declare. Such a parameter imports value-less, and therefore disabled unless required.
- **A declared `""` is treated as no value.** An empty-value row writes as a bare key, so `?q=` is not a shape the Params table can express.

Header parameters are untouched - they are not part of the `url` join and the issue does not name them.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

Behaviour change for *newly imported* collections only; nothing migrates stored requests.

## Testing

- `cd app && pnpm test` - **457 files, 4744 tests, all passing.**
- `cd app && pnpm lint` - clean (0 errors, 0 warnings).
- `pnpm exec prettier --check` on the six touched TS files - clean (each was prettier-clean before; only those were formatted).
- `mkdocs build --strict -f .github/mkdocs.yml` - builds, so the new anchors and the cross-file link resolve.

Engine untouched, so no engine build or `ctest` run per the repo's scale-the-verification rule.

New coverage: a `query parameter enabled state` block in each parser's tests (optional-vs-required split, schema `default`, `example` precedence, `examples` unwrapping, `$ref`'d schema, non-scalar default, declared `""`, description preserved on a disabled row), plus an end-to-end pair in `factory.test.ts` - the fixture's optional stub stays off the URL, and a required + defaulted operation joins as `?tenant&limit=25`.

Mutation-checked both ways, as the repo asks: forcing `enabled: true` fails 15 tests across 3 files; dropping the carried value fails 6.

Note on the run history: attempt 1 of the CI run was cancelled and re-run by me. The jobs were healthy - I misread a lagging status endpoint as a stuck windows runner. Attempt 2 is green across the matrix on the same commit; nothing was changed between attempts.

## Checklist

Docs updated in the same commit: both OpenAPI mapping docs get a `Query parameter values & enabled state` section replacing the current-behaviour line PR #606 left, the Swagger `collectionFormat` section explains why an array `default` is not carried, and the pipeline README's url/params-invariant bullet no longer says the OpenAPI parsers put every declared parameter in the URL.

## Related Issues
Closes #622


athrvk:
Added the **0.17.1** release to this branch (`chore(release): 0.17.1`, 6229b86): version bump across `VERSION`, `engine/CMakeLists.txt`, `engine/include/vayu/version.hpp` and `app/package.json`, plus the curated notes at `.github/release-notes/v0.17.1.md`.

Patch, since the range is fixes only. The notes cover the import fix in this PR and the release-CI work master brought in with #634 (vcpkg cache reuse, winget pruning disabled) under Internal.

`engine/vcpkg.json` is deliberately **not** in the bump: #634 removed its `version` field because it churned the cache key that workflows hash, and `build.py` now refuses to bump if it reappears. The rebase onto the merged master conflicted exactly there, and I resolved it by keeping master's field-less file.

Not tagging from this branch, per the releasing skill: a squash or rebase merge rewrites the hash, so a tag on a pre-merge commit would point at a commit that never reaches master while still triggering the release build. Once this merges, the tag goes on the merged commit:

```
git checkout master && git pull && git tag v0.17.1 && git push origin --tags
```

---

Issue #627, opened 2026-08-15
app/engine(openapi): phase 2 - sync: re-fetch, diff by operation identity, selective apply - with external $ref bundling and the examples origin column


athrvk:
Phase 2 of #625; requires phase 1 (#626). The importer family stops being a one-way door.

## The two prerequisites this phase carries

**1. External `$ref` bundling - the hidden requirement.** Both parsers resolve in-document `#/` refs only, via an identical duplicated closure (`openapi-v3.ts:71-79`, `openapi-v2.ts:68-76` - the repo's hand-rolled-copy pattern); an external ref silently resolves to `undefined` today, no error, no skip tally. Real-world specs at sync's target audience are multi-file. Fix, app-side, before parse:
- **Unify the resolver** (one shared module, both parsers - retire the copies).
- **Bundling step**: walk the raw document for external refs; for a URL-sourced spec, resolve relatives against the source URL and fetch via the existing engine `/import/fetch`; for a file-picked spec, read sibling files via a new gated IPC `specFile:read` modeled exactly on `dataFile:read`'s landed shape (`app/electron/data-file.ts:58-154`: extension allowlist `.json/.yaml/.yml`, byte cap fetched from the engine's `maxSpecDocumentBytes`, bytes-not-text, posture rationale in the module header) with the dirname captured at pick time via `getFilePath`. Inline the resolved subtrees under `#/components` (standard bundling), cap total bundled size at the same engine cap, and **tally unresolvable refs into the skip tally** instead of today's silence. One line per the security-scope rule: this reuses `/import/fetch` as-is and must not widen it; its hardening belongs to the deferred security session.

**2. The examples origin column.** `request_examples` records no origin (`database.cpp:248-261` - "Created by import today"); sync must update spec-derived examples without clobbering user-added ones. Add `origin` TEXT NOT NULL `default_value("import")` (ALTER-friendly backfill; honest for every existing row, which import created), values `import` | `user`; `apply_request_example_fields` accepts it so `/import/apply` and the single routes agree; #588's save-as-example writes `origin: "user"` (its acceptance criterion has been revised accordingly - see the comment there).

## Sync itself

[...]

Issue #625, opened 2026-08-15
OpenAPI-native Vayu - bind a spec to a collection, keep it in sync, validate and measure coverage against it, and export back out - epic


athrvk:
## The positioning

Vayu's OpenAPI import is already deep (2.0/3.0/3.1, YAML/JSON, URL fetch, auth schemes, `op.responses` into the examples/mock pipeline, disclosed skip tallies). This epic makes the spec a **first-class contract bound to the collection** - not a one-way door, and deliberately not a spec editor. When it completes, the honest marketing claim is **"OpenAPI-native: import, sync, validate, coverage, export - 2.0 through 3.1"**, with contract-aware load testing ("held 2,000 RPS and every sampled response matched its schema; 14 of 18 operations covered") as the headline no competitor prints. The word "compliant" is deliberately not claimed - there is no tool certification to claim.

Grounded at `2666e0b`; every decision below is anchored to a verified precedent.

## Decisions (recorded here; phases implement)

1. **Spec-attached, not spec-first.** The collection stays Vayu's native object; the spec is a bound contract. Free-form collections are untouched. Spec *editing* is out of scope permanently - Vayu consumes and verifies contracts.
2. **Storage follows the #598/#599 two-halves law.** Portable state engine-side: a new `spec_documents` table (engine-minted id, `content` TEXT, `source_url`, `fetched_at`, `hash` = `vayu::utils::sha256` hex - the `body_blobs` content+hash shape at `database.cpp:320-322`, table creation needs no migration since `sync_schema` creates new tables outright) plus an object-shaped `openapi` binding field on collections (`{"specId","specHash","syncedAt"}`, default `"{}"` = unbound - the `data_schema` column/applier/serializer precedent at `database.cpp:206-211` / `collections.cpp:229-239` / `json.cpp:313-318`) plus a per-request operation identity (`operationId` + method + path template - the `http_version` TEXT column-add precedent, `database.cpp:231-244`, added in `apply_request_fields` and in BOTH request serializers `json.cpp:325-390` and `:1012`). Machine-local state renderer-side: a spec file's absolute path in a zustand persist store per the `data-file-store.ts` pattern. Spec size gets an engine-authored, named, config-overridable cap (the `maxTraceBodyBytes` + `MAX_DATA_BYTES`-rationale precedents).
[...]

PR #643, opened 2026-08-15
feat(app/examples): save a live response as an example, and remove one

### LINKED ISSUE #588: app(examples): "Save response as example" - the #481 follow-up that lets a live response feed the mock server without a round-trip through curl
## Problem

#481 shipped the examples model and the collection mock server, and named "save a live response as an example" a phase-1 non-goal - "a natural follow-up". The follow-up had no tracker; this is it.

Today the engine's example routes are import-only from the app's perspective: `engine/src/http/routes/examples.cpp` implements POST/PUT/DELETE but no app surface calls them (documented as deliberate at `docs/app/api-integration.md:149-151` - "Read-only from the app: examples arrive by import"). So the only ways to grow a mock's route table are re-importing a collection or hand-curl against the CRUD routes. The natural loop - send a request in the builder, like the response, keep it as the example the mock will serve - does not exist, even though every piece below it does: the response is on screen, the request is saved, the mock serves the first ordered example per route.

## What ships

- A **"Save as example"** affordance on the response viewer of a *saved* request (unsaved/transient requests have no `requestId` to attach to - the affordance is absent, not disabled-with-a-lie, per the repo's honesty discipline; or disabled with the reason stated, decide and record).
- It captures status, headers, body from the rendered response into `POST /requests/:id/examples` (the existing route - check the exact shape in `examples.cpp` and reuse the importers' field mapping so import-created and app-created examples behave identically). **Revised per the #625/#627 OpenAPI epic**: app-created examples are written with `origin: "user"` (the origin column #627 introduces on `request_examples`), so spec sync can update import-derived examples without ever clobbering user-saved ones. If this issue lands before #627's column exists, coordinate: either land the column here first (its ALTER shape is specified in #627) or sequence after it.
- Name defaulting: "&lt;status&gt; &lt;reason&gt;" or the request name + status; editable inline at save time (one small dialog or inline field, match the StartMockServerDialog scale, not a new modal framework).
- The ExamplesPanel (`app/src/modules/request-builder/components/RequestTabs/panels/ExamplesPanel.tsx`) lists it immediately (query invalidation), and gains delete for app-created examples while there - the engine DELETE route also has no caller, and an example you can create but never remove recreates the #553 zombie shape at a smaller scale. Ordering: appended last (the mock serves the first ordered example; saving must not silently change what a running mock would serve after restart - state this in the dialog or docs).
- Body caps: respect whatever budget the trace/body storage already enforces; a truncated stored body must be flagged truncated, never saved-as-complete (truncation-disclosure discipline).

## Explicit non-goals

- No editing of a saved example's body in v1 (the panel is a viewer; editing is its own issue if wanted).
- No auto-save-on-every-response, no "update existing example from response" merge flow.
- Mock hot-reload stays out: a running mock still serves its start-time snapshot (documented #481 decision, `docs/engine/api-reference.md:1642-1645`); the new example is picked up on restart like any other edit.

## Tests

- App: save flow round-trip (mocked api layer), affordance absent/disabled for unsaved requests, panel refresh, delete flow, truncated-body flagging - mutation-checked per discipline.
- Engine: none expected beyond the origin field (routes exist and are tested in `examples_route_test.cpp`); add a case only if the app needs a field the route does not currently accept.

## Docs, same commit

`docs/app/api-integration.md:142-160` - the "Read-only from the app" paragraph becomes false the moment this lands; rewrite it. `docs/app/COMPONENTS.md` ExamplesPanel section. Mock docs already state restart-to-reload; add one line that app-saved examples follow the same rule.

## Sequencing

Coordinates with #627 (the examples `origin` column - see the revised bullet above; either order, the column's shape is fixed there). Otherwise parallel-safe against the open set. Touches the response viewer, ExamplesPanel, and the api service layer.

## Acceptance criteria

- [ ] A saved request's live response can be kept as an example from the response viewer, and the mock serves it after restart.
- [ ] App-created and import-created examples are indistinguishable **in behaviour**; storage records `origin: "user"` for app-created rows (revised 2026-08-15 for the #625 OpenAPI epic - sync must be able to tell them apart); ordering appends.
- [ ] Examples are deletable from the panel; unsaved-request behaviour is honest and its choice recorded.
- [ ] Truncated bodies are flagged, never silently saved as complete.
- [ ] `api-integration.md` no longer claims examples are read-only from the app.

athrvk:
## Description

The #481 follow-up it named a phase-1 non-goal: a live response can now be kept
as the example a mock server answers with, without a round trip through curl.

**The plan (root cause → approach → rejected alternative).** The engine's
example routes were complete and had no app caller (`api-integration.md:149-151`
recorded that as deliberate), so the only ways to grow a mock's route table were
re-importing the collection or hand-curling `POST /requests/:id/examples` -
even though the response is on screen, the request is saved, and the mock
already serves the first ordered example per route. The approach is to call the
routes that exist rather than add any: a **Save as example** action on the
response viewer writing through the existing `POST`, and a delete on the
Examples panel through the existing `DELETE`. The one engine change is #627's
`request_examples.origin` column, landed here per that issue's "whichever of
#588/#627 lands first carries the column" note. **The alternative I rejected is
saving silently with a generated name.** The name is the only field the panel
lists, so "200 OK" three times over is a list you cannot read - and the save is
also the last honest moment to say that a restored run's body may be partial, or
that the row appends rather than taking the position the mock serves from. I
also rejected making delete conditional on `origin`: the engine's route is not
scoped that way, an imported example is as much the user's to prune as a saved
one, and gating it here would mean reading a field no surface displays.

**Verified against the code first**, per the issue's ask: the routes, the
per-request cap and the append-on-absent-`order` rule are all still as #481
described them, so the fix is the one the issue specifies.

### What landed

**Engine - the `origin` column (#627's shape).** `import` | `user`, NOT NULL,
`default_value("import")` so `sync_schema` ALTERs it onto the existing table and
every pre-existing row backfills to `import`, which is what all of them are.
`apply_request_example_fields` validates it and returns a **400** on anything
else rather than falling back to the default - on the `apply_http_version_field`
reasoning: an absorbed typo would store a user-saved example as an import row
for the next spec sync to overwrite, which is the one outcome the column exists
to prevent. Both write paths share that applier, so the single route and
`POST /import/apply` cannot drift; the import path claims no origin and its rows
are `import`, asserted.

**App - Save as example.** At the right of the response tab strip, not beside
Copy and Download: those act on the body and live in the body pane, this acts on
the whole exchange. It is **absent, not disabled**, in two cases (the issue asks
for the choice to be recorded):

- *Unsaved request* - there is no id to nest an example under, and the Examples
  tab beside it already says exactly that in a full sentence. A permanently
  greyed button whose explanation only appears on hover would be a second,
  worse copy of it - and Radix cannot reliably show a tooltip on a natively
  disabled trigger anyway.
- *A stream still open* - the response shown is then the placeholder built from
  the relay's `open` frame, whose body is empty by construction. The button
  returns when the stream ends and the run's stored response takes over.

The payload rules live in `save-as-example.ts`, not the dialog, because they are
the part with a contract: `origin: "user"`, the **Postman importer's**
Content-Type mapping header-for-header (`""` when the response stated none), and
never an `order` - the engine appends, which is what keeps a restarted mock
answering with the same first example. That last rule is also stated in the
dialog, along with the restart-to-reload behaviour.

**Truncated bodies.** A restored run's body may be only the slice the engine
kept, and a mock serves an example as though it were whole. The dialog warns
with both byte counts and points at re-sending, and the default name carries a
`(truncated body)` suffix - the name is the only field the panel lists, so it is
the part of the disclosure that survives the save. The engine has no truncation
column and this did not add one.

**App - delete.** On each panel row, confirmed through the existing
`DeleteConfirmDialog` because removing a row can change what the next mock
restart serves. The row is now a container that paints the hover with a
`self-stretch` activator, per the `drawer-row-hit-area` rule - two controls in
one row is exactly the shape that guard exists for. A refused delete is named
inside the dialog rather than closing over it.

1. **Delete is not scoped to app-created examples.** The issue says "delete for
   app-created examples"; this deletes any of them, for the reasons in the plan
   above. Noted here rather than quietly widened.
2. **The mutation lives in `SaveAsExampleDialog`, not its caller** - the other
   way round from the `StartMockServerDialog` precedent it otherwise follows.
   There the parent owns a success toast and a running/stopped switch either
   way; here success means only "close". It also keeps `ResponseViewer` - a pane
   that redraws on every send, rendered by five unrelated test files - from
   needing a query client for a feature that only exists inside a modal.
3. **`origin` is write-only in the app.** It is not added to the `RequestExample`
   domain type, because no surface displays it and this repo's most repeated
   defect is a field written with no reader. Its reader is #627.

## Type of Change
- [x] New feature
- [x] Documentation update

## Testing

- `python build.py -e -t` then `ctest --preset linux-dev --output-on-failure` -
  **1659/1659 passed**, including **3 new tests** in
  `engine/tests/examples_route_test.cpp` (origin defaults to `import` and keeps a
  claimed `user`; an unknown origin and a non-string origin are both 400s that
  write nothing; the null-vs-absent rule on update) plus origin assertions folded
  into the existing import round-trip.
- `cd app && pnpm test` - **4765/4765 passed** across 460 files, including **7
  new tests** in `ResponseViewer/save-as-example.test.tsx` and **2** added to
  `RequestTabs/examples-tab.test.tsx`.
- `pnpm type-check`, `pnpm lint` and `pnpm format:check` clean;
  `mkdocs build --strict -f .github/mkdocs.yml` builds clean with the docs
  changes. Only the three files this PR touched were formatted.

**Mutation-checked** (broken, confirmed red, restored):

- `origin: "user"` → `"import"` in `exampleFromResponse` - the payload test
  fails, and nothing else does. That is the field with no reader in this app, so
  it is the one a test has to hold.
- `canSaveExample` → `true` - both availability tests fail (unsaved request, live
  stream), which is what proves the gate rather than the button.

The delete tests only pass against the new affordance, so they cannot go green
without it. No pre-existing failures reproduced on this base.

## Checklist

Docs land in the same commit per the repo rule: `docs/app/api-integration.md`
(the "Read-only from the app" paragraph is now false and is rewritten, with the
write-only `origin` contract), `docs/app/COMPONENTS.md` (both surfaces),
`docs/engine/api-reference.md` (the `origin` field on the routes, the import
section, and the mock server's restart rule naming app-saved examples),
`docs/engine/db-schema.md` (the column and its migration shape) and
`engine/CLAUDE.md`.

## Related Issues

Closes #588. Carries the `request_examples.origin` column #627 specifies, so
phase 2 of the OpenAPI epic (#625) consumes it rather than adding it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GLRuwivNtqTFToFqjtK57o)_


athrvk:
fix merge conflict

athrvk:
Conflict resolved - `master` merged in at `967f91e`.

#640 and #641 landed between branching and opening this. The only textual
conflict was one row of the `docs/app/COMPONENTS.md` component table: #641
rewrote the `UrlBar` line for `SendWithRowMenu` while this branch rewrote the
`RequestTabs` and `ResponseViewer` lines beside it. Both sides are kept - #641's
`UrlBar` text verbatim, this branch's other two. Everything else auto-merged,
including the engine files #640 also touched (`types.hpp`, `database.cpp`,
`json.cpp`): `spec_operation` and `origin` are separate columns on separate
tables, and the two serializers now emit both.

Re-verified against the new base, not just re-merged:

- engine **1702/1702** (`python build.py -e -t`, `ctest --preset linux-dev`)
- app **4787/4787** across 461 files
- `pnpm type-check`, `pnpm lint`, `pnpm format:check` and
  `mkdocs build --strict` all clean

The counts in the PR body are from the pre-merge run; these are the current
ones. Nothing in the diff itself changed.

---

Issue #655, opened 2026-08-15
app/engine(openapi): phase 2c - apply a spec diff selectively and atomically, never overwriting what the user touched


athrvk:
The write half of #627, split from the read half (#654) - see #654 for why, and #627/#625 for every decision restated here rather than re-opened. Lands **after** #654, which owns the re-fetch, the three-bucket diff and the user-touched flags this consumes.

## Scope

- **Selection.** The diff renders as a checklist (the `ImportModal` preview's structure is the visual precedent), per item and per bucket. Applying is explicit and partial by design - a user who wants the four new operations and not the removal of the one they still use gets exactly that.
- **Added** creates requests through the normal appliers, in the collection the identity's tag maps to, stamped with their operation identity.
- **Removed** deletes, under the count-naming confirm discipline - never a silent cascade.
- **Changed** updates the spec-derived fields **only**, and never a field #654 flagged as user-touched unless the user ticks that field specifically. The flag is the contract between the two halves; a change that reads it as advisory is the defect this split exists to prevent.
- **Examples**, which #654 deliberately left uncompared: only `origin="import"` rows are replaced, `origin="user"` rows always survive, and the mock-serves-first ordering is preserved with the exact rule stated in the PR. `request_examples.origin` already exists (`database.cpp:274`, via #588).
- **The commit is one transaction.** Engine-side, extend `/import/apply` or add a sibling batched-update path - decide and record; the atomicity is the requirement, not the route. `spec_documents` gains the new document and `Collection.openapi` moves to it (`specId`, `specHash`, `syncedAt`) in the same commit that applies the rows, so a collection is never bound to a document its requests do not reflect.
- **The #622 rule re-applies**: stub-param enabled/disabled is re-derived by the same keyed-on-`required` rule at sync, with the user-touched protection covering manual flips.

## Tests

- Selective subsets: applying one bucket leaves the others exactly as they were.
- A user-touched field survives an apply that changes its operation; mutation-check the protection (drop the flag check and the test reddens).
- `origin="user"` examples survive a sync that replaces the imported ones; ordering rule asserted.
- Atomicity: a failing item rolls the whole apply back (the import transaction precedent), and the binding does not move when the apply fails.
- An unchanged spec applies nothing and touches no row.

## Docs, same commit

[...]

Issue #630, opened 2026-08-15
app(openapi): phase 5 - export: a collection back out as an OpenAPI 3.1 document, with both fidelity directions stated


athrvk:
Phase 5 of #625; requires phase 1 (#626) only. Closes the lifecycle loop and resolves the recorded #599/PR #610 deferral that waited on "a native export format".

## Where it runs (decided in #625)

**App-side assembly** from engine GETs - the codegen precedent (`app/src/services/codegen` builds derived artifacts from engine data; the engine keeps owning canonical resource JSON, and gains no route). A new `app/src/services/exporters/openapi.ts` beside the importers, sharing their types and the unified `$ref`/identity modules from phases 1-2.

## The two fidelity directions, stated up front

1. **Spec-bound collection: update the original document.** Start from the stored `spec_documents.content`, patch what Vayu owns - operations present/absent per the collection's requests (by the phase-1 identity), request-derived fields (params, headers where they came from the spec's vocabulary), examples from `origin="import"` AND `origin="user"` example rows (user examples become `examples` entries - export is where user work flows back to the contract) - and **preserve every member Vayu does not model** (vendor extensions, `info`, `tags`, unmapped `components`) byte-faithfully where untouched. Output validates against the same dialect it arrived in (a 3.0 spec exports as 3.0; do not force-upgrade).
2. **Free-form collection: a skeleton, honestly.** `openapi: 3.1.0`, `info` from the collection name, `servers` from `{{baseUrl}}`-style variables where derivable, one operation per request (method, path with `{param}` templating recovered from path segments that are variables, params/headers as declared), examples from stored example rows and nothing else. **No invented schemas**: request/response schemas appear only where an example body exists to derive a trivial type-shape from, and even then marked with a comment-equivalent (`description` noting derivation). The skeleton claim is in the UI copy: "a starting point, not a contract".

Round-trip invariant worth its test: export a bound collection, re-import the result - the diff (#627's machinery) is empty.

## UX

- Export action on the Spec tab (bound) and the collection context menu (both kinds). Format choice JSON/YAML (`js-yaml` already a dependency - the importers' parser, used in reverse).
- Delivery via the repo's one file-to-disk pattern: Blob + `&lt;a download&gt;` (`ResponseActions.tsx:55-64`); no new IPC. Copy-to-clipboard secondary via `useCopy` (toast on failure honesty).
[...]

PR #656, opened 2026-08-15
feat(app/openapi): phase 2b - re-fetch a bound spec and diff it, before anything is applied

### LINKED ISSUE #654: app(openapi): phase 2b - re-fetch a bound spec and diff it, in three honest buckets, before anything is applied
Extracted from #627 (phase 2 of #625), split for the same reason #626 became #637/#638 and #649 came out of this phase: what is left of #627 after bundling landed is still **re-fetch + diff + a selection UI + a transactional apply**, and a half-applied sync is the worst state this epic can be in. This issue is the read half - work out and show what changed. #TBD-2c is the write half.

Verified against `f8011d0` before writing: `SpecTab.tsx` has no re-fetch of any kind, `spec-operations.ts` reads identities only (it throws the drafts away), and `operation-match.ts:36` already names #627's sync as the thing that will consume a wrong identity.

## Problem

A bound collection knows which document it answers to (`Collection.openapi.specId`/`specHash`) and which operation each request is (`Request.specOperation`), and there is still no way to ask **"has the contract moved?"**. The document is re-fetchable - a URL sits in `spec_documents.source_url`, a picked file's path sits in `spec-file-store` - and nothing re-fetches it. The user's only route to a changed spec is importing it again beside the old copy, which is the one-way door phase 1 was supposed to close.

## Scope - the read half only

**Nothing here writes.** No request is created, deleted or edited, `spec_documents` is not updated and `openapi.syncedAt` does not move. That is #TBD-2c, and it is the boundary: a diff that cannot be trusted is worse than no diff, so the diff earns its trust in a change that cannot damage anything.

- **Re-fetch, by the source the binding already records.** URL-sourced re-fetches through `apiService.importFetch` (reused as-is, not widened - its hardening stays with the deferred security session); file-sourced re-reads through the landed `specFile:read` gate; a document with neither (pasted, or picked on another machine) says so and offers the picker rather than pretending it can look. The fetched text goes through `bundleExternalRefs` with both intakes, exactly as import does, because the stored document is the bundled one and a diff against an unbundled re-fetch would report every external `$ref` as a change.
- **Unchanged short-circuits.** The stored document's bytes are already in hand from `useSpecQuery`, so equality against those bytes is the check - the engine owns hashing (`spec_content_hash`, `specs.cpp:69`) and a second implementation in the renderer is the repo's named hand-rolled-copy defect. State this in the PR: #627 says "hash the fetched content", and comparing the bytes the hash is of is the same answer without a second hasher.
- **Diff by operation identity, three buckets.** `added` (an operation no request claims), `removed` (a request whose identity the new document no longer declares), `changed` (identity matches, and a field a fresh import would produce differs). Identity is matched by `operationId` first, `METHOD path-template` second, per #627.
- **Renames surface as renames.** An `operationId` that persists across a path change follows the `operationId`; a path that persists across an `operationId` change follows the path; both changed is `removed` + `added`, disclosed as such and never guessed at - the rule `operation-match.ts` already states for binding.
- **User-touched fields are flagged, three-way.** The old document is still stored, so for each matched request the comparison is old-spec-derived vs current vs new-spec-derived: a field the user edited away from what the old spec produced is marked, and the mark is what #TBD-2c reads to know it must not overwrite it silently. A field the user did not touch and the spec did not move is neither changed nor flagged.
- **The tab grows a Sync section**, bound state only: one action to check, then the three counts (zeros included, the `MatchSummary` discipline) and the per-item detail - which operation, which fields moved, which of those the user has edited. Read-only, and it says so: the apply is the next phase, named.

**Deliberately not compared here: response examples.** Comparing them means a query per request, and the rule that decides what happens to them (`origin="import"` is replaced, `origin="user"` survives) belongs to the change that applies them. Stated in the docs and in the UI's scope line rather than left as a silent gap.

## Where the code goes

`readSpecOperations` keeps the drafts it currently discards (a sibling reader in `services/openapi/spec-operations.ts`, expressed once - the identities it returns today are derived from the same pass), so the diff compares against what an import *would* build rather than a second opinion about what a document means. The diff itself is a pure module beside it (`services/openapi/spec-diff.ts`), taking two documents plus the requests and returning the buckets - no queries, no React, so the semantics are testable without a DOM.

## Tests

- Each bucket, including the zero cases; the byte-identical short-circuit reports up to date and computes no diff.
- Both rename directions, and both-changed disclosed as removed + added.
- The three-way flags: spec moved a field the user never touched (changed, not flagged); user edited a field the spec did not move (not changed, flagged); both moved (changed **and** flagged). Mutation-check the flag - drop the old-spec side and the "user edited it" test reddens.
- A request whose identity the new document does not declare lands in `removed` and is never matched by path shape by accident.
- Re-fetch: URL and file intakes both reach the bundler; a document with no source states so and offers the picker; an unreachable source surfaces the engine's message.
- Spec tab: the counts render, the per-item detail names fields, and nothing in the section can write (no mutation is called - assert it).

## Docs, same commit

`docs/app/openapi.md` - a sync section (what is compared, what a rename means, why examples are not in this half, and that this half only reports), and the "What is not here yet" list loses the half that ships. `docs/app/COMPONENTS.md` if the tab's paragraph names its sections.

## Sequencing

Wave 2 of #625, after #649 (landed - the bundler this depends on). **Before #TBD-2c**, which applies what this computes. Independent of #628-#630. Parallel-safe against everything outside the epic.

## Acceptance criteria

- [ ] A bound collection re-fetches its document from the source the binding records, or says honestly why it cannot.
- [ ] An unchanged document reports up to date without computing a diff or writing anything.
- [ ] A changed document produces three buckets with rename semantics as stated, and per-request field detail.
- [ ] A field the user edited is flagged against the *old* spec-derived value, not against the new one; mutation-checked.
- [ ] Nothing in this change writes: no request, no collection, no spec document.
- [ ] Docs land same-commit; `mkdocs build --strict` clean.

Part of #625; the read half of #627.

athrvk:
## Description

Phase 2b of #625: a bound collection can ask whether its contract has moved, and see exactly where.

**The plan.** The root cause is a door phase 1 left half-open: a collection records which document it answers to and which operation each request is, and nothing re-reads that document - so the only route to a changed spec was importing it again beside the old copy, which is the one-way door the binding existed to close. The approach splits #627's remaining scope on the read/write line (#654 here, #655 next) and puts every judgement a sync makes - which operation is which, what counts as changed, what the user edited - into a pure function over two documents and the requests, where it is provable in a test that cannot touch a row. That ordering is the point: a diff nobody trusts yet is harmless, a half-applied sync is not. The alternative I rejected was landing #627 whole: re-fetch through two intakes, the diff, a selection UI and an atomic engine commit in one change is past what this repo lands at its bar, and the failure mode is asymmetric - the apply half is the half that destroys work.

**Three decisions worth naming, all recorded in the code:**

- **"Up to date" is byte equality against the stored document, not a second hash.** `useSpecQuery` already holds the bytes the engine hashed (`spec_content_hash`, `specs.cpp:69`); a renderer-side SHA-256 would be the hand-rolled copy of a rule that CLAUDE.md names as this codebase's recurring defect. Same answer, one hasher. #654 records this as a deliberate deviation from #627's "hash the fetched content".
- **The file intake reuses `specFile:read` rather than adding a channel.** The document names *itself* as the file beside itself, so the main process still does the path arithmetic and the extension allowlist and the engine-fetched byte cap are stated once. A second channel would have to repeat both, identically, forever.
- **Renames are followed; ambiguity is not guessed.** `operationId` first, path shape second (through `specPathShape`, the same flattening binding uses, so a renamed path *parameter* is still one endpoint). Both halves moved at once is disclosed as a removal plus an addition, because there is nothing left to follow and a wrong identity is precisely what would make #655 rewrite the wrong request.

**The user-touched flag is three-way** - the request holds neither what the new document produces nor what the bound one did - because a two-way comparison cannot tell an edit apart from a spec change, and #655 reading a two-way flag would quietly revert somebody's work. When the bound document itself cannot be read, that is stated per request rather than guessed at.

`readSpecOperations` now keeps the drafts it used to throw away, so "changed" is measured against what an import would really produce; the identities it returns are derived from the same walk, so the two readers cannot disagree.

**Deliberate scope boundaries** (both stated in the UI and in the docs, not left silent):

- **Nothing in this change writes.** No request, no collection, no spec document; `openapi.syncedAt` does not move. There is a test that asserts no write reaches `apiService` during a check.
- **Response examples are not compared.** It costs a query per request, and the rule that decides their fate (`origin="import"` replaced, `origin="user"` kept) only means anything at apply time, so they land with #655.

## Type of Change
- [x] New feature
- [x] Documentation update

## Testing

- `cd app && pnpm test` - **471 files, 4912 tests passed** (28 new: 13 in `spec-diff.test.ts`, 7 in `spec-refetch.test.ts`, 8 in `SpecSync.test.tsx`).
- `cd app && pnpm type-check` - clean. `eslint` clean on every touched file; `prettier --check` clean on the same set.
- Engine untouched, so no `ctest` run: no C++ file is in this diff and no test could change colour.

**Mutation-checked, run rather than reasoned about:**

- Making the user-touched flag two-way reddens 4 tests across the diff and the UI (`does not flag a field only the document moved`, `makes no claim about who edited what when the bound document cannot be read`, and the two that assert which of two fields is marked) - then green again on restore.
- Dropping the operationId-first lookup reddens `follows an operationId whose path moved, rather than reporting a delete and an add`, which is exactly the data-loss case.

What else the tests lock: each bucket including the zero cases; both rename directions and both-changed as removal + addition; a renamed path parameter as one endpoint; a request with no operation counted rather than reported as removed; URL and file intakes both reaching the bundler with the URL preferred (only it is portable); the two no-origin messages kept distinct, because "pick it again" is wrong advice in a browser; an unreachable `$ref` counted rather than dropped; and the `--rule` declaration guard rendering the section and asserting it scanned more than nothing.

## Checklist

## Related Issues

Closes #654. Part of #625; the read half of #627, whose write half is #655.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Rv3we36NYYgwoNkK5dvrQY)_


athrvk:
fix merge conflict

athrvk:
Conflict resolved - `master` merged in (#651 and #652). The branch is 0 behind and the diff against `master` is the same 11 files.

Both conflicts were in docs, where three phases now describe the same tab:

- **`COMPONENTS.md`, the tab table** - the Spec row gained the Sync sentence *and* kept #630's "Export as OpenAPI"; the two edits were adjacent, not competing.
- **`COMPONENTS.md`, the architecture section** - `SpecSync.tsx` and `ExportSpecDialog.tsx` are separate paragraphs, both kept.
- **`openapi.md`, "What is not here yet"** - both sides removed something from that list (export landed on `master`, re-fetch lands here), so the resolved list is what actually remains: applying what a check found, response validation, coverage.

Neither file is prettier-clean on `master`, so per the repo rule they were left unformatted rather than reflowed.

Re-verified on the merged tree: **473 files, 4935 tests pass**, `pnpm type-check` clean, `mkdocs build --strict` clean. `spec-diff.ts` still imports `operationShapeKey` / `specPathShape` from `operation-match.ts`, which #651 refactored - the exports it uses are unchanged, and the 17 files covering `services/openapi` and `CollectionDetail` pass together.

---

Issue #649, opened 2026-08-15
app(openapi): phase 2a - one $ref resolver, and external refs bundled before parse instead of silently undefined


athrvk:
Extracted from #627 (phase 2 of #625), which carries this as "the two prerequisites this phase carries", item 1. Split for the same reason #626 became #637/#638: phase 2 as written is bundling **plus** the whole sync pipeline - re-fetch, a three-bucket diff with rename semantics, three-way user-touched protection, selective apply, an atomic commit - and a half-applied sync is the worst state this epic can be in. Bundling stands alone and is worth landing alone: it fixes a defect that is live **today**, on every multi-file spec import, whether or not sync ever ships.

Two corrections to #627 while I was in the code, both verified at `5ac2bb2`:

- **Prerequisite 2 has already landed.** `request_examples.origin` exists (`database.cpp:274`, `NOT NULL` + `default_value("import")`, via #588). #627 no longer carries it.
- The duplicated resolver is at `openapi-v3.ts:73-81` and `openapi-v2.ts:69-77` (not `:71-79` / `:68-76`), and `openapi-shared.ts` already exists - it holds `deref`, `resolvePathItem` and `SkipTally`, but not the resolver those call.

## Problem

Both parsers build the same resolver closure by hand:

```ts
const resolveRef = (ref: string): unknown => {
    const path = ref.replace(/^#\//, "").split("/").map(unescapePointer);
    let cur: unknown = spec;
    for (const seg of path) cur = prop(cur, seg);
    return cur;
};
```

It resolves in-document pointers only. Given `"./schemas/pet.yaml#/Pet"` - or an absolute `"https://example.com/common.yaml#/components/schemas/Error"` - the `^#/` strip does nothing, the whole string is split into segments that name no key, and the walk lands on `undefined`. `deref` then hands `undefined` back to `buildOperation`, which reads it as *the spec documented nothing here*: the request body stub is empty, `$ref`-ed parameters vanish, `$ref`-ed responses import no example. **No error, no skip tally, no difference the preview can show** - the user sees a smaller import and no reason for it. Multi-file specs are the norm at the audience sync targets.

The duplication is the repo's named hand-rolled-copy defect: whatever this fix is, it must not be written twice.

## Scope

**One resolver, shared.** `createRefResolver(document)` in `openapi-shared.ts`, used by both parsers; the two closures are deleted. Same single-hop, same pointer unescaping (`~1` / `~0`) - behaviour identical for the `#/` case that works today.

[...]

Issue #628, opened 2026-08-15
engine/app(openapi): phase 3 - response schema validation: design trace, per-step in collection runs, sampled under load - the validator dependency lands here


athrvk:
Phase 3 of #625; requires phase 1 (#626) only - runs in parallel with sync (#627). The contract-aware testing headline starts here.

## The division of labor (decided in #625)

The app pre-extracts **dereferenced per-operation response schemas** from the bound spec at bind/sync time and stores them engine-side beside the spec document (a `response_schemas` JSON on `spec_documents`, or a sibling table keyed by operation identity + status pattern + content-type - decide shape in the PR, record). The engine never parses OpenAPI; it validates a response body against a ready JSON Schema. Extraction reuses the resolver machinery from #627's bundling work where present, or in-document resolution alone until #627 lands (state which at implementation time).

## The dependency

`json-schema-validator` (pboettch, v2.4.0, MIT, nlohmann-json-based) added to `engine/vcpkg.json` per the never-undeclared-deps rule. **Dialect honesty is an acceptance criterion**: the library targets draft-07. OpenAPI 3.0 schemas fit; 3.1 is JSON Schema 2020-12, and any keyword the validator cannot evaluate must be **counted and disclosed as unevaluated** in the verdict (the truncation-disclosure discipline) - never silently passed, never spuriously failed. `valijson` is the recorded fallback candidate if integration fails; switching is a PR-time decision, documented.

## Verdict semantics

A validation verdict is `{checked, valid, failures[], unevaluatedKeywords}` where a failure names the JSON path, the keyword, and a bounded message; `checked=false` (with a reason: no schema for this status/content-type, unparseable body, unbound collection) is distinct from `valid=false`. Failures are capped per the `MAX_FAILURE_MESSAGES` script-replay precedent. Matching: operation identity from the request's `spec_operation`, response status against the spec's status patterns (exact, `2XX` ranges, `default`), content-type against the declared media types - each miss is a named `checked=false` reason.

## The three hooks (each on its verified precedent)

1. **Design mode**: verdict node beside `events`/`scripts` in the stored trace at the `execution.cpp:500-560` store site, after body capping. Both app-side ResponseState funnels gain it **together** - `responseFromExecuteResult` (`execute-mapping.ts:154-205`) and `responseFromRunResult` (`restore-response.ts`, contract-coupled to `store_result`) - the named copy-does-not-receive-the-fix trap; history restore must show the same verdict live execution did.
[...]

Issue #626, opened 2026-08-15
app/engine(openapi): phase 1 - the binding: spec_documents storage, the collection binding, per-request operation identity, and the Spec tab


athrvk:
Phase 1 of #625 (decisions recorded there). The storage foundation everything else stands on - shippable alone: after this, an imported spec is *remembered*, visible, and stamped onto runs.

## Engine

- **`spec_documents` table**: `id` (engine-minted `spec_` prefix via `generate_id`), `content` TEXT, `source_url` TEXT nullable, `fetched_at`, `hash` (hex of `vayu::utils::sha256(content)` - `sha256.hpp:39`, the `body_blobs` content+hash shape at `database.cpp:320-322`). New table = no migration (`sync_schema` creates outright, the `request_examples` comment precedent at `:248-250`; indexes precede tables in the make_storage list, `:161-194`). Routes: `POST /specs` (content + optional sourceUrl; engine computes hash; size-capped), `GET /specs/:id`, `DELETE /specs/:id` (refused while a collection binds it, or cascades the binding to unbound - decide and record; recommend refuse-with-name, the mock-issuer in-use precedent).
- **Size cap**: engine-authored named constant, config-overridable as `maxSpecDocumentBytes` (default 10MB = `json::MAX_FIELD_SIZE` alignment), refusing with count and cap per the `MAX_DATA_BYTES` rationale (`constants.hpp:350-357`) - never the transport's reset.
- **Collection binding**: object field `openapi` (`{"specId","specHash","syncedAt"}`), stored as NOT NULL TEXT default `"{}"` = unbound - column, applier (`apply_json_field` + contents validation after, the `validate_data_schema` shape at `collections.cpp:112-161`), serializer fallback block (`json.cpp:313-318`), all in `apply_collection_fields` so POST/PUT/`/import/apply` cannot drift. `specId` must resolve at bind time except under import's `is_create` path, where it resolves through the temp-id map (below).
- **Per-request operation identity**: one nullable TEXT column `spec_operation` holding `{"operationId"?, "method", "path"}` (the `config_entries.unit` nullable precedent, `database.cpp:379-382`; JSON object because path templates and operationIds are one fact, not two columns). Applied in `apply_request_fields` (`requests.cpp:123-198`), added to **both** serializers (`json.cpp:325-390` AND `serialize_to_stream` `:1012` - the named two-serializer trap), roster comment updated (`requests.cpp:385-386`).
[...]

PR #676, opened 2026-08-16
feat(engine/app/openapi): contract coverage - which operations a run exercised, which declared responses it saw

### LINKED ISSUE #629: engine/app(openapi): phase 4 - contract coverage: which operations a run exercised, which declared responses it saw, as a first-class report block
Phase 4 of #625; requires phase 1 (#626) and phase 2's matured identity mapping (#627). The marketing sentence's second half: "14 of 18 operations covered, every declared 2xx seen, three declared 4xxs never triggered".

## What coverage is

For a run of a spec-bound collection: the set of spec operations (from the bound spec at the run's stamped `specHash` - phase 1 put it in `config_snapshot`, so coverage is computed against **the spec as it was when the run ran**, not as it is now), crossed with what the run actually did - per operation: requests sent, response status codes seen, and which of the operation's *declared* response codes those hit. Rollup: operations covered / total, declared-response coverage, undeclared statuses observed (a 500 the spec never mentions is a finding, not a miss).

## Engine

- Per-request attribution already exists: every result/sample carries `requestId`, every request its `spec_operation` (phase 1). Coverage is an **aggregation at run end**, computed beside the phase-3 deferred pass (design scenario runs: over the step results; load runs: over the full status-code counters which are NOT sampled - `report.status_codes` aggregates everything, so operation-level sent-counts and status sets come from the per-step/per-request tallies the collectors already keep, and sampling limitations apply only to body-dependent facts - state precisely in the PR which numbers are total and which sampled, per the honesty discipline).
- Tallies ride `runs.summary` (`build_run_summary_payload` / `build_scenario_summary_payload` precedents); the report renders a conditional `coverage` block via `ReportExtras` (absent for unbound collections or single-request runs - the not-measured rule). Shape: machine-readable rollup + per-operation rows (operationId/method/path, sent, statusesSeen[], declaredHit[], declaredMissed[], undeclaredSeen[]) - capped and disclosed if the spec is enormous (the caps-are-named rule).
- **Machine verdict for the future CLI gate**: the rollup includes plain numbers a gate can threshold (`operationsCovered`, `operationsTotal`, `declaredResponseCoveragePct`) - shaped like `thresholdValidation` so #473 (parked, stays parked) can adopt it without reshaping. Noted, not built.

## App

- **OverviewTab block** beside `ThresholdVerdict`/`CapacitySummary` (`OverviewTab.tsx:25-50`): the rollup as tiles + a per-operation table (uncovered operations first - they are the finding), rendered only when the report carries the block (the renders-nothing precedent).
- **ScenarioRunView**: design collection runs get the same block in their run view.
- **The Spec tab** (phase 1's surface) gains a "last run coverage" line with a link to the run - the binding page answering "when did we last exercise this contract".
- **MCP free**: `get_run_report` passes the block through verbatim; the docstring (`tools.ts:1256-1257`) names it.

## Tests

Engine: coverage computed against the stamped spec version even after the binding synced to a newer spec (the specHash pin - mutation-check it); rollup math per bucket including undeclared statuses; absent-when-unbound; per-mode (scenario design, scenario load, and absent for single-request load). App: block render states, uncovered-first ordering, Spec-tab line. Fixtures: a small bound collection + spec where one operation is deliberately never called.

## Docs, same commit

`docs/app/openapi.md` coverage section (including the total-vs-sampled honesty table); `api-reference.md` report block; `mcp.md` one line.

## Sequencing

After #626 and #627. Parallel-safe against #628 (shares the summary/report files - trivial rebase; conventions identical) and everything else. #576 shares report-shape files - same trivial rebase either order.

## Acceptance criteria

- [ ] A bound collection's run reports operation and declared-response coverage against the spec version it ran with, machine-readable, absent-not-zeros when unbound.
- [ ] Undeclared observed statuses are surfaced as findings.
- [ ] Which numbers are exact and which sample-derived is stated in the block and the docs.
- [ ] The rollup shape is adoptable by a CLI gate without reshaping (documented, not implemented).

Part of #625.

athrvk:
## Description

**Plan.** Root cause, verified against `c9cf2a5` before touching anything: phase 1 landed the per-request attribution coverage needs (`requests.spec_operation`) and the run stamp it is anchored to (`metadata.openapi`), and phase 2 is fully closed (#649/#654/#655 all merged), so the sequencing gate holds. But **one premise of this issue had gone stale, and it is the one the whole design turns on.** The issue reads coverage as "an aggregation at run end" over facts the run already has. It is not: the rollup needs the operations the *document* declares - crucially including the ones no request ever covers, which is the entire point of "14 of 18" - and `spec_documents.content` is stored **verbatim**, which is YAML as often as JSON. The engine has no YAML parser, and #625 decided it never parses OpenAPI at all.

**Approach:** the side that already parses the document writes down what it found, once, when it stores it. `spec_documents` gains an `operations` index (`{operationId?, method, path, responses[]}`), the two OpenAPI parsers emit it in the same walk that already stamps each request's identity, and the engine counts against it without ever reading the document. This is the same division of labour #628 prescribes for its response schemas, arrived at independently and landing first.

**Rejected alternative:** parsing the stored document engine-side. It works for JSON specs and silently reports *no coverage at all* for YAML ones - a feature that quietly does nothing for half its users is worse than one that says it cannot. **Also rejected:** computing coverage renderer-side, where the parser already lives. The block has to ride `runs.summary` so `get_run_report` and a future CLI gate (#473) get it for free, and a renderer-side figure could not be pinned to the document the run was *planned* with.

### Why the numbers are exact, and why that is a design constraint

Coverage is counted as each request is sent and each response arrives, through a tally that is **not** the run's `results[]` store. Under load that store is a bounded reservoir, so coverage derived from it would report an operation as uncovered whenever the store happened to thin the only request that touched it - the failure mode would be silent and would look exactly like a real finding. This is what the acceptance criterion about total-vs-sampled resolves to: **every number in the block is exact**, stated in the card, the API reference and `docs/app/openapi.md`.

The cost is one relaxed atomic increment per completion. Identity resolution happens **once per plan step**, when the tally is built, rather than per completion - the difference between coverage a load run can afford and one it cannot.

### Blast radius, traced rather than assumed

- **Every writer of a spec document** goes through one `read_spec_operations` helper (`POST /specs`, `/import/apply`'s spec section, `/specs/sync`), so a document cannot acquire or lose an index depending on the route that stored it. Same on the app side: import, bind and sync all carry it.
- **Every reader of the new field** exists before the field does - CLAUDE.md's named "written but never read" defect. `spec_documents.operations` is read by `resolve_scenario`; `ScenarioStep::spec_operation` by the tally; `summary.coverage` by the report route; `RunReport.coverage` by `ContractCoverage`, `SpecCoverageLine` and the MCP docstring.
- Both executors share **one** tally builder (`make_coverage_tally`), so a design run and a load run of one collection cannot resolve their steps differently.

### Decisions worth reviewing

- **A status hits the most specific declared pattern that covers it** (exact, then a `2XX` range, then `default`). An operation declaring both `200` and `2XX` that only ever answered 200 reports `2XX` as **missed** - the document declared two distinct promises and the run produced one.
- **A transport error is a request sent and no response seen.** The operation counts as covered (something was tried), no declared response is hit, and the row says how many failed. Status `0` never appears as a status a server sent.
- **Absent, never zeros.** No block for an unbound collection, a single-request run, or a document stored before the index existed. A run that was never judged against a contract did not cover none of it.
- **The index is pinned at plan resolution**, and a binding whose stored hash no longer matches the document it names yields no index at all - "not measured" rather than measured against a contract the run never ran.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Re-run after merging `master` (a1764dd) in. `python build.py -e -t` then `ctest --preset linux-dev` - **1842/1842 passed**. `cd app && pnpm test` - **5039 passed across 484 files**. `pnpm type-check`, `pnpm lint` and `pnpm format:check` all clean. `mkdocs build --strict -f .github/mkdocs.yml` builds clean.

The 28 tests this branch adds were **1816/1816** against its original base `c9cf2a5`, which runs 1788; the rest of the growth to 1842 is #670/#672 arriving with the merge.

New engine coverage:

| File | What it pins |
|---|---|
| `spec_coverage_test.cpp` (15) | Pattern precedence (`200` beats `2XX`), undeclared-as-finding, uncovered-first ordering, transport errors, off-contract requests, absent-not-zeros, index parse/validate |
| `specs_route_test.cpp` (+6) | The index stored and read back through `POST /specs` and `/import/apply`; `null` not `[]` when absent; refused on every write path; **the hash pin** |
| `runs_route_test.cpp` (+3) | The block passed through verbatim; absent in both its shapes; `metadata.openapi` echoed at the route |
| `scenario_load_test.cpp` (+2), `scenario_runner_test.cpp` (+1) | Per mode: a load run counts every completion against its operation on the real event loop; an unbound run of either mode writes nothing |

**Mutation-checked**, per CLAUDE.md: dropping the `document->hash == spec_hash` check in `resolve_scenario` reddens `ABindingWhoseHashHasMovedIsNotMeasuredRatherThanMismeasured`; dropping the `exchange.sent` guard in the design runner reddens the skipped-step count; flattening the pattern precedence into one predicate reddens `AStatusHitsTheMostSpecificPatternAndLeavesTheRangeUnhit`.

**Two things worth knowing about this environment, neither caused by this change:**

- Two `HttpClientTest` cases that need a DNS lookup to *fail* (`UnreachableHostsRawRequestNamesWhatWasAskedFor`, `HandlesInvalidUrl`) each timed out once under full-suite parallel load and pass standalone in 0.02s. They reproduce on the base commit and touch no code this branch changes; the confirming full runs were green both before and after the merge.
- `clang-format` here is **18.1.3** and the tree is formatted by a newer one: untouched files (`scenario_load.cpp`, `threshold_eval.cpp`) report 22 and 14 warnings unchanged, so the counts on the new files are the same systemic line-wrapping difference rather than a violation introduced here. Same finding recorded in #670.

## Checklist

Docs in the same commit: `docs/app/openapi.md` (a Contract coverage section, including the exact-vs-sampled discussion the acceptance criteria ask for, and the "What is not here yet" list loses the half that ships), `docs/engine/api-reference.md` (`POST /specs`' `operations` with the never-derived rationale, the report's `coverage` block), `docs/engine/db-schema.md` (the column and its migration shape), `docs/engine/mcp.md`, `docs/app/COMPONENTS.md`.

## Related Issues

Closes #629.

Two notes for the phases still open:

- **#628 inherits the index.** Its "decide the shape in the PR" for storing pre-extracted response schemas is now half-decided: `spec_documents.operations` exists, is validated by one helper on all three write paths, and its rows are keyed by the same operation identity a schema map would be. Adding `responseSchemas` to those rows is the smaller change; a sibling table would be a second thing to keep in step.
- **The `metadata.openapi` echo now has its direct test**, as noted on this issue when #626 closed - `ReportEchoesTheSpecTheRunWasPlannedAgainst`, in `runs_route_test.cpp`, asserting both the bound and unbound cases.

One deliberate deviation from the issue spec: `pm.response` / script-surface access to coverage is not added, and the issue did not ask for it - noting it only because #628 names the equivalent as an explicit non-goal and the same reasoning applies.


athrvk:
Fix merge conflicts

athrvk:
Conflicts resolved and pushed (`4b3d6bc`) - `master` had moved seven commits ahead with #671, #672 and #675 landing.

**One conflict, in `engine/CMakeLists.txt`:** #672 added `tests/version_isolation_test.cpp` to the test source list and this branch added `tests/spec_coverage_test.cpp` on the same line. Both are kept, and both *have* to be - #670's registration guard is on `master` now, so an unlisted `tests/*_test.cpp` fails configure by name.

Everything else merged cleanly, but I checked the three semantic overlaps rather than trusting that:

- **#672 moved the version string out of `constants.hpp`** behind `user_agent.hpp`, so a bump no longer recompiles the tree. The coverage constants added here sit beside that without pulling `version.hpp` back in - `VersionIsolationTest` passes.
- **#672's `request_examples.body_truncated` and this branch's `spec_documents.operations`** are columns on different tables; both are `default_value` migrations, so `sync_schema` ALTERs each independently.
- **#671's importer work and this branch's declared-operation index** touch the same two OpenAPI parsers but different code paths - the index is built in the `paths` walk that already stamps identity, not in the file-intake layer #671 changed.

**Re-verified on the merge result:** engine **1842/1842**, app **5039 passed across 484 files**, `type-check` / `lint` / `format:check` clean, `mkdocs build --strict` clean. The PR body's testing section is updated to these numbers.

One note on the `CI gate` failure you may see against `ac25857`: it reads `app: success, engine: cancelled`. That is the merge push cancelling the older commit's in-flight run, not a test failure - the run for `4b3d6bc` is the live one.

---

Issue #682, opened 2026-08-16
engine/app(openapi): phase 3c - validate a load run's sampled responses at run end, never on the hot path


athrvk:
Extracted from #628 (phase 3 of #625) - the third of its three hooks. Design mode landed with phase 3a (the dependency, the stored index, the verdict shape); collection runs are #681; this is load mode.

Split for the reason the other two were: the hooks share only the verdict, and this one is the only one with a performance constraint that governs its whole design.

## What already exists

- `core/schema_validation.{hpp,cpp}` - the verdict, the matching rules, the failure cap, the dialect disclosure. Settled; this issue calls it, it does not reshape it.
- `spec_documents.response_schemas`, and the plan-time resolution coverage already performs (`resolve_scenario`).

## Scope

- **Deferred, over the sampled reservoirs, inside `run_replay`** (`run_manager.cpp:78-130`, call sites `:283/:308`) - the deferred script-replay precedent, and the same reason: the load loop refills concurrency per completion, so anything on that path costs throughput for the whole run. Validation happens once, at run end, over the samples that were kept.
- **This makes the numbers sampled, and that has to be said.** Coverage is exact and validation is not - they sit in the same report, and a reader who assumes the second is exact would read "0 failures" as "no response failed" when it means "no *sampled* response failed". `docs/app/openapi.md` already carries the exact-vs-sampled table; this adds the row.
- **Tallies ride `runs.summary`** via `build_run_summary_payload`: `samplesChecked` / `valid` / `failed` / `unevaluated`, plus capped failure examples.
- **The report block is gated by a `ReportExtras` flag** (`runs.cpp:983-992`, the `testValidation` template) - **absent** when the collection is unbound, never zeros. Mutation-check the gate: that is the assertion that reddens if absent-becomes-zeros.
- **App**: the block in the run report's `OverviewTab` beside the coverage block, on the same renders-nothing-when-not-measured terms.

## Tests

- Tallies from seeded reservoirs: all-valid, all-invalid, mixed, and one whose body is not JSON (unchecked, counted as such rather than as a failure).
- Summary/report round-trip, and the absent-when-unbound gate, mutation-checked.
- A run whose samples were thinned still reports honestly - the tallies describe the samples, and the report says so.
- **Proof the hot path is untouched**: no validation call reachable from the completion path; assert structurally (the call site is inside `run_replay`) rather than by timing, which a shared CI runner cannot measure.

## Sequencing

[...]

Issue #680, opened 2026-08-16
app(openapi): re-importing a bound spec silently makes a second collection - detect it at import time and offer the Sync flow instead


athrvk:
Carries the implementation of #677 item 2, whose decision is recorded there: option (a), detect at import time and offer to sync.

## Problem

Importing an OpenAPI document that a collection is **already bound to** creates a second collection from scratch. It is PR #665's recorded deviation, and the Sync section of the Spec tab is the documented path for exactly this - a user who re-imports gets the collection they did not want, with no line anywhere saying the first one exists.

The two then diverge, and nothing marks which is real: the bound one keeps its `spec_operation` identities, its saved examples and its coverage history, while the fresh one has none of that and looks newer.

## Approach

No new machinery is needed to notice it. `spec_documents` stores `source_url` and an engine-computed content hash, and a collection's binding already names its `specId` - so after a document is parsed in the import flow, a bound collection whose document matches by `source_url` **or** by content hash is a lookup, not an inference.

On a match, offer one dialog rather than importing: "this spec is bound to `<collection>` - sync instead?", with Sync routing into the Spec tab's Sync section for that collection and the other choice importing as it does today (a user who deliberately wants a second copy is a real case and must not be blocked).

Match by both keys, not one: a spec re-fetched from the same URL whose bytes have moved is the *most* important case to catch (it is precisely what Sync exists for), and a spec pasted or picked from a file has no URL to match on but the same bytes.

## App changes (required)

- Detection in the import flow (`modules/collections/ImportModal.tsx` and the spec path around `readSpecOperations`), reading the stored bindings and documents rather than re-deriving anything.
- The dialog, built from the existing primitives - see `app/CLAUDE.md` on surface and border rules, and `ImportModal.surface-rule.test.tsx` for the guard the dialog family already carries.
- Routing to the bound collection's Spec tab Sync section on the Sync choice.

## Acceptance criteria

- [ ] Re-importing a document bound to a collection, matched by `source_url` **or** content hash, offers Sync instead of importing straight through.
- [ ] Choosing to import anyway still imports, unchanged - the dialog is a fork, never a block.
- [ ] An unbound document, and a document bound to nothing this database knows, import exactly as today with no dialog.
- [ ] Tests cover all three outcomes (bound and identical, bound and moved, not bound), mutation-checked: remove the detection and the bound cases import silently again.
- [ ] `docs/app/openapi.md` and the import pages describe the fork, replacing the current silence.

## Sequencing

[...]

PR #687, opened 2026-08-16
feat(engine/app/openapi): per-step schema verdicts in collection runs

### LINKED ISSUE #681: engine/app(openapi): phase 3b - a per-step schema verdict in collection runs, and the opt-in that can fail one
Extracted from #628 (phase 3 of #625), which carries three hooks - design mode, collection runs, load mode - plus the validator dependency and five app surfaces. The first of those landed with the dependency, the stored index and the design-mode verdict; this is the second, and #TBD-3c is the third.

Split for the reason #626 became #637/#638 and #627 became #649/#654/#655: the three hooks share nothing but the verdict they carry, and one change holding all of them cannot be reviewed against what each run mode actually does with it.

## What already exists

Verified at the phase-3a commit before writing:

- `core/schema_validation.{hpp,cpp}` - `ResponseSchemaIndex::check(spec_operation, status, contentType, body)` returns a `ValidationVerdict`, and `build_validation_payload` renders the node clients read. **Nothing here needs re-deciding**: matching, the failure cap, the dialect disclosure and the reason codes are settled.
- `spec_documents.response_schemas` - the stored index, written by all three spec write paths through `read_spec_indexes`.
- `resolve_design_schema_index` / `validate_design_response` (specs.cpp) - the design-mode resolution, walking a request's collection ancestry for the nearest binding.
- App: `ResponseValidation` on `ResponseState`, the `ValidationChip` and the Tests-tab `SchemaValidation` section.

## Scope

- **Per-step verdicts.** A scenario run's steps carry `spec_operation` already (`stamp_step_identity`, `scenario_runner.cpp:125-136`), and the run resolves its bound document when the plan resolves (`resolve_scenario`, which reads `operations` for coverage). Read `response_schemas` in the *same* place and for the same reason - the run must be judged against the document it was planned against, not whatever the binding points at when it ends - and stamp a verdict per step, on the same terms coverage is counted.
- **The SSE step frame carries it** (`:251-263`), so a run being watched shows verdicts as they happen rather than only in the report.
- **`failOnSchemaError`, default false.** A schema failure is its own verdict channel and does **not** change a step's outcome classification by default; the flag is per-run, recorded in the run config, and validated like every other run field. Both ways tested.
- **App surfaces**: the per-step verdict in `ScenarioStepCard`, and a block in `OverviewTab` beside `ThresholdVerdict` - which already models rendering nothing when not configured, and is the precedent to follow rather than a new pattern.
- **MCP**: `run_collection_smoke` folds the verdict into its per-row `ok` the way `testResults` folds (`tools.ts:1917-1931`) **and its declared `outputSchema` gains the field** (`:1851`) - the declared-schema-must-be-updated trap. Docstrings and `docs/engine/mcp.md` in the same commit.

## Tests

- Per-step stamping: a step whose response matches, one that does not, one whose operation the document does not declare.
- `failOnSchemaError` both ways, mutation-checked: with it off, a failing schema leaves the step outcome exactly as it was.
- The SSE frame carries the verdict, and a run of an unbound collection carries no verdict key at all - absent, never an empty object.
- `run_collection_smoke`'s output validates against its own declared schema (the trap above is only caught by asserting that).

## Docs, same commit

`docs/app/openapi.md` (the collection-run half of the validation section - this PR's "what is not here yet" list loses it), `docs/engine/api-reference.md` (the step frame, the run flag, the report block), `docs/engine/mcp.md`, `docs/app/COMPONENTS.md` if the step card's paragraph names its sections.

## Sequencing

Wave 3 of #625, after phase 3a (landed). Independent of #TBD-3c - they touch different run modes and share only the core. Parallel-safe against everything outside the epic; `scenario_runner.cpp` is shared with #576's landed work, rebase-level only.

## Acceptance criteria

- [ ] A collection run of a bound collection carries a per-step verdict, live on the SSE frame and in the stored report.
- [ ] A schema failure does not change a step's outcome unless `failOnSchemaError` is set; mutation-checked.
- [ ] An unbound collection's run carries no verdict anywhere - absent, never zeros or empty objects.
- [ ] `run_collection_smoke`'s declared outputSchema matches what it returns.
- [ ] Docs land same-commit.

Part of #625; extracted from #628.

athrvk:
## Description

Phase 3b of #625, extracted from #628. Phase 3a landed the validator, the stored response-schema index and the design-mode verdict; this carries the same verdict through a collection run.

> **Rebased onto #682.** Phase 3c and #680 landed while this was open. 3c is the substantial overlap - it shipped its own `schemaValidation` block, its own tally struct and its own component - so this branch **dropped its parallel versions and built on 3c's**. See *Reconciling with #682* below; the merge commit carries the same reasoning.

**Plan.** The root cause is that the collection runner had no access to the bound document's response schemas at all - `SpecBinding` carried `declared_operations` for coverage and nothing else, so a run that could count which operations it touched could not say whether what came back matched them. The approach mirrors the decision `spec_coverage.hpp` already records: read the schemas **once, when the plan resolves**, carry them on `SpecBinding`, and check per step - so a run is judged against the document it was *planned* against and a sync landing mid-run cannot rewrite it. The alternative I rejected was reusing the design-mode hook (`validate_design_response`) per step: it re-walks the collection ancestry, re-reads the document and re-parses the whole schema index on every response, which for a 200-step run is 200 parses of the largest column in the schema - and is free to return a different answer partway through a run.

### Engine

- `SpecBinding` keeps #682's `response_schemas` **text**; the collection runner parses it once before its first send, so the load pass stays lazy and neither mode reparses per response.
- `SpecBinding::schema_reason` is added beside it, and **only the collection runner reads it**. Coverage's one spelling of "not measured" is an absent block, and that is also 3c's whole answer; a per-step verdict needs more, because `checked: false` on a step has to name something a reader can act on - "the document has moved under the binding" is fixed by a sync, and a chip that silently never appears is how they never learn of it. Unbound produces nothing on either side.
- Each step that sent is stamped with a verdict on its stored trace **and** on the live `step` SSE frame - the same object from the same `build_validation_payload`, so a run watched live and the same run read back cannot disagree. A step that sent nothing (skipped, or a `{{data.column}}` that would not bind) carries no verdict at all: a response nobody made was not judged against a contract.
- **`failOnSchemaError`**, top-level on `POST /runs`, default `false`, type-checked by `validate_run_config` before the run row exists. Off, a schema verdict changes no step outcome - it is its own channel. On, only a step that passed everything else is failed by it; a step already failing an assertion keeps the error that named it.
- Coverage behaviour is byte-for-byte unchanged: `declared_operations` is still populated only when the document exists **and** its hash matches.

### App

- The shared `ValidationChip` in the step row (beside the outcome chip, never folded into it) and the Tests-tab `SchemaValidation` section in its expansion - both reused rather than re-laid-out, so the three-state wording and the unevaluated-keyword disclosure cannot drift from the response pane's.
- #682's `SampledSchemaValidation` rollup now also renders for a collection run, in the history Overview and the collection run's own view.
- `run_collection_smoke` folds a checked-and-failed verdict into its per-row `ok` the way `testResults` folds, and its declared `outputSchema` gains the field. `checked: false` is reported and never counted against the run.

### Reconciling with #682

3c owns the `schemaValidation` shape, and its version is richer than the one this branch had written (per-reason unchecked accounting, named unevaluated keywords, bounded failure examples). So:

- **`ValidationTally` and `build_validation_summary_payload` are deleted.** The collection runner records into `SampledValidationTotals` and stores through `build_sampled_validation_payload`. One shape, one component - "how many responses failed their schema" cannot come to mean two things.
- **`SchemaVerdict.tsx` and its test are deleted** in favour of 3c's component, whose doc comment now describes both callers.
- **The denominator genuinely differs, and now says so.** A load run walks a bounded reservoir; a collection run checks every step. The collection run writes `exact: true` and the shared component picks its scope sentence from it. Absent reads as sampled - that is what a report written before the flag was, and overclaiming exactness is the worse error of the two. Both branches are asserted in `SampledSchemaValidation.test.tsx`.

**One change to #682's own test, and it needs a reviewer's eye.** `SchemaValidationHotPathTest.NoValidationIsReachableFromTheCompletionPath` listed `scenario_runner.cpp` among the completion-path files, so this branch failed it. That listing was wrong: `scenario_runner.cpp` is the design-mode **sequential** runner - one request at a time, nothing refilling concurrency behind it - which is exactly why 3b can check every step where 3c must defer. The load-mode scenario executor is `scenario_load.cpp` and it stays listed, as do `load_strategy.cpp` and `metrics_collector.cpp`. The guard's expected-callers list gains `scenario_runner.cpp`, which is that list doing precisely its job: it caught the new hook and sent someone to decide rather than letting a reviewer maybe notice.

- **The rollup is in `ScenarioRunView` as well as `OverviewTab`.** The issue names `OverviewTab` only, but `ContractCoverage` - the block this sits directly under, and half of the same question - already renders in both. A rollup present in one surface and not the other reads as a missing block rather than a scoping decision.
- **A stored step's verdict is read through `responseFromRunResult`, not onto `ScenarioStepRow`.** My first pass filled `ScenarioStepRow.validation` from `trace.validation` via `validationFromTrace`. That turned a type-only import into a runtime one, pulling `restore-response.ts` and the whole response-viewer barrel into the step list's module graph - which broke `timing-phases.test.tsx` and `execute-stream-branch.test.tsx` with `vi.mock` hoisting errors. The row now carries the **live** half only; the card takes `step.validation ?? response?.validation`, so `trace.validation` still has exactly one reader and it is the funnel the response itself comes from. `ScenarioStepCard.validation.test.tsx` asserts both paths render the same verdict.

## Type of Change
- [x] New feature
- [x] Documentation update

## Testing

All green locally on `linux-dev`, **after** the merge with master.

- **Engine:** `python build.py -e -t && ctest --preset linux-dev` - **1899/1899 passed, 0 failed.**
- **App:** `pnpm test` - **492 files, 5120 tests passed**; `pnpm type-check`, `pnpm lint` and `prettier --check` on every touched file clean.
- New engine coverage: per-step stamping (matched / failed / an operation the document does not declare), the unbound run carrying no verdict anywhere, `no_index` and `hash_mismatch` said rather than left silent, the skipped step carrying none, the run tally including its `exact` flag and per-reason accounting, the SSE frame's absent-vs-present verdict, `read_fail_on_schema_error`, and the summary section's absent-when-nothing-judged gate.
- **Mutation-checked:** removing `fail_on_schema_error &&` from the gate reddens `ScenarioRunnerTest.ASchemaFailureDoesNotChangeAStepsOutcomeByDefault` and nothing else; verified by building the mutation and restoring it.
- C++ formatting checked per file against master's own baseline, so no unrelated reflow rides along.
- One flake seen and dismissed: `HttpClientTest.UnnegotiatedConnectionReportsEmptyHttpVersion` timed out once while the app suite ran concurrently, and passes in 0.02s on its own and in every clean full run.
- Baseline: `python build.py -e -t` was run against `origin/master` first; no pre-existing failures. (Needs `vcpkg-fix-port valijson` in this environment, per `engine/CLAUDE.md`.)

## Checklist

## Related Issues

Closes #681

Part of #625; extracted from #628. With #682 already merged, this is the last of phase 3's three hooks - #628 should close once this lands.


athrvk:
fix merge conflict

Issue #760, opened 2026-08-17
app(mcp): load-run knob parity (recording, redirects, comment), OAuth2 token tools, scripting types resource, and the capacity omission in suggest_load_profile


athrvk:
Part of epic #753 (S6). A batch of smaller parity items sharing the start_load_run region and the read surface.

## Problem (verified at 46b115d)

1. **Recording knobs**: the UI's load dialog sets `success_sample_rate`, `slow_threshold_ms`, `save_timing_breakdown`, and `comment` on the run payload (`app/src/types/api.ts` StartRunRequest ~:518-:562), plus `followRedirects`/`maxRedirects`. `start_load_run` (`app/electron/mcp/tools.ts:2062`) exposes none of them - an agent cannot ask for slow-request capture or stamp a run comment, and cannot control redirect policy on an ad-hoc load target. (No per-run timeout exists engine-side - the UI's config strip shows the config snapshot - so timeout is NOT a gap; recorded here so nobody re-derives it.)
2. **OAuth2 tokens**: the UI manages tokens explicitly (fetch/status/clear, `engine/src/http/routes/oauth.cpp:143/:213/:220`); MCP passes an oauth2 auth block but an agent cannot pre-fetch or clear a cached token, so an expired cache mid-flow is undiagnosable. The interactive authorize flow (browser loopback) stays out of scope - agent-driven browser auth is not a parity target; refusals should name the app as the place to authorize.
3. **Scripting types**: `GET /scripting/types` (`engine/src/http/routes/scripting.cpp:1468`) serves the sandbox's TypeScript definitions; MCP serves only completions (`vayu://scripting/completions`). Agents writing test scripts benefit from the real d.ts.
4. **Prompt drift (small bug)**: `suggest_load_profile` enumerates four modes and omits `capacity` (`app/electron/mcp/prompts.ts:139-140`), which `start_load_run` supports - the prompt steers agents away from a mode the tool has.

## Fix pathway

1. Extend `start_load_run` inputs: `successSampleRate` (1-100), `slowRequestThresholdMs`, `saveTimingBreakdown`, `comment`, `followRedirects`, `maxRedirects` - mapped to the payload field names above (snake_case where the engine expects it; follow the existing mapping table in the handler). No new gating: none of these change blast radius.
[...]

Issue #753, opened 2026-08-17
epic: MCP feature parity with the app - an agent should be able to do (almost) everything the UI can


athrvk:
## Context

Owner-requested parity audit (2026-08-17, verified at master 46b115d): compare the MCP surface against the full UI feature surface and close the gap. The MCP server (PR #56) was scoped as "inspect collections/runs/config, send requests, run/compare load tests, read live metrics, and create/update saved data". Since then, releases 0.16.0-0.18.1 shipped four major feature stacks the MCP surface never followed: services (webhook inboxes, mock servers), scenario/collection runs with data files, OpenAPI binding/sync/export, and richer run housekeeping (baselines, samples, time series). The result today: 24 tools that cover single-target execution and basic CRUD well, and several entire subsystems with zero MCP coverage.

## Verified coverage matrix (at 46b115d)

Covered well: single-request execution incl. SSE, ad-hoc scripts, all 7 body modes, auth passthrough (`run_request`); collection smoke (`run_collection_smoke`, direct requests only); load runs incl. capacity mode, thresholds, monitor vitals, SSE-under-load (`start_load_run`); run reports and bounded live metrics; baseline-aware `compare_runs`; basic collection/request CRUD; environment variable updates (metadata-preserving); engine config read/write; OAuth mock issuer lifecycle.

Zero or partial coverage (each owned by a sub-issue below):

| Gap | UI has | MCP has | Sub |
|---|---|---|---|
| Scenario/collection runs (design + load), data rows, flow control | Run Collection dialog, full runner | smoke only; scenario runs explicitly scoped out in #454 | #754 |
| Run housekeeping: list filters/pagination/search, delete, baseline pin, samples, time series, monitor series | full History | first 100 runs, report only; baseline read-only | #755 |
| Webhook inboxes | full lifecycle + captures + canned responses | nothing | #756 |
| Mock servers; issuer live-edit | start/stop/routes; live failure-mode edit | issuer start/list/stop only | #757 |
| Environments create/delete/activate, per-variable flags, globals, cookie jar | full | update_environment only | #758 |
| Saved-request auth/settings, examples CRUD, collection auth/variables/scripts, move/reorder | full | name/url/method/headers/body/scripts only | #759 |
| Load-run recording knobs (sampling, slow capture, timing breakdown, comment), redirect policy, OAuth2 token ops, scripting types resource | full | none of these params | #760 |
[...]

Issue #454, opened 2026-08-10
app(mcp): two loose ends from the 08-09/10 wave - start_load_run says nothing about scenario runs, and #417's promised graphql run_request test was never added


athrvk:
From the 2026-08-10 closure verification of #357 and #417 (verified at master `054059a`). Two small MCP-surface items that their parent issues required and their closing PRs did not deliver. Both are five-minute fixes batched here because they touch the same surface (`app/electron/mcp`); neither affects behaviour today, so this is hygiene, not urgent.

## Problem

1. **`start_load_run` is silent about scenario runs.** #357's App-changes section required: MCP's `start_load_run` "cannot start a scenario load run - say so in its description rather than leaving an agent to discover it". The shipped description (`app/electron/mcp/tools.ts:1722-1726`) covers URL/requestId targets only. Verified: `rg -i scenario app/electron/mcp/tools.ts` returns zero hits. An agent reading the tool list has no way to know collection/scenario load runs exist but are app-dialog-only.
2. **The #417 MCP integration test is missing.** #417's Tests section promised an MCP-level case: "run_request with bodyType graphql and a bare query succeeds against the mock server - the exact owner scenario". Verified: `rg -i graphql` matches nothing in `app/electron/mcp/tools.test.ts` (whose run_request cases at :1183-1280 cover form/script only) nor in `protocol.integration.test.ts`. The behaviour itself is pinned engine-side (`engine/tests/graphql_body_test.cpp:155-161` asserts the exact MCP payload shape on the wire), so this is a test-scope gap, not a behaviour gap - but the MCP layer's own serialization path between tool input and engine payload is what that engine test cannot cover.

## Fix / solution pathway

1. Extend the `start_load_run` description string (`tools.ts:1722-1726`) with one sentence: scenario/collection load runs are started from the app's Run Collection dialog and are not reachable through this tool. Keep it to the description - no schema change, no new tool.
2. Add one test beside the existing run_request cases in `tools.test.ts`: `bodyType: "graphql"` with a bare query (no envelope), asserting the engine-bound payload carries the body as stored and the response round-trips - mirror the shape of the adjacent form-body case at `:1183-1280`. If the harness stubs the engine, assert on the captured `/execute` payload (that is the layer the engine test cannot see).

## App changes

`app/electron/mcp/tools.ts` (description string), `app/electron/mcp/tools.test.ts`. Engine: none. Renderer: none, verified.

## Tests

[...]

PR #765, opened 2026-08-17
feat(app/mcp): run a collection from MCP, in design mode and under load

### LINKED ISSUE #754: app(mcp): scenario runs - run a collection (design mode and load mode, with data rows) from MCP, reversing #454's deferral
Part of epic #753 (S1). This reverses a recorded deferral: #454 (closed) scoped scenario load runs out of `start_load_run` with a description sentence ("started from the app's Run Collection dialog and cannot be started from here", `app/electron/mcp/tools.ts` ~2067, `docs/engine/mcp.md` ~500-506). The engine side is complete and the UI uses it daily; the MCP layer just never followed.

An agent can smoke a collection serially (`run_collection_smoke`) but cannot run it as the product means collections to be run: the design-mode runner (ordered steps, iterations, `pm.execution` flow control, per-step traces) and scenario load runs (VUs against the sequence) are UI-only. Everything rides one engine route with no file I/O:

- `POST /runs` with a `scenario` block = scenario run (`engine/src/http/routes/execution.cpp:1540`; handler ~1280-1538).
- Scenario schema (`engine/src/core/scenario_plan.cpp:166`): `source: "collection"` (required), `collectionId` (required), `recursive` (bool), `data` (inline array of row objects - capped by `maxScenarioDataRows`/`maxScenarioDataBytes`, never persisted), `iterations` (default = data row count, else 1). There is no step list: the collection tree IS the sequence.
- Load mode = scenario block + non-empty `mode` (`engine/src/core/scenario_load.cpp:41`). Allowed modes exactly `constant_concurrency`, `ramp_up`, `iterations`; `constant_rps`, `capacity`, and any `rps`/`targetRps` &gt; 0 are refused with explanatory 400s (`scenario_load.cpp:54`).
- Design-mode step results land in the run report as `results[]` (capped 100) plus a `scenario` totals object; load-mode reports carry `scenario.steps[]` per-step breakdown (`engine/src/http/routes/runs.cpp:1038-1152`). Bodies via `GET /runs/:id/samples`.

## Fix pathway

Two additions in `app/electron/mcp/tools.ts`, following the registry pattern (category, invalidates, gating helpers, structuredResult; see any recent tool):

1. **New tool `run_collection`** (design-mode scenario; category `execute`, invalidates `["run", "cookie"]`). Inputs: `collectionId` (req), `environmentId`, `recursive` (default false), `iterations` (positive int, optional), `data` (array of flat row objects, optional). Handler: pre-flight allowlist walk exactly like smoke's (list the collection's requests - recursive when asked - compose each via `POST /compose`, refuse the whole run naming the first un-allowlisted host; reuse smoke's compose-walk helpers). Then `POST /runs` with the scenario block. Return `runId` plus how to follow (poll `get_run_report`; state that step rows are capped at 100 by the engine and how many steps the plan has).
2. **Extend `start_load_run`** with an optional `scenario: {collectionId, recursive, data, iterations}` input, mutually exclusive with `url`/`requestId` (refuse both). Schema restricts scenario mode to the three the engine accepts, with the engine's own reasoning in the refusal for `constant_rps`/`capacity`. Existing gates all apply: allowlist via the same pre-flight walk, `checkLoadCaps` (VUs against maxConcurrency, duration, iterations), `confirmDestructive`, and the caveat that pre-request scripts do not run under load.

Rows come inline from the agent (the engine never opens files - `app/src/types/api.ts` StartRunRequest.scenario comment); do not add file reading. Let the engine's own row/byte caps refuse oversize payloads; surface its 400 verbatim.

None beyond what ships automatically: new tool appears in the Settings MCP panel via the existing category grouping; `run` invalidation entity already exists. Verified: no renderer enum change needed for these two tools.

## Tests

`app/electron/mcp/tools.test.ts` (fake engine client): scenario payload forwarded intact (rows, recursive, iterations); allowlist walk refuses on first bad host and nothing is started; mutual-exclusion refusals (url+scenario, both-ID); load caps applied to scenario VUs/duration/iterations; mode refusals for constant_rps/capacity name the engine's reason; confirm flow on the load variant. Run with `cd app &amp;&amp; pnpm test`, plus `pnpm lint` and `pnpm format:check` (CI-enforced).

## Docs - same commit

`docs/engine/mcp.md`: replace the #454 "cannot be started from here" sentences (both the tool table row and the Scenario runs section) with the new surface; add `run_collection` to the tool table with its engine routes (`GET /requests`, `POST /compose` xN, `POST /runs`).

## Sequencing

**S1 - first in wave M-parity** (epic #753). Lands before S6 (#760), which also edits `start_load_run`'s schema. Parallel-safe against everything outside the epic; the engine files named here are read-only references (no engine change).

**Coordinate with #767** (filed separately, not part of this epic): `get_run_report`'s MCP handler forwards the engine's report verbatim with no bound on embedded response bodies - reproduced live at up to ~1.3MB of tool-result text from a single ordinary request. Engine-side, `engine/src/core/scenario_runner.cpp:678-689` builds each scenario step's trace with the same `build_result_trace` + `cap_trace_bodies` pair `record_design_result` uses for a single `/execute` - so once this issue's `run_collection` ships, a scenario report's `results[]` will carry that same unbounded-in-practice body once per step, up to `maxScenarioSteps` (200 by default) times per report. #767 should land before or alongside this issue so `run_collection`'s reports are bounded from the start rather than needing a follow-up.

## Acceptance criteria

- [ ] An agent can run a collection in design mode with inline data rows and read per-step outcomes from the report.
- [ ] An agent can start a scenario load run in the three engine-supported modes; the other two refuse with the engine's reasoning.
- [ ] No scenario request is sent to an un-allowlisted host; the refusal names the host and the step.
- [ ] Caps, confirmation, and write-toggle semantics unchanged for existing tools.
- [ ] mcp.md no longer says scenario runs cannot be started from MCP.

---

athrvk:
## Description

**Plan.** Root cause: the engine has taken a `scenario` block on `POST /runs` since #357 and the Run Collection dialog uses it daily, but the MCP layer stopped at `run_collection_smoke` - #454 scoped scenario runs out with a sentence in the tool description. Nothing engine-side is missing, so the fix is entirely MCP-layer plumbing: a new `run_collection` tool posting the block with **no** `mode` (which is what selects the design-mode runner) and an optional `scenario` argument on `start_load_run` posting it **with** one (virtual users). The alternative I rejected was a single tool with a `loadTest` flag: the two shapes accept disjoint argument sets - a design run reads `scenario.iterations` and no load fields, a load run reads `concurrency`/`duration` and never reads `scenario.iterations` - so one schema would have had to accept every field and refuse most of them per branch, which is exactly the "argument written and never read" shape the repo keeps finding. Verified against master before writing: the engine files the issue anchors to are unchanged, and `POST /compose` by id is the same call `resolve_scenario` makes per step, which is what makes the pre-flight gate read the URL the run would actually send to.

Both tools carry the pre-flight walk the issue specifies: list the collection's requests (recursive when asked), compose each by id, gate the resolved URL. Unlike the smoke matrix, which skips one request and runs the rest, a scenario is one run - so a single un-allowlisted step refuses the whole run with nothing sent, naming the step the way the engine names it (`step 1 (request 'checkout', id 'r2')`). The recursive walk mirrors `collect_requests` - each subtree before that level's own requests, same cycle guard - so the step it names is the step the run would have reached first.

Every argument that describes a *single* target (`url`, `requestId`, `method`, `headers`, `body`, `auth`, `httpVersion`, `postRequestScript`, `collectionId`) is refused **by name** beside a `scenario`, as are `maxInFlight`, `sloMs`/`stepDuration` and `stream` with its two caps. Each was traced to the engine first: `maxInFlight` is warned about and ignored in `scenario_load.cpp`, `sloMs`/`stepDuration` belong to `capacity` (which a scenario cannot use), and `request.stream_bounds` is attached inside `if (!context->scenario)` in `run_manager.cpp`. Silently forwarding them would be the codebase's most-repeated defect arriving through the front door.

**One deliberate deviation from the issue spec.** The issue lists `iterations` inside `start_load_run`'s scenario block; it is not there. A scenario *load* run reads the **top-level** `iterations` (`scenario_load.cpp:251`, `config.value("iterations", 1000)`), while `scenario.iterations` is the design runner's per-run pass count that the load executor never reads - so offering it there would be an argument an agent believes bounds the run that nothing looks at. `run_collection` offers it, the top-level one is unchanged for iterations mode, and both are documented.

**Deferred, filed rather than dropped:** `failOnSchemaError` (the contract-as-gate flag the Run Collection dialog exposes for design-mode runs) is outside this issue's stated input list, so `run_collection` does not take it and the engine's default (off) applies. Filed as #766 with its sequencing.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Testing

App-only change (no engine source touched), so the verification is the app half of the issue's stated commands:

- `pnpm test` - **5216 passed, 498 files**, of which `electron/mcp/tools.test.ts` is 178 (23 new). New coverage: scenario payload forwarded intact (rows, recursion, iterations); design mode sends no `mode`; iterations/data omitted so the engine keeps its own defaults; first un-allowlisted step refuses the run with nothing started and the walk stopping there; a step that will not compose refuses it; recursive order matches the engine; a `parentId` cycle terminates; run id returned with plan size and the 100-row report bound; the engine's own 400 surfaced verbatim; mutual-exclusion refusals across nine single-target fields; capacity/constant_rps/targetRps refusals carrying the engine's reasoning; caps applied to VUs, duration and iterations; capped duration injected when omitted; preview-then-confirm and declined elicitation both starting nothing; the scenario block offering no `iterations` of its own.
- `pnpm type-check`, `pnpm lint`, `pnpm format:check` - clean (only files touched here were formatted).
- Mutation-checked the two subtlest guards: reversing the walk's emit order fails the ordering test (`['root_a','child_a','grand_a']` vs expected), and stubbing the allowlist gate to always-pass fails the refusal test. Both restored.
- `data-changed.test.ts`'s registry guard needed `run_collection: ["run", "cookie"]` added - that is the guard working as designed (the design-mode runner is the one executor handed the cookie jar).
- `mkdocs build --strict` not run: mkdocs is not installed in this environment. The docs change adds no page, link or heading - one table row and prose inside existing sections.

## Checklist
- [x] I have added tests

Docs updated in the same commit: `docs/engine/mcp.md` (tool table row, `start_load_run`'s route/mode cell, the scenario section that said this could not be started from MCP, the data-rows note, the cookie-jar paragraph - `run_collection` shares the jar, a scenario load run gives each VU its own), plus the traffic-sending tool enumerations in `SECURITY.md` and `config.ts`'s `allowWrites` comment.

## Related Issues
Closes #754 · part of epic #753 (S1, landing before #760, which also edits `start_load_run`'s schema) · defers #766

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -29,10 +29,10 @@ ships with safe-by-default guardrails. See `docs/engine/mcp.md` for the design.
   `enableDnsRebindingProtection` + `allowedHosts`), so a malicious web page
   cannot drive the endpoint through a forged `Host`.
 - **Target allowlist (empty by default).** The traffic-sending tools
-  (`run_request`, `run_collection_smoke`, `start_load_run`) refuse any host that
-  is not explicitly allowlisted. The check runs against the **resolved** host
-  (after `{{variable}}` substitution), so a variable-built URL cannot slip past
-  it. A fresh install cannot be used to send traffic anywhere until the user opts
+  (`run_request`, `run_collection_smoke`, `run_collection`, `start_load_run`)
+  refuse any host that is not explicitly allowlisted. The check runs against the
+  **resolved** host (after `{{variable}}` substitution), so a variable-built URL
+  cannot slip past it. A fresh install cannot be used to send traffic anywhere until the user opts
   in per host.
 - **Hard load caps (enforcement).** `start_load_run` rejects requests whose RPS,
   concurrency, or duration exceed configured ceilings. Together with the
@@ -46,8 +46,8 @@ ships with safe-by-default guardrails. See `docs/engine/mcp.md` for the design.
 - **Data writes off by default.** The data-mutating tools (`create_request`,
   `update_environment`, `update_engine_config`) are refused unless the user
   enables write access in Settings. Traffic-sending tools (`run_request`,
-  `run_collection_smoke`) and load runs are not affected by this toggle - they
-  are governed by the allowlist and caps.
+  `run_collection_smoke`, `run_collection`) and load runs are not affected by
+  this toggle - they are governed by the allowlist and caps.
 - **Per-tool control.** Any tool (or a whole read/execute/write/load category) can
   be switched off; a disabled tool is removed from `tools/list` and rejected by
   `tools/call`.
--- a/app/electron/mcp/config.ts
+++ b/app/electron/mcp/config.ts
@@ -51,10 +51,10 @@ export interface McpSafetyConfig {
 	 * Gates every tool in the `write` category - the collection and saved-request
 	 * CRUD verbs, `update_environment` and `update_engine_config`. When false
 	 * (default), those tools refuse; the two deletes additionally require
-	 * confirmation even with it on. It does
-	 * **not** gate traffic-sending tools (`run_request`, `run_collection_smoke`)
-	 * or load runs - those are governed by the allowlist, the hard caps, and the
-	 * load-run confirmation gate independently.
+	 * confirmation even with it on. It does **not** gate traffic-sending tools
+	 * (`run_request`, `run_collection_smoke`, `run_collection`) or load runs -
+	 * those are governed by the allowlist, the hard caps, and the load-run
+	 * confirmation gate independently.
 	 */
 	allowWrites: boolean;
 	/**
--- a/app/electron/mcp/data-changed.test.ts
+++ b/app/electron/mcp/data-changed.test.ts
@@ -97,6 +97,9 @@ describe("the registry declares its effects", () => {
 			update_engine_config: ["config"],
 			run_request: ["run", "cookie"],
 			run_collection_smoke: ["run", "cookie"],
+			// The design-mode runner is the one executor handed the cookie jar
+			// (`start_scenario_run`), so its steps refill it the way a Send does.
+			run_collection: ["run", "cookie"],
 			start_load_run: ["run"],
 			stop_run: ["run"],
 		};
--- a/app/electron/mcp/tools.test.ts
+++ b/app/electron/mcp/tools.test.ts
@@ -1304,6 +1304,452 @@ describe("run_collection_smoke", () => {
 	});
 });
 
+/**
+ * Scenario runs from MCP (issue #754), reversing #454's deferral. Two surfaces
+ * over one engine route: `run_collection` posts a `scenario` block with no
+ * `mode` (design-mode runner), `start_load_run` posts the same block *with* one
+ * (virtual users). What these tests own is the same thing every other tool test
+ * owns - what reaches `POST /runs`, and that no traffic is arranged before the
+ * gates have run. The plan resolution itself is engine-side.
+ */
+describe("run_collection", () => {
+	/** A client whose collection `c1` holds two allowlisted requests. */
+	function scenarioClient(overrides: Partial<Record<keyof EngineClient, unknown>> = {}) {
+		return fakeClient({
+			listRequests: vi.fn().mockResolvedValue([
+				{ id: "r1", name: "login" },
+				{ id: "r2", name: "checkout" },
+			]),
+			composeRequest: vi
+				.fn()
+				.mockImplementation(({ requestId }: { requestId: string }) =>
+					Promise.resolve({ method: "GET", url: `https://api.example.com/${requestId}` })
+				),
+			...overrides,
+		});
+	}
+
+	test("is an execute tool that invalidates runs and the cookie jar", () => {
+		const tool = TOOLS.find((t) => t.name === "run_collection");
+		expect(tool?.category).toBe("execute");
+		// Steps share the environment's jar, the way a Send does.
+		expect(tool?.invalidates).toEqual(["run", "cookie"]);
+	});
+
+	test("posts the scenario block with rows, recursion and iterations intact", async () => {
+		const client = scenarioClient();
+		const rows = [{ id: "1" }, { id: "2" }];
+		const res = await dispatchTool(
+			"run_collection",
+			{
+				collectionId: "c1",
+				environmentId: "env_1",
+				recursive: false,
+				iterations: 3,
+				data: rows,
+			},
+			ctxWith(client, { allowlist: ["api.example.com"] })
+		);
+		expect(res.isError).toBeFalsy();
+		expect(client.startRun).toHaveBeenCalledTimes(1);
+		const payload = (client.startRun as ReturnType<typeof vi.fn>).mock.calls[0][0];
+		expect(payload).toEqual({
+			scenario: {
+				source: "collection",
+				collectionId: "c1",
+				recursive: false,
+				iterations: 3,
+				data: rows,
+			},
+			environmentId: "env_1",
+		});
+		// The absence of `mode` is what makes this a design-mode run: a mode
+		// beside the block would hand the same plan to the load executor.
+		expect(payload).not.toHaveProperty("mode");
+	});
+
+	test("omits iterations and data when the caller named neither", async () => {
+		const client = scenarioClient();
+		await dispatchTool(
+			"run_collection",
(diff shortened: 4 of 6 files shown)


athrvk:
Note from the review side: #754's Sequencing section was updated (possibly after this PR was opened) with a new dependency note worth checking against this PR:

`get_run_report`'s MCP handler is a generic passthrough across every run type, and a separate issue (#767, PR #768) found its embedded response bodies unbounded on the design/load side - reproduced live at ~1.3MB from a single ordinary request. Engine-side, `engine/src/core/scenario_runner.cpp:678-689` builds each scenario step's trace with the exact same `build_result_trace` + `cap_trace_bodies` pair `record_design_result` uses, so `run_collection`'s scenario reports would inherit the same defect - multiplied per step, up to `maxScenarioSteps` (200 by default) - unless #768 already guards it generically.

#768's description suggests it does bound "every row of a multi-row report... the scenario multiplier," which would cover this PR's reports too regardless of merge order, since neither PR touches the other's code region. Flagging so it gets an explicit check rather than an assumed one - worth confirming with a test here (or in #768) that a scenario report with an oversized per-step body actually comes back bounded once both PRs are in.

athrvk:
Checked it rather than assumed it, and the check turned up two things - one that confirms your read, one that does not, plus a correction to this PR that fell out of it.

**1. #768 does cover the scenario multiplier at the node level - confirmed empirically, not by reading.** Its `boundRunReport` keys off `results[]` and `row.trace` with no branch on run type, and a scenario step's trace is the same shape by construction: `scenario_runner.cpp:678` calls the same `build_result_trace`, and `stamp_step_identity` adds only `iteration`/`stepIndex`/`stepName`/`requestId`/`outcome`/`validation` - nothing that carries a body. I merged #768 into this branch locally and ran a fixture shaped as the engine emits one (200 rows, all three body-carrying nodes populated): every row came back cut to exactly 32,768 bytes per node with `bodyTruncated`/`bodyBytes` set, `rawRequest` headers intact and step identity preserved. So no ordering dependency between the two PRs, and nothing this PR needs to add for that case.

**2. But the per-node bound does not bound the report, and the scenario case is where that starts to matter.** `results[]` is capped at 100 rows by the report route, and each row keeps up to 32 KB per node, so the ceiling is row count x nodes x 32 KB - which nothing checks. Measured on the merged branches, response body only:

| Report | Result size |
|---|---|
| 1 step x 1 MB body (#767's single-row case) | 33,365 chars - working exactly as intended |
| 100 steps x 40 KB body (each cut to 32 KB) | **3,328,454 chars** |
| 100 steps x 8 KB body (under the bound, untouched) | 845,354 chars |
| 200 rows, all three nodes populated | 19,834,336 chars |

#767's live blowout was 1,334,938 characters, so a *bounded* 100-step report is ~2.5x over the size that already failed - and at 8 KB bodies, which #768 never touches, the row count alone gets most of the way there. That is not a defect in #768 (its scope is the per-body bound and it delivers it); it is the next layer, which #754 shipping this tool is what makes reachable. Filed as **#769** with the measurements and a fix pathway (a total budget beside the per-node one, disclosed in band, non-passing rows kept first to match the engine's own thinning rule).

**3. The check found a real error in this PR, now fixed in `e3ac53f`.** `run_collection`'s follow-up note told an agent "Response bodies are not on the report - GET /runs/:id/samples has them." I had taken #754's "Bodies via `GET /runs/:id/samples`" as covering both executors; it holds for the load-mode scenario, whose results never build a trace, and is false for the design-mode runner this tool starts - those rows carry the request and response bodies inline under `trace`. An agent told bodies were absent would size a 100-step report as metadata, which is exactly the wrong expectation given the numbers above. The result and `docs/engine/mcp.md` now say what the rows actually carry, pinned by the existing follow-up test. CI was green on `c9054be` and is re-running on the correction.

---

Issue #575, opened 2026-08-14
engine/app(sse): phase 3 - integrations: pm.response.events, a budgeted MCP stream consume, curl -N import/codegen, and an SSE test fixture


athrvk:
Phase 3 of the SSE epic #572, consuming phase 1's engine surface (grounded at master `539b0a1`). Four independent integration legs, one issue because each is small and they share the wave.

## Fix / solution pathway

1. **Scripts: `pm.response.events`** - a buffered, post-close array (the sandbox is synchronous with no event loop per `ABSENT_GLOBALS`; live callbacks are impossible by design - state this in the docs). Each entry `{event, id, data}`; on a streaming run the post-request script runs after stream termination against the bounded stored list, with `pm.response.eventsTruncated`/`totalEvents` mirroring the trace markers so an assertion can refuse to run over a partial list. New members go ONLY in the `get_script_completions()` table (`scripting.cpp` - `script_types.cpp` derives the `.d.ts`; the completions guard and docs-compile corpus pick them up via one docs example). Non-stream runs: `events` is absent, not empty - `typeof` distinguishes per the `pm.request.body` precedent.
2. **MCP: `run_request` with `stream: true`** - streamed-then-buffered under a hard budget, copying `getLiveMetricsSnapshot` verbatim (AbortController budget, SSE frame parsing) against `/runs/:id/events`; result carries the events array plus in-band budget/cap disclosure per the `aaba1d8` discipline (the tool description names its own budget and limits BEFORE the payload). The allowlist gate is unchanged (resolved-URL check already runs pre-execute).
3. **Codegen + import**: the curl importer moves `-N`/`--no-buffer` out of `CURL_NOARG`'s silent-skip set (`parseCurl.ts:337-360`) into the stream flag; curl codegen emits `-N` + the Accept header for stream-flagged requests; other codegen targets emit their idiomatic streaming form or a comment stating non-support (decide per target, state it).
4. **Fixture**: an SSE endpoint in the C++ in-process test harness (phase 1 builds it - this leg extends it with scripted multi-event/slow/malformed cases shared by app-side integration tests if any run against a live engine; the Go `mock-server.go` is NOT the home - its 5s WriteTimeout kills streams, documented there).

## App changes

MCP `tools.ts` + `engine-client.ts`; `parseCurl.ts` + codegen targets; no renderer UI (phase 2 owns it).

## Tests

[...]

PR #768, opened 2026-08-17
fix(app/mcp): bound the response bodies run_request and get_run_report return

### LINKED ISSUE #767: app(mcp): run_request and get_run_report return response bodies unbounded - live-reproduced context blowout on an ordinary request
**Problem (verified live against this session's own connected Vayu MCP server, engine 0.18.1, and in code at master 46b115d)**

`run_request` and `get_run_report` are the two most heavily used tools in the MCP surface - the epic #753 coverage matrix calls single-request execution and run reports "covered well" - and both return an engine response body **completely unbounded**, with no size disclosure and no truncation. This is not a theoretical edge case: it reproduced on the *first* ordinary request tried in this session.

Live reproduction 1 - `run_request` against an ordinary page:

```
run_request({ url: "https://cloudflare.com", method: "GET" })
→ Error: result (1,334,938 characters across 48 lines) exceeds maximum allowed tokens.
```

The saved output shows the exact shape `POST /execute` returns and the MCP tool forwards verbatim:

```json
{
  "body": null,
  "bodyRaw": "<!DOCTYPE html>... (the entire page, ~1.3MB of text)",
  "bodySize": <n>,
  "headers": { ... },
  "rawRequest": "GET / HTTP/2\r\nHost: www.cloudflare.com\r\n...",
  "requestHeaders": { ... },
  "status": 200,
  "...": "..."
}
```

For a JSON API the doubling is worse - both `body` (parsed) and `bodyRaw` (raw text) carry the *same* payload, verified with a small response:

```json
run_request({ url: "https://jsonplaceholder.typicode.com/todos/5" })
→ { "body": {"completed": false, "id": 5, ...}, "bodyRaw": "{\n  \"userId\": 1,\n ...}", "bodySize": 128, ... }
```

`bodySize` is already computed and present on every response - nothing currently reads it to decide whether to include the full body.

Live reproduction 2 - `get_run_report` against a stored design-mode run of the same request:

```
get_run_report({ runId: "run_947e2a0b-..." })
→ Error: result (1,338,027 characters across 129 lines) exceeds maximum allowed tokens.
```

The saved report embeds the identical full body a second time, under `results[].trace.response.body`, and - checked directly against the saved file - carries **none** of the size-disclosure fields the engine already uses elsewhere for exactly this situation: zero occurrences of `bodySize`, `bodyBytes`, or `bodyTruncated` anywhere in the 1.3M-character report. Contrast with `GET /runs/:id/samples` (`run_samples_response`, `engine/src/http/routes/runs.cpp:594-677`), which already does this correctly for load-run captures: every sample carries `response.bodyBytes`, and a body over the run's capture budget comes back with `response.bodyTruncated: true` rather than silently full or silently absent (`runs.cpp:626,642-646`). Design-mode report bodies never go through that convention. `results[].trace.request.rawRequest` carries the same risk on the request side for a large POST body, unverified live only because the reproduction request happened to be a GET.

Verified in code, both handlers are raw passthroughs with zero post-processing:

- `run_request` (non-streaming path): `app/electron/mcp/tools.ts:1481-1483` - `return callEngine(() => ctx.client.executeRequest(payload, signal));`
- `get_run_report`: `app/electron/mcp/tools.ts:1362-1365` - `callEngine(() => ctx.client.getRunReport(requireStr(args, "runId"), signal))`

`callEngine` (`tools.ts:239-245`) is `jsonResult(await fn())` - nothing between the engine's answer and the tool result.

Confirmed **not** affected, so the fix should not touch these: `run_collection_smoke` reads only `status`/`testResults`/the schema verdict off each `/execute` response and discards the rest (`tools.ts:2028-2062`), so its per-row results stay small regardless of body size. `compare_runs` fetches two full reports internally (`tools.ts:2486-2489`) but returns only the derived `compareReports(...)` delta, never the raw bodies - it pays the cost engine-side without leaking it to the agent (a secondary memory-cost note, not part of this fix). Load-mode run reports are unaffected, and not merely by observation of one run: a load run's per-request results never go through the trace-building path at all. `build_result_trace` (`engine/src/http/request_exchange.cpp:41`) is called only from `record_design_result` (`engine/src/http/routes/execution.cpp:606`, the `/execute` completion path) and the scenario runner (`engine/src/core/scenario_runner.cpp:678`) - never from the load hot path (`load_strategy.cpp` / `scenario_load.cpp`), which is exactly the split `engine/CLAUDE.md` already documents for schema validation ("a source-scan test fails if validation reaches load_strategy.cpp / scenario_load.cpp") and holds here too. A load run's captured bodies, when capture is on, are written to the separate `result_bodies` table and reachable only through `GET /runs/:id/samples` (`run_samples_response`, `runs.cpp:594-677`) - which already truncates and discloses correctly (below) - never inlined into `/report`. So "unaffected" is a routing fact, not a config-dependent one: no capture setting makes a load report grow this node.

Root cause context, now pinned to two distinct engine code paths with two distinct (and both wrong-for-this-purpose) ceilings:

- **`run_request`'s path has no ceiling at all.** The `/execute` response `run_request` returns comes from `serialize(const Response&)` (`engine/src/utils/json.cpp:846-889`), a plain field-by-field dump - `json["bodyRaw"] = response.body;` at line 871, unconditional, no cap, no truncation call anywhere in that function. `maxResponseBodyBytes`, the config entry that *does* cap a single response, documents itself as load-only: *"Largest response body a single load-test request will read into memory... Design-mode sends are not affected."* A design-mode `/execute` reads the entire body into memory and returns every byte of it, whatever the size.
- **`get_run_report`'s path inherits a 5MB storage cap that was never meant for this.** The trace `get_run_report` reads back was capped once, at write time, by `cap_trace_bodies` (`engine/src/utils/json.cpp:263`), called from `record_design_result` right after `build_result_trace` (`execution.cpp:606-610`) at `maxTraceBodyBytes` (default 5,242,880 bytes / 5MB) - and that function only *adds* `bodyTruncated`/`bodyBytes` when a body actually exceeds that cap (confirmed: zero occurrences of either key in this session's 1.3MB reproduction, because 1.3MB is under 5MB - the disclosure never activates below the cap it guards, which is a size no tool result should reach in the first place). 5MB is sized for "a human is looking at one full trace in the UI," not for what fits in an agent's context, and the MCP layer does nothing to narrow that gap.

**Fix pathway**

In `app/electron/mcp/tools.ts`:

1. Add one bounding helper near the other result helpers (`jsonResult`/`structuredResult`/`errorResult`, `tools.ts:175-193`). Neither existing engine cap is the right number to reuse as-is - `maxResponseBodyBytes` doesn't apply to design sends at all, and `maxTraceBodyBytes`'s 5MB is a storage bound, not a context bound - so this is a new, MCP-layer-only budget, not a borrowed one:
   ```ts
   // A tool-result budget, not an engine one: neither maxResponseBodyBytes (load
   // runs only) nor maxTraceBodyBytes (5MB, sized for the DB) bounds what an agent
   // should receive inline. 32KB is picked to match the codebase's own nearest
   // precedent for "how much of a body an automated reader gets" - maxSampleBodyBytes,
   // the load-run capture default - rather than inventing an unrelated number.
   const MAX_INLINE_BODY_BYTES = 32_768;
   ```
   with a small function that truncates a string to that many UTF-8 bytes and reports whether it did, mirroring the engine's own `bodyBytes`/`bodyTruncated` naming (`runs.cpp:626,642-646`) so an agent that has seen one convention recognizes the other. (A read-time fetch of the live `maxSampleBodyBytes` value was considered and rejected: `run_request` is already latency-sensitive - it just sent a real HTTP request - and a hardcoded constant is simpler to test and reason about than a per-call config round-trip for a number that only needs to be "reasonable," not user-tunable.)
2. Apply it in `run_request`'s non-streaming return (`tools.ts:1481-1483`): after the engine answers, bound `bodyRaw` and `rawRequest`; when `bodyRaw` was truncated, drop or null the parsed `body` too rather than leaving it holding the untruncated full payload the truncated `bodyRaw` claims not to have (the JSON-doubling case above) - state which choice was made in the PR. `bodySize` already exists on the response and is exactly what a truncation flag needs to reference; use it rather than recomputing the byte length.
3. Apply it in `get_run_report` (`tools.ts:1362-1365`): map over `results[]` (design-mode only; a load-mode report has no matching field and the map is a no-op for it) and bound `trace.response.body` and `trace.request.rawRequest` on each row, adding `bodyTruncated`/`bodyBytes` the same way.
4. Tool descriptions (`tools.ts:1354-1355` for `get_run_report`, and `run_request`'s description) should state the bound and the truncation field, the way `list_runs` already states its 100-row bound and `get_live_metrics` states its tick-count bound - an agent needs to know a body might be a slice before it reasons from it as if it were the whole thing.

**Why this belongs at the MCP layer, not the engine**: the engine's per-request full-fidelity capture is correct for its actual purpose (a human clicking Send and expecting to see everything); the constraint being violated here is specific to an MCP tool result feeding an LLM context window, which is what every other size-bounded MCP read (`get_live_metrics`, `list_runs`, and `get_spec`'s planned "size-disclosed, off by default" content field in #761) already treats as an MCP-layer concern, not an engine one.

**App changes**

None. This is `app/electron/mcp/` only - no engine change, no renderer change. Verified: the renderer never consumes this shape; it talks to the engine's HTTP API directly (`app/src/services/api.ts`), and production code under `app/electron/` cannot import `app/src` (the build-boundary rule already documented in `engine/CLAUDE.md` and enforced by `app/tsconfig.node.json`), so nothing on the renderer side reads an MCP tool's JSON output.

**Tests**

`app/electron/mcp/tools.test.ts`, `dispatchTool` against a fake engine client (existing pattern, e.g. `tools.test.ts:156-246`):

- Mutation-check the bound itself: a fake `/execute` response with `bodyRaw` over `MAX_INLINE_BODY_BYTES` comes back truncated with `bodyTruncated: true` and the real `bodySize`; a response under the bound round-trips byte-for-byte unchanged (guards against over-truncating ordinary small responses, which would be its own regression).
- JSON-doubling case: a large JSON body's parsed `body` field is not left holding the full untruncated payload once `bodyRaw` has been cut.
- `get_run_report`: a design-mode fixture with an oversized `results[].trace.response.body` truncates that row; a load-mode fixture (no body field present) passes through the map untouched, asserted by exact equality against the fixture.
- Run with `cd app && pnpm test`, `pnpm lint`, `pnpm format:check` (CI-enforced per #189).

**Docs - same commit**

`docs/engine/mcp.md`: `run_request` and `get_run_report` tool-table rows/descriptions gain the bound and the truncation-flag behavior, next to the other tools that already document their bounds (`list_runs`'s 100-row cap, `get_live_metrics`'s tick count).

**Sequencing**

Independent of the MCP parity wave (epic #753, subs #754-761): this is a bug fix to two *existing* tools, not new capability. It touches the same file (`tools.ts`, `docs/engine/mcp.md`) as every sub in that wave, but a disjoint region - none of #754-761 modify `run_request` or `get_run_report`'s handler bodies (#755 extends `list_runs` and adds new samples/timeseries/monitor reads, but does not touch `get_run_report` itself). Rebase-level conflict only; safe to land at any point relative to the wave.

One forward-looking note worth recording here rather than rediscovering later: **#754 (scenario/collection runs) is the multiplier case, confirmed in code, not speculative.** `get_run_report`'s MCP handler is one generic passthrough for every run type. Engine-side, `engine/src/core/scenario_runner.cpp:678-689` builds each step's trace with the *exact same pair* this issue is fixing around - `build_result_trace` then `cap_trace_bodies` - that `record_design_result` uses for a single `/execute`. Once #754 ships `run_collection`, a scenario run's report will carry one such trace per step, capped at `maxScenarioSteps` (200 by default): the identical unbounded-in-practice body, multiplied up to 200x per report instead of the 1x reproduced here. Land this issue before or alongside #754 for that reason, even though there is no file-level dependency forcing the order - #754's own PR should not have to rediscover this.

**Acceptance criteria**

- [ ] `run_request` against `https://cloudflare.com` (or any response over the bound) returns a truncated `bodyRaw`, a truncated or dropped `body`, `bodyTruncated: true`, and the real `bodySize` - and the overall tool result stays well inside normal MCP token limits.
- [ ] `run_request` against `https://jsonplaceholder.typicode.com/todos/5` (or any response under the bound) is byte-for-byte unchanged from today.
- [ ] `get_run_report` on a design-mode run with a large stored body returns a bounded `results[].trace.response.body` with the same disclosure fields; a load-mode report is unaffected.
- [ ] Both tool descriptions state the bound.
- [ ] `pnpm lint`, `pnpm format:check`, `pnpm test` are green.

---

athrvk:
## Description

**Plan.** Root cause: both tools are raw passthroughs (`callEngine` was `jsonResult(await fn())`, nothing between the engine's answer and the agent), and *neither* engine cap covers a tool result - `maxResponseBodyBytes` documents itself as load-only ("Design-mode sends are not affected") and `maxTraceBodyBytes` is 5MB, a storage bound sized for a human opening one full trace in the UI. Between them a design send returns every byte it read, which is how an ordinary page fetch became 1.3M characters and exceeded the token limit outright. The approach is one shaping seam at the MCP layer: `callEngine` takes an optional `shape` callback, and the two handlers pass a bounding function. The alternative I rejected was capping engine-side - the engine's full-fidelity capture is correct for its actual purpose (a person clicks Send and expects to see everything), and the constraint being violated is specific to an MCP result feeding a context window, which is how every other bounded MCP read (`get_live_metrics`, `list_runs`) already treats it. I also rejected reading `maxSampleBodyBytes` from live config per call: `run_request` has just sent a real HTTP request and should not pay a second round trip for a number that only has to be reasonable, and a constant is what the tests can pin.

Verified against master (`21e5eb1`) before writing: both handlers are still the passthroughs the issue anchors to, `serialize(const Response&)` still emits `bodyRaw` unconditionally with no cap, and `/runs/:id/report` still parses `result.trace_data` into `results[].trace` for up to 100 rows.

**The bound is 32KB** - `DEFAULT_MAX_SAMPLE_BODY_BYTES`, this codebase's own answer to "how much of a body does an automated reader get", not a new number. **The disclosure is the engine's existing vocabulary**, so an agent that has read one convention recognises the other:

- `bodyTruncated: true` beside the full size - `bodySize` on a `run_request` response (already present and already correct, so it is used rather than recomputed) and `bodyBytes` on a stored trace node, the pair `cap_node_body` and `run_samples_response` use.
- **A cut `bodyRaw` nulls the parsed `body`** rather than leaving it holding in full the bytes `bodyRaw` now says were dropped - the JSON-doubling case. Null rather than deleted: `serialize(Response)` always emits that key and `null` is already its "nothing parsed here" value, so a truncated response keeps the shape every reader already parses. (The issue asked for this choice to be stated.)
- **A cut wire message keeps its header block whole**, the engine's own rule for that field (`cap_node_raw_request` - the `Cookie` line libcurl attached appears nowhere else, so a cut that ate the headers would take exactly what the reader opened it for), and is flagged `rawRequestTruncated` / `rawRequestBytes`. Unlike the engine it carries its own marker, because in an `/execute` answer `rawRequest` sits beside the *response*'s size fields, not the request body's.
- **A trace the engine already truncated keeps the size the engine recorded.** Overwriting `bodyBytes` with the length of the 5MB slice re-cut here would replace a true number with a smaller one that looks just as real.
- **Under the bound nothing is added and nothing changes** - the report is rebuilt only when a row was actually cut, so an unaffected answer comes back as the object the engine returned.

Load-run reports are unaffected by routing, not by luck: their results never go through `build_result_trace` (called only from `record_design_result` and the scenario runner, never the load hot path), so they carry no `trace` node and the walk is a no-op. Their captured bodies live behind `GET /runs/:id/samples`, which has always truncated and disclosed correctly.

**One deliberate extension beyond the issue's spec.** Step 3 names `trace.response.body` and `trace.request.rawRequest`; this also bounds `trace.request.body`. It is the same node, the same defect and the same helper, and bounding a large POST body in the wire copy while leaving it whole in the field beside it would leave the context blowout open through the one field still uncapped.

**Not touched, as the issue asks:** `run_collection_smoke` (reads only status/testResults off each response), `compare_runs` (returns a derived delta, never raw bodies), and `run_request`'s streaming path (returns events under #575's bounds, not an exchange).

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

App-only change (`app/electron/mcp/` plus one doc), so the verification is the app half of the issue's stated commands:

- `pnpm test` - **5210 passed, 498 files**, of which `electron/mcp/tools.test.ts` is **172 (13 new)**. New coverage: an oversized body cut to exactly the bound with the engine's real `bodySize` preserved; a small response byte-for-byte identical to the fixture by exact equality (over-truncating an ordinary response would be its own regression); the JSON-doubling case nulling the parsed body; a large `rawRequest` cut with its `Cookie` header intact and both flags set; a headers-only `rawRequest` untouched; a multi-byte body cut without producing a replacement glyph; every row of a multi-row report bounded, not just the first (the scenario multiplier); the engine's original `bodyBytes` preserved; the request side (`body` and `rawRequest`) bounded; a load-mode report and an under-bound design report passing through by exact equality; both descriptions stating the bound.
- `pnpm type-check`, `pnpm lint`, `pnpm format:check` - clean. Only the two TS files were formatted; `docs/engine/mcp.md` was already not prettier-clean on master and was deliberately left alone per the repo's "format only files you touched that were clean before" rule.
- **Mutation-checked** the two subtlest guards: deleting `body: null` reddens the JSON-doubling test, and replacing the `bodyBytes`-preservation guard with an unconditional overwrite reddens the already-truncated-trace test. Both restored and re-run green (172/172).
- Engine untouched, so no engine build or `ctest` run - no C++ source, test or CMake file is in the diff.
- `mkdocs build --strict` not run: mkdocs is not installed in this environment. The docs change adds no page, link or heading anchor - two table cells and one bullet inside an existing section.

## Checklist

`docs/engine/mcp.md` in the same commit: the `run_request` and `get_run_report` tool-table rows now state their bound in the Gate column, beside `list_runs`' 100-row cap and `get_live_metrics`' tick bound, and a note records the bound, the flag names, what a cut looks like on each shape, and why this belongs at the MCP layer rather than the engine.

## Related Issues
Closes #767

---
_Generated by [Claude Code](https://claude.ai/code/session_014ZokMjGaCAJxsbVHreLpxy)_

### DIFF
--- a/app/electron/mcp/tools.test.ts
+++ b/app/electron/mcp/tools.test.ts
@@ -1574,6 +1574,68 @@ describe("inline body bounds", () => {
 		expect(await runReport(reported(report))).toEqual(report);
 	});
 
+	/**
+	 * The scenario shape, pinned ahead of `run_collection` (#754 / PR #765).
+	 *
+	 * A scenario step is written through the same pair a single `/execute` uses
+	 * - `build_result_trace` then `cap_trace_bodies`
+	 * (`scenario_runner.cpp:678-689`) - into the same `results.trace_data`
+	 * column (`scenario_runner.cpp:715`) that `/runs/:id/report` parses back
+	 * into `results[].trace`. So one scenario report carries one such trace per
+	 * step, up to the report route's 100-row cap, instead of the single trace
+	 * reproduced in #767. This fixture is that shape: step identity stamped on
+	 * the trace beside the nodes, and a skipped step whose `response` the
+	 * runner erased.
+	 */
+	test("a scenario report is bounded per step, skipped steps included", async () => {
+		const out = await runReport(
+			reported({
+				summary: {},
+				scenario: { iterations: 1, steps: [{ index: 0, name: "login", executed: 1 }] },
+				results: [
+					{
+						id: 1,
+						trace: {
+							iteration: 0,
+							stepIndex: 0,
+							stepName: "login",
+							outcome: "passed",
+							request: { url: "u" },
+							response: { body: huge },
+						},
+					},
+					{
+						id: 2,
+						trace: {
+							iteration: 0,
+							stepIndex: 1,
+							stepName: "checkout",
+							outcome: "skipped",
+							// No `response` node at all: the runner erases it for a
+							// step that never sent.
+							request: { body: huge },
+						},
+					},
+				],
+			})
+		);
+
+		const rows = out.results as Array<Row & { trace: Record<string, unknown> }>;
+		expect(Buffer.byteLength(rows[0].trace.response!.body as string, "utf8")).toBe(
+			MAX_INLINE_BODY_BYTES
+		);
+		expect(rows[0].trace.response!.bodyBytes).toBe(huge.length);
+		// Identity survives the rebuild - the step list reads these.
+		expect(rows[0].trace.stepName).toBe("login");
+		// The skipped step's request is bounded, and no empty `response` is
+		// invented for it: an erased node must stay erased, or the step's
+		// expanded view reads as a server that answered with nothing.
+		expect(Buffer.byteLength(rows[1].trace.request!.body as string, "utf8")).toBe(
+			MAX_INLINE_BODY_BYTES
+		);
+		expect(rows[1].trace).not.toHaveProperty("response");
+	});
+
 	test("both tool descriptions state the bound", () => {
 		for (const name of ["run_request", "get_run_report"]) {
 			const description = TOOLS.find((t) => t.name === name)!.description;
--- a/app/electron/mcp/tools.test.ts
+++ b/app/electron/mcp/tools.test.ts
@@ -10,6 +10,7 @@ import { z } from "zod";
 import {
 	dispatchTool,
 	MAX_IN_FLIGHT_BOUND,
+	MAX_INLINE_BODY_BYTES,
 	toolCatalog,
 	TOOLS,
 	type ToolContext,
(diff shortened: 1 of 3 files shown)


athrvk:
Note from the review side: #767's issue body was edited *after* it was first filed, adding sharper root-cause grounding - precise citations for why `run_request`'s `/execute` path (`serialize(const Response&)`, `engine/src/utils/json.cpp:846-889`) has no cap at any size, versus `get_run_report`'s inherited 5MB storage cap (`cap_trace_bodies`, `json.cpp:263`, wired at `execution.cpp:606-610`) - and reframed the 32KB constant as a deliberate MCP-layer tool-result budget rather than borrowed engine semantics.

This PR's own description already tracks that reasoning closely (including the scenario-multiplier row bounding), so it looks like it was written against the updated body - but flagging per the issue's edit so it gets a quick diff against #767's current text before merge, particularly:

- The #754 cross-reference: `engine/src/core/scenario_runner.cpp:678-689` builds each scenario step's trace with the same `build_result_trace` + `cap_trace_bodies` pair `record_design_result` uses for a single `/execute`, which is why #754's own Sequencing section now notes this issue as a dependency. Worth confirming the multi-row bounding this PR added actually covers that per-step shape once `run_collection` (PR #765) ships, not just today's design/load report shapes.

No action needed if the diff confirms it's already covered - just don't want a stale read of the issue to merge past something that changed under it.

athrvk:
Diffed against #767's current text - this PR was written against the edited body, so nothing is stale. Both citations it added are the ones the description and the code comments carry: `serialize(const Response&)` (`json.cpp:846-889`) dumping `bodyRaw` unconditionally at any size, versus `cap_trace_bodies` (`json.cpp:263`) wired at `execution.cpp:606-610` with a 5MB storage ceiling that only discloses *above* itself. The 32KB constant is framed as an MCP-layer tool-result budget in both the code comment and the doc note, explicitly saying neither engine cap applies - `maxResponseBodyBytes` is load-only and `maxTraceBodyBytes` is sized for the database - with `maxSampleBodyBytes` cited as precedent for the number, not as borrowed semantics.

**On the #754 cross-reference: confirmed covered, and now pinned by a test** (`72cae11`).

Traced it rather than assuming. A scenario step goes through the identical pair you name - `build_result_trace` then `cap_trace_bodies` at `scenario_runner.cpp:678-689` - and lands in the same column a single `/execute` uses: `row.trace_data = record.trace.dump(...)` at `scenario_runner.cpp:715`, one `results` row per step. `/runs/:id/report` parses that column back into `results[].trace` in one loop for every run type (`runs.cpp:1141-1150`), so a scenario report is structurally today's design report with up to 100 trace rows instead of one. The walk this PR added maps every row, which is why it already covers it - the multi-row test was there for exactly this reason, but it used the plain design shape, so nothing pinned the scenario shape itself.

The new test uses the real thing: step identity stamped on the trace beside the nodes (`stamp_step_identity` - `iteration`, `stepIndex`, `stepName`, `outcome`), and a skipped step whose `response` node the runner erases at `scenario_runner.cpp:686`. It asserts the identity survives the rebuild, the skipped step's *request* body is still bounded, and no empty `response` is invented for the erased one - mutation-checked by defaulting an absent node to `{}`, which reddens it. That last one matters beyond tidiness: an invented empty node reads in the step's expanded view as a server that answered with nothing, which is the same reason the runner erases it.

Two related things I checked while in there, both clean:

- **`scenario.steps` carries no bodies.** It is `build_step_breakdown` (`scenario_load.cpp:173`) - index, name, requestId, method, executed, errors, latency percentiles off the histograms - so a scenario report has no second place for a body to land unbounded.
- **The 100-row report cap is the engine's, not something this PR narrows.** A 200-step scenario report is already truncated to 100 rows by `runs.cpp:1126` before MCP sees it; this bounds the bodies within whatever rows arrive.

So once #765 ships `run_collection`, its reports come back bounded with no further change, and if someone later narrows the walk to the design-only shape, the new test goes red.

CI: ubuntu and macOS app jobs green on the first commit, windows still running; the test commit is building now. `pnpm test` (5211), `type-check`, `lint` and `format:check` are green locally.

---

athrvk:
Cross-checked this against #754 / PR #765 (the scenario multiplier your multi-row test names), by merging the two branches locally and measuring - posting the result because one half of it affects how this PR's coverage reads.

**Confirmed working, generically:** `boundRunReport` needs no scenario awareness. A scenario step's trace is built by the same `build_result_trace` (`scenario_runner.cpp:678`) and `stamp_step_identity` adds no body-carrying field, so a 200-row scenario fixture came back with every node cut to exactly 32,768 bytes, `bodyTruncated`/`bodyBytes` set, `rawRequest` headers intact. Your walk covers it as written - no change needed for that, and no ordering dependency between the PRs.

**The gap the measurement exposed:** the per-node bound does not bound the result. At the report route's own 100-row cap, 100 steps x 40 KB bodies (each cut to 32 KB) comes back as **3,328,454 characters** - ~2.5x the 1,334,938 that #767 reproduced as exceeding the token limit - and 100 steps x 8 KB bodies, which this bound never touches, is still 845,354. So for a scenario report the binding constraint is row count, not body size, and an agent still gets a token-limit error rather than a bounded answer.

That is outside what this PR set out to do and I would not hold it up for that - filed as **#769** with the numbers and a fix pathway (a total budget beside the per-node one, disclosed in band, non-passing rows kept first to match the engine's `stepsStored` thinning). Flagging here only because this PR's description reads as covering "the scenario multiplier", which is true per node and not yet true per report.

---

athrvk:
CI is green - gate passed, app suite green on ubuntu, macOS and windows; engine jobs correctly skipped (no C++ in the diff).

Two red marks on the way there, neither a test failure, recorded so the history is not a mystery:

1. **`CI gate` on `6662428`: `app: cancelled`.** Self-inflicted - I pushed the test commit while that run's windows job was still in flight, so the matrix was superseded. Ubuntu and macOS had already passed on that commit.
2. **`App on macos-latest` on `72cae11`: failed after 70s.** Infrastructure, not code: `codeload.github.com` answered `429 Too Many Requests` three times while downloading the `pnpm/action-setup` action itself, so the job died before checkout - no test body ran, and ubuntu was green on the same commit. Re-ran that job alone; it passed in 5m22s.

No code changed for either.

---

PR #770, opened 2026-08-17
fix(app/titlebar): drag the window by the row's slack right of the search bar

### LINKED ISSUE #764: app(titlebar): right-cluster wrapper blanks the drag region across the row's slack - window cannot be dragged right of the search bar
**Problem**

The title row is a `-webkit-app-region: drag` surface (`app/src/components/layout/TitleBar.tsx:282`), and every interactive control opts out *individually*: `AppIcon` sets `no-drag` on itself (`TitleBar.tsx:136`), `WindowControls` sets it on its own root (`TitleBar.tsx:58`), and `CommandSearchBar`'s trigger button sets it on itself (`app/src/components/layout/CommandSearchBar.tsx:43`).

The right-hand cluster breaks that pattern. The wrapper `div` that holds `EnvSwitcher` (and, on Linux, `WindowControls`) sets `WebkitAppRegion: "no-drag"` on **itself**, not on the button inside it (`TitleBar.tsx:303-315`):

```tsx
<div
	className="flex min-w-0 items-center justify-end gap-2 px-3 h-full"
	style={{
		WebkitAppRegion: "no-drag",
		...(isWindows && { paddingRight: "calc(100vw - env(titlebar-area-width, 100vw) + 0.5rem)" }),
	} as React.CSSProperties}
>
	<EnvSwitcher />
	{isLinux && <WindowControls />}
</div>
```

That wrapper is the entire third `1fr` grid column (`TitleBar.tsx:281`, `grid-cols-[1fr_auto_1fr]`), laid out `justify-end`. Whatever empty space sits between the centred search bar and the right-aligned `EnvSwitcher` pill - the row's "slack" on that side - is inside this `no-drag` wrapper, so it can't drag the window either, even though it looks like plain title-bar chrome.

This is the exact anti-pattern the project's own docs warn against. `docs/app/COMPONENTS.md:81`:

> Opting out per control or per cluster is fine; opting out a wrapper that spans the row's slack is not, because that slack is what the window is dragged by.

...and cites `TitleBar.search-bar.test.tsx` as the thing enforcing it. It doesn't, for this case: the `title-row drag regions › drags by the row and not by its controls` test (`TitleBar.search-bar.test.tsx:161-182`) only asserts `effectiveRegion` (own-or-inherited) on the controls themselves - `searchBar()`, the `switch environment` button, `minimize`, `close`. It never asserts that the *wrapper* around them keeps its own declaration undeclared, so a wrapper-wide `no-drag` passes every existing assertion silently.

Reported symptom (matches the layout exactly): dragging the window works to the left of the centred search bar - column 1 mostly inherits `drag` from the header, aside from the small `AppIcon` hit target on Windows - and fails everywhere to the right of it, because column 3 is `no-drag` in its entirety. Reproduces on all three platforms; Windows and macOS make it most visible because `EnvSwitcher` is the column's only element and sits well away from the window edge (Windows additionally reserves `paddingRight` for the native overlay at `TitleBar.tsx:311-313`, widening the dead strip further). Linux has the same defect, just less visible because `WindowControls` already fills more of that column.

Verified: `rg AppRegion app/src` returns only `TitleBar.tsx` and `CommandSearchBar.tsx` as producers of the style (the other two hits, `app-icon-system-menu.test.tsx` and `TabStrip.overflow.test.tsx`, are pre-existing tests reading it, not new surface) - this is a self-contained renderer defect, not something engine or MCP code touches.

**This predates the doc rule, and the same file already shows the correct pattern next to it - so this reads as an accidental gap, not a deliberate exception.** `git log -p -S 'opting out a wrapper that spans' -- docs/app/COMPONENTS.md` shows the current wording landed in `b2b2e612` (2026-08-13, the B2 shell rework, #529), but `git show b2b2e612 -- app/src/components/layout/TitleBar.tsx` shows the right-hand wrapper's `WebkitAppRegion: "no-drag"` was **already present before that commit** - it is a context line in the diff, carried forward with only className tweaks (`justify-end`, `h-full` added), not introduced by it. The *pre-B2* wording of the same doc section made the identical point using `TabStrip` as the illustration: *"the `TabStrip` wrapper flexes to fill the bar, so marking it `no-drag` turns all the empty space right of the last tab into a dead zone... `TabStrip` sets `no-drag` on its own tab row for this reason."* And the pre-B2 `TitleBar.tsx` followed that correctly for `TabStrip` (wrapper stayed `drag`, only the tab row opted out) while getting it wrong, in the same file, at the same time, for the right-hand cluster. So this has been a live inconsistency since before the rule was even reworded - not something B2 introduced and not a considered trade-off for the Windows case addressed next.

**The Windows native-overlay padding strip is not a reason to keep the wrapper-wide `no-drag`.** The reserved `paddingRight` region (`TitleBar.tsx:311-313`, sized from `env(titlebar-area-width)`) is where Electron's Window Controls Overlay draws native min/max/close controls *above* the web content - by the WCO API's own contract, that rectangle is hit-tested by the OS before it ever reaches the DOM, which is precisely what `env(titlebar-area-width/-x)` exist to let a page know to stay out of. Whatever `-webkit-app-region` value the page declares underneath it is moot in that specific strip. This can't be fully confirmed without a packaged Windows build (jsdom has no real hit-testing), so the acceptance criteria below ask for a manual check on Windows rather than asserting it - but there is no platform-API reason for the fix to special-case that strip.

**Fix / solution pathway**

1. `TitleBar.tsx`, the right-hand wrapper div (~303-315): drop `WebkitAppRegion: "no-drag"` from the style object; keep the Windows overlay `paddingRight` reservation as-is. The wrapper then inherits `drag` from the `<header>` (`TitleBar.tsx:282`), the same way column 1 already does.
2. `EnvSwitcher()`'s trigger button (`TitleBar.tsx`, the `<button>` under `DropdownMenuTrigger asChild>`, ~209-246): add `style={{ WebkitAppRegion: "no-drag" } as React.CSSProperties}` directly on that button, matching the established per-control pattern (`AppIcon` at `TitleBar.tsx:136`, `CommandSearchBar`'s trigger at `CommandSearchBar.tsx:43`). `DropdownMenuTrigger` is `DropdownMenuPrimitive.Trigger` (`app/src/components/ui/dropdown-menu.tsx:19`) used with `asChild`, so it renders the `<button>` directly with no extra wrapping DOM node - the style lands on the real element.
3. `WindowControls` (Linux-only, `TitleBar.tsx:46-87`) already sets `no-drag` on its own root (`TitleBar.tsx:58`) - no change needed; removing the wrapper's blanket `no-drag` doesn't affect it.
4. Leave `docs/app/COMPONENTS.md:81` alone - it already states the target invariant correctly. It describes what should be true, not what the code currently does; once the code matches it, no doc edit is needed. (Flagging this explicitly so the implementer doesn't skip the docs step assuming it's forgotten - it's the opposite, the doc already got this right and the code needs to catch up.)

**App changes**

N/A beyond the fix itself - this is a pure renderer change scoped to one component. Verified no engine, MCP, or import-path code sets or reads `WebkitAppRegion` (see the `rg` result above).

**Tests**

Add an assertion to the existing `describe("title-row drag regions")` block in `app/src/components/layout/TitleBar.search-bar.test.tsx`, alongside the current per-control test, that checks the *wrapper* rather than the control:

```ts
it("does not blank out the drag region around the env switcher", async () => {
	await renderFor("win32");
	const envButton = screen.getByRole("button", { name: /switch environment/i });
	// DropdownMenuTrigger asChild renders the button in place, so its DOM
	// parent is the row's right-hand flex wrapper, not a Radix-inserted node.
	expect(appRegion(envButton.parentElement)).not.toBe("no-drag");
	expect(effectiveRegion(envButton)).toBe("no-drag");
});
```

This is a mutation-check case per the repo's testing convention: revert the fix (put `no-drag` back on the wrapper) and confirm this new assertion fails, then restore. The pre-existing "drags by the row and not by its controls" test passes against both the buggy and the fixed code, which is exactly why it didn't catch this - it only ever checked the controls.

Run: `cd app && pnpm vitest run src/components/layout/TitleBar.search-bar.test.tsx`, then `pnpm lint` and `pnpm format:check` (CI-enforced repo-wide per #189). No engine suite involved.

**Sequencing**

Parallel-safe with everything currently open. Verified: this touches only `app/src/components/layout/TitleBar.tsx` and `TitleBar.search-bar.test.tsx`; no other open issue and the one open PR (#763, transport-policy phase 2 - custom CA certs / verifySSL, engine + app settings/network code) reference either file. Standalone, no phasing needed, no wave-map position beyond "safe to pick up any time."

**Acceptance criteria**

- [ ] Dragging the title bar anywhere in the empty space to the right of the search bar (below the reserved native-overlay padding on Windows) moves the window, on Windows, macOS and Linux.
- [ ] The `EnvSwitcher` button, and on Linux the `WindowControls` buttons, remain clickable - not swallowed by the drag region.
- [ ] Manually verified on a packaged (or `pnpm electron:dev`) Windows build: the native min/max/close overlay still responds correctly with the reserved padding strip now inheriting `drag`, confirming the WCO hit-testing reasoning above rather than assuming it.
- [ ] The new vitest assertion in `TitleBar.search-bar.test.tsx` fails against the pre-fix code and passes after the fix (mutation-checked).
- [ ] `pnpm lint`, `pnpm format:check`, and `pnpm vitest run src/components/layout/TitleBar.search-bar.test.tsx` are green.

---

athrvk:
## Description

**Plan.** Root cause: the right-hand cluster opts out of the drag region on its **wrapper** (`TitleBar.tsx:303-315`) rather than on the control inside it, and that wrapper is the entire third `1fr` grid column (`grid-cols-[1fr_auto_1fr]`, laid out `justify-end`). Everything between the centred search bar and the right-aligned pill - the row's slack on that side, which looks like plain title-bar chrome - is therefore `no-drag` and cannot move the window. The approach is the per-control pattern the same file already uses three times over (`AppIcon:136`, `WindowControls:58`, `CommandSearchBar.tsx:43`): drop the wrapper's declaration so it inherits `drag` from the `<header>` the way column 1 does, and put `no-drag` on `EnvSwitcher`'s trigger button. The alternative I rejected was keeping the wrapper `no-drag` and adding a sibling drag spacer to fill the slack - that reintroduces the same dead strip whenever the column's contents change width (a long environment name, the Linux button cluster), and it hand-rolls what inheritance already gives correctly.

Verified against master (`a8690df`) before writing: the wrapper still carries `WebkitAppRegion: "no-drag"` exactly as the issue anchors it, `rg AppRegion app/src` still returns only `TitleBar.tsx` and `CommandSearchBar.tsx` as producers, and no engine, MCP or import path reads the property - this is self-contained to the renderer.

**The Windows overlay padding stays as-is.** The reserved `paddingRight` (sized from `env(titlebar-area-width)`) is where Electron's Window Controls Overlay draws native controls above the web content; that rectangle is hit-tested by the OS before it reaches the DOM, so whatever the page declares underneath it is moot there. The reasoning is now a comment beside the padding rather than an unstated assumption. Per the issue this is the one claim a jsdom test cannot settle - see the deviation note below.

**`docs/app/COMPONENTS.md:81` is deliberately untouched.** It already states the target invariant ("opting out per control or per cluster is fine; opting out a wrapper that spans the row's slack is not, because that slack is what the window is dragged by"). The doc was right and the code had drifted from it, so the fix is the code catching up - the issue calls this out explicitly so the docs step is not skipped as forgotten.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

App-only renderer change (two files, no engine source), so the verification is the app half of the repo's gate:

- `pnpm vitest run src/components/layout/TitleBar.search-bar.test.tsx` - **7 passed (1 new)**.
- `pnpm test` - **5222 passed, 499 files**. Nothing else moved; the two other tests that read `-webkit-app-region` (`app-icon-system-menu.test.tsx`, `TabStrip.overflow.test.tsx`) are unaffected.
- `pnpm type-check`, `pnpm lint`, `pnpm format:check` - clean. Both touched files were already prettier-clean, so formatting them changed nothing else.
- **Mutation-checked**, as the issue asks: putting `no-drag` back on the wrapper reddens the new case (`right column is dead on win32: expected "no-drag" to be "drag"`) while the other six stay green - including the pre-existing "drags by the row and not by its controls", which is precisely why it never caught this. Restored and re-run green (7/7).
- Engine untouched, so no engine build or `ctest` run - no C++, CMake or test file is in the diff.

The new test asserts what the old one structurally could not. The existing case reads `effectiveRegion` (own-or-inherited) on the *controls*, which a blanket wrapper satisfies exactly as well as a per-control opt-out; the new one asserts the **wrapper** inherits `drag` and that the button carries its **own** `no-drag`, looped over `win32`/`darwin`/`linux` rather than pinned to the host platform.

## Checklist
- [x] I have added tests
- [x] I have updated documentation (n/a - the doc already states the invariant; see above)

## Deviation from the issue spec

One acceptance criterion asks for a manual check on a packaged (or `pnpm electron:dev`) **Windows** build, confirming the native min/max/close overlay still behaves with the reserved padding strip now inheriting `drag`. **This run could not perform it** - the environment is headless Linux with no Windows build and no real hit-testing (jsdom has none, which is why the issue asked for a manual check rather than an assertion). The code change matches the issue's stated pathway exactly and the WCO reasoning is documented at the site; the Windows overlay check and the three-platform drag-by-the-slack check remain for you at review time. Everything else in the acceptance criteria is done and verified above.

Also noted while in the file, deliberately **not** touched (adjacent, not this diff): `docs/app/COMPONENTS.md` names the right-hand control `EnvPill` while the component is `EnvSwitcher` - a naming drift in the same paragraph, worth a one-word fix in whatever next edits that file.

## Related Issues
Closes #764

---
_Generated by [Claude Code](https://claude.ai/code/session_011BpCNjz5vErxBpMs5bxZqB)_

### DIFF
--- a/app/src/components/layout/TitleBar.search-bar.test.tsx
+++ b/app/src/components/layout/TitleBar.search-bar.test.tsx
@@ -180,4 +180,34 @@ describe("title-row drag regions", () => {
 			).toBe("no-drag");
 		}
 	});
+
+	/**
+	 * The test above passes against the defect it is paired with: it reads
+	 * `effectiveRegion`, which a *cluster* opting out satisfies, so a wrapper
+	 * marked `no-drag` in its entirety looks identical to a control that opted
+	 * out on itself. The difference is everything the wrapper covers and the
+	 * control does not - here the whole third `1fr` column, i.e. the row's slack
+	 * right of the search bar, which is what the window is dragged by.
+	 *
+	 * So this asserts the wrapper's own inherited value instead, and the button's
+	 * own declaration rather than an inherited one.
+	 */
+	it("keeps the slack around the env switcher draggable, opting out only the control", async () => {
+		// Every platform: the right column holds the switcher alone on Windows and
+		// macOS and the HTML window buttons too on Linux, and the slack has to
+		// drag on all three.
+		for (const platform of ["win32", "darwin", "linux"]) {
+			await renderFor(platform);
+			const envButton = screen.getByRole("button", { name: /switch environment/i });
+			// `DropdownMenuTrigger asChild` renders the button in place, so its DOM
+			// parent is the row's right-hand wrapper, not a Radix-inserted node.
+			const wrapper = envButton.parentElement;
+			expect(wrapper, `no right-hand wrapper on ${platform}`).not.toBeNull();
+			expect(effectiveRegion(wrapper), `right column is dead on ${platform}`).toBe("drag");
+			expect(appRegion(envButton), `env switcher not clickable on ${platform}`).toBe(
+				"no-drag"
+			);
+			cleanup();
+		}
+	});
 });
--- a/app/src/components/layout/TitleBar.tsx
+++ b/app/src/components/layout/TitleBar.tsx
@@ -235,6 +235,11 @@ function EnvSwitcher() {
 								"bg-primary/10 text-primary-text border-primary/30 hover:bg-primary/20"
 							: "border-transparent text-muted-foreground hover:bg-accent hover:text-foreground"
 					)}
+					// Per-control opt-out, the pattern AppIcon and CommandSearchBar's
+					// trigger follow. `DropdownMenuTrigger asChild` renders this button
+					// in place with no wrapper node, so the declaration lands on the
+					// real element.
+					style={{ WebkitAppRegion: "no-drag" } as React.CSSProperties}
 					aria-label="Switch environment"
 				>
 					<Cloud className="w-3 h-3 shrink-0" />
@@ -299,15 +304,22 @@ export default function TitleBar() {
 			    remaining area stays draggable. */}
 			<CommandSearchBar className="w-[min(44vw,28rem)]" />
 
-			{/* Right controls */}
+			{/* Right controls.
+			    No `no-drag` here: this wrapper is the whole third 1fr column, so
+			    opting it out would take the row's slack right of the search bar with
+			    it - the largest drag surface on this side. Each control opts out on
+			    itself instead (EnvSwitcher's trigger, WindowControls' own root), the
+			    way AppIcon and CommandSearchBar already do. */}
 			<div
 				className="flex min-w-0 items-center justify-end gap-2 px-3 h-full"
 				style={
 					{
-						WebkitAppRegion: "no-drag",
 						// Windows paints native min/max/close as an overlay on top of the
 						// web content in the top-right corner. Reserve its width (exposed by
 						// the Window Controls Overlay API) so the env switcher isn't covered.
+						// The strip stays a drag region: the overlay rectangle is hit-tested
+						// by the OS before the DOM sees it, so what the page declares under
+						// it does not apply there.
 						...(isWindows && {
 							paddingRight: "calc(100vw - env(titlebar-area-width, 100vw) + 0.5rem)",
 						}),


athrvk:
CI is green - gate passed, app suite green on ubuntu, macOS and windows; engine and installer jobs correctly skipped (no C++ or script in the diff).

One red mark on the way there, recorded so the history is not a mystery: **`Analyze JavaScript/TypeScript` failed on `35b7a2f`** at 15:31 UTC. Not a finding and not code - the analysis itself completed and exported `javascript.sarif` cleanly, then the *upload* to code scanning got `No server is currently available to service your request` from GitHub and the job failed at that step. Nothing was flagged; no alert exists to address. Re-ran the failed job alone and it passed in 2m12s.

No code changed for it.

---

PR #771, opened 2026-08-17
fix(app/dialog): clamp the dialog panel's grid track so wide content scrolls

### LINKED ISSUE #701: app(collections): the Run dialog's layout blows out past its painted panel when the data-file preview table is wide - every right-aligned control floats over the backdrop
## Problem

**Severity: high - the dialog is visibly broken the moment a realistic data file is picked, on the exact flow the data-contract work exists to serve.**

Observed on a collection run with a 7-column CSV (userId, email, plan, quantity, orderRef, city, delayMs - emails and "Portland, OR"-style cells): the Run dialog's *painted panel* stays at its configured width, but its *content* lays out roughly 50% wider. The result on screen:

- The "Include sub-folders" and "Load test" toggles, the Iterations input, the data file's Remove button, and the Cancel/Run footer buttons all render **outside the panel**, floating over the dimmed backdrop far to the right of the dialog's visible edge.
- The header description and every full-width child (the data-file card with its preview table) likewise extend past the panel's right border.

The component states the invariant this breaks, in its own comment (`app/src/modules/collections/DataFilePicker.tsx:312-313`):

> The grid scrolls rather than widening the dialog: a file with twenty columns must not push the footer off-screen.

Seven columns is enough.

## Mechanism

- `RunCollectionDialog.tsx:321` renders `DialogContent className="sm:max-w-md"`; the primitive (`app/src/components/ui/dialog.tsx:62`) is `grid w-full max-w-lg ... bg-background p-6`. So the panel's *box* (and its painted background) is width-capped.
- The preview table's cells are `whitespace-nowrap` (`DataFilePicker.tsx:319`, `:329`), giving the table a large min-content width. Its wrapper is `max-h-56 overflow-auto` (`:314`) - but **a scroll container still contributes its content's min-content width to intrinsic sizing**. Making a box scrollable does not stop its intrinsic width from propagating upward; only a width constraint on the chain does.
- The wrapper sits inside the card div (`:278`) which is a child of `DialogContent`'s `grid`. A grid item's default `min-width` is `auto` (content-based). The table's min-content therefore inflates the grid track past `max-w-md`, and **every** sibling row in that grid lays out at the track width - which is why controls that are right-aligned with `justify-between` (toggles, Remove, the footer) land at the track's right edge, outside the capped, painted panel. The panel box cannot grow (max-width), the children overflow it visibly, and the background does not follow them.

## Fix pathway

The standard cure for min-content escape: give the chain permission to shrink, so `overflow-auto` can do its job.

- `min-w-0` on the grid items between `DialogContent` and the scroll wrapper (the card div at `DataFilePicker.tsx:278`, plus any flex/grid ancestor inside `RunCollectionDialog` that wraps the picker), and `max-w-full` on the `overflow-auto` wrapper itself if needed. After the clamp, a wide file scrolls horizontally inside the wrapper - which is the behavior the `:312` comment already promises.
- Audit the **other consumer**: `DataFilePicker` is also rendered by the collection Data tab (`mode="declare"`, issue #599 surface). Same component, same table, different ancestor chain - verify that surface with the same wide file and clamp its chain too if it exhibits the same escape.
- Verify in the running app (a 7-column CSV like the one above is sufficient; also try ~20 columns to exercise the horizontal scroll), not only in jsdom - jsdom has no layout and cannot show a blowout.

Secondary item, same component, worth fixing in passing: with a file of 10 rows or fewer the preview renders all rows inside `max-h-56` with **no disclosure line** (the "Showing the first N of M rows" text at `:339-343` only appears past `PREVIEW_ROWS = 10`), and on macOS overlay scrollbars an unscrolled container shows no scrollbar - so an 8-row file shows ~7.5 rows with the last row half-clipped and nothing saying more exist. Either state the row visibility ("8 rows - scroll for the rest") whenever the table can exceed `max-h-56`, or size the container to whole rows so a clip is never ambiguous.

## Tests

- jsdom reports 0 for every layout measurement, so assert the *declarations*, not geometry - the `boxed-surfaces.test.tsx` / `surface-rule.test.tsx` precedent in app/CLAUDE.md: render `RunCollectionDialog` with a selected wide parsed file and assert `min-w-0` (and the wrapper's clamp) are present on the ancestry via `element.className`.
- Mutation check: remove one `min-w-0` from the chain, confirm the guard fails, restore.

## Docs - same commit

None obligated - no doc describes this dialog's layout. The `:312` comment should be extended to name the trap it now actually survives (overflow-auto does not stop min-content propagation; the `min-w-0` chain is load-bearing), so the next person does not remove the "redundant" classes.

## Sequencing

Standalone. Touches the same component family as #599's Data tab but conflicts with nothing open.

## Acceptance criteria

- [ ] With a 7-column CSV of realistic cell widths, every control of the Run dialog renders inside the painted panel, and the preview table scrolls horizontally within its own wrapper.
- [ ] The Data tab's picker surface verified with the same file, clamped if affected.
- [ ] Class-declaration guard in place and mutation-checked.
- [ ] A file with more rows than fit in `max-h-56` but fewer than `PREVIEW_ROWS` is not silently clipped: either a visibility line or whole-row sizing.

athrvk:
## Description

**Plan.** Root cause: `DialogContent` is a `grid`, an implicit track is `auto`, and an auto track's minimum is its items' min-content - so the data-file preview table's min-content grew the panel's single track past the panel's own `max-w`, which cannot follow it. The panel stayed the painted width while every row in that grid laid out at the track's, which is why controls with nothing to do with the table (the toggles, Iterations, Remove, the Cancel/Run footer) were the ones hanging over the backdrop. Making the table's wrapper `overflow-auto` does not prevent this: it bounds the box, not the min-content width the box contributes upward.

The approach is to clamp the track on the **surface** rather than the call site: `grid-cols-1` is `repeat(1, minmax(0, 1fr))`, and the `0` lets the track be narrower than its content's min-content, which is what gives the scrollers inside it room to scroll. The alternative I rejected is the one the issue names - `min-w-0` on the grid items between `DialogContent` and the scroll wrapper. It works, but it is a per-call-site remedy for a trap the surface creates: the items are written by all seventeen `DialogContent` consumers and the width cap is the panel's own promise, so the next dialog that puts a wide table in one has the same bug and no reason to expect it. See the deviation note below, with the measurements behind it.

Verified against master before writing: `DialogContent` is still the `grid w-full max-w-lg` the issue anchors to, the picker's wrapper is still `max-h-56 overflow-auto`, and its cells are still `whitespace-nowrap`.

**Measured in a real engine, not reasoned about.** jsdom reports 0 for every layout measurement, so the mechanism and the fix were reproduced in headless Chromium against a harness mirroring the real class chain, using the issue's own 7-column CSV (`userId, email, plan, quantity, orderRef, city, delayMs`, with `Portland, OR`-style cells):

| Chain | Painted panel | Content track | Worst control spill |
|---|---|---|---|
| Run dialog, master | 448px | **851px** | **428px outside the panel** |
| Run dialog, with the clamp | 448px | 398px | 0px (preview scrolls, 823px inside 355px) |
| `min-w-0` on the card only | 448px | 851px | 428px - inert |
| `min-w-0` on the picker root only | 448px | 851px | 428px - inert |

That last pair is why the clamp is not in `DataFilePicker`: the component **cannot** defend itself, because the box that escapes is its host's, not one it owns. The comment at the preview now says so, so the next reader does not try.

**The Data tab's picker is unaffected, measured on the same harness** (acceptance criterion 2). Its chain is `max-w-[720px] flex flex-col`, and a stretched flex item gets a definite width from the container, so the preview already scrolled there (823px inside 677px, 0px spill). No clamp added to a surface that does not need one.

**One thing worth flagging, because it nearly shipped inert.** I first wrote the clamp as `grid-cols-[minmax(0,1fr)]`, which says it far more directly. **Tailwind emits no rule for that class** - checked against the built CSS, where `grid-cols-[1fr_auto_1fr]` and friends are present and this one is absent. It would have read correctly in review, passed the `className` assertion in the test, and styled nothing: this repo's most-repeated defect wearing a CSS hat. The stock `grid-cols-1` compiles to exactly `repeat(1, minmax(0, 1fr))`; both the rule (`.grid-cols-1{grid-template-columns:repeat(1,minmax(0,1fr))}`) and its use on the panel are verified present in `vite build` output. The reason not to "simplify" it back is recorded at the declaration and in `docs/design-system.md`.

**Secondary item, as the issue asks.** The preview scrolls inside `max-h-56`, and on overlay-scrollbar platforms an unscrolled container draws no scrollbar at all, so a file of a handful of rows ended at a half-drawn row with nothing saying the rest were there - the disclosure line only appeared past `PREVIEW_ROWS`, which is precisely the range where it was least needed. It now speaks in both directions ("Showing the first 10 of 24 rows - the table scrolls" / "Showing all 8 rows - scroll the table if they do not fit"), and stays quiet for a single row, which a box this tall cannot cut off.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

App-only renderer change (one primitive, one component, two docs), so the verification is the app half of the repo's gate:

- `pnpm test` - **5247 passed, 501 files** (8 new across two files).
- `pnpm type-check`, `pnpm lint`, `pnpm format:check` - clean. Only the files touched here were formatted, and they were prettier-clean before.
- `npx vite build` - used as the check that the utility is real, per the note above.
- **Mutation-checked** both guards, as the issue asks. Removing `grid-cols-1` from `DialogContent` reddens the clamp guard (`expected 'dialog-panel fixed left-[50%] top-[50…' to contain 'grid-cols-1'`); reverting the disclosure to the old `rowCount > PREVIEW_ROWS &&` form reddens the short-file case. Both restored and re-run green.
- Engine untouched, so no engine build or `ctest` run - no C++, CMake or engine test file is in the diff.
- `mkdocs build --strict` not run: mkdocs is not installed in this environment. The docs change adds no page and no heading anchor; the one relative link it adds (`../design-system.md` from `docs/app/COMPONENTS.md`) is a path that file already links by.

New coverage: the panel carries the clamp and is still width-capped (the bug was never that the dialog was too narrow); the footer is a child of the panel it is painted in; the preview box declares `max-h-56 overflow-auto` and its cells `whitespace-nowrap`, which is what makes it need to scroll; and the three disclosure cases (24 rows, 8 rows, 1 row).

## Checklist
- [x] I have added tests

Docs updated in the same commit: `docs/design-system.md` gains the grid half of "Flex Items Must Be Told They May Shrink" - the section already carried the flex rule and the scroller exemption, and the exemption is exactly what misleads here (it reaches the item itself, never a scroller nested below one). `docs/app/COMPONENTS.md`'s `DataFilePicker` paragraph now states the disclosure and where the clamp has to live.

## Deviation from the issue spec

The issue's fix pathway asks for `min-w-0` on the grid items between `DialogContent` and the scroll wrapper, plus `max-w-full` on the wrapper. **This clamps the track on the primitive instead**, for the reason in the plan above and because the measurements show the per-item version is only correct at the one item that happens to be a grid child - a clamp on the card or the picker root, both of which the issue's pathway also suggests, changes nothing at all. If you would rather keep the primitive untouched and pay the `min-w-0` at each call site, say so and I will swap it - the single-line change is `min-w-0` on `RunCollectionDialog.tsx`'s content div, and it measured identically (398px track, 0px spill).

One acceptance criterion asks for verification **in the running app** rather than only jsdom. **This run could not launch it** - the environment is headless Linux with no display and no packaged build. What I did instead is stronger than a jsdom assertion and weaker than your eyes on the real dialog: the mechanism and the fix were measured in real Chromium against a harness reproducing the exact class chain, and the compiled CSS rule was verified in the build output. The 7-column and 20-column look on a real window remains for you at review time.

Noted while in the file, deliberately **not** touched (adjacent, not this diff): the other sixteen `DialogContent` consumers now inherit the clamp. That is the intent, and it changes nothing for a dialog that was not already overflowing - an auto track and `minmax(0, 1fr)` size identically whenever min-content fits.

## Related Issues
Closes #701

---
_Generated by [Claude Code](https://claude.ai/code/session_01TZmxxQpCfG8ZnRoYzdJMFU)_


athrvk:
a slightly wider and larger modal would be better UI/UX
check all the other modals as well

athrvk:
Done in `9b05281`, and checking the others turned it into a scale rather than a bump.

**What the audit found.** Eleven call sites carrying **five** widths: `sm:max-w-md` (448), `sm:max-w-lg` (512), `sm:max-w-xl` (576), `sm:max-w-[560px]` and a hard `w-[500px]`. Nothing separated them - the same kind of dialog came out a different size depending on who wrote it, and the *narrowest* bucket held the run dialog with its preview table, the delete confirm, and the mock-server form alike.

So there are now two, with a rule you can apply without asking:

| Size | Class | For |
|---|---|---|
| Standard | `sm:max-w-lg` (512) | a form or a decision |
| Wide | `max-w-xl` (576), the primitive's default | a dialog holding something with a shape of its own - a table, a diff, a preview, a dense config |

Standard, 448 -> 512: delete-confirm, move-to, start-mock-server, welcome collection picker, save-as-example, new-issuer. Wide -> 576: **the run dialog**, export-spec (was the one-off 560), the load-test config, and the import modal.

**The import modal's `w-[500px]` went too**, which is worth a line: a *fixed* width is one the panel keeps on a viewport narrower than it, so that dialog could not fit a small window. `w-full` under a cap gives the same stable band and fits.

**One thing the measurement is worth saying plainly: widening is not a fix, and I did not let it stand in for one.** Re-ran the harness with the panel at 576 and the track left `auto` - the same seven-column CSV still spills **300px** past the painted edge. The clamp is what stops it; the width is what makes the result pleasant. Together at 576 the preview now shows **483px** of table instead of 355px.

```
576px panel, auto track   : content 851px, worst control 300px outside  <-- still broken
576px panel, clamped      : content 526px, 0px outside, preview scrolls 823 inside 483
```

**Held closed by a test.** `dialog-width-scale.test.tsx` scans every `<DialogContent>` and fails on a sixth value - mutation-checked by putting `sm:max-w-md` back on `MoveToDialog`, which reddens it by name. Being a source scan it asserts it actually read something (both the call sites and the widths, per the empty-string lesson in CLAUDE.md), and it strips `//` comments inside the opening tag before matching - without that it flagged my own comment explaining a width, and `overflow-hidden` read as the width `w-hidden`.

**On "larger" in the vertical sense - I did not do it, deliberately, and it needs your call.** Most dialogs set no height cap at all: taller than the viewport, a panel is simply cut off top and bottom with no way to reach the footer, since it is `fixed` and centred by translate. Only the load-test config (`max-h-[90vh] overflow-y-auto`) and the import modal (`max-h-[82vh]`) handle it, each ad hoc. The systemic version - `max-h-[85vh] overflow-y-auto` on the primitive - is the same shape of fix as the width clamp, but it is a real behaviour change across all eleven, and it has a visible cost: the corner close button is `absolute right-4 top-4` *inside* the panel, so it scrolls away with the content on a tall dialog. Doing it properly means the scroll living on the body band with the header and close button pinned, which is a restructure of the primitive rather than a class. Happy to do it as its own PR, or file it - say which and I will.

Full gate green on the branch: **5250 tests, 502 files** (3 new), `type-check`, `lint`, `format:check` clean. `docs/design-system.md` carries the scale table and the "clamp the track, widen only for the reading" note.

---

athrvk:
Filed the deferred vertical half as **#773**, since it merged with only a comment behind it and CLAUDE.md's rule is that deferred work gets an issue number, not a "later".

It carries the measurement rather than the assertion: at a 613px viewport an eighteen-row dialog is 1058px tall, clipped 223px above, with the primary action 198px below the screen and nothing to scroll - the panel is `fixed`, so the overflow is unreachable by pointer. Two of the fourteen dialogs guard against this today, each with its own number (90vh and 82vh).

Also recorded there is the reason it is not a one-liner: `max-h-[…] overflow-y-auto` on the primitive makes the panel the scrollport, and the close button is `absolute` *inside* it, so it scrolls away on exactly the dialogs long enough to need it. The structural version puts the scroll on a body band with the header and footer pinned, which needs a `DialogBody` boundary the primitive does not have - a primitive API addition, so it wants your call rather than a drive-by.

No action needed on this PR; it is merged and the width work stands on its own.

---

PR #772, opened 2026-08-17
fix(app/mcp): bound get_run_report's traces as a set, not only one at a time

### LINKED ISSUE #769: app(mcp): get_run_report's per-node body bound does not bound the report - a 100-step scenario report is still 3.3M characters, 2.5x the size #767 reproduced as over the limit
Found while cross-checking #754 (PR #765) against #767 (PR #768), as asked for on #765 - measured, not reasoned.

## What is already covered

#768's `boundRunReport` is **generic over run type**, so the scenario multiplier the #767 note predicted *is* handled at the node level: the walk keys off `results[]` and `row.trace` and bounds `trace.request.body`, `trace.request.rawRequest` and `trace.response.body` on **every** row, with no branch on run type. A scenario step's trace is built by the same `build_result_trace` a single `/execute` uses (`scenario_runner.cpp:678`, then `stamp_step_identity` adds only `iteration`/`stepIndex`/`stepName`/`requestId`/`outcome`/`validation` - no body-carrying field), and the report route parses `trace_data` into `results[].trace` for every run type. Verified empirically by merging PR #768 into PR #765 locally and running a fixture shaped as the engine emits one: all 200 rows came back cut to exactly 32,768 bytes per node, `bodyTruncated`/`bodyBytes` set, `rawRequest` headers intact, step identity preserved. **No follow-up is needed for the per-node case.**

## Problem

The per-node bound does not bound the **result**. `results[]` is capped at 100 rows by the report route (`runs.cpp`, `max_results = 100`), and each row keeps up to 32 KB per node - so the ceiling is row count x nodes x 32 KB, which the tool never checks. Measured on the merged branches (`get_run_report` through `dispatchTool`, response body only, one field per row):

| Report | Result size |
|---|---|
| 1 step x 1 MB body (the #767 single-row case) | **33,365 chars** - the fix working exactly as intended |
| 100 steps x 40 KB body (each cut to 32 KB) | **3,328,454 chars** |
| 100 steps x 1 MB body (each cut to 32 KB) | **3,328,654 chars** |
| 100 steps x 8 KB body (under the bound, untouched) | **845,354 chars** |
| 200 rows x all three nodes populated | **19,834,336 chars** |

For reference, #767's live reproduction blew the tool-result token limit at **1,334,938 characters**. So a bounded 100-step scenario report is **~2.5x over** the size that already failed, and with the request-side nodes populated (a collection of POSTs) it is far worse. Two consequences worth separating:

- **Truncation stops being the binding constraint.** At 100 rows the row count is, and a body of 8 KB - well under the bound, so #768 touches it at all - still totals 845 K characters, most of one blowout on its own.
- **The failure is silent in the same way #767 was.** Every row honestly says it was truncated; nothing says the *report* is too large to read, so an agent gets the token-limit error rather than a bounded answer, which is precisely the shape #767 exists to remove.

This is not a defect in #768: its stated scope is the per-body bound and it delivers that, correctly and generically. It is the next layer, made reachable by #754 shipping the tool that produces 100-row reports in the first place.

## Fix pathway

In `app/electron/mcp/tools.ts`, on top of #768's helpers:

- Give `boundRunReport` a **total** budget as well as the per-node one - the same shape `run_collection_smoke`'s scope caveat and `get_live_metrics`' tick bound already use: fill rows until the budget is spent, then stop embedding traces and say so in band (`tracesOmitted: <n>`, `traceBudgetBytes`), rather than letting the sum run. Non-passing steps should be kept first if rows must be dropped, matching the engine's own thinning rule for `stepsStored` so the two do not disagree about which steps matter.
- Consider a `results` row cap of its own for the MCP result (the engine's 100 is a report-route bound, not a context one), disclosed the same way `list_runs` discloses its 100-row page.
- An explicit `includeTraces: false` / `stepLimit` argument is the alternative worth weighing: it puts the choice with the agent, at the cost of a knob. The budget-with-disclosure needs no argument and is the default that cannot be got wrong, which is why it is listed first.

## Tests

`app/electron/mcp/tools.test.ts`, against `dispatchTool` with the fake client: a 100-row report of over-bound traces comes back under the total budget with the omission disclosed and non-passing rows retained; a small report is byte-for-byte unchanged (exact equality, as #768's under-bound tests do); the per-node behaviour #768 pinned still holds. Mutation-check the budget by removing it and confirming the size assertion reddens.

## Docs - same commit

`docs/engine/mcp.md`: the bodies-are-bounded note gains the total budget beside the per-node one, and the scenario bullet's line about `results` carrying bodies inline points at it.

## Sequencing

- Requires #768 (the helpers this extends) - open at filing.
- Independent of #754 / PR #765, which only made the case reachable; that PR now states in its result and in mcp.md that step rows carry bodies inline, so an agent is at least not told the opposite.
- Land after #768; no conflict with the rest of wave M-parity beyond a rebase on `tools.ts`.

## Acceptance criteria

- [ ] A 100-step scenario report with oversized bodies returns a result inside normal tool-result limits, with what was omitted disclosed in band.
- [ ] A report already small enough is byte-for-byte unchanged.
- [ ] The measurements above are re-run and recorded in the PR, not assumed.

athrvk:
## Description

**Root cause.** #768 bounds a *node*; nothing bounds the *result*. `/runs/:id/report` returns up to 100 rows and each row may keep `MAX_INLINE_BODY_BYTES` on each of three body-carrying nodes, so the ceiling is `rows x nodes x 32 KB` and no code was checking it. Verified against the live code before writing anything: `runs.cpp:1126` still caps `results` at 100, `boundRunReport` still walks per node only, and the scenario runner still writes one `build_result_trace` trace per step into `trace_data` - the shape the issue describes is the shape that is there.

**Approach.** A second bound at the same seam: the traces together are held to **96 KB**, rows past the budget keep every scalar and carry `traceOmitted: true` instead of their trace, and the report says `tracesOmitted` / `traceBudgetBytes`. Non-passing steps spend the budget first, mirroring `ScenarioStepStore::add`'s own rule for `stepsStored` (`outcome != Passed`) so the tool and the engine cannot disagree about which steps matter; rows come back in run order regardless, since the budget decides what a row carries, never where it sits. The first trace is always embedded whatever it costs - a design run's report is one row, and a budget that dropped it would answer nothing at all.

96 KB is `MAX_INLINE_BODY_BYTES * 3`: the largest single row the per-node bound can produce, so it is derived from the existing bound rather than invented beside it, and a row is never dropped for being merely as big as one row is allowed to be.

**The alternative I rejected.** The issue offers `includeTraces: false` / `stepLimit` as an explicit argument. It puts the choice with the agent, but it is a knob that has to be *known about* to help - and the failure mode here is an agent that asked for a report and got a token-limit error instead, which is exactly the case where it would not have passed the argument. The budget-with-disclosure needs no argument and cannot be got wrong. I also **did not** add the second `results` row cap the issue says to consider: row scalars are ~200 bytes each, so 100 of them are noise beside one body, and dropping them would cost the run's shape - the part still worth reading once the payloads cannot come along.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`cd app && pnpm test` - **5260 tests, 499 files, all passing** (7 new tests; the suite was 5250/499 before). `pnpm type-check`, `pnpm lint` and `pnpm format:check` clean. No engine change, so no C++ build or ctest run - the diff is TypeScript and Markdown.

**Measurements re-run, not assumed** (acceptance criterion 3). Fixtures driven through `dispatchTool("get_run_report")` with the fake client, sizes are the tool result's own text:

| Report | Unbounded | Bounded result |
|---|---|---|
| 1 step x 1 MB body (the #767 single-row case) | 1,048,874 | **33,131** |
| 100 steps x 40 KB body | 4,122,310 | **76,360** |
| 100 steps x 1 MB body | 104,883,910 | **76,364** |
| 100 steps x 8 KB body (under the per-node bound) | 845,510 | **102,234** |
| 200 rows x all three nodes populated | 24,657,310 | **119,588** |

The 8 KB row is the case worth reading twice: every body is under the per-node bound, so #768 never touches it, and it was still 845 K on its own - the total budget is what brings it down. The 8 KB case lands *higher* than the 40 KB one because small traces fit more rows inside the same budget, which is the budget working rather than a cap leaking.

**Mutation-checked, both halves.** Removing the budget line reddens three tests by name (`expected 3289980 to be less than or equal to 98304`, matching the issue's 3.3M measurement); collapsing the priority ordering back to run order reddens exactly the non-passing test. Two tests pin what must *not* change: a report inside both bounds comes back byte-for-byte equal, and a load report (no traces at all) is never given an invented disclosure. #768's own per-node tests are untouched and still green.

## Checklist

## Related Issues

Closes #769

`docs/engine/mcp.md` in the same commit: the bounded-bodies section gains the total budget beside the per-node one with the measurements, the tool table row states both caps, and `run_collection`'s bullet about step rows carrying bodies inline now points at what bounds them.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KorRQUifig7RTufrLS7P5C)_

### DIFF
--- a/app/electron/mcp/tools.test.ts
+++ b/app/electron/mcp/tools.test.ts
@@ -11,6 +11,7 @@ import {
 	dispatchTool,
 	MAX_IN_FLIGHT_BOUND,
 	MAX_INLINE_BODY_BYTES,
+	MAX_REPORT_TRACE_BYTES,
 	toolCatalog,
 	TOOLS,
 	type ToolContext,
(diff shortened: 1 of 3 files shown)


athrvk:
CI note, so the red check is not read as this diff: **"Apply path labels" is failing on a GitHub API outage, not on anything here.** `actions/labeler@v7` does not check the repo out, so it fetches its config over the API - and that call is what dies:

```
The configuration file (path: .github/labeler.yml) was not found locally, fetching via the api
##[error]HttpError: No server is currently available to service your request.
```

It never evaluated a path rule, so there is no rule to fix. I re-ran the failed job once and it returned the same 5xx 25 seconds later. The same check passed on #771 twenty minutes earlier against an unchanged `.github/labeler.yml`, which is what places it on GitHub's side rather than the repo's.

Everything that reads the diff is green: **Build**, **Detect changes**, **Analyze JavaScript/TypeScript**, **Conventional title**; the three **App** matrix jobs were still running when this was written. I will re-run the labeler once the API recovers and report back here - no push is needed for it.

---

athrvk:
The API recovered and the re-run passed: **"Apply path labels" is green**, and so is everything else - **Build**, **Detect changes**, **Analyze JavaScript/TypeScript**, **Conventional title**, **App** on ubuntu / macos / windows, and the **CI gate**. CodeQL is `neutral`, as it is on #771. No code changed for any of it; the outage was the whole story.

---

Issue #757, opened 2026-08-17
app(mcp): mock servers over MCP, and live-editing a running mock issuer


athrvk:
Part of epic #753 (S4). Distinct from the OAuth mock issuer tools that already exist (#509): the examples-driven mock server (#481, shipped 0.16.0) has zero MCP coverage, and the issuer tools themselves lag the UI's live-edit capability.

## Problem (verified at 46b115d)

- Mock servers: `POST /mock/start` (`engine/src/http/routes/mock_server.cpp:790`; body `collectionId` required, `port` 0-65535, `latencyMs`, `errorRatePct` 0-100 - parse at :248), `GET /mock` (:835, running only), `GET /mock/:id/routes` (:849 - method, path template, has-example flag), `POST /mock/:id/stop` (:824). None reachable from MCP. For an agent, "start a mock of this collection, point the client under test at it" is the natural pairing with the examples it can already read about in smoke results - and #570's history shows these knobs used to be curl-only for the app too.
- Issuer live-edit: the UI edits `failureMode` and `slowMs` on a running issuer (`ServicesPanel.tsx` IssuerDelayControl) via `PUT /mock-issuer/:id` (`engine/src/http/routes/mock_issuer.cpp:808`); MCP can only start/list/stop (`app/electron/mcp/tools.ts:2478-2585`), so an agent testing retry behavior must destroy and recreate an issuer (invalidating minted tokens' issuer) to flip failure modes.

## Fix pathway

Four tools + one extension in `app/electron/mcp/tools.ts`, registry pattern, gating posture identical to the issuer tools (loopback-bound engine-side, no allowlist, no write toggle - restate the mcp.md ~716-725 rationale):

1. `start_mock_server` (execute): `collectionId` (req), `port` (0=auto), `latencyMs`, `errorRatePct`. Engine limit values deliberately not restated in the schema (the tools.ts:1216 posture); surface its 400s verbatim. Result: mockId, base URL, route count, and the no-example-serves-501 caveat.
2. `list_mock_servers` (read): running mocks with collection name and port.
3. `get_mock_routes` (read): `mockId` - the route table incl. has-example flags; description notes the table is a start-time snapshot (same fact the UI tooltip states).
4. `stop_mock_server` (execute): unknown id is an error, not silent success (the stop_mock_issuer precedent at tools.ts:2571).
[...]

Issue #502, opened 2026-08-11
app(services): one home for local services - a Services drawer view for inboxes, OAuth issuers and future mock servers, with a running-state indicator in the Dock


athrvk:
**Owner direction, 2026-08-11**: "the inbox needs a UI, it's currently only available via welcome screen. think UI/UX - also for the OAuth mock and other mocks." Design grounded in a full shell-navigation audit at master `fe6a7af`; every anchor below verified.

## Problem - three services, three different non-answers

Vayu now runs local *services* - things that keep listening after you switch tabs - and the shell has no concept for them:

1. **Webhook inbox (#480, shipped)**: reachable from exactly one place, the welcome Launcher tile (`app/src/modules/welcome/WelcomeScreen.tsx:165` is the sole non-test `openTab({type:"inbox"})` call site). No keyboard shortcut (`Shell.tsx:111-161` has no inbox branch), no menu item, no drawer view - PR #499 flagged this placement as a deviation and offered to move it. Worse: a **running** inbox is invisible outside its own tab - the state badge lives at `modules/inbox/index.tsx:183-185`, `Dock.tsx` has no inbox reference, and the tab icon is a static glyph (`TabStrip.tsx:129-131`). Close the tab and nothing anywhere says a listener is recording.
2. **OAuth2 mock issuer (#479, shipped)**: engine + docs only. Verified: `rg -i issuer app/src` returns zero hits - no endpoint constant, no query module, no Settings card, nothing. The feature exists (`engine/src/http/routes/mock_issuer.cpp:772-816`: start/list/stop/update, each issuer on its own loopback listener) and a user cannot reach it without curl. The MCP tool is deliberately deferred (`docs/engine/mcp.md:757-759`), so the UI gap is total.
3. **Mock server (#481, open)**: plans a per-collection chip and explicitly "not a surface" - which is right contextually, but a third service with a third ad-hoc entry and no shared visibility repeats the pattern.

The shell's own structure says where this belongs: the Dock's left group is a row of **drawer-view switchers** (`Dock.tsx:220-232`, `DRAWER_BUTTONS` :28-81) over `DrawerView = "collections" | "history" | "variables" | "settings"` (`app/src/stores/layout-store.ts:21`), and the Dock's middle is precisely the ambient-status region (`EngineStatus` at `Dock.tsx:142-189` - the app's one live service indicator, with its own guard test).

### 1. A fifth drawer view: `services`

[...]

PR #779, opened 2026-08-17
feat(app/mcp): webhook inboxes over MCP - full lifecycle, captures, and the live canned response

### LINKED ISSUE #756: app(mcp): webhook inboxes - full lifecycle, captures, and canned responses over MCP
Part of epic #753 (S3). The webhook inbox (#480, shipped 0.16.0) is one of the highest-value MCP gaps: "start an inbox, trigger the webhook, assert on the capture" is a natural agent loop, and today none of it is reachable - MCP covers only the OAuth issuer third of the Services drawer.

Zero MCP coverage of a complete engine surface: `POST /inbox/start` (`engine/src/http/routes/inbox.cpp:760`; body all-optional: `port`, `bind`, `confirmNonLoopback`, `response:{status, body, headers, delayMs}`), `GET /inbox` (:838), `PUT /inbox/:id` (canned-response merge-patch, live, :851), `POST /inbox/:id/stop` (:793 - stop keeps record and captures), `DELETE /inbox/:id` (:819, returns `capturesDeleted`), `GET /inbox/:id/requests` (paginated captures, rows are the full captures - no separate detail route, :884), `DELETE /inbox/:id/requests` (clear captures keep listener, :906), `GET /inbox/:id/live` (SSE, single watcher, 409 for a second, :932).

## Fix pathway

Six tools in `app/electron/mcp/tools.ts`, registry pattern:

1. `start_webhook_inbox` (category `execute`, like the issuer - loopback services are ungated by allowlist, rationale mirrors mcp.md ~716-725): inputs `port` (0-65535, 0=auto), optional canned `response` {status, headers, body, delayMs}. **Never send `bind` or `confirmNonLoopback`** - loopback-only over MCP is a stated non-goal of epic #753; say so in the description.
2. `list_webhook_inboxes` (read): running and stopped, URL, port, capture counts.
3. `stop_webhook_inbox` (execute): stop, keep captures - description states the distinction.
4. `delete_webhook_inbox` (write category + `confirmDestructive` when the record holds captures - the engine returns `capturesDeleted`; fetch the inbox first so the prompt states the real count, matching the delete_collection read-before-prompt pattern at tools.ts:1124).
5. `get_inbox_captures` (read): `inboxId` + bounded `limit`/`offset`, newest first; rows are complete captures (method, path, headers, body, remote addr, truncation flag). Also `clear_inbox_captures` (execute or write - match whichever delete_webhook_inbox lands as, minus confirmation; it is recoverable-by-retrigger state).
6. `update_inbox_response` (execute): live merge-patch of the canned response, mirroring `PUT /inbox/:id` fields.

Do NOT expose the live SSE stream: it is single-watcher (409) and the app's inbox tab may hold it - polling `get_inbox_captures` is the MCP shape, and the description should say so (the get_live_metrics bounded-snapshot precedent).

New invalidation entity: `McpDataEntity` today is `collection | request | environment | run | cookie | config` (`tools.ts:64-71`, mirrored in `app/src/types/domain.ts` - production electron code cannot import app/src, the two copies are kept honest by `data-changed.conformance.test.ts`). Add an `inbox` (or broader `service`) entity to BOTH copies + the conformance test, and wire the renderer listener so the Services drawer and inbox tab queries refetch when an MCP call changes inbox state (the drawer polls, but the #363 invalidation channel exists precisely so MCP writes are not invisible until the next poll). Grep note: `useRunningServices` and the ServicesPanel queries are the readers to invalidate - verified they exist at `app/src/modules/services/`.

## Tests

tools.test.ts: canned-response fields forwarded; bind/confirmNonLoopback never present in any payload (assert the absence - this is the mutation-check for the loopback guarantee); delete confirm names the capture count; captures pagination bounded; every new tool declares the new entity where it mutates. Conformance test updated for the new entity. `pnpm test`, `pnpm lint`, `pnpm format:check`.

## Docs - same commit

`docs/engine/mcp.md`: tool table rows + a short "Webhook inboxes" section stating loopback-only and the no-live-stream/poll-captures decision; update the Settings-panel category counts if the doc states them.

## Sequencing

**S3 - after S2 (#755)** in wave M-parity (epic #753). **Lands before S4 (#757)**, which reuses the invalidation entity this issue introduces. Parallel-safe outside the epic.

## Acceptance criteria

- [ ] An agent can start a loopback inbox with a canned response, receive a webhook, read the capture, and clean up - without the app UI open on the Services drawer.
- [ ] No MCP path can bind an inbox to a non-loopback interface.
- [ ] Inbox state changes via MCP become visible in an open Services drawer without waiting for the next poll.
- [ ] Deleting an inbox with captures requires confirmation naming the real count.

---

athrvk:
## Description

The webhook inbox (#480, shipped 0.16.0) had **zero** MCP coverage, so the loop it exists for - start an inbox, trigger the sender, assert on what arrived - was unreachable from an agent; MCP covered only the OAuth-issuer third of the Services drawer.

**Plan.** Root cause: the inbox routes are a complete engine surface (`POST /inbox/start`, `GET /inbox`, `POST /inbox/:id/stop`, `DELETE /inbox/:id`, `GET|DELETE /inbox/:id/requests`, `PUT /inbox/:id`) that the tool registry never followed, and the renderer's `mcp:data-changed` union had no family for engine-hosted services at all - so even a wired-up tool would have been invisible in an open Services drawer until the next poll. Approach: seven tools over those routes in the registry pattern, plus a new `service` entity in **both** copies of the union and an `inboxId` scope hint. The alternative rejected was giving inboxes their own entity name (`inbox`): epic #753 has #757 reusing this family for mock servers and the issuers, and the renderer surfaces that read it (the drawer, the Dock's running-services count) ask "what is listening", not "which kind" - one family with a scope hint is what those readers actually want.

Verified at `682cc9a`: the routes, the field names and the bounds in the issue all still match `engine/src/http/routes/inbox.cpp`, and the two renderer readers it names (`useRunningServiceCount`, the drawer/tab queries under `app/src/queries/inbox.ts`) exist as described.

**Tools** (`app/electron/mcp/tools.ts`, client methods in `engine-client.ts`):

| Tool | Category | Notes |
|---|---|---|
| `start_webhook_inbox` | execute | canned `response` {status, headers, body, delayMs}; loopback-only |
| `list_webhook_inboxes` | read | running and stopped, URL, port, capture count |
| `stop_webhook_inbox` | execute | frees the port, **keeps** the captures - the description says so |
| `delete_webhook_inbox` | write | + confirm; reads `GET /inbox` first so the prompt names the real capture count |
| `get_inbox_captures` | read | 25/call default, 100 max (refused, not clamped), bodies cut to 32 KB |
| `clear_inbox_captures` | write | destroys recorded data, so the toggle; listener survives, so no confirm |
| `update_inbox_response` | execute | live merge-patch; refuses a patch that names no field |

**The loopback guarantee is a guard, not a note.** `inboxStartPayload` never emits `bind` or `confirmNonLoopback`, whatever a call carries, and the schema offers neither - so the non-goal epic #753 records lives in the one function that builds the body, mutation-checked by a test that sends both fields and asserts their absence.

**No live SSE stream**, deliberately: `GET /inbox/:id/live` is single-watcher (a second is a `409`) and the app's own inbox tab may hold it, so polling `get_inbox_captures` is the MCP shape - the `get_live_metrics` bounded-snapshot precedent. A registry-level test locks that as a decision rather than an omission.

**Capture bodies are bounded** through the existing `boundTraceNode` helper rather than a copy of it: a capture carries the same `body` / `bodyTruncated` / `bodyBytes` trio a stored trace node does, `inboxMaxBodyBytes` reaches 8 MB, and a webhook payload is whatever the sender sent - this is #767's case arriving by another route. The engine's original-size count survives the cut.

- New `service` entity in `MCP_DATA_ENTITIES` (main) and `McpDataEntity` (renderer), plus the conformance test and the preload payload type.
- New `inboxId` scope hint, read at the dispatch chokepoint like the other three.
- `INVALIDATORS.service` invalidates `inbox.list()` and **removes** `inbox.captures(inboxId)` for a named inbox. The removal is load-bearing, not a nicety: `useInboxCapturesQuery` *merges* its fetched page into the cache (three writers share that entry - fetch, load-more, the SSE stream - so every write is a union by capture id), so invalidating after a `clear_inbox_captures` unions the destroyed rows straight back. The app's own clear mutation writes an empty page first for exactly this reason; from the main process, dropping the entry is the equivalent. A delete needs the same drop for the reason `useDeleteInboxMutation` already removes rather than invalidates - refetching an id the engine now 404s leaves an error state describing a list that no longer exists.

**Deliberate deviation from the issue spec:** the entity is named `service`, not `inbox`, for the reason above (the issue offered either). Also, the issuer tools' `invalidates: []` comments claimed "no renderer surface reads issuers" - #502's drawer landed and does, so the comments now point at the `service` entity and at #757, which owns the wiring. Comment-only; the issuer tools' behaviour is unchanged and stays #757's.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

App suite, full: **502 files, 5340 tests, all passing** (26 new). `pnpm type-check` clean, `pnpm lint` clean (zero warnings), `prettier --check` clean on every touched file, `mkdocs build --strict` clean. No engine or C++ change - nothing under `engine/` is touched and no engine test could change colour - so no `build.py -t` / `ctest` run. No pre-existing failures observed on the base commit.

Mutation-checked (fix reverted, confirmed red, restored):

| Guard | Mutation that turns it red |
|---|---|
| no inbox payload carries `bind` / `confirmNonLoopback` | forward `bind` in `inboxStartPayload` |
| a capture body past 32 KB is cut, with the engine's `bodyBytes` intact | drop the `boundInboxCaptures` shape |
| an inbox event names the inbox it acted on | drop `inboxId` from `notifyDataChanged` |
| a named inbox has its captures **removed**, not invalidated (spied, and again against a real `QueryClient`) | swap `removeQueries` for `invalidateQueries` |

Also covered: pagination forwarded and refused above the cap / below 1, the bounded-page caveat, the delete preview naming the real capture count (and `0 captured requests` for an empty one), an id the engine does not list refusing without deleting, both write-toggle refusals, the empty merge-patch refusal, categories and destructive hints, and that no tool exposes the live stream.

## Checklist

Docs in the same commit: `docs/engine/mcp.md` (seven tool-table rows, a webhook-inbox notes bullet, the loopback-services and write-toggle/confirmation lists in the safety model, the `mcp:data-changed` families and hints) and `docs/app/state-management.md` (the `service` row and the removal rationale).

## Related Issues
Closes #756

---
_Generated by [Claude Code](https://claude.ai/code/session_01SxPU9xBjgPzXPofErVPkJJ)_

### DIFF
--- a/app/electron/mcp/data-changed.conformance.test.ts
+++ b/app/electron/mcp/data-changed.conformance.test.ts
@@ -37,6 +37,7 @@ const RENDERER_ENTITIES: Record<McpDataEntity, true> = {
 	run: true,
 	cookie: true,
 	config: true,
+	service: true,
 };
 
 describe("McpDataEntity mirror", () => {
--- a/app/electron/mcp/data-changed.test.ts
+++ b/app/electron/mcp/data-changed.test.ts
@@ -47,6 +47,16 @@ function fakeClient(overrides: Partial<Record<keyof EngineClient, unknown>> = {}
 		getRun: vi.fn().mockResolvedValue({ id: "run_1", type: "load", status: "completed" }),
 		deleteRun: vi.fn().mockResolvedValue({ message: "Run deleted successfully" }),
 		setRunBaseline: vi.fn().mockResolvedValue({ id: "run_1", baseline: true }),
+		startInbox: vi
+			.fn()
+			.mockResolvedValue({ inboxId: "inbox_1", url: "http://127.0.0.1:45001" }),
+		listInboxes: vi.fn().mockResolvedValue({
+			data: [{ inboxId: "inbox_1", url: "http://127.0.0.1:45001", captureCount: 2 }],
+		}),
+		stopInbox: vi.fn().mockResolvedValue({ inboxId: "inbox_1", running: false }),
+		deleteInbox: vi.fn().mockResolvedValue({ inboxId: "inbox_1", capturesDeleted: 2 }),
+		clearInboxCaptures: vi.fn().mockResolvedValue({ inboxId: "inbox_1", cleared: 2 }),
+		updateInboxResponse: vi.fn().mockResolvedValue({ inboxId: "inbox_1" }),
 		composeRequest: vi
 			.fn()
 			.mockImplementation((body: { request?: object }) =>
@@ -109,6 +119,14 @@ describe("the registry declares its effects", () => {
 			// lists, so both invalidate `run` exactly as the runners do.
 			set_run_baseline: ["run"],
 			delete_run: ["run"],
+			// Webhook inboxes (#756): every one of these changes what the Services
+			// drawer and the Dock's running-services count answer, and the two
+			// that touch captures change what an open inbox tab is showing.
+			start_webhook_inbox: ["service"],
+			stop_webhook_inbox: ["service"],
+			delete_webhook_inbox: ["service"],
+			clear_inbox_captures: ["service"],
+			update_inbox_response: ["service"],
 		};
 		for (const [name, entities] of Object.entries(expected)) {
 			const tool = TOOLS.find((t) => t.name === name);
@@ -330,6 +348,32 @@ describe("dispatch emits mcp:data-changed", () => {
 		// With the hint, because a stopped run's own report and series are what
 		// changed - it went terminal.
 		expect(onDataChanged).toHaveBeenCalledWith({ entity: "run", runId: "run_1" });
+	});
+
+	test("an inbox call names the inbox it acted on, so the captures cache can be dropped", async () => {
+		const { ctx, onDataChanged } = ctxWithNotifier(fakeClient(), WRITES_ENABLED);
+		const res = await dispatchTool("clear_inbox_captures", { inboxId: "inbox_1" }, ctx);
+		expect(res.isError).toBeFalsy();
+		// The hint is what makes a clear correct renderer-side: an invalidation
+		// alone refetches into a cache that still holds the cleared rows and
+		// unions them straight back (see `lib/mcp-invalidation.ts`).
+		expect(onDataChanged).toHaveBeenCalledWith({ entity: "service", inboxId: "inbox_1" });
+	});
+
+	test("starting an inbox reports the family with no id to narrow by", async () => {
+		const { ctx, onDataChanged } = ctxWithNotifier(fakeClient(), WRITES_ENABLED);
+		const res = await dispatchTool("start_webhook_inbox", { port: 0 }, ctx);
+		expect(res.isError).toBeFalsy();
+		// The engine assigns the id, so there is none in the arguments - and a new
+		// inbox has no capture cache to drop anyway.
+		expect(onDataChanged).toHaveBeenCalledWith({ entity: "service" });
+	});
+
+	test("an unconfirmed inbox delete emits nothing - nothing changed", async () => {
+		const { ctx, onDataChanged } = ctxWithNotifier(fakeClient(), WRITES_ENABLED);
+		const res = await dispatchTool("delete_webhook_inbox", { inboxId: "inbox_1" }, ctx);
+		expect(res.isError).toBeFalsy();
+		expect(onDataChanged).not.toHaveBeenCalled();
 	});
 
 	test("a context without a notifier dispatches normally", async () => {
--- a/app/electron/mcp/engine-client.ts
+++ b/app/electron/mcp/engine-client.ts
@@ -741,6 +741,100 @@ export class EngineClient {
 			signal
 		);
 	}
+
+	// --- Local services: the webhook inbox -----------------------------------
+	//
+	// The inbox routes take a `bind` and a `confirmNonLoopback` flag; nothing
+	// here sends either, so every inbox an MCP call starts listens on the
+	// engine's loopback default. That is a stated non-goal of epic #753 rather
+	// than an omission, and it is enforced by the payload builder in `tools.ts`
+	// (`inboxStartPayload`) - this client takes the payload it is given.
+
+	/**
+	 * Start an inbox: `POST /inbox/start`. Every field of the body is optional;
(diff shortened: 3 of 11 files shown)


athrvk:
Verified against a **live engine** (built at this branch, `vayu-engine 0.18.1` on `:9876`), driving the real tools through `dispatchTool` with a real `EngineClient` - no fakes - since the route paths and the payload shapes are the part unit tests with a fake client cannot prove. Throwaway spec, not committed.

What ran, in one pass: `start_webhook_inbox` with a canned `{status: 202, body: "queued", headers: {X-Mock}}` → a real `POST` to the returned URL, which received exactly that reply → `get_inbox_captures` returned the capture with `method`, `path`, `query` and the JSON body verbatim → `update_inbox_response {status: 503}` and the next trigger got `503` **with the body and headers it did not name still in place** (the merge-patch behaving as documented) → `clear_inbox_captures` reported `cleared: 2` → the delete preview said `0 captured requests` → the confirmed delete removed it from `list_webhook_inboxes`. A second case confirmed a stop leaves the captures readable.

Against the acceptance criteria:

- [x] Start → receive → read → clean up, with no app UI open - the run above.
- [x] No MCP path binds off-loopback - the started record came back `bind: "127.0.0.1"`, `loopback: true`, from a call that passed `bind: "0.0.0.0"` in the unit test and had it dropped.
- [x] Inbox changes become visible without waiting for the poll - the renderer half, covered by `mcp-invalidation.test.ts` against a real `QueryClient` (the emitted events were asserted in the live run too: `{entity: "service", inboxId}` for the id-bearing calls, `{entity: "service"}` for the start).
- [x] Deleting with captures requires confirmation naming the real count.

---

athrvk:
fix merge conflict

Issue #709, opened 2026-08-16
engine(import): a spec bound at import carries no specHash/syncedAt - contract coverage and response validation silently never engage on the primary binding path, and the Spec tab shows "-" for Hash and Bound


athrvk:
## Problem

**Severity: high.** Two of the OpenAPI epic's flagship features - contract coverage (#629) and response-schema validation (#628) - are silently disabled for any collection bound the way nearly every collection gets bound: **by importing a spec**. Live repro: importing GitHub's REST description by URL (`api.github.com.2026-03-10.yaml`, 9.7 MB) succeeds fully - all requests, operation stamps, a working Spec tab - but the tab's Hash and Bound fields render "-", and every run of that collection is planned as "not measured against a contract".

### Mechanism, end to end

1. The import orchestrator sends the binding as `openapi: { specTempId }` (`app/src/services/importers/orchestrator.ts:186`) - app-side, this is the **only** live binding writer.
2. Engine-side, `resolve_spec_binding` (`engine/src/http/routes/import.cpp:323-348`) rewrites `specTempId` into `specId` and stores exactly that: **no `specHash`, no `syncedAt`** - even though the document was stored (and its hash engine-computed) in the same apply.
3. The shared applier validates `specHash`/`syncedAt` *if present* (`collections.cpp:188-206`) and computes nothing. So the import-bound binding is `{"specId": "..."}` forever.
4. The Spec tab reads `binding?.specHash` (`SpecTab.tsx:232`) and renders `shortHash(undefined)` = `"-"` (`:519-521`), and Bound from the absent `syncedAt` (`:465-467`). That is the visible tell.
5. The functional damage is in plan resolution: `scenario_plan.cpp:307` reads `binding.value("specHash", "")`, and `:336` requires `document->hash == spec_hash` before loading declared operations and response schemas. Empty never equals the real hash, so the run takes the `UncheckedReason::HashMismatch` branch: **no coverage block, no schema validation, on every run, forever** - and the collection runner's `checked: false` reason tells the user "the document has moved under the binding," which is false; it was never stamped.
6. The advertised self-heal does not fire: a sync of an unchanged document short-circuits on byte equality as "up to date" **without touching the binding** (this is the reporter's exact trace - sync "worked", hash stayed "-"). Only a sync that *applies changes* stamps the binding (`spec_sync.cpp` updates content/hash/syncedAt together), so a freshly imported, current spec - the healthiest possible state - is the one that can never self-repair.

### Why no test caught it

[...]

PR #785, opened 2026-08-18
fix(app/engine/openapi): clear stamps a re-bind orphans, and reclaim documents nothing can reach

### LINKED ISSUE #718: app/engine(openapi): binding-lifecycle pair - unbind/re-bind leaves stale specOperation stamps lying to the new spec, and spec_documents rows accrete forever with no reclamation path
Two CONFIRMED findings from the adversarial review at `ac27b97`, both lifecycle holes in the binding state machine; filed together because the fixes share the unbind/re-bind/sync code paths.

## Problem A - stale stamps after re-bind

`handleUnbind` clears only the collection binding and file memory (`SpecTab.tsx:208-217`); **no code path anywhere writes null to a request's `specOperation`** (all writers enumerated: bind stamps matches only via `useBindSpecMutation` `specs.ts:183-195`; sync patches matched requests; importers set on create). Re-binding to a different document stamps only the requests that *match* the new spec (`SpecTab.tsx:175-178`) - non-matching requests keep identity stamps from the OLD spec. Coverage resolves stamps operationId-first (`spec_coverage.cpp:227-235`), so a stale stamp whose id happens to exist in the new document claims the wrong operation; export's claim-matching has the same exposure. The refetch error message itself instructs the exact flow that triggers this ("Bind it again from the file or URL it came from", `spec-refetch.ts:120`): unbind, re-bind v2, and the request still calling the v1 path carries a v1 stamp into every v2 run.

## Problem B - unreachable document rows accrete monotonically

Every sync mints a new `spec_documents` row and moves the binding (`spec_sync.cpp:401-404,652-657`); the transaction never deletes the superseded row (`database.cpp:1080-1124`). Deleting a bound collection cascades requests/examples but not the document (`database.cpp:788-832`); unbind leaves it; every re-bind creates a fresh row (`specs.ts:171`). `DELETE /specs/:id` has **no caller** (`api-endpoints.ts:52-55` says so itself) and there is no `GET /specs` list route to even enumerate orphans (`specs.cpp:410-463`). Weekly syncs of a Stripe-sized 12MB document strand ~120MB/quarter of rows nothing can read, reach or reclaim - and #712's meta-read makes the per-row cost more visible, not smaller.

## Fix pathway

- **A**: the bind mutation clears `specOperation` on subtree requests that did NOT match the new document (the match result already enumerates both sides), and unbind gains an explicit choice or recorded rule about stamps (the current "leaves recorded operations exactly as they are" copy is the right default for unbind-then-rebind-same-doc; the *re-bind to a different doc* is where clearing must happen). Coverage's id-first resolution then never sees a cross-document stamp.
- **B**: an engine-side sweep - at startup or post-sync - deletes spec_documents rows bound by zero collections AND referenced by no retained run's `config_snapshot` (runs pin documents for coverage history; the sweep must honor run retention, so tie it to the existing run-pruning pass in `database.cpp`). Record the decision that run-referenced documents live as long as the runs do. No new UI needed; a `GET /specs` list route is optional and separate.

## Tests

- A: bind v1, edit nothing, re-bind v2 where one request matches and one does not - assert the non-matcher's stamp is cleared and coverage counts it as unstamped, mutation-checked.
- B: sync twice, prune runs, assert the orphaned first document is swept and the run-referenced one survives exactly as long as its run; deleting a bound collection sweeps the final orphan.

## Docs - same commit

`docs/engine/db-schema.md` (document lifetime rule), `docs/engine/api-reference.md` (unbind semantics if they change), the Spec tab's unbind copy if the stamp rule gains a choice.

## Sequencing

- A touches the same bind mutation as #709's stamping. (updated 2026-08-17: #709 shipped in 0.18.1, so the gate is met - A can land now against the shipped bind path.)
- B's sweep must not race #709's backfill. (updated 2026-08-17: the backfill shipped with #709, so B's sweep no longer waits - the recorded ordering, backfill first, sweep second, is already satisfied on master.)

## Acceptance criteria

- [ ] After re-bind to a different document, no request carries a stamp from a document its collection is not bound to.
- [ ] Orphaned spec rows are reclaimed on the recorded lifetime rule; a year of weekly syncs holds one live document plus only run-pinned history.
- [ ] Both mutation-checked.

athrvk:
## Description

Two lifecycle holes in the OpenAPI binding state machine, filed together because they share the bind / sync / delete paths.

**Plan.** Root cause, both halves: **nothing in the codebase ever un-references a document.** No writer anywhere sets a request's `specOperation` back to null, and no code path deletes a `spec_documents` row that stopped being reachable - so identity and storage both accumulate against documents the collection has moved off. Approach: give each a single statable rule and put it in one place. **A** - after a bind, a request's `specOperation` is the operation it matched *in the bound document*, or nothing; the bind mutation now writes `null` for the non-matchers, and the tab discloses the count before the user agrees. **B** - a `spec_documents` row lives while a collection binds it *or* a retained run names it, enforced by one sweep hung off the pass that releases run references.

The alternative rejected for **B** was reference-counting documents, or cascading a delete from unbind. Counting needs a column every one of four write paths must maintain correctly forever and is wrong the first time one forgets; cascading from unbind is simply incorrect, since several collections may bind one document and a *run* may still need it. A sweep asking "can anything still reach this?" cannot drift out of step with the writers, because it reads the same two columns they write.

The alternative rejected for **A** was clearing only a stamp the new document does not *declare* (rather than every non-matcher). It sounds gentler and is the actual bug: coverage resolves stamps `operationId`-first, so a v1 stamp surviving because v2 happens to declare the same id is exactly the wrong-operation claim #718 describes.

**Verified at `682cc9a`** (the issue's anchors were written at `ac27b97`): `handleUnbind` still cleared only the collection binding (`SpecTab.tsx:208-217`), `useBindSpecMutation` still wrote stamps and nothing else (`specs.ts:183-195`), `spec_sync_apply` still never deleted the superseded row (`database.cpp`), and `DELETE /specs/:id` still had no caller. Both sequencing gates the issue records are met: #709 shipped in 0.18.1, so the bind path is the shipped one and the backfill-before-sweep ordering is already satisfied on master.

### A - stale stamps after a re-bind

`useBindSpecMutation` gains a required `clearStamps: string[]`, and writes `specOperation: null` for each - the engine's existing null-vs-absent rule, so no engine change was needed. `SpecTab` computes it as *the unmatched requests that carry a stamp*, and:

- the match summary states the count **before** Bind is pressed ("N requests record an operation this document does not have - that identity is cleared"), because this is the one thing a bind rewrites;
- the "nothing is created, deleted or rewritten" copy - now false - says what is actually true;
- a clear that fails is reported **apart from** a stamp that failed, because the state it leaves is the worse of the two: a surviving stamp still reads as identity, and it is identity in a document nothing is bound to.

Unbind is untouched, deliberately, per the issue: it still leaves stamps exactly as they are, so unbind-then-bind-the-same-document is lossless.

### B - unreachable `spec_documents` rows

New `Database::sweep_orphaned_spec_documents()`, returning how many rows went, implementing one rule:

> A document lives while a collection binds it, **or** while a retained run names it in `runs.config_snapshot` under `scenario.openapi.specId`.

Called from three places, each with its own reason: `prune_runs_configured()` (the pass that *releases* run references - which is what puts the sweep on a startup and the end of every run with no schedule of its own), `spec_sync_apply()` (the accretion source), and `delete_collection()` (the last binder going away).

Three properties are load-bearing rather than incidental:

- **A 10-minute grace window** (`SPEC_DOCUMENT_SWEEP_GRACE_MS`). Binding is three writes over three requests and the *document is the first of them*, so between `POST /specs` and the `PUT /collections/:id` that names it a document is bound by nobody and pinned by no run - indistinguishable from an orphan. Without this, a run finishing in that window deletes the document a bind is midway through storing.
- **Cheapest question first, stopping at the first that leaves nothing at stake.** It rides on every run completion, so what it costs when there is nothing to reclaim is what it costs: age, then bindings, then - only if a candidate survives both - the runs table, which is the wide read. It never reads `content` on any path, which is the one column that reaches the 10 MiB cap.
- **It never throws.** All three callers reach it as the tail of an operation whose success does not depend on it; a failed sweep must not turn into a failed sync.

Blast radius traced before writing: every reader of a stored document reaches it through a *binding* (`useSpecQuery(binding?.specId)`, `useBoundSpecReader`, `scenario_plan.cpp` at plan time, the specs routes). Nothing reads a document by a run's stamped `specId`, and a run's coverage is computed at plan time and stored on the run - so a swept document breaks no reader. The scenario manifest is stamped into `config_snapshot` at run *creation*, so a run pins its document from the moment its row exists, with no window where an in-flight run is unprotected.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

**Engine:** `python build.py -e -t` clean, `ctest --preset linux-dev` **1996 tests, all passing** (9 new). **App:** **503 files, 5317 tests, all passing** (7 new; 502/5310 at the base commit). `pnpm type-check` clean, `pnpm lint` clean (zero warnings), `mkdocs build --strict` clean.

`prettier --check` clean on every touched source file - all four were clean at `682cc9a` and were formatted. The three Markdown files under `docs/` were **not** prettier-clean at the base commit and are left as they are, per the repo rule about formatting only files that were clean before. C++ formatted with `git clang-format` against the base, so only the lines this PR adds moved.

Each guard below was mutation-checked (mutation applied, whole suite run, confirmed red, reverted) - and in the engine's case the mutations were run in two batches so each failing set names the line it belongs to, rather than one build failing everything at once:

| Guard | Mutation that turns it red | Result |
|---|---|---|
| a bind clears identity the new document does not account for, and stamps only the matches | `clearStamps: staleStamps` -> `[]` in `SpecTab` | 1 red |
| the clear is written as `null`, not an omitted key | send the patch without `specOperation` | 1 red |
| a failed clear is reported apart from a failed stamp | fold `failedClears` into `failedStamps` | 2 red |
| pruning runs reclaims what the last pruned run held; deleting the last binder reclaims the document; a sync reclaims what it superseded | drop all three `sweep_orphaned_spec_documents()` call sites | exactly 3 red, one per call site (`PruningRunsReclaims...`, `DeletingTheLastBinder...`, `SyncingReclaims...`) |
| a document a retained run names survives (and a run naming a *different* document does not shield it) | drop the run-pin filter | 3 red |
| a document inside the grace window is spared | drop the `fetched_at` gate | 1 red |

Also covered, behaviourally rather than by mutation: the sweep takes an unreachable document and leaves a bound one; a document goes with the *last* run that named it (asserted by deleting that run and sweeping again); deleting one of two binders reclaims nothing and deleting the second reclaims it; a corrupt (unparseable) `collections.openapi` binds nothing rather than pinning a document forever; the disclosure line is absent when nothing is stale; a request with no stamp is left alone rather than written to; and the binding still holds when a clear fails, rather than being rolled back.

## Checklist

Docs in the same commit: `docs/engine/db-schema.md` gains the **lifetime rule** beside the "no cascade" section (the rule, the three callers, the grace window, and why run-referenced documents live as long as the runs); `docs/engine/api-reference.md` states it as API-visible behaviour under **Specs**, since a client that stores a document and never binds it will eventually see a `404`; `docs/app/openapi.md` gains **"Identity from another document is cleared"** beside the binding section and the document-lifetime paragraph beside the ownership one. The `SpecTab` header comment and the tab's own copy both carried the now-false "rewrites nothing" claim and say what happens instead.

## Related Issues

Closes #718

---
_Generated by [Claude Code](https://claude.ai/code/session_014Ljj1KuTAstpT4S9Rz39qg)_

### DIFF
--- a/app/src/modules/collections/CollectionDetail/SpecTab.test.tsx
+++ b/app/src/modules/collections/CollectionDetail/SpecTab.test.tsx
@@ -42,7 +42,7 @@ const bindSpec = {
 	isPending: false,
 	isError: false,
 	error: null as Error | null,
-	data: undefined as { failedStamps: string[] } | undefined,
+	data: undefined as { failedStamps: string[]; failedClears: string[] } | undefined,
 };
 
 const syncSpec = {
@@ -169,7 +169,7 @@ async function pickFile(name: string, text: string) {
 /** Resolve the bind the way TanStack would, running the caller's onSuccess. */
 function bindSucceeds() {
 	const [, options] = bindSpec.mutate.mock.calls[bindSpec.mutate.mock.calls.length - 1];
-	options?.onSuccess?.({ stamped: 0, failedStamps: [] });
+	options?.onSuccess?.({ stamped: 0, failedStamps: [], failedClears: [] });
 }
 
 beforeEach(() => {
@@ -256,6 +256,67 @@ describe("binding a collection that already has requests", () => {
 		expect(useSpecFileStore.getState().locations.col_1).toBeUndefined();
 	});
 
+	/*
+	 * Re-binding to a different document (issue #718). Nothing writes null to a
+	 * request's `specOperation` anywhere else in the app, so a stamp the new
+	 * document does not account for would otherwise be permanent - and coverage
+	 * resolves stamps by `operationId` first, so it would claim whichever
+	 * operation of the new document shares the id rather than simply going
+	 * unread.
+	 */
+	it("clears identity recorded against a document this one is not", async () => {
+		requests = [
+			request("r1", "{{baseUrl}}/pets"),
+			// Matched the *old* document, matches nothing here: its URL is not a
+			// path this spec declares.
+			request("r2", "{{baseUrl}}/v1/legacy", {
+				operationId: "legacyPing",
+				method: "GET",
+				path: "/v1/legacy",
+			}),
+			// No stamp to clear - an unmatched request that never had identity is
+			// left exactly as it was.
+			request("r3", "{{baseUrl}}/health"),
+		];
+		render(<SpecTab collection={collection()} />);
+		await pickFile("petstore.json", OPENAPI);
+
+		fireEvent.click(screen.getByRole("button", { name: /bind this spec/i }));
+
+		const [payload] = bindSpec.mutate.mock.calls[0];
+		expect(payload.clearStamps).toEqual(["r2"]);
+		// The matched request is stamped, not cleared - the two lists are disjoint,
+		// which is what lets the mutation write both at once.
+		expect(payload.stamps.map((s: { requestId: string }) => s.requestId)).toEqual(["r1"]);
+	});
+
+	it("discloses the clearing before it happens, and stays silent when there is none", async () => {
+		requests = [
+			request("r1", "{{baseUrl}}/pets"),
+			request("r2", "{{baseUrl}}/v1/legacy", {
+				operationId: "legacyPing",
+				method: "GET",
+				path: "/v1/legacy",
+			}),
+		];
+		const { unmount } = render(<SpecTab collection={collection()} />);
+		await pickFile("petstore.json", OPENAPI);
+
+		// The one line on this screen that describes a write to something already
+		// there, so the user reads it before pressing Bind.
+		expect(
+			screen.getByText(/1 request records an operation this document does not have/i)
+		).toBeTruthy();
+		unmount();
+
+		// A collection with nothing stale says nothing - a zero here would read as
+		// a warning about a bind that rewrites nothing.
+		requests = [request("r1", "{{baseUrl}}/pets"), request("r3", "{{baseUrl}}/health")];
+		render(<SpecTab collection={collection()} />);
+		await pickFile("petstore.json", OPENAPI);
+		expect(screen.queryByText(/does not have/i)).toBeNull();
+	});
+
 	it("refuses a document that is not a spec, by name, and offers no bind", async () => {
 		render(<SpecTab collection={collection()} />);
 
@@ -325,8 +386,18 @@ describe("a bound collection", () => {
 	});
 
 	it("says which requests kept no identity when some stamps failed", async () => {
-		bindSpec.data = { failedStamps: ["req_9"] };
+		bindSpec.data = { failedStamps: ["req_9"], failedClears: [] };
 		render(<SpecTab collection={bound()} />);
 		expect(screen.getByText(/1 request kept no operation identity/i)).toBeTruthy();
+	});
+
+	// The other half of the same report, and the worse state of the two: a stamp
+	// that survived still reads as identity, in a document nothing is bound to.
+	it("says which requests still record another document's operation", async () => {
+		bindSpec.data = { failedStamps: [], failedClears: ["req_9"] };
+		render(<SpecTab collection={bound()} />);
+		expect(
+			screen.getByText(/1 request still records an operation of another document/i)
+		).toBeTruthy();
 	});
 });
--- a/app/src/modules/collections/CollectionDetail/SpecTab.tsx
+++ b/app/src/modules/collections/CollectionDetail/SpecTab.tsx
@@ -24,12 +24,18 @@
  *   `spec-file-store` and never reaches the engine. A URL-sourced spec keeps its
  *   origin portably instead, in `spec_documents.source_url`.
  *
- * **Binding** deliberately creates, deletes and edits nothing: it matches what
- * is already there and stamps identity on the matches - the operations with no
+ * **Binding** deliberately creates and deletes nothing: it matches what is
+ * already there and stamps identity on the matches - the operations with no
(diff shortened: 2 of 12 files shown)


athrvk:
CI gate went red on the first run because **Engine on ubuntu-latest was cancelled at the 60-minute cap**, not because anything failed.

The job died in dependency install, before cmake ever ran: the Azure apt mirror was unreachable (a run of `Ign: http://azure.archive.ubuntu.com/...`), it fell back to `archive.ubuntu.com`, and then produced no output at all from `02:44:04` until the runner was killed at `03:43:33`. No compile, no test body, and the junit step then reported "No test results found" because there were none to find.

The same engine suite passed on the other two platforms in that run - **Engine on macos-latest** (12 min) and **Engine on windows-latest** (27 min) both green, each having built and run the 9 new C++ tests this PR adds. All three app jobs were green too.

So this is the one case where a re-run is the fix rather than a way of avoiding a diagnosis - the job never reached the code. I have re-queued the failed jobs and will report where they land.

---

Issue #795, opened 2026-08-18
app(mcp): `verifySSL` is stored on every request and writable from no MCP tool - the CRUD half of #706 that its own PR deferred


athrvk:
Filed from #759 (document CRUD parity), which was told in its Sequencing section to leave this field to the transport epic's own PR: "when transport epic phases #706/#707 land per-request `verifySSL`/client-cert fields, extend the same CRUD inputs in that PR, not this one". #706 has merged and the CRUD inputs were not extended, so the field is now stored, sent on every transfer, editable in the app - and unreachable over MCP.

## Problem (verified at a48d08d)

`apply_request_fields` (`engine/src/http/routes/requests.cpp:230`) stores `verifySSL` on the row beside `followRedirects` / `maxRedirects` / `httpVersion` / `stream`, and `request_composer.cpp:567` emits it on every composed payload under the never-elided rule - "the one field where eliding the default would be a security bug rather than a surprise". The app edits it in the request builder's **Settings** tab.

MCP writes the other four (#759 added them to `create_request` / `update_request`) and not this one. So:

- An agent cannot save a request against a host with a self-signed certificate, which is the entire case the setting exists for - it can only ask a human to tick the box.
- An agent cannot *read back* what it wrote and see the security-relevant half of it, because the field also appears in `list_requests` output while no tool can restate it.
- `run_request` has no per-call `verifySSL` either, so an agent cannot even send once with verification off; the composed payload always carries the stored value.

## Fix pathway

1. **`create_request` / `update_request`** (`app/electron/mcp/tools.ts`): add `verifySSL` to `requestSettingsInput()` and to `REQUEST_SETTINGS_KEYS`, which is the one list the payload builder reads - both verbs then carry it under the merge-patch rule the other four follow. Delete the comment in that function that defers it, and the `verifySSL` clause in `docs/engine/mcp.md`'s saved-request bullet.
2. **Decide, and record, whether `run_request` takes a per-call `verifySSL`.** The engine's `POST /execute` accepts one; the argument against offering it is that turning off verification for one agent-issued send is exactly the kind of thing the safety model exists to gate, and unlike the stored field there is no UI where a human ever saw it. If it is offered, it belongs beside the allowlist in the safety config rather than as a free-form argument - a `allowInsecureTls` toggle, default off, with the tool refusing the argument when it is off. Either answer is fine; an undecided one is not, and this issue is where it gets written down.
[...]

Issue #706, opened 2026-08-16
app/engine(tls): transport policy phase 2 - additive custom CAs per platform, and verify-SSL made real end-to-end (stored, composed, exposed, imported from -k)


athrvk:
Phase 2 of #704, on the phase-1 (#705) substrate. This issue also **retro-files the untracked deferral** from commit `964c220a` ("verifySSL stays engine-only; exposing it weakens transport security and was deferred" - no issue number cited, contra root CLAUDE.md's deferral rule; canonized at `engine/CLAUDE.md:226`).

## Problem

Two halves of one trust story:

**Custom CAs.** Verification today is fully native (verified per platform at the pinned vcpkg baseline: Windows = Schannel, macOS = SecTrust via `USE_APPLE_SECTRUST=ON`, Linux = OpenSSL defaults + `CURL_CA_FALLBACK`) and there is **no hook for a user-supplied CA** - `CURLOPT_CAINFO`/`CAPATH` appear nowhere. A corporate TLS-inspecting proxy or a self-signed internal CA leaves every HTTPS request failing with no in-app recourse.

**verify_ssl.** The knob exists on the wire (`utils/json.cpp:800-802`, default `true` at `types.hpp:265`, honored identically by all three drivers) and is unreachable: no column in `requests`, absent from both stored-request serializers, absent from `POST /compose`'s `payload_from_stored` (`request_composer.cpp:500-502`), no UI writer. It is documented as a public payload field (`docs/engine/api-reference.md:2445`) that no Vayu surface can set.

## Fix pathway

**1. Custom CAs - additive, per-backend-honest.**
- A settings entry (category `network_performance`) holding user-supplied PEM(s) - pasted content stored in config, engine materializes a bundle file under its data dir. Content, not a path: a path breaks when the file moves and cannot be shown in the UI.
- **Native trust is preserved; the user's CAs extend it** (epic decision 4). The mechanism differs per backend, and the implementer must verify and document each rather than assume uniformity: on OpenSSL-backed platforms `CURLOPT_CAINFO` *replaces* the default bundle, so the materialized file is system-bundle + user PEMs merged on Linux; on macOS SecTrust and Windows Schannel, verify what curl 8.21 actually does when CAINFO is set alongside the native verifier, and disclose the resulting behavior in the docs per platform - a wrong claim here is a security claim.
[...]

PR #797, opened 2026-08-18
feat(app/mcp): load-run recording knobs, OAuth2 token tools, scripting types resource

### LINKED ISSUE #760: app(mcp): load-run knob parity (recording, redirects, comment), OAuth2 token tools, scripting types resource, and the capacity omission in suggest_load_profile

athrvk:
## Description

The S6 batch of epic #753: four smaller parity items sharing `start_load_run`'s region and the read surface.

**Plan.** The root cause is not one omission but three of a kind - the load dialog has always sent knobs that decide what a run *keeps* (rather than what it sends), the app has always driven the engine's OAuth 2.0 token cache, and the engine has always served the sandbox's `.d.ts`; the MCP layer followed none of them, so an agent could shape a run but not ask it to capture the slow tail, could pass an oauth2 block but not find out why it was 401ing, and could read the sandbox's names but never a signature. The approach is additive at the tool/resource layer over endpoints that already exist: extend `start_load_run`'s schema and map the four knobs onto the engine's own config keys, add three tools over `/oauth2/token`, and add one resource over `GET /scripting/types`. The alternative I rejected for the token tools was returning the access token the engine hands back - it is what `POST /oauth2/token` answers with, and passing it through would have been the smaller diff - but the engine is what applies a token to a request, so the bytes buy an agent nothing it can use through Vayu, while `acquire_token` answers a **cache hit before it looks at the grant**, which would have made `fetch_oauth2_token` a way to read back a credential the user acquired in a browser flow no agent could complete. Redaction plus an up-front `authorization_code` refusal closes that shape instead.

**What changed**

- **Recording knobs.** `successSamplePeriod`, `slowRequestThresholdMs`, `saveTimingBreakdown` and `comment`, forwarded as `success_sample_rate` / `slow_threshold_ms` / `save_timing_breakdown` / `comment`. The argument is named for what the value means and the payload key stays the engine's: `success_sample_rate` is a **period** (keep 1 in N), not the percentage its name suggests - the app's own `StartLoadTestRequest` carries a comment saying so, and `LoadTestConfigDialog` converts from the slider's percentage before sending. Bounds mirror `validate_run_config` (1-100000, 0-86400000), not the dialog's narrower sliders, so a value the schema accepts is one `POST /runs` accepts. All four apply to a **scenario** run as well - both executors read them off the one `RunContext`, and `comment` is lifted into the run summary and `metadata.configuration` whichever produced the run - so forwarding on one path only would have been a knob that silently did nothing on the other.
- **Redirect policy.** `followRedirects` / `maxRedirects` on `start_load_run`, through `readRequestOverrides` and so through `POST /compose`, beside `httpVersion`: they describe the *request*, the engine emits them on every composed payload under the never-elided rule, and a URL-only run has no stored row whose policy it could inherit. A scenario run refuses both by name, as it does every other single-target field. `maxRedirects` is bounded 0-100 because `POST /runs` has **no** guard of its own and the value reaches `CURLOPT_MAXREDIRS`, where a negative means unlimited.
- **OAuth 2.0 token tools.** `fetch_oauth2_token` (execute; allowlist checked on `accessTokenUrl` **and** `refreshTokenUrl` - a gate that read one of two URLs is a gate a config can walk around), `get_oauth2_token_status` (read; an absent entry is `found: false`, not an error), `clear_oauth2_token` (write toggle, idempotent, no confirmation - the same posture `clear_cookies` takes for the same reason). Neither returns access-token bytes; the projection keeps the entry's shape and states `accessTokenWithheld`, because an agent that finds no token and is not told why concludes the acquisition half-failed. The interactive grant is refused with the app's Auth tab named as where to do it.
- **`vayu://scripting/types`.** The `.d.ts` the app's editor feeds to its TypeScript worker, on the completions resource's non-truncation posture (throw rather than serve a partial surface, `resources.ts` precedent). Two resources because a name and a signature are different questions.
- **Prompts.** `suggest_load_profile` now names all five modes - it omitted `capacity` for as long as the tool has had it - and says when to reach for it rather than only listing it; the test measures the enumeration against `safety.ts`'s `KNOWN_LOAD_MODES`, the one list that cannot fall behind the engine. `compare_runs`' `baseRunId` is optional and resolves the target's pinned baseline through the **same** `resolveBaseline` the tool uses (now exported rather than copied), so the prompt stops demanding an id Vayu already knew.

**App changes (required by the issue).** The issue said "none - verify the token-status card polls or listens". It does both: `useOAuth2TokenStatusQuery` is keyed at `queryKeys.oauth.token(cacheKey)` and polls at 30s. That is long enough for an agent to clear a token, report it, and leave the row saying the entry is valid, so the two writing tools declare a new **`oauth`** data family and the renderer invalidates `queryKeys.oauth.all` on it. Invalidating `config` (the issue's fallback suggestion) would have been an unrelated cache. The prefix rather than the one key, because the event carries no cache key: an agent names the key it clears, but the key a fetch writes under is derived engine-side and appears only in the answer - a hint present for one tool and absent for the other is the shape that leaves a stale row exactly when it matters.

**Deliberate deviations from the issue spec**

- The issue named the argument `successSampleRate` with a range of 1-100. The stored field is a period with an engine range of 1-100000, and the app converts a percentage into it before sending; a 1-100 "rate" would have refused legal values and invited an agent to pass 50 meaning half. Named `successSamplePeriod` and described as the period it is.
- The issue offered "align the `compare_runs` prompt to accept an omitted base, or state why not". Aligned.
- No new safety-config entry. The issue anticipated one only for a per-call `verifySSL` on `run_request`, which belongs to #795 (and to #706's transport epic), not here.

**Rebased onto #796.** Document CRUD parity landed while this was open and both branches added a block after `authInput` in `tools.ts` - kept both, in master's order then this branch's. One de-duplication the merge made necessary: `requestSettingsInput`'s `maxRedirects` bound and this branch's `MAX_REDIRECTS_BOUND` were the same engine clamp written twice in one file, so the schema now reads the constant.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Local, on the merged tree (base `f79fd6b`):

- `cd app && pnpm test` - **508 files, 5536 tests passed**. Per-file test counts were checked against both merge parents to prove the merge dropped nothing: every changed test file equals `master + this branch - base`.
- `cd app && pnpm type-check` - clean. `data-changed.conformance.test.ts` failed to compile until the renderer's union named `oauth`, which is the guard doing its job.
- `cd app && pnpm lint` and `pnpm format:check` - clean, zero warnings. `docs/engine/mcp.md` and `docs/app/state-management.md` were already not prettier-clean before this branch and were left unformatted per the repo rule.
- `mkdocs build --strict -f .github/mkdocs.yml` - clean, before and after the merge.
- No engine build or ctest run: the diff is TypeScript and Markdown only, so no C++ test could change colour.

New coverage worth naming, each written so reverting the fix fails it:

- The three recording knobs arrive under the engine's snake_case keys **and** the camelCase argument names do not also reach the engine; a knob the caller did not name stays absent rather than being defaulted.
- `successSamplePeriod: 0` is refused at the schema (it is a modulo divisor engine-side), `slowRequestThresholdMs: 0` is accepted and `-1` refused, `maxRedirects: -1` refused.
- A scenario load run carries all four knobs, and refuses `followRedirects` with the reason naming the saved request.
- `fetch_oauth2_token` never returns the bearer: the result is asserted whole, so a passthrough fails it. `force: false` is not forwarded; an `authorization_code` config is refused with nothing reaching the engine; a `refreshTokenUrl` on an off-list host is refused after the token URL passed.
- `clear_oauth2_token` is refused while writes are off with nothing reaching the engine.
- `vayu://scripting/types` serves a declaration the app has never heard of, and throws on an empty or missing `typeDefinitions` rather than serving a partial surface.
- The `oauth` invalidator reaches a status query cached under a key this side never saw.

## Checklist

## Related Issues

Closes #760

### DIFF
--- a/app/electron/mcp/data-changed.conformance.test.ts
+++ b/app/electron/mcp/data-changed.conformance.test.ts
@@ -38,6 +38,7 @@ const RENDERER_ENTITIES: Record<McpDataEntity, true> = {
 	cookie: true,
 	config: true,
 	service: true,
+	oauth: true,
 };
 
 describe("McpDataEntity mirror", () => {
--- a/app/electron/mcp/data-changed.test.ts
+++ b/app/electron/mcp/data-changed.test.ts
@@ -74,6 +74,11 @@ function fakeClient(overrides: Partial<Record<keyof EngineClient, unknown>> = {}
 		startMockIssuer: vi.fn().mockResolvedValue({ issuerId: "issuer_1" }),
 		stopMockIssuer: vi.fn().mockResolvedValue({ stopped: true }),
 		updateMockIssuer: vi.fn().mockResolvedValue({ issuerId: "issuer_1", failureMode: "slow" }),
+		fetchOAuth2Token: vi
+			.fn()
+			.mockResolvedValue({ cacheKey: "key_1", accessToken: "bearer", expiresIn: 3600 }),
+		getOAuth2TokenStatus: vi.fn().mockResolvedValue({ found: true, expired: false }),
+		clearOAuth2Token: vi.fn().mockResolvedValue({ deleted: true }),
 		composeRequest: vi
 			.fn()
 			.mockImplementation((body: { request?: object }) =>
@@ -162,6 +167,11 @@ describe("the registry declares its effects", () => {
 			start_mock_issuer: ["service"],
 			stop_mock_issuer: ["service"],
 			update_mock_issuer: ["service"],
+			// The OAuth 2.0 token cache (#760): both change what the auth tab's
+			// token-status row reports, and it polls at 30s - long enough to keep
+			// showing an entry an agent has already cleared.
+			fetch_oauth2_token: ["oauth"],
+			clear_oauth2_token: ["oauth"],
 		};
 		for (const [name, entities] of Object.entries(expected)) {
 			const tool = TOOLS.find((t) => t.name === name);
@@ -456,6 +466,37 @@ describe("dispatch emits mcp:data-changed", () => {
 		const res = await dispatchTool("update_mock_issuer", { issuerId: "issuer_1" }, ctx);
 		expect(res.isError).toBe(true);
 		expect(client.updateMockIssuer).not.toHaveBeenCalled();
+		expect(onDataChanged).not.toHaveBeenCalled();
+	});
+
+	test("a token fetch and a token clear report the oauth family", async () => {
+		const acquired = ctxWithNotifier(fakeClient(), { allowlist: ["id.example.com"] });
+		const fetched = await dispatchTool(
+			"fetch_oauth2_token",
+			{
+				config: {
+					grantType: "client_credentials",
+					accessTokenUrl: "https://id.example.com/token",
+					clientId: "client_a",
+				},
+			},
+			acquired.ctx
+		);
+		expect(fetched.isError).toBeFalsy();
+		// No scope hint: the key a fetch writes under is derived engine-side and
+		// appears only in the answer, so the family is invalidated at its prefix.
+		expect(acquired.onDataChanged).toHaveBeenCalledWith({ entity: "oauth" });
+
+		const cleared = ctxWithNotifier(fakeClient(), WRITES_ENABLED);
+		const res = await dispatchTool("clear_oauth2_token", { cacheKey: "key_1" }, cleared.ctx);
+		expect(res.isError).toBeFalsy();
+		expect(cleared.onDataChanged).toHaveBeenCalledWith({ entity: "oauth" });
+	});
+
+	test("reading a token's status emits nothing", async () => {
+		const { ctx, onDataChanged } = ctxWithNotifier(fakeClient());
+		const res = await dispatchTool("get_oauth2_token_status", { cacheKey: "key_1" }, ctx);
+		expect(res.isError).toBeFalsy();
 		expect(onDataChanged).not.toHaveBeenCalled();
 	});
 
--- a/app/electron/mcp/engine-client.ts
+++ b/app/electron/mcp/engine-client.ts
@@ -430,6 +430,18 @@ export class EngineClient {
 		return this.request("GET", "/scripting/completions", undefined, signal);
 	}
 
+	/**
+	 * The same surface as TypeScript declarations: `GET /scripting/types` answers
+	 * `{version, engine, libUri, typeDefinitions}`, the `.d.ts` text generated
+	 * from the completion table by `script_types.cpp`. The app feeds it to
+	 * Monaco's TypeScript worker; MCP serves it as a resource, because a
+	 * signature an agent can read is what turns a name into a call it gets right
+	 * the first time.
+	 */
+	getScriptTypeDefinitions(signal?: AbortSignal): Promise<unknown> {
+		return this.request("GET", "/scripting/types", undefined, signal);
+	}
+
 	updateConfig(payload: unknown, signal?: AbortSignal): Promise<unknown> {
 		return this.request("POST", "/config", payload, signal);
 	}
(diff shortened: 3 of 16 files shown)


athrvk:
fix merge conflicts

athrvk:
Conflicts resolved - merged `master` (now at `f79fd6b`, with #796 in it) into the branch and pushed.

One real conflict, in `tools.ts`: both branches added a block immediately after `authInput` - #796's `storedAuthInput` / `requestSettingsInput` / `applyRequestSettings`, and this one's OAuth 2.0 helpers. Kept both, in master's order then this branch's. That left the same engine clamp written twice in one file (`requestSettingsInput`'s `maxRedirects` max and this branch's `MAX_REDIRECTS_BOUND`, both 0-100 from `apply_request_fields`), so the schema now reads the constant.

Everything else auto-merged. Verified on the merged tree: `pnpm test` 508 files / 5536 tests passed, `type-check` / `lint` / `format:check` clean, `mkdocs build --strict` clean. I also compared per-file test counts against both merge parents - every changed test file equals `master + this branch - base`, so the merge dropped no coverage from either side.

---

PR #809, opened 2026-08-18
feat(app/engine/spec): a metadata-only spec read, and a Spec tab that shows a skeleton instead of blanks (#712)

### LINKED ISSUE #712: app/engine(spec): the Spec tab's first open renders blanks while it downloads the entire document - add a real loading state, and a metadata-only read so a 12 MB spec is not the price of a URL and a date
## Problem

Opening the Spec tab for the first time on a bound collection shows a half-rendered card: the source row is blank, "Fetched" shows "…", and "Operations mapped" flashes 0 before the real counts arrive. On real-world specs the wait is long enough to notice - and the reason is structural, not cosmetic.

**What renders during load:** `SpecTab.tsx` passes `specLoading` into `BoundSpec`, where it gates exactly one cell - `loading ? "…" : formatRelative(...)` on Fetched (`:461`). The source line (`:440-446`) reads `sourceUrl`/`fileName` off the still-absent document and renders an icon beside an empty string; the match summary counts arrive from separate requests queries and pop in from zero. There is no skeleton, no unified pending state - only the already-good *failure* state (`:469-473`).

**Why the load is slow at all:** `useSpecQuery` → `apiService.getSpec` (`api.ts:245-252`), whose own comment states the cause: "One stored document, **`content` included - the engine has no metadata-only read**, and the Spec tab needs `sourceUrl` and `fetchedAt`, which live on the document." So first open transfers the full stored document - **12 MB for Stripe's spec, 9.7 MB for GitHub's** - to paint a URL and a date. `queries/specs.ts:44` sets `staleTime: Infinity`, which is why it is precisely a first-open experience.

## Fix pathway

**1. Engine: a metadata-only read.** `GET /specs/:id/meta` (or a `?meta=1` variant on the existing route - implementer's pick, disclosed in the API reference) returning everything but `content`: id, `sourceUrl`, `fetchedAt`, hash, content type, and sizes. The full-content read stays for the consumers that genuinely need the document (export via `useOpenApiExportSource`, the sync/diff flow, bound-spec matching at `queries/specs.ts:96`) - those already fetch on user action, not on tab open.

**2. App: the tab's info card queries the meta shape.** The 12 MB transfer disappears from first open entirely; the query comment at `queries/specs.ts:33` ("the engine has no metadata-only read, so this pulls content too") comes out, having stopped being true.

**3. A real pending state.** While the meta query (and the requests queries feeding the match summary) are unsettled, the card renders `Skeleton` rows (`app/src/components/ui/skeleton.tsx` - the primitive exists; the hand-rolled-copy rule applies) in place of the source line, the hash/mapped/fetched/bound grid, and the match summary - never a blank string beside an icon, never a 0-of-0 flash. The existing failure state stays as-is. The coverage line (`SpecCoverageLine`) manages its own fetch - give it the same treatment if it currently pops.

## Tests

- App: render the tab with the spec query pending and assert the skeleton state (class-assertion style per app/CLAUDE.md - jsdom has no layout); settle the query and assert the real card; the failure state's existing test stays green.
- Engine: the meta route returns the document row minus `content` (assert `content` absent, not empty - absent-not-empty is the repo's spelling), 404 shape matching the full route.
- The flash is the mutation check: with the skeleton wired, a pending query must never render the empty-string source row - assert the blank-row markup is unreachable while pending.

## Docs - same commit

`docs/engine/api-reference.md` (the new read), `docs/app/api-integration.md` (the tab's query shape change).

## Sequencing

- Standalone. (updated 2026-08-17: #709 shipped in 0.18.1, so the coordinate-merges clause is resolved - `SpecTab.tsx` no longer has a #709 fix in flight to merge around.)
- (updated 2026-08-17) With #709's fix shipped, the card's Hash/Bound cells show real values today; this issue's skeleton covers their *loading* moment, the shipped fix covers their *content*.
- Residual coordination: this issue brushes `SpecTab.tsx` near #718's item A (stale specOperation stamps), which is still open - coordinate merges, do not couple.

## Acceptance criteria

- [ ] First open of the Spec tab on a freshly imported large spec shows a skeleton card, then complete data - no blank source row, no 0-of-0 flash, at no point a full-document transfer.
- [ ] The full-content read still backs export, sync and matching, fetched on action.
- [ ] Engine meta read tested absent-not-empty; app pending-state test proves the blank markup unreachable.

athrvk:
## Description

**Root cause.** The Spec tab's card needs `sourceUrl` and `fetchedAt`, and both live on the *document* rather than on the collection's binding - so `useSpecQuery` -> `GET /specs/:id` pulled the whole stored document on first open (12 MB for Stripe's spec, 9.7 MB for GitHub's) to paint a URL and a date, with `staleTime: Infinity` making it precisely a first-open experience. While it was in flight the card was not merely blank: the source line rendered `sourceUrl ?? fileName ?? "a document stored with this collection"`, which for a URL-imported spec is a false statement until the read lands, and the mapped count flashed `0 of 0`.

**Approach.** Split describing a document from reading one.

- Engine: `GET /specs/:id/meta` returns `id`, `sourceUrl`, `fetchedAt`, `hash` and `contentBytes`, with `content`, `operations` and `responseSchemas` **absent, not empty**. `serialize_meta` builds the meta shape by narrowing the full serialization rather than spelling the fields a second time, and the test asserts every key it carries equals the full read's value for that key - so the two views of one row cannot drift when a field is added. `404` is the same body as the full read.
- App: the card reads meta (`useSpecMetaQuery`); the full read stays for the callers that need the text - export, the import dialog's bound-spec match, and the Sync section's comparison - each fetched on the action that needs it.
- Sync: `Check for changes` was disabled until the tab's background full-document query had answered ("The stored document has to load before it can be compared"). It now takes the binding's `specId` and reads the stored bytes **on the click**, through `fetchQuery` on the document's own key so an export or a second check reuses the read. The gate and its note are gone with the transfer they waited on; the existing `checking` spinner already covers the wait honestly.
- Pending state: cells whose input is unknown render the `Skeleton` primitive - the source line and `Fetched`/`Size` while the meta read is in flight, `Operations mapped` while the subtree's request lists are. `Hash` and `Bound` come off the collection's binding, which the tab has before it renders, so they are **not** skeletoned: hiding a value that is already known would be a second way of describing the document wrongly.

**Alternative rejected.** The issue offered `?meta=1` on the existing route as an option. A separate path gives the app its own cache key (`queryKeys.specs.meta(id)`), which is what stops a cached *description* from ever satisfying a reader that needs `content` - a query-string variant would have collided on the same route in the API surface and made that distinction a convention instead of a shape. I also considered a DB-level column select so the engine never materializes `content` at all; it is not in this diff because SQLite's `length()` counts characters rather than bytes and `contentBytes` has to be in the unit `maxSpecDocumentBytes` refuses a write by. The transfer was the cost this issue names, and an in-process row read is not it.

**Deviation from the issue spec, disclosed.** `SpecCoverageLine` is untouched. It is silent until it has both a run and a report and then appears - it never states a wrong number, so it has nothing to skeleton; the issue asked for the same treatment "if it currently pops".

`contentBytes` is part of the meta shape the issue asked for, and it has a reader: a new **Size** cell on the card, formatted through the existing settings `formatBytes` so a document's size and the limit it is stored under read in one unit.

## Type of Change
- [x] Bug fix (a pending card that made a false claim; a button gated on a transfer nobody asked for)
- [x] New feature (`GET /specs/:id/meta`)
- [ ] Breaking change
- [x] Documentation update

## Testing

All green on this branch, all commands from the issue:

- **Engine** - `python build.py -e -t` then `ctest --preset linux-dev --output-on-failure`: **2039 tests, 100% passed** (354s). Two new tests in `specs_route_test.cpp`: meta carries the full read's values with the three heavy fields absent (not empty), and a missing id answers the same 404 body as the full route.
- **Mutation check (engine)**: dropping `json.erase("content")` from `serialize_meta` reddens `SpecMetaCarriesTheSameValuesAsTheFullDocumentWithoutTheHeavyFields`; restoring it returns the suite to green.
- **App** - `pnpm test`: **518 files, 5649 tests, 100% passed**. `pnpm type-check`, `pnpm lint` and `prettier --check` on the touched files are clean (only files that were already clean were formatted).
- New app tests: the card renders skeletons for source/fetched/size while pending **and the fallback source string is unreachable there** (the mutation check the issue asked for), the mapped count cannot flash `0 of 0`, the size renders once known, Check is pressable before any document has been read and reads it on the click, and a stored document that cannot be read reports the failure without re-fetching anything.
- `mkdocs build --strict` passes for the docs edits.

## Checklist
- [x] I have updated documentation - `docs/engine/api-reference.md` (the new read), `docs/app/api-integration.md` (which caller reads which shape, and why), `docs/app/state-management.md` (the two spec keys and their cache policy), `engine/CLAUDE.md` (endpoint table)

## Related Issues
Closes #712

---
_Generated by [Claude Code](https://claude.ai/code/session_01VXb4gtK8CckiTSFixmZ9ZH)_

### DIFF
--- a/app/src/config/api-endpoints.ts
+++ b/app/src/config/api-endpoints.ts
@@ -55,6 +55,10 @@ export const API_ENDPOINTS = {
 	// whoever else binds it.
 	SPECS: `/specs`,
 	SPEC_BY_ID: (id: string) => `/specs/${id}`,
+	// The same document without the document (issue #712): what the Spec tab's
+	// card needs, so opening the tab does not transfer a 12 MB spec to paint a
+	// URL and a date. The full read above still backs export, sync and matching.
+	SPEC_META: (id: string) => `/specs/${id}/meta`,
 	// Applying a re-fetched document to the collection bound to it (issue #655):
 	// the new document, the moved binding and the created/updated/deleted
 	// requests in one engine transaction.
--- a/app/src/modules/collections/CollectionDetail/SpecSync.test.tsx
+++ b/app/src/modules/collections/CollectionDetail/SpecSync.test.tsx
@@ -37,6 +37,7 @@ const createRequest = vi.fn();
 const deleteRequest = vi.fn();
 const createSpec = vi.fn();
 const syncSpec = vi.fn();
+const getSpec = vi.fn();
 
 vi.mock("@/services/api", () => ({
 	apiService: {
@@ -47,6 +48,8 @@ vi.mock("@/services/api", () => ({
 		deleteRequest,
 		createSpec,
 		syncSpec: (payload: SpecSyncRequest) => syncSpec(payload),
+		// The stored document is read by the check, not by the tab (issue #712).
+		getSpec: (id: string) => getSpec(id),
 	},
 }));
 
@@ -129,11 +132,14 @@ function renderSync(props: {
 	const wrapper = ({ children }: { children: ReactNode }) => (
 		<QueryClientProvider client={client}>{children}</QueryClientProvider>
 	);
+	// The section is given the binding's id and reads the document itself, so a
+	// test supplies the document by answering that read.
+	if (props.spec) getSpec.mockResolvedValue(props.spec);
 	return render(
 		<SpecSync
 			collection={collection()}
 			collections={props.collections ?? [collection()]}
-			spec={props.spec}
+			specId="spec_1"
 			specFile={undefined}
 			requests={props.requests ?? [request()]}
 		/>,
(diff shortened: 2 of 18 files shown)


athrvk:
master moved, update this PR, and verify everything works

athrvk:
Updated: merged `master` (`d588701`, which brought in #808) into this branch - **no conflicts**, and the diff against master is still exactly the 18 files this PR owns.

Re-verified everything on the merged tree (`19218d8`):

- **Engine** - `python build.py -e -t` (full rebuild, since #808 touched `tests/main.cpp`, `temp_database.hpp` and the presets) then `ctest --preset linux-dev --output-on-failure`: **2039 tests, 100% passed**, now 55.9s rather than ~355s, since the preset this branch inherited from #808 runs `-j`. Worth noting for this PR specifically: the two new `SpecsRouteTest` cases pass under the parallel scheduling and per-process scratch directories #808 introduced - they use the fixture's relative `DB_PATH`, so they inherit that isolation rather than needing their own.
- **App** - `pnpm test`: **518 files, 5649 tests, 100% passed**. `pnpm type-check` and `pnpm lint` clean.
- `mkdocs build --strict` clean.

No source changes were needed for the merge; the only new commit is the merge itself.

---

Issue #810, opened 2026-08-18
engine(scripting): a pm.test in a pre-request script fails its step and is named nowhere - counted by describe_failed_tests, dropped by build_script_result_node


athrvk:
Filed from #724 (the step-detail work), which hit this while deciding what a step's
tally is allowed to count.

## Problem

`pm.test` is registered in **both** script phases - `js_pm_test` pushes onto the
context's list whichever event is running (`engine/src/runtime/script_engine.cpp`),
and nothing gates it on `ScriptEvent::PreRequest`. Two readers then disagree about
whether those assertions exist:

- **`describe_failed_tests`** (`engine/src/core/scenario_runner.cpp:43-68`) walks
  `{ &pre, &post }`, so a failing assertion in a *pre-request* script makes its
  collection-run step `failed` and supplies the `results.error` summary line
  ("Test failed - <name>").
- **`build_script_result_node`** (`engine/src/http/routes/execution.cpp:518+`)
  serializes `post_script_result.tests` **only**, with the stated reason that "a
  pre-request script runs before there is anything to assert about".

So for a collection-run step the reader gets a step marked `failed`, an error line
naming an assertion, and - after #724 - a Tests list and a tally that do not
contain it. The list is not wrong about what it shows; it is silently narrower
than the outcome it sits beside. On a **design send** the same assertions vanish
entirely: nothing there classifies an outcome, so a pre-request `pm.test` runs,
passes or fails, and is reported nowhere at all.

This is the repo's "written but never read" shape in its computed-and-half-read
form, and it is small: two functions, one of two answers.

## The decision this issue is

One of the two readers is wrong and the fix is to pick which, not to widen both
by reflex:

**(a) Pre-request assertions are real.** Then `build_script_result_node` lists
them, with the phase on each entry (`source: "pre" | "test"`, the way
`consoleLogs` already carries `source`) so a reader can tell an assertion made
before the request from one made after. The renderer's Tests tab and
`TestResults` show the phase; the #724 tally counts both, and its "test script
alone" comment goes away. Cost: a shape change to a node three surfaces read
(live `/execute` body, stored trace, step trace).

**(b) Pre-request assertions are not a supported thing.** Then
`describe_failed_tests` stops walking `pre`, the pre-request phase stops
registering `pm.test` (or registers it as a loud script error - "pm.test is not
available before the request"), and `docs/engine/scripting.md` says so where it
lists what each phase can do. Cost: a behaviour change for anyone who writes one
[...]

Issue #725, opened 2026-08-16
engine(execution): a buffered Send's test results are returned live but never stored - a restored response's Tests tab is empty, an asymmetry acknowledged in a code comment and tracked by no issue


athrvk:
From the adversarial data/scripting review at `ac27b97`, CONFIRMED - and it retro-files an acknowledged, untracked deferral.

## Problem

`record_design_result` stores `trace['scripts']` only when `stream != nullptr` (`execution.cpp:625-631`); an ordinary buffered `POST /execute` returns testResults/consoleLogs in the live body (`:433`) and stores none of it. The restore funnel documents the asymmetry as fact: "A buffered send's results arrive in the /execute body and are not stored, which is why an ordinary restored run still shows no Tests pane" (`restore-response.ts:100-104`) - a comment saying "later" with no issue number behind it, which root CLAUDE.md forbids.

User consequence: Send shows "5 passed"; restart the app or reopen from History; same run row restores body and headers but the Tests tab shows the empty-state "Assertions written in the request's Tests script show up here" - **tests-passed is indistinguishable from tests-never-ran** in any audit of yesterday's work. The two funnels the repo requires to agree (`app/CLAUDE.md:52-60`) disagree by design here: a *streaming* send's restored response shows its Tests, a buffered one's does not.

## Fix pathway

Store the same `build_script_result_node` output on the buffered design path that the streaming path already stores (`execution.cpp:625-631` drops the `stream` condition for the scripts node, or attaches it in `record_design_result` unconditionally). The restore funnel already parses it - `restore-response.ts:110-122` - so the app half is zero new code, which makes the asymmetry's survival purely a storage-side conditional. Body-size economics are unchanged: script results are small relative to the `maxTraceBodyBytes`-capped bodies beside them.

## Tests

Buffered send with pm.test → restored ResponseState carries the same testResults (the `validation-funnels.test.ts` shape: drive both funnels, compare) - mutation-check by restoring the `stream` condition. The restore comment comes out in the same commit.

## Docs - same commit

`docs/engine/api-reference.md` trace shape note; the `restore-response.ts:100-104` comment replaced by the new truth.

## Sequencing

Same trace shape as #724 (collection-run steps) - land together or in either order; #724's step rendering benefits immediately.

## Acceptance criteria

- [ ] A restored buffered response shows the identical Tests list the live send showed (mutation-checked funnel-parity test).
- [ ] No untracked "not stored" comment remains.

athrvk:
[...]

PR #811, opened 2026-08-18
fix(engine/app): a collection run's step keeps the assertions it made

### LINKED ISSUE #724: engine/app(runs): a collection run computes per-test pm.test results for every step, then discards them - no scripts node on the stored trace or the SSE frame, so the step detail can never show a Tests tab
From the adversarial data/scripting review at `ac27b97`; found independently by three dimensions, CONFIRMED. This is the live user report: a run's step detail shows only Response/Headers/timing while the same request's single Send shows a full Tests tab.

## Problem

The scenario runner executes both scripts per step and holds the itemized results in `exchange.pre_script_result`/`post_script_result` - but uses them only to classify the outcome (`classify_step`, `scenario_runner.cpp:113-118`, folding failures into one `record.error` summary string). The stored step row is `build_result_trace(request, response)` + `stamp_step_identity` (`scenario_runner.cpp:671-682`) - **no scripts node**; `build_script_result_node`'s only callers are the single-send live body and the streaming path (`execution.cpp:433,1148`). The step SSE frame (`build_step_payload`, `scenario_runner.cpp:299-317`) carries no testResults either. So:

- A passing step shows **zero evidence** its assertions ran at all.
- A step with 3 failing tests shows one summary line ("3 tests failed - first name"), not the per-test list the Tests tab gives the identical request on Send.
- Unlike a buffered Send (results at least arrive live in the `/execute` body), a collection run has **no live path at all** - every step is only ever viewed through the stored trace, so storing is the only delivery.

The counts machinery is healthy (verified clean: a failing pm.test fails the step; the header tallies count step executions correctly) - it is the *detail* that is computed and thrown away. The repo's signature defect, in its computed-but-never-stored form.

## Fix pathway

- Engine: attach the existing `build_script_result_node` output to the step trace (`trace["scripts"]`, the exact shape the restore funnel already parses - `restore-response.ts:110-122` reads it today, which is why the streaming design path's stored steps would already render), and carry test tallies (or the full list, size-permitting) on the step SSE frame beside the schema validation node it already has.
- App: the step detail (`ScenarioStepCard.tsx`) renders the Tests section from the restored trace - the component work is small because the response viewer's Tests rendering already exists; follow the two-funnel discipline (this IS the second funnel).
- Respect `maxScenarioStoredSteps` economics: per-test lists ride only stored steps (thinned steps lose them with the rows, already disclosed); the tallies on SSE frames are constant-size.

## Tests

Engine: stored step trace carries the scripts node (mutation-check by reverting the attach); SSE frame tallies. App: step card renders per-test rows from a stored trace; funnel-parity style assertion against the single-send rendering of the same script results.

## Docs - same commit

`docs/engine/api-reference.md` step-trace shape; `docs/engine/scripting.md` if it describes where results appear.

## Sequencing

(Updated 2026-08-17 at `46b115d`.) The buffered-send companion filed alongside was #725 - now shipped via PR #740: `record_design_result` attaches `trace["scripts"]` for every design send that ran scripts (`engine/src/http/routes/execution.cpp:625-632`). The trace shape is therefore no longer co-designed here - this issue's engine half extends the `scripts` node #725 landed to collection-run step traces, and the restore funnel already reads it.

## Acceptance criteria

- [ ] A collection-run step's detail shows the same per-test list its single Send shows, for passing and failing steps alike (mutation-checked).
- [ ] Live run view gains at least per-step test tallies as steps stream.

athrvk:
## Description

**Root cause.** The scenario runner executes both scripts per step and holds
every itemized `pm.test` result in `exchange.pre_script_result` /
`post_script_result` - and then uses them only to classify the outcome
(`classify_step`, folding failures into one `record.error` summary string). The
stored step row was `build_result_trace` + `stamp_step_identity` with **no
`scripts` node**, and `build_step_payload` carried no test data either. So a
step with 3 failing assertions showed one line ("3 tests failed - <first
name>"), a passing step showed *zero* evidence its assertions had run, and
neither could ever become a Tests tab - while the identical request sent on its
own showed the full list. Verified still true at `dcfe406`; #725 (PR #740) had
already landed the trace shape for design sends, so this is the extension it
left open.

**Approach.** Attach the node the design path already builds
(`build_script_result_node`) to the step trace, which is the exact shape
`restore-response.ts:114` parses today - so the app half is rendering, not
re-mapping. The `step` SSE frame gains `tests: {passed, failed}` rather than the
list: the tick ring is fixed-size and replayed by every watcher on
`Last-Event-ID` resume, so a step that made 400 assertions would push the rest
of the run out of it. On the app side `ScenarioStepCard` renders the response
pane's own `TestResults` component from the restored trace and a new shared
`TestsChip` beside the outcome; the chip prefers the stored list (tallied by
`tallyTests`) over the frame's numbers, so a row does not change what it says
when its stored copy lands.

**Alternative rejected.** Putting the full per-test list on the SSE frame, which
would have given the live view the same detail. Rejected on the ring: the frame
must stay constant-size, and the issue's own economics note says the same. Also
rejected: a second reader of `trace.scripts` in the step list - the card takes
the list from `responseFromRunResult`, the funnel its response already comes
from, so there is one parser and not two to keep in step.

**One decision worth flagging.** Both the stored list and the tally count the
**test script's** assertions, which is what `build_script_result_node` has
always serialized. `describe_failed_tests` (the summary line) counts a
*pre-request* script's assertions too, so a `pm.test` written there fails its
step and is named in the error line but appears in no list. That disagreement
predates this PR and is a user-visible behaviour call, so it is filed as **#810**
and documented in `scripting.md` rather than settled here - widening the node
would change the shape three surfaces read, and narrowing `describe_failed_tests`
would change what fails a step.

`build_script_result_node`'s declaration moves from `routes.hpp` to
`request_exchange.hpp`, beside `build_result_trace` - the trace-building sibling
the runner already calls - so `core/scenario_runner.cpp` can write the same node
without including the route header. The definition stays in `execution.cpp`;
every existing caller compiles unchanged, since `routes.hpp` includes
`request_exchange.hpp`.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Full suites, both green on this branch:

- **Engine**: `python build.py -e -t` then `ctest --preset linux-dev` -
  **2040 tests, 100% passed** (333s serial).
- **App**: `pnpm test` - **5654 tests in 519 files, all passed**;
  `pnpm type-check`, `pnpm lint` and `pnpm format:check` clean.
- `mkdocs build --strict` clean.

New coverage, all mutation-checked (reverted the fix, confirmed red, restored):

- `scenario_runner_test.cpp` (+3): `EveryStepStoresTheAssertionsItsTestScriptMade`
  (a passing step *and* a mixed one - a failure-only test would miss the "no
  evidence it ran" half), `AStepWhoseScriptsSaidNothingStoresNoNodeAtAll`,
  `StepFramesCarryTheTallyOnTheSameTermsAsTheStoredList` (reads the run's own
  ticks; asserts the tally is present with numbers on the step that asserted and
  **absent** on the one that did not).
- `ScenarioStepCard.script-results.test.tsx` (+6), including a funnel-parity
  case that drives `TestResults` directly with one list and asserts the card's
  expansion contains the same rendering, and a live-vs-stored case asserting the
  chip reads identically from the frame's tally and from the stored list.
- `scenario-steps.test.ts` (+2) for `tallyTests`, `sse-client.test.ts` (+1) for
  the frame parse (absent stays absent; a half-tally is rejected rather than
  rendered as `NaN`).

**Mutation check detail.** Dropping the trace attach reddens
`EveryStepStoresTheAssertionsItsTestScriptMade`; dropping `record.tests =
tally_tests(...)` reddens `StepFramesCarryTheTallyOnTheSameTermsAsTheStoredList`;
hiding either app half reddens 5 of the 6 card cases (the sixth asserts absence
and correctly stays green).

No pre-existing failures observed on this base.

## Checklist
- [x] My code follows the style guidelines (clang-format checked on every C++ file touched; prettier/eslint clean)
- [x] I have updated documentation (`docs/engine/api-reference.md` - the step-trace `scripts` node and the frame's `tests` node; `docs/engine/scripting.md` - where a collection run's results appear; `docs/app/COMPONENTS.md` - the shared chip and the step card)

## Related Issues

Closes #724
Files #810 (the pre-request-script assertion disagreement this deliberately did
not settle).

---
_Generated by [Claude Code](https://claude.ai/code/session_01AV9snvr6xrKb4m2b1tGdKp)_


athrvk:
master moved, update this PR, and verify everything works

athrvk:
Updated - `master` (now `d588701`, carrying #808) merged in at `15b5223`. Clean merge, no conflicts: #808 touched `CMakePresets.json`, `tests/main.cpp`, `tests/temp_database.hpp` and the build docs, none of which this PR edits.

Re-verified on the merged tree:

- **Engine**: `python build.py -e -t`, then `ctest --preset linux-dev` - **2040 tests, 100% passed in 100.7s**. That is now #808's `-j8` preset, so the three new `ScenarioRunnerTest` cases are proven parallel-safe rather than assumed: they take their scratch DB from the fixture, and their mock server binds an ephemeral port.
- **App**: `pnpm test` - **5654 tests in 519 files, all passed**. `pnpm type-check`, `pnpm lint`, `pnpm format:check` clean.
- `mkdocs build --strict` clean.

One thing worth stating rather than hiding: the *first* app run after the merge reported `1 failed | 5653 passed` while the engine build was compiling alongside it (617s wall, against 365s for a clean run), and my capture kept only the summary, so I do not have the test's name. An immediate clean re-run of the whole suite passed 5654/5654. I am reading that as contention on a loaded box rather than a real failure - nothing in this diff is timing-dependent, and the same suite was green on this branch before the merge - but since I cannot name the test, it is a claim about a flake I did not reproduce, not a diagnosis. If CI shows a red app test here, that is the thing to look at first.

---

Issue #802, opened 2026-08-18
engine/tests(tls): a live mTLS fixture server - the handshake #707 could only assert structurally


athrvk:
Filed from #707 (PR #801), which deferred it deliberately.

## Problem

#707's acceptance criterion 1 - "a registered host+cert makes an mTLS endpoint callable from a design send, a load run, an SSE stream, `pm.sendRequest` and an OAuth token fetch" - is satisfied *structurally* in the shipped code (one registry, one applier, every driver through it) and asserted three ways in `engine/tests/client_certificates_test.cpp`:

- the lookup unit-tested at the boundary, mutation-checked on precedence,
- the lookup's **outcome** asserted end-to-end through the design and SSE drivers (both record the entry they matched on the response),
- a per-platform probe that this build's TLS backend accepts `CURLOPT_SSLCERT` / `SSLKEY` / `KEYPASSWD` at all - the shape `TlsBackend.AcceptsACustomCaBundleOnThisPlatform` set in #706.

What none of that covers is a **real handshake**: that the bytes curl loads from those paths are accepted by a server demanding a client certificate, that a wrong passphrase fails with the setting named, and that a server's rejection surfaces as `SslError` rather than `ConnectionFailed`.

## Why it was not done in #707

Two reasons, both worth recording because they are the work this issue actually is:

1. **There is no TLS server in the suite at all.** `cpp-httplib` is declared without the `openssl` feature in `engine/vcpkg.json`, so `httplib::SSLServer` does not exist in this build. Adding the feature is a dependency change on all three CI platforms.
2. **The backends disagree on client-certificate formats.** OpenSSL (Linux) takes a PEM certificate and key pair; Schannel (Windows) expects a PKCS#12 file or a certificate-store reference and does *not* take a PEM pair; Apple SecTrust (macOS) takes PKCS#12 or PEM. A fixture built on checked-in PEMs would go red on Windows while proving nothing about the code under test - the opposite of a useful guard.

## Fix pathway

- Add `cpp-httplib[openssl]` (note the cloud-environment `vcpkg-fix-port` step in `engine/CLAUDE.md` when a cold cache 403s).
- A fixture server that requires a client certificate, with checked-in test material. Decide per platform what material that is - a PEM pair plus a PKCS#12 built from the same key is the obvious answer, selected by backend at test time - and write the matrix down rather than assuming one format travels.
[...]

Issue #704, opened 2026-08-16
epic: Vayu behind real networks - proxy support, TLS trust, and client certificates across every outbound path


athrvk:
**Status (updated 2026-08-19, evening): all four phases shipped (4/4), and the comparison-row rider landed via PR #850 (closure-verified - row on all four compare pages with both mandatory asterisks). The epic is HELD OPEN (owner decision) on exactly ONE item: Windows mTLS, which closes when **#851** (the approved OpenSSL backend switch - implementation-ready, nothing blocks it) ships. No other work remains here.**

## Why

Behind a corporate proxy, Vayu's UI works perfectly and every request fails - the most confusing failure shape possible. The Electron shell (and the updater) rides Chromium's networking, which picks up the OS proxy on its own; the engine's libcurl stack, which carries every request, load run, stream, OAuth token fetch and spec import, had no proxy, no custom CA and no client-certificate support at all. Postman, Insomnia, Bruno and k6 all cover this set. This epic makes Vayu usable on the networks real API teams actually sit on.

## Remaining work

**Criterion 3 waits on #851** (the OpenSSL backend switch - approved 2026-08-19, decision record in closed #843, implementation order in #851): when it lands, `mutual_tls_test.cpp` runs both identity formats on all three legs, the Schannel skip dies, and the mTLS box below gets checked. No work here beyond what #851 already specifies - this epic just does not close before it.

(Delivered riders, for the record: the comparison row shipped via PR #850 - proxy/certificates row on all four `docs/compare/` pages under the checkable-claims rule, with the Windows-mTLS qualification and the PAC-once/headless-fallback note, and the `vayu-vs-postman.md` firewall line extended to both directions. Wildcard cert hosts #803 shipped via PR #847. No satellites remain open.)

## Current state - every claim verified at `ac27b97` (historical; all resolved by the phases)

- **`proxy_url` was born dead.** Declared in both transfer configs (`engine/include/vayu/http/client.hpp:29`, `event_loop.hpp:71`), read behind `if (!empty)` at `client.cpp:428` and `curl_utils.cpp:639` - and no writer has ever existed in the repo's history (full-history pickaxe: the field arrived in the first commit and nothing ever assigned it). `ClientConfig::ca_bundle_path` (`client.hpp:30`) is a second dead field: read by nothing, written by nothing. The repo's most-repeated defect class, inverted.
[...]

PR #817, opened 2026-08-18
test(engine/tls): verify a custom CA on a wire, and correct the claim that said we already did (#812)

### LINKED ISSUE #812: engine(tls): CLAUDE.md claims custom-CA verification is CI-tested per platform; the cited test only checks the curl option is accepted
Filed on closure re-verification of #706 (PR #763, merged). Not a security-pass item - filing normally, the same way #802 was filed for the analogous #707 gap.

**Problem**

`engine/CLAUDE.md:248` states, about the custom-CA-bundle feature #706 shipped:

> The per-backend claim is tested on each CI platform (`TlsBackend.AcceptsACustomCaBundleOnThisPlatform`) rather than reasoned about, because a wrong claim here is a security claim.

The cited test (`engine/tests/transport_policy_test.cpp:428-441`) does not test the claim the doc says it tests. It calls:

```cpp
CURLcode result = curl_easy_setopt(curl, CURLOPT_CAINFO, "/nonexistent/ca-bundle.pem");
EXPECT_EQ(result, CURLE_OK)
```

That asserts only that the backend *accepts the option syntactically* - i.e. that this build's curl (OpenSSL on Linux, Schannel on Windows, Apple SecTrust on macOS) doesn't refuse `CURLOPT_CAINFO` outright with `CURLE_NOT_BUILT_IN`. It never stands up a TLS server, never signs a certificate with a user-supplied CA, and never checks that a request against that server actually verifies once the bundle is set (or fails before it is). "The per-backend claim" in context is the *additive-trust* claim two paragraphs earlier - that adding a custom CA extends the platform's trust store rather than replacing it - and this test exercises neither the additive behavior nor basic verification success/failure.

This is the same shaped gap as #707's, which was caught and disclosed correctly: #707's commit message named the missing live-mTLS-handshake test explicitly, and it was filed as #802. #706's version of the same gap was not disclosed anywhere - no comment, no follow-up issue - and the doc text actively asserts stronger coverage ("tested... rather than reasoned about") than exists, which is worse than silence: a reader has no reason to doubt a claim phrased that confidently.

Everything else about #706 is well covered: bundle materialization (system anchors + user paste), the additive-not-replacing merge, atomic writes, and `POST /config` PEM validation (rejecting private keys and truncated blocks) are all genuinely unit-tested. Only the live "does verification actually work" claim is unsubstantiated.

**Fix pathway**

Two independent parts - the doc claim can be corrected immediately regardless of when the real test lands:

1. **Correct `engine/CLAUDE.md:246-248` now.** Reword to state what `AcceptsACustomCaBundleOnThisPlatform` actually checks (the backend accepts the option without erroring) and name the still-missing live-verification test as a known gap, pointing at this issue - the same honest pattern #707's commit used for #802.
2. **Add the live test**, likely alongside whatever fixture #802 builds for the mTLS handshake (#707) - a self-signed CA, a server certificate it signs, an httplib (or equivalent) TLS listener using that certificate, and two assertions: a request against it fails verification with `customCaCertificates` unset, and succeeds once the CA is pasted into the setting. If #802's fixture work lands first, this should reuse it rather than building a second one - coordinate in whichever issue's PR lands second.

**App changes**

None - engine test/doc only.

**Tests**

The new test itself is the deliverable; mutation-check it the way `engine/CLAUDE.md`'s own testing conventions ask (comment out the `customCaCertificates` application and confirm the "fails before the setting is applied" half actually fails without it).

**Docs - same commit**

`engine/CLAUDE.md:246-248`, corrected as above.

**Sequencing**

Companion to #802 (both are "the live TLS test #70x could only assert structurally" gaps off the same transport epic, #704). Fully parallel-safe with everything else open - touches only `engine/CLAUDE.md` and adds a new test file/fixture, no shared region with any other open issue. If #802's fixture lands first, prefer reusing it here rather than duplicating server-fixture setup.

**Acceptance criteria**

- [ ] `engine/CLAUDE.md`'s claim matches what the cited test actually verifies.
- [ ] A live test proves a host signed by a user-added CA verifies once `customCaCertificates` is set, and fails before it is set (or against a CA that didn't sign it).
- [ ] Runs on all three CI platforms (Linux/OpenSSL, macOS/SecTrust, Windows/Schannel) - the reason the original claim needed to be a real per-backend test in the first place.

---

athrvk:
## Description

**Plan.** The root cause is a doc claiming coverage that does not exist.
`engine/CLAUDE.md:246-248` said the custom-CA feature's per-backend claim was
"tested on each CI platform rather than reasoned about, because a wrong claim
here is a security claim", citing `TlsBackend.AcceptsACustomCaBundleOnThisPlatform`.
That test sets `CURLOPT_CAINFO` to `/nonexistent/ca-bundle.pem` and asserts
`CURLE_OK` - it proves the backend does not refuse the *option*, and nothing
more: no server, no signed certificate, no verification. Verified against the
current code before touching anything; the gap is exactly as the issue
describes. The approach is both halves of the issue's fix pathway - the claim
rewritten to say what each test checks, and the missing live test added,
built on a fixture written so #802 extends it rather than standing up a second
listener. **Alternative rejected:** correcting the doc alone and leaving the
test to #802's fixture work. The doc would then have to describe an open hole
in a security claim with no date on closing it, and #802 is the larger job -
it needs the per-backend PKCS#12/PEM matrix, which server-side verification
does not. The smaller question stands on its own.

**What is new.**

- `engine/tests/tls_server.hpp` - `TestCertificateAuthority` mints a self-signed
  CA and signs leaves for it; `TlsServer` serves one route from an
  `httplib::SSLServer` holding a leaf for `IP:127.0.0.1`. Material is generated
  **per run, in memory**: a checked-in certificate expires, and the suite would
  then go red on a date rather than on a change - naming TLS while the cause is
  the calendar. It also keeps a private key out of the repository and makes a
  *second*, unrelated CA cheap, which one of the cases below needs.
- `engine/tests/tls_verification_test.cpp` - `CustomCaVerificationTest` drives
  the ordinary design-send path (`Client::send` through the shipped
  `apply_transport_policy`) at that listener, changing only
  `customCaCertificates`.

**Four cases, because the happy path alone proves little.** A bundle libcurl
reads and ignores would pass it on any machine with a permissive store. So:
the host verifies once its CA is pasted in; it does **not** verify before that;
it does **not** verify against a CA that signed nothing here; and it *still*
verifies when a second anchor is added beside the first - #704 decision 4's
additive rule observed on a handshake instead of by reading the file back.

**One thing the fixture had to pin.** `proxyMode` is set to `off` in `SetUp`.
The default is `environment`, so a machine exporting `https_proxy` - a
corporate laptop, a sandboxed runner - would send the handshake to a proxy
instead of to loopback, and an assertion that passes or fails on an
environment variable is not an assertion about this engine.

**Dependency.** `cpp-httplib` gains the `openssl` feature; `httplib::SSLServer`
does not exist without it. OpenSSL was already in the tree on Linux and macOS
(vcpkg's `curl[ssl]` resolves to `curl[openssl]` on `!windows`); Windows gains
it. The first CI run on this branch missed the vcpkg cache **by construction** -
the key is the manifest hash - and cost ~13 min per leg cold; the second run hit
the new key and the engine legs came back in 2-4 min.

## The backends disagreed, which is the point

The first CI run (`ffe1c19`) passed all four cases on ubuntu and macOS and
failed **exactly the two positive ones** on Windows, with both negatives
passing. That split is the diagnosis: if the pasted anchor were not being
consulted, the *negative* cases would be the ones misbehaving.

curl's Schannel path validates a `CURLOPT_CAINFO` chain itself and asks Windows
to revocation-check it - `CERT_CHAIN_REVOCATION_CHECK_CHAIN` unless
`CURLSSLOPT_NO_REVOKE` is set, with an unknown status treated as failure unless
`CURLSSLOPT_REVOKE_BEST_EFFORT` is (`lib/vtls/schannel_verify.c`, pinned
baseline). A CA minted seconds ago publishes no CRL, so that answer can only be
"unknown". The OpenSSL-backed legs make no such check.

`e90a546` makes the positive cases skip where that happens, built so the skip
cannot rot into the quiet no-op #812 exists to end:

- it fires **only** when the backend's own error text names revocation - any
  other refusal, on any platform, still fails the suite;
- it **prints** that text, so the reason comes from the run;
- it matches on the words, **not** on `_WIN32`. "This backend demands revocation
  information" is the property that matters, and a test asserting the host
  platform is what `CLAUDE.md` forbids;
- the **negative cases keep no carve-out** and are asserted on every leg. They
  are what proves the anchor is consulted, and they were green on Windows from
  the first run.

**That gate also turned the diagnosis into evidence.** Windows now reports
`2047 tests run, 2045 passed, 2 skipped, 0 failed` - two skipped and zero failed
is only reachable if the refusal named revocation, since anything else fails.

**Deliberately not done:** setting `CURLSSLOPT_REVOKE_BEST_EFFORT` when a custom
bundle is in force. It is one line beside the existing `CURLSSLOPT_NATIVE_CA`
and it would turn every leg green - which is exactly why it does not belong in a
test-and-doc change. It weakens revocation enforcement as a side effect of a
user pasting a certificate, a security posture change wearing a one-line diff.

## Type of Change
- [x] Bug fix - a false claim in a doc future sessions are told to trust
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update
- [x] Test coverage

## Testing

Engine-only diff (C++ tests, CMake, vcpkg manifest and three docs). No app
file is touched, so per `CLAUDE.md`'s "scale the verification to what the
change could actually break" the app suite was not run - no vitest test could
change colour.

- `python build.py -e -t` then `ctest --preset linux-dev --output-on-failure` -
  **2046 tests, 100% passed** (2042 on the base commit; the four new ones are
  the delta).
- CI on `e90a546`: **all legs green, CI gate success.** ubuntu and macOS run all
  four cases; Windows runs two and skips two, loudly.
- `clang-format` on the two new files (both new, so nothing pre-existing was
  reflowed).

Mutation-checked in both directions, since the positive and negative cases fail
to different mutations:

| Mutation | Result |
|---|---|
| `CURLOPT_CAINFO` always null (the bundle never applied) | `AHostSignedByTheAddedCaVerifies` and `TheAddedCaExtendsTheStoreRatherThanReplacingIt` fail; both negative cases still pass |
| `CURLOPT_SSL_VERIFYPEER` forced to `0` | `TheSameHostDoesNotVerifyWithoutIt` and `ACaThatSignedNothingHereDoesNotVerify` fail; both positive cases still pass |
| `refused_for_want_of_revocation` inverted to fire on success | the positive cases report `[  SKIPPED ]` with the full reason appended, rather than reading as a pass |
| baseline, unmutated | 4/4 pass |

All mutations reverted; `git diff` on `engine/src/` is empty in this branch.

The fixture itself was checked outside the engine, so a green suite could not be
the fixture being permissive: standing it up in a standalone binary and calling
it with the system `curl` gives `curl: (60) SSL certificate problem: unable to
get local issuer certificate` without `--cacert`, and `200` with it.

## Checklist
- [x] I have updated documentation - `engine/CLAUDE.md` (the corrected claim,
      naming what each test checks, what the backends disagreed about, and what
      is still structural), `docs/engine/building.md` (the dependency table row
      and the `vcpkg install` line in Linker Errors),
      `engine/tests/client_certificates_test.cpp`'s header (it said this suite
      has "no TLS server at all... deliberately", which is now half wrong -
      corrected to name what remains for #802 and to point at `TlsServer` as
      the thing to extend rather than duplicate)

## Related Issues
Closes #812

Two issues filed from what this turned up, rather than folded in:

- **#819** - the fixture CRL that would let the positive cases run on a
  revocation-checking backend too (and the skip be deleted), plus the
  user-facing half: someone on Windows pasting an internal CA with no reachable
  distribution point hits this same refusal today, worded so it names neither
  the setting they changed nor the reason. That second half needs your call, not
  a patch.
- **#818** - `engine/CLAUDE.md` and `transport_policy.hpp` both describe macOS as
  Apple SecTrust-backed, while the pinned vcpkg baseline maps curl's default
  `ssl` feature to `curl[openssl]` on `uwp | !windows`. This PR leaves that
  claim alone rather than trading one unverified statement for another - it is
  about a build I cannot run here.

**One note for review.** The error shape is worth a look: a refused handshake is
not a `Result` error - `Client::send` reserves that arm for a request it could
not attempt at all - so a verification failure arrives as an ok `Result` holding
a `Response` with `status_code == 0` and `error_code == SslError`. The tests
assert that shape, the same one the proxy tests in `transport_policy_test.cpp`
read and the same one a stored trace holds. My first draft asserted `is_error()`
and went red against entirely correct engine behaviour.

### DIFF
--- a/docs/engine/building.md
+++ b/docs/engine/building.md
@@ -139,7 +139,7 @@ Dependencies are managed via vcpkg and specified in `engine/vcpkg.json`:
 | libsodium | SHA-256, HMAC-SHA256, base64 and hex (PKCE, Basic/OAuth credentials, `pm.crypto`) |
 | nlohmann-json | JSON parsing/serialization |
 | valijson | JSON Schema validation of responses against a bound OpenAPI document |
-| cpp-httplib | HTTP server library |
+| cpp-httplib | HTTP server library - built with the `openssl` feature, which the test suite's HTTPS listener needs (custom-CA verification is asserted on a real handshake, not reasoned about) |
 | sqlite3 | Embedded database |
 | sqlite-orm | C++ ORM for SQLite |
 | gtest | Unit testing framework |
@@ -397,8 +397,8 @@ Set `VCPKG_ROOT` environment variable or install vcpkg in a standard location.
 
 ### Linker Errors
 
-- Ensure all vcpkg dependencies are installed: `vcpkg install curl[http2] libsodium nlohmann-json valijson cpp-httplib sqlite3 sqlite-orm gtest`
-  (the `http2` feature is required - without it libcurl is built without nghttp2 and the HTTP/2 support test fails)
+- Ensure all vcpkg dependencies are installed: `vcpkg install curl[http2] libsodium nlohmann-json valijson cpp-httplib[openssl] sqlite3 sqlite-orm gtest`
+  (the `http2` feature is required - without it libcurl is built without nghttp2 and the HTTP/2 support test fails; the `openssl` feature is required for the same reason one step further along - without it `httplib::SSLServer` does not exist and the TLS-verification tests do not compile)
 - On Windows, ensure Visual Studio C++ tools are installed
 
 ### Build Script Issues
--- a/engine/CLAUDE.md
+++ b/engine/CLAUDE.md
@@ -244,9 +244,25 @@ Three things worth knowing before you design around them:
   `customCaCertificates` setting beside the database and **extends** the
   platform's trust rather than replacing it: on an OpenSSL build CAINFO *is*
   the whole store, so the file is the system anchors plus the user's, while
-  Schannel and Apple SecTrust keep their OS store. The per-backend claim is
-  tested on each CI platform (`TlsBackend.AcceptsACustomCaBundleOnThisPlatform`)
-  rather than reasoned about, because a wrong claim here is a security claim. A proxy-hop failure is **`ErrorCode::ProxyError`**, distinct from
+  Schannel and Apple SecTrust keep their OS store. That claim is checked on
+  each CI platform rather than reasoned about, because a wrong claim here is a
+  security claim - by **two** tests answering two different questions, and the
+  distinction matters because for a while only the first existed and the docs
+  read as though it covered both (#812).
+  `TlsBackend.AcceptsACustomCaBundleOnThisPlatform` asks the narrow one:
+  whether this build's backend refuses `CURLOPT_CAINFO` outright
+  (`CURLE_NOT_BUILT_IN`). It stands up no server and verifies nothing.
+  `CustomCaVerificationTest` (`tests/tls_verification_test.cpp`) asks the one a
+  user cares about, on a wire: an in-process HTTPS listener holds a certificate
+  a per-run CA signed (`tests/tls_server.hpp`), and the send verifies once that
+  CA is pasted into `customCaCertificates`, fails before it is, fails against a
+  CA that signed nothing here - the case that separates real verification from
+  a bundle read and ignored - and still verifies when a second anchor is added
+  beside it, which is the additive rule observed rather than asserted. That
+  listener is why `cpp-httplib` carries the `openssl` feature in
+  `engine/vcpkg.json`. Still structural, not on a wire: the *client*
+  certificate's handshake, tracked by #802.
+  A proxy-hop failure is **`ErrorCode::ProxyError`**, distinct from
   the target's `ConnectionFailed` - and `curl_to_error` now takes the handle,
   because a 407 answered to a CONNECT is a plain `CURLE_RECV_ERROR` and only
   `CURLINFO_HTTP_CONNECTCODE` remembers a proxy said no.
--- a/engine/CLAUDE.md
+++ b/engine/CLAUDE.md
@@ -260,8 +260,17 @@ Three things worth knowing before you design around them:
   a bundle read and ignored - and still verifies when a second anchor is added
   beside it, which is the additive rule observed rather than asserted. That
   listener is why `cpp-httplib` carries the `openssl` feature in
-  `engine/vcpkg.json`. Still structural, not on a wire: the *client*
-  certificate's handshake, tracked by #802.
+  `engine/vcpkg.json`. **The two backends do not answer this the same way, and
+  the wire is how we found out.** Where curl revocation-checks the chain itself
+  - the Schannel path does, passing `CERT_CHAIN_REVOCATION_CHECK_CHAIN` unless
+  told not to - a certificate authority minted for one test run is refused for
+  publishing no CRL, with the anchor loaded and the signature good. So the two
+  *positive* cases skip there, loudly and only on an error text naming
+  revocation, and #819 carries both halves of that: a fixture that serves a CRL,
+  and the user-facing question it exposes (someone pasting an internal CA with
+  no reachable distribution point gets the same refusal, and no doc says so).
+  The negative cases are asserted on every platform. Still structural, not on a
+  wire: the *client* certificate's handshake, tracked by #802.
   A proxy-hop failure is **`ErrorCode::ProxyError`**, distinct from
   the target's `ConnectionFailed` - and `curl_to_error` now takes the handle,
   because a 407 answered to a CONNECT is a plain `CURLE_RECV_ERROR` and only
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -469,6 +469,7 @@ if(VAYU_BUILD_TESTS)
         tests/load_stream_events_test.cpp
         tests/sse_stream_test.cpp
         tests/transport_policy_test.cpp
+        tests/tls_verification_test.cpp
         tests/client_certificates_test.cpp
         tests/requests_route_test.cpp
         tests/examples_route_test.cpp
--- a/engine/tests/client_certificates_test.cpp
+++ b/engine/tests/client_certificates_test.cpp
@@ -3,12 +3,15 @@
  * @brief The client-certificate registry (issue #707): what a row may say, how
  *        a target picks one, and that the pick reaches every driver.
  *
- * **What is not here, and why.** There is no live mTLS handshake. This suite
- * has no TLS server at all - `cpp-httplib` is built without OpenSSL support in
- * `engine/vcpkg.json`, deliberately - and the certificate *formats* the three
- * CI backends accept differ (Schannel wants a PKCS#12 or a store thumbprint
- * where OpenSSL wants a PEM pair), so a fixture built on checked-in PEMs would
- * fail on Windows while proving nothing about the code under test. What that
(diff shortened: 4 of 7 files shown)


athrvk:
**CI found something, which is the point of the change - recording it here so the PR body is not the only account.**

First run (`ffe1c19`, run 32199309562): ubuntu and macOS passed all four new cases; **Windows failed exactly the two positive ones** and passed both negatives.

That split is the diagnosis rather than a puzzle. If the pasted anchor were not being loaded and consulted, the *negative* cases would be the ones misbehaving - they assert a refusal, and a backend ignoring the bundle refuses just fine. Failing only where the CA is supposed to make the call succeed points at something else in the chain.

curl's Schannel path validates a `CURLOPT_CAINFO` chain itself, and asks Windows to revocation-check it (`lib/vtls/schannel_verify.c`, pinned baseline `94a5411`):

```c
if(!CertGetCertificateChain(cert_chain_engine, pCertContextServer, NULL,
                            pCertContextServer->hCertStore, &ChainPara,
                            (ssl_config->no_revoke ? 0 :
                             CERT_CHAIN_REVOCATION_CHECK_CHAIN),
                            NULL, &pChainContext)) {
```

and treats an unknown status as failure unless `CURLSSLOPT_REVOKE_BEST_EFFORT` is set. A CA minted seconds ago publishes no CRL, so that answer can only be "unknown". The OpenSSL-backed legs make no such check.

**What `e90a546` does about it.** The two positive cases skip where that happens, built so the skip cannot rot into the quiet no-op #812 exists to end:

- it fires **only** when the backend's own error text names revocation - any other refusal, on any platform, still fails;
- it **prints** that text, so the reason comes from the run rather than from my reading of curl's source;
- it matches on the words, **not** on `_WIN32` - "this backend demands revocation information" is the property that matters, and a test asserting the host platform is what `CLAUDE.md` forbids;
- the **negative cases keep no carve-out** and are asserted on every leg. They are what proves the anchor is consulted, and they were green on Windows already.

The skip path itself was exercised locally before pushing (helper temporarily inverted to fire on success): it reports `[  SKIPPED ]` with the full reason and the backend's message appended, rather than reading as a pass.

**What I deliberately did not do.** Setting `CURLSSLOPT_REVOKE_BEST_EFFORT` when a custom bundle is in force is one line beside the existing `CURLSSLOPT_NATIVE_CA`, and it would turn every leg green. That is exactly why it is not in this PR: it weakens revocation enforcement as a side effect of a user pasting a certificate, which is a security posture change wearing a one-line diff, and it is not mine to make.

#819 carries both halves of what is left - a fixture that serves a CA-signed CRL so the positive case can run on a revocation-checking backend too, and the user-facing question underneath it: someone on Windows pasting an internal CA with no reachable distribution point hits this same refusal today, worded so it names neither the setting they changed nor the reason.

Local state on `e90a546`: `ctest --preset linux-dev` **2046 tests, 100% passed**; `mkdocs build --strict` clean.

---

PR #824, opened 2026-08-19
fix(app/electron): refuse a navigation off the app, and deny window.open (#822)

### LINKED ISSUE #822: app(electron): the main window still has no will-navigate refusal or setWindowOpenHandler - the anchor defense is single-layer by design, and #711 made it load-bearing for HTML too</title>
<parameter name="labels">["component:app", "type:enhancement"]
Filed from #711, which names this as a "recommended defense-in-depth rider, in
scope if small, its own issue if not". It is not small enough to ride along -
reasoning below - so it is filed rather than folded in.

## Problem

`markdown-view.tsx`'s decision 1 rests on a fact about the main window, stated
in three places (the component header, `markdown-view.test.tsx`'s header,
`docs/app/COMPONENTS.md`):

> The main window has no `will-navigate` handler, no `setWindowOpenHandler` and
> no CSP - the only `setWindowOpenHandler` in the tree is in `oauth.ts`.

Still true at `main.ts:210-251`: `createWindow` sets `contextIsolation: true`
and `nodeIntegration: false`, subscribes `page-title-updated`,
`render-process-gone` and `did-finish-load`, and nothing else. `oauth.ts:56`
denies window-opens for the *auth* window only.

So the renderer is one successful navigation away from handing
`window.electronAPI` to a third-party origin - the preload re-runs there - and
the only thing standing in front of that today is a single application-level
convention: the `a` renderer in `MarkdownView` emits a `<button>` instead of an
anchor. One component, one override, no second layer.

That convention is now doing more work than when it was written. #711 turned on
`rehype-raw`, so descriptions off the internet are parsed as markup rather than
printed as text. The sanitiser makes that safe on its own terms - `href`
survives only on an `a`, only for http(s), and the `a` renderer never puts it in
the DOM - and the hostile suite pins it. But the layer *underneath* is still
absent: if any surface ever renders an anchor by another path, nothing at the
window level refuses the navigation.

## Fix pathway

Follow the file's own precedent rather than adding handlers inline: `main.ts`
registers IPC and creates windows at import time, so anything testable is
extracted (`quit-signals.ts`, `oauth-window.ts`, `renderer-recovery.ts`), driven
through a fake, and the *wiring* is source-scanned.

- New `electron/window-navigation.ts` exporting an installer taking the
  `webContents`-shaped surface it needs, plus the predicate for what counts as
  internal. `will-navigate` is refused unless the target is the app's own
  origin, and `setWindowOpenHandler` returns `{ action: "deny" }`.
- **The predicate is the whole design question, and it is why this is not a
  one-liner.** In production the renderer is a `file://` load; in development it
  is `DEV_SERVER_URL`, and Vite's full-reload path must keep working - a refusal
  that catches an HMR reload breaks the dev loop for everyone, which is a worse
  outcome than the hole it closes. Decide deliberately, in one line the test can
  state, which navigations are the app's own.
- Whether a CSP belongs beside this is a related question with a different blast
  radius (the renderer loads its own bundle, styles and fonts); out of scope
  unless it turns out to be free.

## Tests

- `window-navigation.test.ts`, driving a fake `webContents` the way
  `quit-signals.test.ts` drives a fake `process`: an external `https://` target
  is refused, the app's own origin is not, a `window.open` is denied.
- A source scan of `main.ts` asserting the installer is actually called on the
  main window - the same "everything above this line would still pass with the
  handler never subscribed" gap `oauth-window.test.ts` closes by reading
  `oauth.ts`. It must assert it read a non-empty file.

## Docs - same commit

The three places quoted above all state the absence as a current fact and would
become false together: `app/src/components/ui/markdown-view.tsx` (decision 1),
`app/src/components/ui/markdown-view.test.tsx` (header), and
`docs/app/COMPONENTS.md` (Markdown rule 1). The anchor-as-button behaviour does
**not** change - it stays as the layer above.

## Sequencing

Standalone; nothing blocks it. Independent of #711 (merged or not): that PR's
sanitiser and this window guard defend different things, and neither needs the
other to be correct. No open issue touches `main.ts`'s window creation.

## Acceptance criteria

- [ ] The main window refuses a navigation to anything that is not the app's own
      origin, and denies `window.open`, through an extracted module driven by a
      fake in tests.
- [ ] The dev server's own reload path still works, asserted by the predicate's
      test rather than by hand.
- [ ] A guard fails if `main.ts` stops installing it.
- [ ] The three statements naming the absence say what is true instead.

---

athrvk:
## Description

**Plan.** The root cause is a defense that is one layer deep. `contextIsolation`
and `nodeIntegration` decide what a renderer may reach; neither decides *which
document the renderer is*, and the preload re-runs on whatever the main window
navigates to - so a navigation off the app hands `window.electronAPI` to that
origin. Verified against the current code before touching anything: `createWindow`
subscribed `page-title-updated`, `render-process-gone` and `did-finish-load` and
nothing else, and the only `setWindowOpenHandler` in the tree was `oauth.ts:56`,
for the auth window alone. The approach is the issue's: a new
`electron/window-navigation.ts` following the file's own precedent for anything
testable (`quit-signals.ts`, `oauth-window.ts`, `renderer-recovery.ts`) - an
installer taking the `webContents`-shaped surface it needs, driven through a
fake, with the wiring source-scanned. **Alternative rejected:** handlers written
inline in `createWindow`. It is fewer files and untestable - `main.ts` creates
windows and starts the engine at import time, so the predicate, which is the part
that can be wrong, would have no test at all.

### The predicate is the whole design question

The two builds are the app in two forms, and the identity that distinguishes each
is not the same one:

- **Production is a `file:` load, where an origin cannot be the test.** Every
  `file:` URL has the opaque origin `"null"`, so comparing origins would accept
  `file:///etc/passwd` as the app's own. The document is the identity there,
  which costs nothing the app wants - it never navigates between local files.
  Asserted directly, including the `new URL(...).origin` equality that is the
  reason for the branch.
- **Development is the Vite dev server, where the origin is exactly the test.** A
  full reload (an HMR update the client cannot apply in place) is page-initiated,
  so it arrives as a real navigation and may land on any path under that origin.
  A rule tighter than the origin would refuse it and break the dev loop for
  everyone, which is worse than the hole being closed.

Anything unparseable is refused rather than guessed at. `main.ts` passes the URL
it *actually* loads (`rendererEntry` is computed once and used by both the guard
and `loadFile`), because a second spelling of the entry path would drift from the
one the window is on - and a pinned test reads that.

**`window.open` is denied unconditionally.** Nothing in this app calls it -
checked, there is no `window.open` in `src/` or `electron/` outside tests - and
an outbound link goes through the scheme-validated `openExternalUrl` IPC to the
user's browser, which is where a third-party page belongs. A child window would
carry the preload the same way a navigation does.

**Deliberately not done: a CSP.** The issue scopes it out unless it turns out to
be free, and it is not - the renderer loads its own bundle, styles and fonts, so
it has a different blast radius and belongs in its own change. The statements
corrected below therefore still say a CSP is absent, because it is.

**The anchor-as-button behaviour does not change.** It stays as the layer above,
and the reason is now written next to it in all three places: a refused
navigation is a link that silently does nothing, where a button reaches the
user's browser, which is what the user meant.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

App-only diff (TypeScript plus two docs pages); no engine file is touched, so per
`CLAUDE.md`'s "scale the verification to what the change could actually break"
the engine build and ctest were not run - no C++ test could change colour.

**Re-run in full after merging `master` (`63e7815`) into this branch**, since
that merge brought in #823, which touches the same component:

- `pnpm test` - **5715 tests in 521 files, 100% passed**

The new file is 18 of those tests (`vitest run
electron/window-navigation.test.ts` on its own reports 18). Before the merge, on
`2c0715d`, the same suite was 5677 in 520 files.

Every guard mutation-checked (mutation applied, confirmed red, reverted -
`git diff` against the pushed tree is empty):

| Mutation | Result |
|---|---|
| predicate always returns `true` (guard refuses nothing) | **9 of 18 fail** - every refusal case, in both builds |
| the `file:` branch dropped, so production compares origins | **3 fail**, including `file:///etc/passwd` being accepted as the app's own |
| `setWindowOpenHandler` not installed | the `window.open` case fails |
| the installer never called in `main.ts` | **3 wiring tests fail** |
| installed *after* the load instead of before | the ordering test fails |
| baseline, unmutated | 18/18 pass |

The wiring scan asserts it read a non-empty `main.ts` and that the file contains
`createWindow` before anything else, per the precedent in `oauth-window.test.ts`
- a scan of an empty string would otherwise pass every assertion under it.

One note on what the tests deliberately do **not** assert: nothing here reads
`process.platform`. The predicate's two branches are the two builds, not the
three operating systems, so there is no host-platform assertion for
`CLAUDE.md`'s rule to catch.

**`mkdocs build --strict` could not run here** - mkdocs is not installed in this
environment. The docs edits are prose inside existing sections and add no new
links, headings or nav entries, so the failure modes that gate catches are not
reachable from this diff; CI's docs job covers it.

## Checklist

Docs in the same commit - the three places the issue names, all of which stated
the absence as a current fact and would have become false together:
`app/src/components/ui/markdown-view.tsx` (decision 1),
`app/src/components/ui/markdown-view.test.tsx` (header), and
`docs/app/COMPONENTS.md` (Markdown rule 1). Plus a **Window Navigation** bullet
in `docs/app/architecture.md`'s Security Considerations list, which described the
Electron posture and would otherwise have been the one place still implying this
layer does not exist.

## Related Issues

Closes #822

### The #823 conflict, resolved

#823 merged first and rewrote the same decision-1 paragraphs this corrects, so
both sides changed the same sentence for different reasons. `master` is merged in
and resolved here (`131ed67`):

- **`markdown-view.tsx` and `docs/app/COMPONENTS.md`** keep #823's newer
  four-decision structure. The clause replaced is the one that had become false
  on merge - #823 wrote that the anchor override "is a single layer by design,
  and stays one until #822 puts a `will-navigate` refusal underneath it", and
  #822 is this branch. Both now say the layer exists and why the override stays
  anyway.
- **`markdown-view.test.tsx`** merged cleanly; nothing to resolve.
- The guard module and its tests never overlapped #823 at all.

Nothing else in the merge needed a decision, and no behaviour changed in the
resolution - both conflicts were comment and prose only.

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -7,7 +7,7 @@
 
 import { app, BrowserWindow, dialog, ipcMain, nativeTheme, Menu, shell } from "electron";
 import path from "path";
-import { fileURLToPath } from "url";
+import { fileURLToPath, pathToFileURL } from "url";
 import { EngineSidecar } from "./sidecar.js";
 import { resolveAppPaths } from "./app-paths.js";
 import { readDataFile } from "./data-file.js";
@@ -16,6 +16,7 @@ import { setupOAuthIpcHandlers } from "./oauth.js";
 import { loadWindowState, trackWindowState } from "./window-state.js";
 import { initAutoUpdater, checkForUpdatesNow, disposeAutoUpdater } from "./updater.js";
 import { installQuitOnSignal } from "./quit-signals.js";
+import { installWindowNavigationGuard } from "./window-navigation.js";
 import { createSaveFlusher } from "./save-flush.js";
 import { createRendererRecovery } from "./renderer-recovery.js";
 import { createQuitShutdown } from "./quit-shutdown.js";
@@ -263,11 +264,21 @@ function createWindow() {
 		mainWindow?.show();
 	});
 
+	// The preload re-runs on whatever this window navigates to, so a navigation
+	// off the app hands `window.electronAPI` to that origin. Installed before the
+	// load, so no navigation can land ahead of the guard. See
+	// window-navigation.ts.
+	const rendererEntry = path.join(__dirname, "../dist/index.html");
+	installWindowNavigationGuard(
+		mainWindow.webContents,
+		isDev ? DEV_SERVER_URL : pathToFileURL(rendererEntry).toString()
+	);
+
 	if (isDev) {
 		mainWindow.loadURL(DEV_SERVER_URL);
 		// mainWindow.webContents.openDevTools();
 	} else {
-		mainWindow.loadFile(path.join(__dirname, "../dist/index.html"));
+		mainWindow.loadFile(rendererEntry);
 	}
 
 	mainWindow.on("page-title-updated", (event) => {
--- a/app/electron/window-navigation.test.ts
+++ b/app/electron/window-navigation.test.ts
@@ -1,0 +1,169 @@
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * What is at stake here is the preload, not tidiness: it re-runs on whatever the
+ * main window navigates to, so a navigation off the app hands
+ * `window.electronAPI` - the engine, the filesystem readers, the OAuth surface -
+ * to a third-party origin. Before this guard the only thing in front of that was
+ * one component rendering links as buttons.
+ *
+ * So the assertions worth holding are the two directions of the predicate. A
+ * guard that refuses everything closes the hole and breaks the app; a guard that
+ * refuses nothing passes any test that only ever checks the happy path. Both
+ * builds are therefore asserted from both sides, and the dev server's own reload
+ * path is stated as a case rather than assumed.
+ *
+ * These drive a fake `webContents` rather than Electron, which is the whole
+ * reason the module lives outside main.ts.
+ */
+
+import { describe, it, expect, vi } from "vitest";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+import {
+	installWindowNavigationGuard,
+	isAppOwnUrl,
+	type NavigableContents,
+} from "./window-navigation.js";
+
+// main.ts creates windows and starts the engine at import time, so the wiring
+// itself can only be read. Everything above this line would still pass with the
+// guard never installed, which is the whole of the bug being fixed.
+const mainSource = readFileSync(join(dirname(fileURLToPath(import.meta.url)), "main.ts"), "utf8");
+
+const DEV_URL = "http://localhost:5173";
+const PROD_URL = "file:///opt/Vayu/resources/app.asar/dist/index.html";
+
+/** A WebContents that never existed: records the listener so a test can fire it. */
+function fakeContents() {
+	let navigate: ((event: { preventDefault(): void }, url: string) => void) | undefined;
+	const openHandler = vi.fn();
+
+	const contents: NavigableContents = {
+		on(_event, listener) {
+			navigate = listener;
+			return this;
+		},
+		setWindowOpenHandler(handler) {
+			openHandler.mockImplementation(handler);
+			return this;
+		},
+	};
+
+	return {
+		contents,
+		openHandler,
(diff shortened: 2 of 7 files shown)


athrvk:
fix merge conflict

Issue #805, opened 2026-08-18
ci/tests: parallelize the fully-serial engine test suite and take the serial legs off CI's critical path - ctest -j behind unique scratch DBs, Windows vitest sharding, release-job splits


athrvk:
Owner-requested (2026-08-18), verified at master 6fdde30 with real timings from the Actions API (runs 956/958/968, 2026-08-17/18). Companion: #700 owns the engine sccache warm - the single biggest *build*-time lever (21-22.5 min cold Windows compiles) - and is deliberately out of scope here. This issue owns everything else: test parallelism and CI job structure.

**Status (2026-08-19): phases 1-2 shipped in PR #808, phase 3 in PR #816 - measured results in the comments. Remaining: phase 4 (release-job splits) and phase 5 (hybrid Windows ctest parallelism, added 2026-08-19 from the research pass in the comments).**

## Problem (verified)

Measured CI wall times at filing: an engine-touching PR took **28-31 min**, an app-only PR **~9.5 min**. Where it went:

| Leg | Duration | Dominant cost |
|---|---|---|
| Engine windows | 27m28s-29m50s | build 21m03s-22m34s (#700), **ctest 5m34s-6m19s serial** |
| Engine ubuntu | 17m00s-18m16s | build 11-14m (#700), **ctest 3m11s-5m12s serial** |
| Engine macos | 10m56s-15m14s | build 6-9.5m (#700), **ctest 4m02s-4m47s serial** |
| App windows | 9m03s-9m12s | **vitest 7m35s-7m58s, one unsharded run** - the entire wall of every app-only PR |
| App ubuntu/macos | 6m03s-6m50s | vitest ~5m, plus ubuntu runs lint 39s + format 13s + type-check 22s serially *before* vitest starts |

1. **The engine suite is 100 percent serial.** One binary (`vayu_tests`, `engine/CMakeLists.txt:446-571`, 94 test files) registered per-test via `gtest_discover_tests` (`:672-675`, ~1,600 tests), and nothing anywhere passes parallelism: no `execution.jobs` in any testPreset (`engine/CMakePresets.json:257-330`), no `-j` in `build.py:1221`/`:1261`, none in `pr-tests.yml:337` or `release.yml:163`. Serial ctest also pays ~1,600 process spawns + curl `global_init`s (`tests/main.cpp:12`).
[...]

Issue #700, opened 2026-08-16
ci(cache-warm): warm engine sccache - the exit condition the workflow named shipped in 0.18.0, so the release's 0% hit rate is now a structural gap, not a header regression


athrvk:
## Problem

The v0.18.0 release run reported **219 engine compile requests, 0 sccache hits (0.00%)** on Linux, with all 220 sccache warnings reading `read failed NotFound (permanent)`. The releasing skill (`.claude/skills/releasing/SKILL.md:44-51`) says zero hits means `vayu/version.hpp` has been pulled back into a widely included header. That diagnosis is wrong for this failure, and the header work is fine:

- `DEFAULT_USER_AGENT` lives in `engine/include/vayu/core/user_agent.hpp` as a declaration only; `engine/src/core/user_agent.cpp` is the sole TU that sees `VAYU_VERSION_STRING` for it, and `engine/include/vayu/core/constants.hpp:32` merely points there.
- `version_isolation_test.cpp` is registered (`engine/CMakeLists.txt:536`) and the engine suite passed on all three platforms in the release run.

The actual cause is that **no workflow ever writes engine sccache entries to a scope a tag build can read**. Actions caches are ref-scoped: a run reads its own ref plus the default branch. `pr-tests.yml` runs only on `pull_request` (`pr-tests.yml:29-32`), so its sccache entries land in `refs/pull/N/merge` where nothing else can reach them; `codeql.yml` compiles deliberately without sccache (`codeql.yml:146-149`, a hit would starve the tracer); and `cache-warm.yml` warms vcpkg and the pnpm store only. A tag build therefore misses every key no matter what the headers do - the scope it reads is empty by construction.

`cache-warm.yml` knows this and excludes an engine compile **on purpose**, with a stated exit condition (`cache-warm.yml:27-38`):

&gt; Until that include is narrowed (move `DEFAULT_USER_AGENT` out of `constants.hpp` so a version bump touches ~5 translation units instead of ~68), warming sccache here would burn ~25 minutes of runner time per merge to speed up nothing.

That narrowing shipped in `a672f79` (#659 item 5) and rode the v0.18.0 release. The precondition is met, the follow-on step was never taken, and the comment now argues from a fact that is no longer true. There is a second copy of the same stale reasoning at `cache-warm.yml:277-284` ("Configure only - no `buildPreset`... see the note at the top about sccache and the version header").

[...]

PR #826, opened 2026-08-19
fix(engine/docs): the TLS backend spread is two-way, and the build says so (#818)

### LINKED ISSUE #818: engine/docs(tls): two comments call macOS Apple SecTrust-backed, but the pinned vcpkg baseline resolves curl's default TLS to OpenSSL off Windows
Filed from #812 (PR #817), which found this while rewriting the neighbouring
claim and deliberately did not touch it - one unverified statement should not
be traded for another.

**Problem**

Two places describe the TLS backend spread as three-way, with macOS on Apple's
native store:

- `engine/CLAUDE.md` (the `TransportPolicy` bullet): "on an OpenSSL build
  CAINFO *is* the whole store, so the file is the system anchors plus the
  user's, while Schannel and Apple SecTrust keep their OS store."
- `engine/include/vayu/http/transport_policy.hpp`, on `system_ca_bundle_pem`:
  "Empty when they cannot be read as a file - which is the normal case on
  Windows (Schannel) and macOS (Apple SecTrust), where the anchors live in an
  OS store rather than a bundle."

The same three-way split is stated in `tests/transport_policy_test.cpp`'s
`TlsBackend` comment ("OpenSSL on Linux, Apple SecTrust on macOS, Schannel on
Windows") and in #812's own text.

The pinned baseline's `curl` port does not appear to agree. `curl`'s
`default-features` are `["non-http", "ssl"]`, and the `ssl` feature reads:

```json
"ssl": {
  "dependencies": [
    { "name": "curl", "default-features": false, "features": ["openssl"],
      "platform": "uwp | !windows" },
    { "name": "curl", "default-features": false, "features": ["sspi"],
      "platform": "windows & !uwp" }
  ]
}
```

`platform: "uwp | !windows"` covers macOS, so a vcpkg-built curl on macOS looks
OpenSSL-backed exactly as on Linux, and only Windows is on a native store. If
that is right, the spread is two-way, and the consequences are not cosmetic:

- `system_ca_bundle_pem`'s doc says macOS "normally" returns empty. On an
  OpenSSL build `curl_version_info()->cainfo` points at a real file, so it
  would normally return *content* there - and the additive merge in
  `materialize_ca_bundle` is doing real work on macOS rather than the no-op
  the comment implies.
- Conversely, if the merge is silently *not* finding a system bundle on the
  macOS runner (vcpkg's OpenSSL has its own default cert paths, which may not
  exist on a mac), then `CURLOPT_CAINFO` replaces the store with the user's
  anchors alone - and the additive-trust promise of #704 decision 4 would be
  broken on macOS while the docs say it holds. That is the failure worth
  ruling out, and it is the security-shaped one.

This is the same defect as #812: a claim phrased confidently enough that a
reader has no reason to doubt it. It is filed separately because settling it
needs a fact from a macOS build, which #812's PR could not produce.

**Fix pathway**

1. **Get the fact rather than reasoning from the port file.** The cheapest
   probe is a test that prints what this build actually is:
   `curl_version_info()->ssl_version` and whether `system_ca_bundle_pem()`
   comes back empty, asserted or logged on each CI leg. `TlsBackend` in
   `tests/transport_policy_test.cpp` is already the per-platform probe and is
   the natural home - it prints `ssl_version` on failure today, so making it
   *record* the backend rather than only mention it in a message is a small
   change with the whole answer in it.
2. **Then correct whichever statements are wrong**, in the same commit:
   `engine/CLAUDE.md`, `transport_policy.hpp`'s `system_ca_bundle_pem` doc, and
   the `TlsBackend` comment. If macOS is OpenSSL-backed, `SYSTEM_CA_BUNDLES`
   and the `cainfo` fallback in `system_ca_bundle_path` should be checked for a
   macOS-shaped candidate path while you are there.
3. **If the probe shows the merge finds nothing on macOS**, that is a bug, not
   a doc fix, and deserves its own issue rather than being folded in here.

**App changes**

None - engine doc/test only, unless step 3 turns up a real merge gap.

**Tests**

Whatever step 1 becomes. It must fail loudly on a leg whose backend it cannot
determine, rather than skipping quietly - a per-platform probe that quietly
does nothing on one platform is the shape #812 exists to end.

**Docs - same commit**

`engine/CLAUDE.md` and `engine/include/vayu/http/transport_policy.hpp`, per
step 2. `docs/architecture.md`'s trust paragraph does not name backends and
needs no change.

**Sequencing**

Standalone, and parallel-safe with everything open. Overlaps #812's PR (#817)
in `engine/CLAUDE.md` and `tests/transport_policy_test.cpp`, so land after
#817 to avoid a conflict in the same paragraph. Nothing blocks it and it
blocks nothing; #802 is unaffected.

**Acceptance criteria**

- [ ] What TLS backend each CI leg actually builds against is recorded by a
      test, not inferred from a port file.
- [ ] Every statement in the repo about the backend spread matches that
      record.
- [ ] Whether `system_ca_bundle_pem()` returns content on each leg is known,
      so the additive-trust claim of #704 decision 4 is checked on macOS rather
      than assumed.

---

athrvk:
## Description

**Plan.** The root cause is a claim nobody had read off a build. Six statements
across the engine and the docs described a three-way TLS backend spread with
macOS on Apple SecTrust; the pinned vcpkg baseline's `curl` port has no SecTrust
feature at all (its feature list is `openssl`, `ssl`, `ssls-export`, `sspi`,
`wolfssl`), and its default `ssl` feature maps to `openssl` on `uwp | !windows`
and to `sspi` only on Windows. So the spread is two-way. Verified the problem
still exists as described before touching anything - all six statements were
present at `63e7815`, and `#817` (which the Sequencing section says to land
after) is merged.

The approach is the issue's step 1 rather than a doc rewrite: get the fact from
the build, then correct whatever disagrees. **Alternative rejected:** correcting
the six statements from the port file alone. That trades one unverified claim
for another, which is the defect #812 and #818 share, and it would leave the
macOS row re-breakable by any baseline bump with nothing to catch it.

The correction is not cosmetic. On an OpenSSL build `CURLOPT_CAINFO` *replaces*
the trust store, so `materialize_ca_bundle` keeps #704 decision 4's additive
promise only by concatenating the platform's own anchors into the file it
writes. The old comments described that merge as a no-op on macOS; it is in fact
the one thing standing between a pasted corporate CA and a silently narrowed
trust store there.

### What the tests record

Three tests in the `TlsBackend` suite, each answering a different question:

- **`ClassifiesTheBackendsThisRepoCanBeBuiltAgainst`** exercises the classifier
  with both answers stubbed (`OpenSSL/3.6.3`, `Schannel`, plus two it must
  refuse to classify). This is `CLAUDE.md`'s rule about not asserting a
  per-platform branch only where it happens to be taken - without it, a
  classifier that answered "bundle file" to everything stays green on Linux and
  macOS and is caught only by the Windows leg.
- **`IsTheBackendEveryTrustStatementHereAssumes`** reads `ssl_version` off the
  build and pins the documented per-leg answer. A backend outside the set this
  repo has a statement about fails with an `ASSERT`, naming it - the issue's
  "fail loudly on a leg whose backend it cannot determine".
- **`FindsTheSystemAnchorsTheMergeExtends`** is the security-shaped half: where
  `CURLOPT_CAINFO` replaces the store, the system anchors must actually be
  found. `PastedCertificatesAreMaterializedBesideTheDatabase` already asserts
  the merge, but only when there are anchors to merge, so before this that
  failure mode was an unasserted branch on a green leg.

**Observed on this leg** (`linux-dev`, pinned baseline): `ssl_version` is
`OpenSSL/3.6.3` and `cainfo` is `/etc/ssl/certs/ca-certificates.crt`. The macOS
and Windows answers are what this PR's CI legs produce - that is the point of
the tests, and if macOS comes back as anything other than an OpenSSL family
build, `IsTheBackendEveryTrustStatementHereAssumes` will say so instead of this
PR quietly shipping a second wrong claim.

### Two deliberate notes

1. **`FindsTheSystemAnchorsTheMergeExtends` skips on an OS-store backend**, with
   a message naming it. That follows the revocation skip in
   `tls_verification_test.cpp` ("narrow and loud") - an OS-store backend has no
   file to merge, so there is nothing there to be right or wrong about. The
   "cannot determine the backend" case is a hard `ASSERT`, not a skip. Asserting
   `system_ca_bundle_pem()` is *empty* on Windows was considered and rejected: a
   user who exports `CURL_CA_BUNDLE` legitimately gets content, and the test
   would fail a correct configuration.
2. **One adjacent correction, from the same probe.** The `SYSTEM_CA_BUNDLES`
   comment said the pinned port is built without a compiled-in bundle path
   (`CURL_CA_FALLBACK`). The linux-dev build reports a real `cainfo` and never
   reaches that list, so which of the three probes answers is a property of the
   build, not the platform. It is in the same comment this change edits and I
   had direct evidence it was misleading, so leaving it would have been the
   exact defect being fixed. The test asserts that *some* probe answers rather
   than naming which.

Issue step 3 ("if the probe shows the merge finds nothing on macOS, that is a
bug and deserves its own issue") did not trigger here and cannot be settled
until this PR's macOS leg reports. If it comes back red, that is the finding,
and it gets its own issue rather than a loosened assertion.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`python build.py -e -t` then `ctest --preset linux-dev --output-on-failure`:
**2057 tests, 100% passed, 0 failed** (61.97s). Baseline on `63e7815` was 2054;
the three new tests are the difference. No pre-existing failures.

Engine-only diff (C++ plus two Markdown files), so per `CLAUDE.md`'s "scale the
verification to what the change could actually break" the app suite was not run
- no TypeScript file is touched and no vitest test could change colour.

Every guard mutation-checked (mutation applied, rebuilt, confirmed red,
reverted; `git diff` against the pushed tree is empty):

| Mutation | Result |
|---|---|
| classifier answers "bundle file" to everything | `ClassifiesTheBackendsThisRepoCanBeBuiltAgainst` fails on 3 of its 4 cases - and the other two tests stay **green on Linux**, which is precisely why that test exists |
| `OpenSSL` dropped from the classifier's family list | **3 of 4 fail**, including both live probes |
| `system_ca_bundle_path()` returns `{}` (no probe answers) | `FindsTheSystemAnchorsTheMergeExtends` fails with `TLS backend 'OpenSSL/3.6.3' verifies against the file CURLOPT_CAINFO names, but no system bundle was found` |
| baseline, unmutated | 4/4 `TlsBackend` pass, 2057/2057 suite |

A separate probe built directly against the vcpkg curl in the build tree is
where the `ssl_version` / `cainfo` values above come from; it is not committed,
since the test now carries the assertion.

**`mkdocs build --strict` could not run here** - mkdocs is not installed in this
environment. The `docs/` edits are prose and table-cell text inside existing
sections, adding no link, heading or nav entry, so the failure modes that gate
catches are not reachable from this diff; CI's docs job covers it.

**No clang-format run.** The system `clang-format` (18.1.3) disagrees with all
three touched C++ files wholesale at `HEAD` (1039, 600 and 117 violations before
my changes), so formatting them would reflow files that were never clean under
this binary - `CLAUDE.md` says format only files that were clean before. New
code matches the surrounding style by hand. No CI job gates C++ formatting.

## Checklist

Docs in the same commit - all six statements, plus the adjacent correction:
`engine/CLAUDE.md` (the `TransportPolicy` bullet),
`engine/include/vayu/http/transport_policy.hpp` (`system_ca_bundle_pem`),
`engine/src/http/transport_policy.cpp` (the merge comment and
`SYSTEM_CA_BUNDLES`), `engine/tests/transport_policy_test.cpp` (the `TlsBackend`
comment), and `docs/engine/api-reference.md` (the trust-anchor table and the
client-certificate format paragraph, which named SecTrust for macOS too).

`docs/architecture.md`'s trust paragraph names no backend and needed no change,
and `tls_verification_test.cpp`'s comment was already hedged correctly by #812.

## Related Issues

Closes #818

Acceptance criteria:

- [x] What TLS backend each CI leg builds against is recorded by a test, not
      inferred from a port file.
- [x] Every statement in the repo about the backend spread matches that record -
      swept with `rg` for `SecTrust`/`Schannel`/`OpenSSL` across the tree, six
      hits, all corrected.
- [x] Whether `system_ca_bundle_pem()` returns content on each leg is known -
      asserted, not assumed, so #704 decision 4 is checked on macOS.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QcpbNBV2jeDCdjjrqJaNRA)_


athrvk:
CI result, and the fact this PR existed to get.

**The macOS leg is green, which settles #818.** Both new probes ran there, so:

- macOS builds against an OpenSSL-family backend, not Apple SecTrust -
  `IsTheBackendEveryTrustStatementHereAssumes` would have failed otherwise, naming
  what it found. The port-file reading is confirmed by a build.
- `system_ca_bundle_pem()` returns **content** on macOS -
  `FindsTheSystemAnchorsTheMergeExtends` asserts it and passed. So #704 decision 4's
  additive-trust promise holds there: a pasted CA extends the platform's anchors
  rather than replacing them. That was the security-shaped failure the issue asked to
  rule out, and it is ruled out rather than assumed. Step 3 of the fix pathway ("if
  the merge finds nothing on macOS, that is a bug needing its own issue") does not
  trigger - no follow-up issue needed.

Windows is green too (Schannel, so the anchors test skips loudly with the backend
named). Per-leg record now: Linux and macOS OpenSSL with anchors found, Windows
Schannel with an OS store.

| leg | result |
|---|---|
| Engine on macos-latest | success (05:00:33) |
| Engine on windows-latest | success (05:16:32) |
| Engine on ubuntu-latest | **still running - see below** |
| App legs, quality checks | skipped (no app file in the diff) |
| CodeQL JS/TS, conventional title, labels | success |

**The ubuntu leg is stuck in GitHub's runner setup, not in this change.** It has sat
in step 3, `Install system dependencies (Linux)` - an `apt-get` - for 68 minutes,
having completed only `Set up job` and `Checkout`. It has never reached `Setup
vcpkg`, `Build engine` or `Run engine tests`, so nothing in this diff has been
compiled there, let alone run. Not re-running it while it is still `in_progress`;
if it fails or is cancelled, one re-run is the right response, and a second failure
would be real.

What stands in for it in the meantime: the identical commit compiled and ran the full
suite green on the other two platforms, and locally on Linux `python build.py -e -t`
plus `ctest --preset linux-dev --output-on-failure` reports **2057/2057 passing**
(61.97s) with the three `TlsBackend` tests among them. The local Linux probe reads
`ssl_version=OpenSSL/3.6.3`, `cainfo=/etc/ssl/certs/ca-certificates.crt`, which is
what the ubuntu leg should confirm once it gets a runner that installs packages.

---

athrvk:
Resolved - **CI is green on all three engine legs**, so the blocker in my last comment is cleared and everything there about the macOS finding stands.

The ubuntu leg's hang was the runner, exactly as read: attempt 1 sat 60 minutes in `apt-get` and was killed by the step timeout at 05:52:14, never reaching `Setup vcpkg`, `Build engine` or `Run engine tests`. That is the one case worth a re-run (died before any test body ran), so I re-ran failed jobs once - attempt 2 on the same commit:

| step | attempt 1 | attempt 2 |
|---|---|---|
| Install system dependencies (Linux) | 60m00s, cancelled | **13s** |
| Build engine | never reached | 55s |
| Run engine tests | never reached | **53s**, green |

2m24s end to end. Worth noting *why* that is credible rather than suspicious: the engine sccache warm from #700/#806 means the compile is a cache read against the master scope, and #805/#808's `ctest -j` is what turns a 3-5 minute serial ctest into 53s. Both landing is the difference between this leg taking 17 minutes and taking two - a fair independent datapoint for those issues.

Also merged `master` (`f9a4cee`, the 0.19.0 release) into this branch, which was `behind` after #825 landed while CI ran. No conflict - the release commit touches `VERSION`, `version.hpp`, `CMakeLists.txt`, `app/package.json` and the notes file, none of which this PR goes near. CI is re-running against the merge.

---

Issue #833, opened 2026-08-19
engine/app(tls): a client certificate can only go out as a PEM pair - Schannel (Windows) takes PKCS#12, and the registry has no way to say which


athrvk:
Filed from #802's live mTLS fixture, which cannot run on the Windows leg for this reason and skips there rather than asserting a shape the engine cannot produce.

## Problem

`detail::apply_transport_policy` (`engine/src/http/event_loop/curl_utils.cpp`) writes `CURLOPT_SSLCERT`, `CURLOPT_SSLKEY` and `CURLOPT_KEYPASSWD` for the matched registry entry, and **no `CURLOPT_SSLCERTTYPE`** - so libcurl reads both files in its default format, PEM.

That is the right and only shape on every OpenSSL-backed leg, which since #818 means Linux **and** macOS (`TlsBackend.IsTheBackendEveryTrustStatementHereAssumes` reads the backend off the build on each CI leg). Windows is the exception: curl there is Schannel-backed, and Schannel takes a client identity as a PKCS#12 file - which libcurl only reads when it is told `CURLOPT_SSLCERTTYPE` is `P12` - or as a certificate-store reference. So on Windows a registered entry has no working shape today: a PEM pair is a format the backend does not accept, and a `.p12` is read as PEM and fails on the parse.

`docs/engine/api-reference.md` already tells the user the second half of this and not the first:

&gt; The format the certificate files must be in is the TLS backend's business, and the two backends differ - OpenSSL (Linux and macOS) takes a PEM certificate and key pair; Schannel (Windows) expects a PKCS#12 file or a certificate store reference. The engine passes the paths through unchanged […]

"Passes the paths through unchanged" is exactly what makes the PKCS#12 half not work: libcurl needs to be *told* the type, and nothing tells it.

[...]

Issue #819, opened 2026-08-19
engine/tests(tls): the custom-CA positive case cannot run on a revocation-checking backend - the fixture CA publishes no CRL, and the user-facing half of that needs a decision


athrvk:
Filed from #812 (PR #817), where CI found it. Two things live here - a test gap
and a question about shipped behaviour - because they have one cause and fixing
the first tells you the answer to the second.

**What happened**

#817 added `CustomCaVerificationTest`: an in-process HTTPS listener holding a
certificate signed by a CA minted for that run, driven through the ordinary
design-send path while `customCaCertificates` changes. Four cases. On its first
CI run, ubuntu and macOS passed all four; **Windows failed exactly the two
positive ones** - `AHostSignedByTheAddedCaVerifies` and
`TheAddedCaExtendsTheStoreRatherThanReplacingIt` - while both negative cases
(no CA pasted, wrong CA pasted) passed. Run 32199309562, job 95909543582:
2045/2047 passed.

That split is the diagnosis. The anchor is being loaded and consulted - if it
were not, the *negative* cases would be the ones misbehaving. Something else
about the chain is being refused.

**Why**

curl's Schannel path validates a `CURLOPT_CAINFO` chain manually, and asks
Windows to revocation-check it (`lib/vtls/schannel_verify.c`, at the pinned
baseline `94a5411`):

```c
if(!CertGetCertificateChain(cert_chain_engine, pCertContextServer, NULL,
                            pCertContextServer->hCertStore, &ChainPara,
                            (ssl_config->no_revoke ? 0 :
                             CERT_CHAIN_REVOCATION_CHECK_CHAIN),
                            NULL, &pChainContext)) {
```

and then treats an unknown revocation status as a failure unless
`CURLSSLOPT_REVOKE_BEST_EFFORT` is set:

```c
if(ssl_config->revoke_best_effort) {
  dwTrustErrorMask &= ~(DWORD)(CERT_TRUST_REVOCATION_STATUS_UNKNOWN |
                               CERT_TRUST_IS_OFFLINE_REVOCATION);
}
```

The engine sets neither option. A certificate authority created seconds ago
publishes no CRL and names no distribution point, so the revocation answer can
only be "unknown" - and the chain is refused despite a loaded anchor and a good
signature. The OpenSSL-backed legs do no such check, which is why only Windows
split.

This is a diagnosis from the pinned source plus the observed split, **not** from
a read of the failure text: the message is buried mid-log on a leg no one can
[...]

PR #834, opened 2026-08-19
test(engine/tls): a registered client certificate on a wire, and the refusals that read as internal errors (#802)

### LINKED ISSUE #802: engine/tests(tls): a live mTLS fixture server - the handshake #707 could only assert structurally

athrvk:
## Description

**Plan.** The root cause is that nothing in the suite had ever put a client certificate on a wire: `client_certificates_test.cpp` registers paths whose contents are the bytes `not-a-real-certificate`, so every assertion about mTLS was about the *registry*, not about the handshake it exists to complete. Verified against the current tree before touching anything, and one half of the issue's premise had gone stale: #802 says "there is no TLS server in the suite at all", but #812/#819 added one (`tests/tls_server.hpp`, a per-run CA, a served CRL) and `cpp-httplib[openssl]` is already in `engine/vcpkg.json`. So the dependency phase of the fix pathway is not needed, and the approach is the one `engine/CLAUDE.md` names: extend `TlsServer` with a client-certificate store rather than stand up a second listener, reusing that CRL. **Alternative rejected:** a second, mTLS-only listener. It would need its own CA, its own CRL and its own distribution point, and the CRL apparatus is exactly the part #819 had to fix once already - two copies is two things to keep current.

`httplib::SSLServer::PemMemory` already carries a `client_ca_pem` field the old fixture left null, and filling it makes the library load that authority and switch the context to `SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT`. So the fixture change is one string, and the refusal - the thing every positive assertion here rests on - is the library's, not a re-implementation of one.

### What the wire said that memory could not

Six cases pass through drivers that were previously only asserted structurally: a design send, an SSE stream, a load run (three transfers over pooled handles) and `pm.sendRequest` all complete a real handshake with a registered certificate; a port-specific entry holding the accepted identity beats a catch-all holding one signed by a **different** CA, so precedence is proven by which connection opens rather than by which pointer a lookup returned; and an encrypted key opens with its stored passphrase and stops connecting with the wrong one, which is what tells a passphrase that is *used* from one that is merely stored.

Every row is registered through `POST /client-certificates` and read back through `resolve_transport_policy`, so what is under test is the shipped path end to end - no handle is assembled in the test.

### The finding: three refusals read as `INTERNAL_ERROR`

The first run of these tests turned up a real defect, which is why this PR is not test-only. `curl_to_error`'s `default:` arm mapped to `ErrorCode::InternalError`, and **every client-certificate refusal lands in that arm**:

| What happened | CURLcode | Message libcurl gave | Was | Now |
|---|---|---|---|---|
| server demanded a certificate, none registered | 56 `CURLE_RECV_ERROR` | `tlsv13 alert certificate required` | `INTERNAL_ERROR` | `SSL_ERROR` |
| identity signed by an authority the server does not trust | 56 `CURLE_RECV_ERROR` | `tlsv1 alert unknown ca` | `INTERNAL_ERROR` | `SSL_ERROR` |
| stored passphrase does not open the key | 43 `CURLE_BAD_FUNCTION_ARGUMENT` | `unable to set private key file` | `INTERNAL_ERROR` | `SSL_ERROR` |

The first two are receive errors because under TLS 1.3 the server's verdict about the *client* arrives after the client's own handshake has finished - so libcurl reports the ordinary code of a connection that went away mid-request, with the alert only in the message text. `INTERNAL_ERROR` tells a user that Vayu broke when the truth is that the endpoint refused their certificate.

The fix follows the precedent `proxy_refused_tunnel` set two lines above it: consult the handle, because one event is spread across several CURLcodes and a code list would be a list to extend forever. `tls_connection_never_answered` asks whether an `https` transfer produced **no response line at all**, and it is consulted **only in the `default:` arm** - after every code with a meaning of its own has been matched. So a timeout, a refused connection, a resolve failure, a malformed URL, a proxy refusal and any `4xx` all keep exactly the code they had. What it costs is stated in the code and the docs rather than hidden: an https endpoint that accepts a connection and hangs up before answering for a reason of its own now reads as a TLS error - which is still a better answer than "internal", since nothing inside the engine failed, and libcurl's own message travels either way.

### Windows: skipped, loudly, with a guard on the skip

The per-backend format matrix is the second half of the issue, and it is recorded as a tested classifier rather than a claim. OpenSSL (Linux **and** macOS since #818) takes a PEM certificate and key pair; Schannel takes PKCS#12 or a certificate-store reference. The engine writes `CURLOPT_SSLCERT` / `CURLOPT_SSLKEY` and **no `CURLOPT_SSLCERTTYPE`**, so a PEM pair is the only shape it can present at all - which means the Windows leg cannot host this fixture, and it skips with that sentence printed.

Two guards keep that honest. `ClassifiesTheBackendsThisRepoCanBeBuiltAgainst` exercises the classifier with **both** answers stubbed, so the Windows branch is asserted on every leg rather than only where it is taken. `NoBackendIsSkippedForAReasonThatWentAway` reads `curl_utils.cpp` and fails the moment `CURLOPT_SSLCERTTYPE` appears in it - a skip cannot outlive its cause. The gap itself is filed as **#833** (engine + app: a format on the row, no key path for a PKCS#12 entry, the card showing it), rather than widened into this PR.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Engine-only diff, and it changes shared error mapping, so the whole engine suite ran rather than the new file alone. No app file is touched.

- `python build.py -e -t` - clean
- `ctest --preset linux-dev --output-on-failure` - **2069 tests, 100% passed, 0 failed** in 79s (`-j8`; the baseline on `2eaeee7` is 2058, and the 11 are this file's)
- `mkdocs build --strict -f .github/mkdocs.yml` - clean (mkdocs installed here for the docs edits)

Mutation checks (mutation applied, confirmed red, reverted; `git diff` against the pushed tree is empty):

| Mutation | Result |
|---|---|
| the `default:` arm maps to `InternalError` again (the state before this PR) | **3 fail** - every refusal: no certificate, wrong authority, wrong passphrase |
| the new rule is consulted *before* the switch instead of inside `default:` | **1 fail** - an unreachable https endpoint stops reading as `ConnectionFailed` |
| the listener stops asking for a client certificate (`client_ca_pem` back to null) | **2 fail** - both refusals, since anyone is now let in |
| the applier stops writing `CURLOPT_KEYPASSWD` | **1 fail** - the encrypted key no longer opens |
| `match_client_certificate` returns the first host match (no port precedence) | **1 fail** - the catch-all's refused identity goes out on a port another entry names |
| `CURLOPT_SSLCERTTYPE` appears in the applier | **1 fail** - the guard on the Windows skip, as designed |
| baseline, unmutated | 2069/2069 |

Formatting, per `CLAUDE.md`'s "format only files you touched that were clean before": `curl_utils.cpp` and `transport_policy_test.cpp` were **not** clang-format-clean at `HEAD`, so both are edited by hand and left otherwise untouched - running the formatter over `curl_utils.cpp` reflowed six unrelated hunks, which were reverted. The two new files are formatted, and `tls_server.hpp` and `client_certificates_test.cpp` were clean before and are clean after.

Local `clang-tidy` is 18.1.3 here and the repo's own pre-commit probe refuses anything below 19, so it did not lint - CI's toolchain does.

## Checklist

Docs in the same commit:

- `docs/engine/api-reference.md` - the client-certificates section said the engine "passes the paths through unchanged" and left the reader to conclude PKCS#12 works on Windows; it now says what actually goes out (a PEM pair), that the handshake is asserted on a wire on the OpenSSL legs, and that Windows cannot present one until #833. The TLS-trust section gains the new `SSL_ERROR` rule and what it deliberately does not cover.
- `engine/CLAUDE.md` - the client-certificate bullet ("still structural, not on a wire" is no longer true), the PEM-only constraint with its reader, and the `curl_to_error` rule beside the `ProxyError` one it follows.
- `engine/tests/client_certificates_test.cpp` - its header's "what is not here, and why" note pointed at a deferral that has now landed; it now states the split between the two files.

## Related Issues

Closes #802

Acceptance criteria:

- [x] A registered certificate is proven to complete a real mTLS handshake on at least one CI platform, on the design, SSE, load and `pm.sendRequest` paths. Proven on both OpenSSL legs (Linux and macOS), each driver asserted separately.
- [x] Per-backend certificate format is recorded as a tested matrix, not a claim - and any platform that cannot run the fixture skips with its reason printed. Both branches of the classifier asserted on every leg; the Schannel skip prints the reason and names #833.
- [x] A server-side rejection is asserted to surface as `SslError`. It did not before this PR - see the table above.
- [x] The deviation note in `client_certificates_test.cpp`'s header is updated.

Filed rather than folded in:

- **#833** - a certificate format on the registry row, so the Schannel build can present a PKCS#12 and the skip can go away. Engine plus app (the Settings card stops demanding a key path for a format that has none), which is a feature rather than the test this issue asked for.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NLxt1xHtRR24tXQqJuAc1B)_

### DIFF
--- a/docs/engine/api-reference.md
+++ b/docs/engine/api-reference.md
@@ -691,6 +691,19 @@ own `SSL_ERROR` - the validation on the way in is about the *shape* of the
 paste, deliberately, so a chain curl would accept is never refused by a parser
 of ours.
 
+**A TLS connection that never answers is an `SSL_ERROR` too**, whatever code
+libcurl put on it. The mapping needs that rule because the interesting
+refusals do not arrive as TLS errors: under TLS 1.3 the server's verdict about
+the *client* - a certificate it will not accept, or one it demanded and did not
+get - reaches us after our own handshake has finished, so libcurl reports the
+generic receive error of a connection that went away, carrying the alert in its
+message. A key the stored passphrase does not open arrives as a generic
+argument error in the same way. All three used to surface as `INTERNAL_ERROR`,
+which says the failure was ours. The rule applies only where the mapping has no
+answer of its own, and only to an `https` transfer that produced no response
+line at all - so a `4xx`, a timeout, a refused connection and a proxy failure
+all keep the code they had (issue #802).
+
 Per-request, `verifySSL: false` turns verification off for one endpoint
 entirely (see [POST /execute](#post-execute)); trusting the authority here is
 the answer that keeps verification on everywhere else.
@@ -2068,12 +2081,19 @@ fails at handshake time as an SSL error against the endpoint, which reads as
 A cert-authenticated exchange says so. `POST /execute` returns
 `clientCertificate` on the response (`""` when none matched) and the stored
 trace carries the same value under the same name, so a restored response names
-the entry a live one named. The format the certificate files must be in is the
-TLS backend's business, and the two backends differ - OpenSSL (Linux and macOS)
-takes a PEM certificate and key pair; Schannel (Windows) expects a PKCS#12 file
-or a certificate store reference. The engine passes the paths through
-unchanged, and a format a backend rejects fails at handshake time with
-libcurl's own error.
+the entry a live one named.
+
+**A registered pair goes out as PEM.** The engine writes the two paths and no
+certificate *type*, so libcurl reads them in its default format - a PEM
+certificate and a PEM key, optionally encrypted under the stored passphrase.
+That is what every OpenSSL-backed build takes, which is Linux and macOS, and
+the handshake is asserted on a wire there (a design send, an SSE stream, a load
+run and a `pm.sendRequest` against a listener that demands a client
+certificate). **Windows cannot present one yet**: that build is Schannel-backed
+and wants a PKCS#12 file or a certificate-store reference, neither of which
+libcurl will read without being told the type. Issue #833 tracks giving a row
+its format; until it lands, an entry registered on Windows fails at handshake
+time with libcurl's own error.
 
 ### GET /client-certificates
 
--- a/engine/CLAUDE.md
+++ b/engine/CLAUDE.md
@@ -295,14 +295,22 @@ Three things worth knowing before you design around them:
   so the skip and its helpers are gone and a refusal naming revocation is an
   ordinary failure again. What #819 keeps is the user-facing half, which is a
   decision rather than a fix: someone pasting an internal CA with no reachable
-  distribution point gets the same refusal, and no doc says so. Still
-  structural, not on a wire: the *client* certificate's
-  handshake, tracked by #802 - which should reuse this CRL rather than build a
-  second one.
+  distribution point gets the same refusal, and no doc says so. **The client
+  certificate's handshake is on a wire too now** (#802): the same `TlsServer`,
+  built with a second CA, demands a certificate that authority signed, so it
+  reuses this CRL rather than standing up a second listener.
   A proxy-hop failure is **`ErrorCode::ProxyError`**, distinct from
   the target's `ConnectionFailed` - and `curl_to_error` now takes the handle,
   because a 407 answered to a CONNECT is a plain `CURLE_RECV_ERROR` and only
-  `CURLINFO_HTTP_CONNECTCODE` remembers a proxy said no.
+  `CURLINFO_HTTP_CONNECTCODE` remembers a proxy said no. The handle answers a
+  second question for the same reason (#802): **an https transfer that failed
+  for a code with no mapping and produced no response line is an `SslError`**,
+  not the `InternalError` the default arm used to give it. Every
+  client-certificate refusal lands there - under TLS 1.3 the server's verdict
+  about the client arrives after the client's handshake finishes, so curl
+  reports `CURLE_RECV_ERROR` with the alert in its message. The rule is the
+  *shape*, never a list of codes, and it is consulted only after every mapping
+  with a meaning of its own has been tried.
 - **A client certificate belongs to a host, not to a request** (#707,
   `client_certificates`). The registry rides *inside* the `TransportPolicy`, so
   it reaches every outbound path with no per-site wiring and is read once per
@@ -316,7 +324,13 @@ Three things worth knowing before you design around them:
   entry is recorded on `Response::client_certificate` and travels both design
   funnels (live body and stored trace) under `clientCertificate`, deliberately
   *not* on the load path: it is a per-transfer string for a fact that is
-  constant for the run.
+  constant for the run. **What goes out is a PEM pair and nothing else**
+  (`tests/mutual_tls_test.cpp`, #802): the applier writes no
+  `CURLOPT_SSLCERTTYPE`, so libcurl reads both files in its default format -
+  right on every OpenSSL leg, and the reason the Schannel build, which wants
+  PKCS#12, can present nothing at all (#833). The live fixture skips that leg
+  with the reason printed and scans the applier so the skip cannot outlive its
+  cause.
 
 ## Request composition (engine-owned - POST /compose)
 
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -471,6 +471,7 @@ if(VAYU_BUILD_TESTS)
         tests/transport_policy_test.cpp
         tests/tls_verification_test.cpp
         tests/client_certificates_test.cpp
+        tests/mutual_tls_test.cpp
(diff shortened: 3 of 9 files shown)


athrvk:
CI green on all three legs (run 32236703057), and the numbers are worth recording because the Windows one is the claim this PR makes about itself:

| Leg | Result |
|---|---|
| ubuntu-latest | 2069 run, 2069 passed, 0 skipped |
| macos-latest | 2069 run, 2069 passed, **0 skipped** - the mTLS cases really ran here, which is the second OpenSSL leg proving the handshake rather than inheriting Linux's word for it |
| windows-latest | 2070 run, 2061 passed, **9 skipped**, 0 failed |

The 9 Windows skips are the 8 `MutualTlsTest` cases, each printing why (Schannel takes PKCS#12, the applier sends no certificate type - #833), plus one pre-existing platform skip; Windows also runs one test the others do not, which is where the 2070 comes from. The three tests that are *not* fixture-bound - the two classifier assertions and `AnUnreachableHttpsEndpointIsStillAConnectionFailure` - ran and passed on every leg, so the Windows branch of the matrix and the error-mapping rule are both asserted where they are not taken.

`CI gate` green.

---

Issue #838, opened 2026-08-19
engine/tests: the scratch-database lock caps Windows ctest at ~80% of its serial wall - making those tests cheap to run concurrently is the next lever


athrvk:
Follow-up to #805 phase 5, filed with the measurement that changes what that phase can be expected to deliver.

## Problem

Phase 5 gives every test that opens a scratch `Database` a shared CTest `RESOURCE_LOCK` on Windows, so no two of them overlap while the rest of the suite runs `-j4`. That is the right shape, and it is bounded - the pathological `-j4` (37 min against a ~6 min serial run) cannot come back, because the tests that caused it never run at the same time.

What it cannot do is reach the "~2.5 min or better" figure in #805's acceptance criteria. Measured locally on a 4-core box (Debug, 2070 tests, all four runs green):

| run | wall |
|---|---|
| serial | 299.6s |
| `-j4`, no lock | 84.8s |
| `-j8`, no lock (the `linux-dev` preset) | 58.6s |
| `-j4`, lock forced on (the model Windows now runs) | 243.7s |

The locked group is **817 of 2070 tests but 244.0s of the 300.5s serial wall - 81%**. A locked group can never run faster than its own serial time, so the hybrid model's floor is ~80% of serial: on Windows, roughly 4.5 min against today's 4m40s-6m38s, not 2.5.

The reason the earlier estimate looked better is a sampling artifact worth recording. The six fixtures named in #805's 2026-08-19 research comment (ScenarioRunnerTest, ScenarioPlanTest, TreeOrderTest, ReorderRouteTest, SpecSyncRouteTest, ScenarioLoadTest) measure 197 tests / 57.2s / 19% of the wall on this box - within a point of the Windows figures for the same six (196 tests, 58s, 19%), which is a useful independent check that this box is a fair proxy for the *shape* of the suite. But those six are not the database group. Tracing every suite under `strace` and keeping the ones that open a `*.db` file gives **54 suites**, and the ones that dominate the wall - the route fixtures, `MonitorRunTest`, `MidRunRefreshTest`, `LoadStrategyTest`, `DatabaseConcurrencyTest` - are all in it. Duration-pinned and database-opening are not two bands here; they are mostly the same tests.

## Fix pathway

The lock removes the contention by removing the concurrency. Getting past 80% means making concurrent scratch-database work cheap enough that the group can overlap again. Ranked:

[...]

PR #839, opened 2026-08-19
perf(engine/tests): run the Windows suite -j4 with the database tests locked apart

### LINKED ISSUE #805: ci/tests: parallelize the fully-serial engine test suite and take the serial legs off CI's critical path - ctest -j behind unique scratch DBs, Windows vitest sharding, release-job splits

athrvk:
## Description

#805 phase 5, option A. The Windows test presets ran the suite one test at a time because a plain `-j4` there measured ~37 min against a ~6 min serial run, while the same `-j4` cut ubuntu from 3-5 min to 1m15s.

**Plan.** Root cause (from #808 and the 2026-08-19 research pass on the issue): what costs on Windows is *concurrent SQLite commits* - the flush barrier `synchronous=FULL` issues per commit, plus the 25-250ms retry sleeps `os_win.c` answers a sharing violation with - not the scratch directories, not Defender, not volume placement. Serializing the whole suite is therefore a bigger answer than the problem needs. The approach here is the issue's preferred option A: give every test that opens a scratch `Database` a shared CTest `RESOURCE_LOCK` on Windows, so no two of them overlap while everything else runs 4-wide beside them. The rejected alternative is option B (LLVM-style whole-binary gtest sharding): it would also collapse the ~2000-process model, but it is a much larger change and it loses per-test CTest granularity - a failure would name a shard, not a test - for a win option A already captures. Also rejected: applying the lock on all platforms, which would only slow down Linux and macOS, where the same concurrency is already free.

Mechanically, the binary is discovered twice - once through a gtest filter matching the locked suites, once through its negation - so the two halves partition the suite exactly, whatever the list says, and only the property differs between them. Both halves are discovered on every platform, so a broken partition shows up as a changed test count everywhere rather than on Windows alone. Scratch-directory isolation goes back on for Windows: it is what makes `-j` safe in general, and keeping it means the lock list stays a performance decision rather than becoming load-bearing for correctness.

**The result is smaller than the issue projected, and that is the main finding here.** The locked group is **817 of 2070 tests but 81% of the serial wall**. The six fixtures the 08:00 research comment measured on Windows (196 tests, 58s, 19%) reproduce almost exactly on this box (197 tests, 57.2s, 19%) - which is a good check that the box is a fair proxy for the shape of the suite - but they are a sample of the database group, not the group. Tracing every suite under `strace` and keeping the ones that open a `*.db` file gives 54 suites, and the fixtures that dominate the suite's duration (`MonitorRunTest`, `MidRunRefreshTest`, `LoadStrategyTest`, `DatabaseConcurrencyTest`, the route fixtures) are all in it. So Windows should land near four fifths of its serial time - roughly 4.5 min against today's 4m40s-6m38s - not the ~2.5 min the acceptance criterion names. That criterion needs the database tests to become cheap to run *concurrently*, which is filed as **#838** with the measurements.

**Deliberate deviations from the issue text**, both recorded above and in the docs:
- The locked list is derived by tracing rather than from the six fixture names, as the issue asked ("derive the definitive fixture list by grepping `temp_database.hpp` users"). It is larger than the issue assumed, which is what caps the win.
- The 10x Windows stability loop the criteria ask for is not runnable from this environment. The evidence is this PR's own `windows-latest` legs plus the local hybrid-vs-serial comparison below; the first CI run is also the first real measurement of the model.
- Phase 4 (`release.yml` splits) is untouched and stays open on #805.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update
- [x] Performance / test infrastructure

## Testing

Measured on a 4-core box (Debug, `linux-dev`), **2070 tests, 0 failures in every run**:

| run | wall | note |
|---|---|---|
| serial (`ctest -j1`) | 299.6s | baseline |
| `-j4`, no lock | 84.8s | what Linux does today |
| `-j8`, no lock (`linux-dev` preset) | 58.6s | no regression from this change |
| `-j4`, lock forced on | 243.7s | the model Windows now runs |

The partition itself: 817 + 1253 = 2070 discovered tests, the same total as before the split, and `RESOURCE_LOCK` present on exactly the 817 when forced on and on none of them on Linux.

Mutation checks, all three run and restored:

1. Rename a locked suite (`TreeOrderTest` -> `TreeOrderTests`) - configure fails: *"vayu_scratch_database_suites names 'TreeOrderTests', which no test declares."*
2. Delete the only suite representing a database file - configure fails naming `tests/tree_order_test.cpp`.
3. Point the source scan at a pattern that matches nothing - configure fails: *"Found no TEST/TEST_F declarations ... would pass no matter which suite was missing."* (the non-empty-scan rule, since a guard reading an empty scan passes forever).

No app-side change: #805's "App changes (required)" shipped with phase 3 in #816. No pre-existing failures were seen on this base.

## Checklist
- [x] I have added tests (if applicable) - test infrastructure, verified by the mutation checks and the timing table above

`docs/engine/building.md` rewrites the "Windows presets run serially, on purpose" section into the hybrid model with the measured table and the ceiling it implies; `docs/building.md` and `CLAUDE.md` carry the one-line version; `engine/tests/main.cpp` and `temp_database.hpp` no longer say a preset disables scratch isolation, because none does. No page was added, moved or renamed and no link changed, so the docs build is unaffected.

## Related Issues
Closes #805 - **no**, deliberately: phase 4 and the acceptance criteria that only a release run or a Windows measurement can confirm are still open there. Delivers #805 phase 5, and files #838 for the ceiling it leaves behind.

---
_Generated by [Claude Code](https://claude.ai/code/session_019SksdwtqoSewwcZPhE3wAw)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -109,8 +109,9 @@ pass `ctest -j8`, cutting its wall time on Linux and macOS to roughly a fifth
 open a scratch database sharing a CTest `RESOURCE_LOCK`** - a plain `-j` there
 measured ~6x *slower* than serial, because concurrent SQLite commits cost more
 than the concurrency returns, so those tests never overlap while the rest of the
-suite does. They are 81% of the serial wall, so expect Windows near four fifths
-of its serial time rather than a fifth; see `docs/engine/building.md`. A
+suite does. They are 81% of the serial wall, so Windows lands near its old
+serial time rather than at a fifth of it - the lock makes `-j4` survivable
+there, it does not make it fast; see `docs/engine/building.md`. A
 rebuild ~2.5min; the app suite ~90s. Running
 both after retouching a doc comment reads as diligence and is just latency. Ask
 what a failure would even look like before running anything: **if no test could
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -102,11 +102,15 @@ rename or signature change needs a *build*, to prove it still compiles. Only a
 behaviour change needs the covering tests, and the full suite belongs once
 before committing a substantial piece of work, not after every edit.
 
-The engine suite now runs multi-process on Linux and macOS - those test presets
-pass `ctest -j8`, cutting its wall time to roughly a fifth (each ctest test
-runs in its own process under a private scratch directory, so `-j` is safe;
-`ctest -jN` overrides). **The Windows presets stay serial deliberately** -
-parallelism measured ~6x *slower* there; see `docs/engine/building.md`. A
+The engine suite now runs multi-process on every platform - the test presets
+pass `ctest -j8`, cutting its wall time on Linux and macOS to roughly a fifth
+(each ctest test runs in its own process under a private scratch directory, so
+`-j` is safe; `ctest -jN` overrides). **Windows runs `-j4` with the tests that
+open a scratch database sharing a CTest `RESOURCE_LOCK`** - a plain `-j` there
+measured ~6x *slower* than serial, because concurrent SQLite commits cost more
+than the concurrency returns, so those tests never overlap while the rest of the
+suite does. They are 81% of the serial wall, so expect Windows near four fifths
+of its serial time rather than a fifth; see `docs/engine/building.md`. A
 rebuild ~2.5min; the app suite ~90s. Running
 both after retouching a doc comment reads as diligence and is just latency. Ask
 what a failure would even look like before running anything: **if no test could
--- a/docs/building.md
+++ b/docs/building.md
@@ -236,8 +236,9 @@ cmake --build --preset macos-prod
 ctest --preset macos-prod
 ```
 
-The Linux and macOS test presets run the suite multi-process (`ctest -j8`); the
-Windows ones run serially on purpose, because parallelism measures far slower
+Every test preset runs the suite multi-process - `ctest -j8` on Linux and macOS,
+`-j4` on Windows, where the tests that open a scratch database are additionally
+held apart from each other because concurrent SQLite commits cost far more
 there. Pass `-jN` to override. See
 [Engine build guide](engine/building.md#running-tests) for the numbers.
 
--- a/docs/engine/building.md
+++ b/docs/engine/building.md
@@ -327,10 +327,17 @@ the serial wall**, because the route and load fixtures that dominate the suite's
 duration all open a database. The six fixtures the earlier analysis measured on
 Windows (197 tests, 57.2s, 19% of the wall here - within a point of the Windows
 figures, which is what makes this box a usable proxy for the *shape* of the
-suite) are only the visible tip of that group. Expect Windows around four fifths
-of its serial time, not the halving a fast band would suggest; getting past that
-means making the database tests themselves cheaper to run concurrently, which is
-issue #838.
+suite) are only the visible tip of that group.
+
+**What the first CI run of the hybrid model measured: 5m27s, 2071 tests, 0
+failures** (run 32249239500). That is inside the 4m40s-6m38s band the serial legs
+spanned, so **it is not a speedup** - one sample cannot be, against a baseline
+that noisy. What it does establish is that the mechanism works: the same `-j4`
+that took ~37 min without the lock now finishes in the time serial used to take,
+which is what the 81% number predicts and the reason to hold the expectation
+there rather than at the ~2.5 min this was first scoped for. Getting a real
+speedup means making the database tests cheaper to run concurrently, which is
+issue #838; treat that as the open work, not this paragraph as a win.
 
 Isolation is the **default** on every platform now, so a hand-run `ctest -j` is
 safe wherever it happens; the Windows presets used to set
--- a/docs/engine/building.md
+++ b/docs/engine/building.md
@@ -329,15 +329,16 @@ Windows (197 tests, 57.2s, 19% of the wall here - within a point of the Windows
 figures, which is what makes this box a usable proxy for the *shape* of the
 suite) are only the visible tip of that group.
 
-**What the first CI run of the hybrid model measured: 5m27s, 2071 tests, 0
-failures** (run 32249239500). That is inside the 4m40s-6m38s band the serial legs
-spanned, so **it is not a speedup** - one sample cannot be, against a baseline
-that noisy. What it does establish is that the mechanism works: the same `-j4`
-that took ~37 min without the lock now finishes in the time serial used to take,
-which is what the 81% number predicts and the reason to hold the expectation
-there rather than at the ~2.5 min this was first scoped for. Getting a real
-speedup means making the database tests cheaper to run concurrently, which is
-issue #838; treat that as the open work, not this paragraph as a win.
+**What CI measured on the hybrid model: 5m27s and 4m50s** on two runs of the
+same tree (32249239500 and 32252015024), 2071 tests, 0 failures, 9 skipped. Both
+sit inside the 4m40s-6m38s band the serial legs spanned, so **this is not a
+speedup** - and with a baseline that noisy, two samples could not show one either
+way. What they do establish is that the mechanism works: the same `-j4` that took
+~37 min without the lock now finishes in the time serial used to take, which is
+what the 81% number predicts and the reason to hold the expectation there rather
+than at the ~2.5 min this was first scoped for. Getting a real speedup means
+making the database tests cheaper to run concurrently, which is issue #838; treat
+that as the open work, not this paragraph as a win.
 
 Isolation is the **default** on every platform now, so a hand-run `ctest -j` is
 safe wherever it happens; the Windows presets used to set
--- a/docs/engine/building.md
+++ b/docs/engine/building.md
@@ -274,9 +274,9 @@ ctest -V
 ./vayu_tests
 ```
 
-The Linux and macOS **test presets** run the suite multi-process (`ctest -j8`,
-wired once into the hidden `test-base` test preset in
-`engine/CMakePresets.json`); a bare `ctest` or `./vayu_tests` runs serially.
+Every **test preset** runs the suite multi-process (`ctest -j8`, wired once into
+the hidden `test-base` test preset in `engine/CMakePresets.json`, which the
+Windows presets narrow to `-j4`); a bare `ctest` or `./vayu_tests` runs serially.
 Parallelism is safe because the test binary enters a private per-process scratch
 directory before running, so the relative `test_*.db` files fixtures open never
 collide between concurrently scheduled tests (see `engine/tests/main.cpp` and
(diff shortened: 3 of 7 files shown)


athrvk:
**CI is green on all three engine legs, and the Windows number is in - it is not a speedup.** Recording it here rather than only in the docs, because it changes what this PR is worth merging *for*.

| leg | ctest step | baseline |
|---|---|---|
| Engine windows-latest | **5m27s**, 2071 tests, 2062 passed, 9 skipped, 0 failed | 4m40s-6m38s serial (runs 32220669108 / 32224353428 / 32228952223) |
| Engine ubuntu-latest | 53s | 1m15s (#808) |
| Engine macos-latest | green | 1m06s (#808) |

5m27s sits inside the band the serial legs spanned, so a single sample cannot call it a win or a loss. That is what the 81% measurement predicted: a locked group can never beat its own serial time, and this group is most of the wall. The ubuntu leg confirms the second `gtest_discover_tests` call costs nothing (53s against a 1m15s baseline, if anything faster).

What the run *does* establish is that the mechanism works. The same `-j4` that measured ~37 min without the lock now finishes in the time serial used to take, with the database tests never overlapping and no test skipped, loosened or quarantined to get there. The pathology is gone; the speedup is not here yet.

**My recommendation: merge it, for what it unblocks rather than what it saves.** It removes "Windows must run one test at a time" from the tree, the guards keep the locked list from rotting, and #838 - making concurrent scratch-database work cheap, which is the only lever that reaches the ~2.5 min this was scoped for - builds directly on it. If you would rather see the speedup and the mechanism land together, holding this until #838's approach is decided is entirely reasonable; nothing here expires.

One lever I deliberately did **not** pull, since it is your call: this PR turns per-process scratch isolation back **on** for Windows (it was off while the presets were serial). With the lock in place the directories are not strictly needed - no two database tests overlap, so the relative `test_*.db` names cannot collide - and dropping them would remove ~2071 directory create/delete pairs from the leg. The cost is that the locked list would stop being a pure performance statement and become load-bearing for correctness: a database suite missing from it would mean two tests sharing a working directory and deleting each other's database, which is a red suite rather than a slow one. That trade seemed wrong to make silently for an unmeasured saving, so it is written down here instead.

---

athrvk:
Second Windows sample, from the docs commit re-running the same test tree: **ctest 4m50s**, 2071 tests, 0 failures (run 32252015024, whole workflow green). Beside the first run's 5m27s.

| | ctest | build |
|---|---|---|
| run 32249239500 | 5m27s | 20m59s (cold, 32% sccache) |
| run 32252015024 | **4m50s** | 1m49s (warm) |
| serial baseline (3 runs, #805) | 4m40s - 6m38s | - |

Both samples land inside the serial band - 4m50s within ten seconds of its fastest end - so the reading is unchanged and now rests on two points rather than one: the lock makes `-j4` survivable on Windows, it does not make it faster than serial was. No regression either, which is the other thing two samples buy.

Worth noting the second job as a whole: **7m20s** against the 27-30 min this leg took at filing. That is #700's warm build (1m49s) doing the work, not this PR - but it does mean ctest is now ~66% of a warm Windows engine job, so #838 is where the remaining Windows minutes are.

I've added the second number to `docs/engine/building.md` so the doc states a pair rather than a single figure that would read as the new normal. Nothing else changed; the PR is green on all legs and ready for your call.

---

Issue #856, opened 2026-08-19
engine(windows): the compatibility manifest outlived the ALPN chain it was added for - decide what still needs it


athrvk:
Deferred out of #851 rather than decided in passing, because the change #851 makes is a *backend* change and this is a *manifest* question.

## Problem

`engine/res/vayu-windows.manifest` is embedded into `vayu-engine`, `vayu-cli` and `vayu_tests` by `vayu_embed_windows_manifest()`, and it is guarded twice - `HttpVersionSupport.WindowsOsVersionIsNotShimmed` for the test binary, `.github/check-windows-deps.py` for the shipped one. Both exist for one reason, recorded in `docs/engine/building.md` and traceable to #215:

1. curl's Schannel backend enables ALPN only at OS >= 6.3 (`s_win_has_alpn`).
2. That check falls back to `VerifyVersionInfoW`, which reports 6.2 to an unmanifested process.
3. So HTTP/2 was never negotiated on Windows - silently, with a 200 and a reported `HTTP/1.1`.

**#851 makes the Windows build OpenSSL-backed, and OpenSSL's ALPN is not gated on an OS version at all.** Step 2 no longer runs. HTTP/2 on Windows no longer depends on the manifest, so the stated justification for both guards is now historical.

#851 left all of it in place and said so in the comments and in `building.md`, on the grounds that the guards are cheap and are what would make a return to Schannel survivable. That is a defensible answer but it is not a decided one, and an assertion whose reason nobody can restate is how a suite accumulates tests that pass for no reason.

## What this issue is asking

Not "delete the manifest". Establish what, if anything, still depends on it, and then either write that reason down beside the guards or remove what has no reason left:

- The `supportedOS` ids change what `GetVersionEx` / `VerifyVersionInfoW` report to the *whole process*, not just to libcurl. Anything else in the engine or its dependencies reading an OS version is a live dependent - find them or establish there are none.
- DPI awareness, long-path support and other manifest sections may be in the same file for unrelated reasons; check what the file actually declares before treating it as one thing.
- If the only remaining reason is "a return to Schannel would need it", say that in the guard's failure message and in `building.md` - a reader who hits a red `WindowsOsVersionIsNotShimmed` should learn why they should care in 2026, not read a chain that stopped applying.

## Acceptance criteria

- [ ] The manifest's live dependents are enumerated (or their absence established), in the issue and in whichever doc a maintainer would read first.
[...]

PR #857, opened 2026-08-19
fix(engine/tls): one OpenSSL backend on all three platforms, so mTLS works on Windows

### LINKED ISSUE #842: engine(tls): a client certificate cannot complete a handshake on Windows - curl 8.21's Schannel client-cert path is broken upstream, and the wire cases are dark there
Filed from #833 (PR #840), which gave the registry a certificate format so a Windows build could present a PKCS#12 identity at all - and found that naming the format is necessary but not sufficient.

## Problem (measured)

With `CURLOPT_SSLCERTTYPE` now written from the stored format, the Windows leg gets far enough to *attempt* client auth and then fails every wire case:

```
SSL_ERROR: schannel: next InitializeSecurityContext failed:
SEC_E_INTERNAL_ERROR (0x80090304) - The Local Security Authority cannot be contacted
```

Measured on runs [32250436591](https://github.com/athrvk/vayu/actions/runs/32250436591) and [32254336923](https://github.com/athrvk/vayu/actions/runs/32254336923) (PR #840), on every driver: design send, SSE stream, load run, `pm.sendRequest`, port precedence, passphrase.

**What the failure is not.** Two hypotheses were tested and are dead:

1. *The bundle's algorithms.* The first fixture minted PKCS#12 with SHA-1 + 3DES; the second used OpenSSL's defaults (AES-256-CBC under PBES2, SHA-256 MAC). **Identical failure**, same error, both runs. So this is not a legacy-PBE-into-CAPI import problem.
2. *The engine's option handling.* `AcquireCredentialsHandle` succeeds - curl reports its own failure text when it does not (`schannel.c:690`) - so Schannel accepted the credential with the client certificate attached. The failure is later, where the private key is actually *used* to sign.

**What it is.** curl 8.21.0 - the version `engine/vcpkg.json` pins - ships this in its own `docs/KNOWN_BUGS.md`:

> **Access violation sending client cert with Schannel**
> When using Schannel to do client certs, curl sets `PKCS12_NO_PERSIST_KEY` to avoid leaking the private key into the filesystem. Unfortunately that flag instead seems to trigger a crash.
> [curl issue 17626](https://github.com/curl/curl/issues/17626)

> **Client cert (MTLS) issues with Schannel**
> [curl issue 3145](https://github.com/curl/curl/issues/3145)

`schannel.c:488` is that exact call: `PFXImportCertStore(&datablob, pszPassword, PKCS12_NO_PERSIST_KEY)`. The key it yields lives only in the curl process, and Schannel's credential path needs one the LSA can reach - which is what "The Local Security Authority cannot be contacted" is saying. A second variable rides along and is *not* separated by the evidence so far: the runner is Windows Server 2025, so curl takes the `SCH_CREDENTIALS` branch with **TLS 1.3 enabled**, and Schannel's TLS 1.3 client-auth path is a separate suspect. Nothing here distinguishes them, and this issue should not pretend otherwise.

## Where that leaves Windows

`vayu` on Windows can **store** a client certificate correctly and hands libcurl the right type - what it cannot do is complete the handshake, and no change in this repository fixes that. PR #840 therefore skips the mTLS *wire* cases on a Schannel build, with this issue and the upstream ones named in the skip. Everything that is not a handshake still runs there: the registry rules, the `certFormat` validation and defaulting, the resolver plumbing, and `ClientCertificateBackend.TakesTheCertificateOptionsOnThisPlatform` (which asserts Schannel accepts `CURLOPT_SSLCERTTYPE` for both types).

## Fix pathway

Nothing to implement here yet - this is a watch item on an upstream fix, plus one experiment worth running when someone has a Windows machine:

- **Separate the two variables.** Point a Schannel build at a listener capped at TLS 1.2 with the same PKCS#12 identity. If it completes, the live cause is TLS 1.3 client auth and `PKCS12_NO_PERSIST_KEY` is a second, latent one; if it fails identically, the key-persistence path is the whole story. This is cheap on real hardware and impossible in this repo's CI without capping the fixture for every platform, which would cost the TLS 1.3 mTLS coverage the OpenSSL legs have today.
- **Re-check on each curl bump.** The baseline bump is already a step that reads curl's changelog; add "does KNOWN_BUGS still carry the Schannel client-cert entries?" to it. When they go, delete the skip and the parameterization runs the p12 half on Windows with no other change - that is why #840 left the format matrix in place rather than hard-coding Windows out.
- **A certificate-store reference is the other shape** (`CURLOPT_SSLCERT` as a store expression on Schannel) and may sidestep the file-import path entirely. It is out of scope for #833 by that issue's own text, and it is a different feature - the registry stores file paths, and a store expression is not a file. Worth its own issue if a Windows user actually needs mTLS before upstream lands a fix.

## Docs - same commit as any change here

`docs/engine/api-reference.md` states what Windows can and cannot do with a client certificate; it must not come to claim a working handshake before one is proven. `engine/CLAUDE.md` carries the same statement in its client-certificate paragraph.

## Sequencing

Standalone, after #833 (PR #840), whose skip this issue is the reason for. Blocked on upstream curl; nothing in this repository blocks it.

## Acceptance criteria

- [ ] Either a curl version reaches the pinned baseline with the Schannel client-cert path fixed and the skip is deleted (the p12 wire cases then run on Windows unchanged), or the TLS 1.2 experiment above records which of the two causes is live, so the next person does not re-run it.
- [ ] No skip in `mutual_tls_test.cpp` outlives its cause: whatever gates the Schannel leg must name this issue and be deleted with it.

### LINKED ISSUE #843: engine(tls): move the Windows build to an OpenSSL-backed curl - one backend everywhere, and the only route to working client certificates on Windows
Filed from #842, where reading what the comparable tools ship changed what the fix probably is. This is the decision issue; #842 stays as the record of the defect it would fix.

**Decision (owner, 2026-08-19): APPROVED.** Normal queue - no priority label; it lands with the transport epic's cadence, not ahead of it. When this ships, #842 closes as superseded-as-a-defect-record, with its watch-the-upstream note moving into the docs this issue rewrites (so any future return to Schannel re-reads the defect first). PR #840 (the format work) merges independently once green - its P12 support is a prerequisite of this issue's all-legs acceptance signal.

## Problem

Vayu's Windows build is Schannel-backed, and every other engine that does this job on Windows is not. That single difference costs us mutual TLS entirely: curl's Schannel client-certificate path imports the bundle with `PKCS12_NO_PERSIST_KEY` and cannot then use the key (measured on #840 as `SEC_E_INTERNAL_ERROR ... The Local Security Authority cannot be contacted` on every driver; documented in curl 8.21.0's own `KNOWN_BUGS`, curl 17626 and 3145; still open on curl master as of 2026-08-19, with no newer release to bump to - see #842).

**We did not choose Schannel; it is the vcpkg default we inherited.** `engine/vcpkg.json` asks for `curl` with `http2`, which takes the port's default `ssl` feature, and `ports/curl/portfile.cmake:50-55` sets `CURL_USE_SCHANNEL=ON` when `ssl` is present on Windows. The port's own comment states the rule: *"Defaults to schannel on Windows, unless http3 is also enabled."*

**Insomnia - the closest comparable tool, and libcurl-based like us - takes the other branch.** `JCMais/node-libcurl`'s `vcpkg.template.json` requests `openssl` and `http3` from the same port, so its Windows build is OpenSSL-backed and never enters that code path (note: it lands there *because of* the `http3` this issue rejects - see costs). Its client-certificate handling is otherwise identical to ours (`SSLCERT` + `SSLCERTTYPE` `PEM`/`P12`, `SSLKEY` on the PEM path only, one `KEYPASSWD`), which is what makes the comparison clean: same design, same library, different backend, and only one of the two can do mTLS on Windows. Bruno and Postman sit on Node's own bundled TLS and never had the problem. The field is split rather than unanimous, for honesty's sake: curl.se's own Windows binaries ship LibreSSL (OpenSSL-family), while Git for Windows compiles both backends and *defaults* to Schannel with OpenSSL selectable - so this is a real choice, not a consensus to copy.

## Fix pathway

**1. Ask the port for OpenSSL - and actually drop Schannel.** (Corrected 2026-08-19 by review: the previously written `"features": ["http2", "openssl"]` alone does NOT do it - vcpkg keeps default features on, so the port's default `ssl` feature still fires `CURL_USE_SCHANNEL=ON` at `portfile.cmake:51-56` and the result is a MultiSSL build with both backends compiled in and Schannel still selectable.) The manifest change is:

```json
{ "name": "curl", "default-features": false, "features": ["http2", "non-http", "openssl"] }
```

with `non-http` re-added explicitly because it is also a default feature this engine relies on. Verify from the built artifact rather than the manifest: `TlsBackend.IsTheBackendEveryTrustStatementHereAssumes` already reads `curl_version_info()` per leg and is the test that must change - it currently asserts Schannel on Windows, and after this it asserts **a single OpenSSL backend** on all three (a MultiSSL build reporting several backends is a failed acceptance, not a variant).

**2. Keep the OS trust store, deliberately.** This is the part that must not regress. Today Windows verifies against the OS store because Schannel does; an OpenSSL build makes `CURLOPT_CAINFO` the whole store, and `system_ca_bundle_pem()` returns empty on Windows (there is no bundle file to read). (Sharpened 2026-08-19 by review:) the population a naive switch hurts most is the user with **no** custom CA pasted - empty CAINFO falls to OpenSSL's baked-in default path, which does not exist on Windows, so that user trusts *nothing*. The pasted-CA user is already half-covered: `curl_utils.cpp:616-617` writes `CURLSSLOPT_NATIVE_CA` whenever a custom bundle is set - and explicitly writes `0L` when none is - so the concrete change is that the Windows leg must set `CURLSSLOPT_NATIVE_CA` in **both** branches, not only beside a pasted bundle.

   `lib/vtls/openssl.c:2995` honours it: OpenSSL loads the Windows certificate store. It has to be applied in `detail::apply_transport_policy` alongside the other transport options, and it interacts with the additive-CA rule from #706 - on Windows the shape becomes "native store **plus** the user's pasted PEMs", which needs to be true rather than assumed.

**3. Re-state what #818 established.** That issue found six statements across the engine and docs claiming macOS used SecTrust when it did not, and answered it by asserting the backend per leg rather than reasoning about it. This change makes a new set of those statements, so the same discipline applies: `engine/CLAUDE.md`'s transport paragraph, `docs/engine/api-reference.md`'s per-backend table, `docs/architecture.md`'s security section, and the `system_ca_bundle_pem()` header comment all currently say "Schannel keeps its OS store" about Windows.

## Tests

The bar #812 and #819 set for trust claims: on a wire, per leg, not reasoned about.

- `TlsBackend.IsTheBackendEveryTrustStatementHereAssumes` asserts a single OpenSSL backend on every leg.
- `TlsBackend.FindsTheSystemAnchorsTheMergeExtends` currently **skips on Windows** (there is no bundle file). With native-CA it must either stop skipping or skip for a newly-true reason - decide which, and do not let it stay dark by inertia.
- **A Windows leg that verifies a real public certificate through the native store**, which is the assertion that this change did not narrow anyone's trust. Nothing in the suite covers it today because Schannel made it free.
- `CustomCaVerificationTest`'s four cases must stay green on Windows - the additive rule is now doing work there that Schannel used to do.
- The mTLS suite's Schannel skip (`client_auth_defect()`, added in #840) is **deleted** by this change, and `mutual_tls_test.cpp` runs both formats on all three legs. That is the acceptance signal.

## Docs - same commit

`engine/CLAUDE.md` (transport paragraph, the two-way backend spread), `docs/engine/api-reference.md` (the client-certificate per-backend table and the Windows note added by #840), `docs/architecture.md` (client-certificate keys / trust), `docs/engine/building.md` if the dependency set visibly changes. Plus, per the decision above: the #842 watch-the-upstream note lands in the api-reference Windows section, guarding a future backend change.

## Costs, stated rather than discovered later

- **Binary size and build time.** OpenSSL is compiled and shipped rather than borrowed from the OS. The Windows engine build is already the slowest leg (#805, #700).
- **Security updates become ours.** Today a Schannel CVE is fixed by Windows Update; after this, an OpenSSL CVE needs a vcpkg baseline bump and a Vayu release. That is the real trade, and it is the reason this was a decision issue rather than a one-line PR. Decided: accepted (owner, 2026-08-19).
- **Native-CA is narrower than Schannel's verifier** (added 2026-08-19 by review, from upstream records): `CURLSSLOPT_NATIVE_CA` imports the store's **roots but not its intermediates** (curl 12155, closed upstream as not-a-bug - chains relying on store-installed intermediates the server fails to send will stop verifying); it cannot fetch missing roots on demand the way Windows' own verifier does via Windows Update; and revocation checking via the store is gone (OpenSSL checks none by default where Schannel does best-effort chain revocation). Each is a real, if narrow, trust-behavior delta for Windows users and belongs in the docs this issue rewrites.
- **`http3` is the other way to reach the same branch** (the port skips Schannel when http3 is enabled) and is *not* what this issue proposes - that would pull in a QUIC stack for a side effect.

## Sequencing

- After #833 (PR #840), which is what made the gap visible and whose skip this deletes. No conflict with it: this touches `vcpkg.json`, the applier's TLS options and the backend assertions, not the registry.
- Related to #704's epic decisions on trust; #706's additive-CA rule is the one this must not break.
- **#842 endgame (decided)**: when this ships, #842 closes as superseded-as-a-defect-record; its body remains the record, and the upstream watch note moves into the api-reference docs per above.
- Build note: the vcpkg dependency change invalidates every cached engine object and the vcpkg binary cache entry for curl - the first CI run after this lands pays a cold compile on every leg, and `cache-warm.yml`'s guard job asserts the vcpkg key agreement, so expect one full warm cycle (#700's machinery re-warms automatically on the master merge).

## Acceptance criteria

- [ ] The Windows engine reports a **single** OpenSSL backend from `curl_version_info()` (not a MultiSSL build with Schannel compiled in), asserted per leg.
- [ ] Windows verifies a public certificate through the native store, on a wire, with no custom CA pasted - and the four `CustomCaVerificationTest` cases stay green there.
- [ ] The additive rule holds on Windows: the user's pasted PEMs verify *in addition to* the native store, not instead of it - including the no-pasted-CA branch setting `CURLSSLOPT_NATIVE_CA`.
- [ ] `mutual_tls_test.cpp` runs both client-identity formats on all three legs with no skip, and `client_auth_defect()` is deleted.
- [ ] Every statement naming Schannel in the engine and docs is corrected in the same commit, and the #842 upstream watch note lands in the api-reference.
- [ ] The security trade (OpenSSL updates ride Vayu releases) and the native-CA trust deltas (intermediates, on-demand roots, revocation) are recorded where a future reader will find it, not only in this issue.
- [ ] #842 closed as superseded in the same sitting the closing PR merges.

### LINKED ISSUE #851: engine(tls): switch the Windows curl build to a single OpenSSL backend - APPROVED, ready to implement
**Nothing blocks this issue. The decision was made by the owner on 2026-08-19 (recorded in #843, now closed as decided); this issue is the implementation order.** Why-and-tradeoffs live in #843's body and comments - do not re-litigate them here. Outcome when this ships: mTLS works on Windows, and one TLS backend on all three platforms. This is the last outstanding piece of the (now closed) transport epic #704 - its final criterion transferred here.

## The change, step by step

**1. `engine/vcpkg.json` - replace the curl dependency entry with:**

```json
{ "name": "curl", "default-features": false, "features": ["http2", "non-http", "openssl"] }
```

`default-features: false` is load-bearing: without it the port's default `ssl` feature keeps `CURL_USE_SCHANNEL=ON` on Windows (`portfile.cmake:51-56`) and you get a MultiSSL build with both backends - a failed outcome, not a variant. `non-http` is re-added explicitly because it is also a default feature the engine relies on.

**2. `engine/src/http/event_loop/curl_utils.cpp` (~:616-617) - set `CURLSSLOPT_NATIVE_CA` in BOTH branches on Windows.** Today the flag is written only when a custom CA bundle is set, and explicitly written `0L` when none is. After the switch, the no-custom-CA Windows user must still get the OS store: empty CAINFO on Windows falls to an OpenSSL default path that does not exist, i.e. trusts nothing. The additive rule (#706) becomes "native store plus the user's pasted PEMs" - true in code, not assumed.

**3. Delete the Schannel skip.** `client_auth_defect()` and `cannot_present_a_pem_pair()` in `engine/tests/tls_backend.hpp`, the `GTEST_SKIP` wiring in `engine/tests/mutual_tls_test.cpp`, and the source-scan guard `MutualTlsBackend.NoBackendIsSkippedForAReasonThatWentAway` - all exist only because of Schannel. `mutual_tls_test.cpp` then runs both identity formats (PEM pair and PKCS#12) on all three legs with no skip.

**4. Update the backend assertions.** `TlsBackend.IsTheBackendEveryTrustStatementHereAssumes` (`engine/tests/transport_policy_test.cpp:~469-501`) asserts a **single** OpenSSL backend on every leg (a build reporting multiple backends fails). `TlsBackend.FindsTheSystemAnchorsTheMergeExtends` currently skips on Windows (no bundle file) - decide whether it stops skipping or skips for a newly-true reason; do not leave it dark by inertia.

**5. New Windows wire test: verify a real public certificate through the native store with no custom CA pasted.** This is the assertion that the switch did not narrow anyone's trust - nothing covers it today because Schannel made it free. `CustomCaVerificationTest`'s four cases must stay green on Windows.

**6. Docs, same commit** - every "Schannel" statement corrected: `engine/CLAUDE.md` (transport paragraph, two-way backend spread - becomes one-way), `docs/engine/api-reference.md` (client-certificate per-backend table, the #840 Windows note, and a new upstream-watch note: the Schannel client-cert defect in curl's KNOWN_BUGS 17626/3145 is why a return to Schannel must re-read closed #842 first), `docs/architecture.md` (trust section), `docs/engine/building.md` if the dependency set visibly changes. Also record the native-CA trust deltas where users will find them: roots but not store intermediates (curl 12155), no on-demand root fetch, no revocation via store. The `docs/compare/` pages' Windows-mTLS asterisk (added by PR #850) comes OUT in the same commit - the qualification stops being true the moment this ships, and a stale asterisk understates the product.

## Build/CI expectations (not blockers - just so nobody misreads the first run)

- The vcpkg dependency change invalidates the curl binary-cache entry and every sccache object key: the **first CI run pays a cold compile on every leg** (~15-22 min each). `cache-warm.yml` re-warms the master scope automatically on the merge; the second run is warm again. This is one-time, expected, and not a regression.
- `cache-warm.yml`'s guard job asserts vcpkg-key agreement across workflows - it should stay green untouched; if it reds, the manifests drifted, fix the manifest not the guard.
- OpenSSL adds roughly 5-8 MB to the shipped Windows binary. Accepted in the decision.

## Acceptance criteria

- [ ] Windows engine reports a **single** OpenSSL backend from `curl_version_info()`, asserted per leg.
- [ ] Windows verifies a public certificate through the native store on a wire with no custom CA pasted, and `CustomCaVerificationTest`'s four cases stay green there.
- [ ] The additive rule holds on Windows: pasted PEMs verify in addition to the native store - including the no-pasted-CA branch setting `CURLSSLOPT_NATIVE_CA`.
- [ ] `mutual_tls_test.cpp` runs both client-identity formats on all three legs with no skip; `client_auth_defect()` and the source-scan guard are deleted.
- [ ] Every Schannel statement in engine and docs corrected in the same commit; the upstream-watch note and the native-CA trust deltas recorded in the api-reference; the compare pages' Windows-mTLS asterisk removed.
- [ ] (Transferred from closed epic #704, its final criterion:) mTLS APIs are callable with a registered client certificate on all execution paths, on all three platforms.

## Sequencing

Supersedes #843 (closed - the decision record) as the actionable item; carries the last criterion of closed epic #704. Parallel-safe against everything open; touches `engine/vcpkg.json`, `curl_utils.cpp`'s TLS options, `tls_backend.hpp`/`mutual_tls_test.cpp`/`transport_policy_test.cpp`, and docs. Closed #842 is the defect record this change routes around.

---

athrvk:
## Description

**Root cause.** Windows built libcurl against Schannel, and curl 8.21's Schannel client-certificate path cannot complete a handshake with an identity read from a file - it imports the bundle with `PKCS12_NO_PERSIST_KEY` and the key that yields is one Schannel's credential path cannot use (curl `KNOWN_BUGS` 17626 and 3145; measured on every driver in closed #842). Nothing in Vayu's configuration reaches it, so the wire half of `mutual_tls_test.cpp` skipped that leg through `client_auth_defect()` and mTLS was unavailable to Windows users.

**Approach.** Get Schannel out of the *decision* rather than skip around it. Two halves, and the second was not in the plan:

1. `engine/vcpkg.json` pins curl's `openssl` feature with `default-features: false`, so an OpenSSL is present to use on every platform.
2. The engine **names** that backend at startup - `pin_tls_backend()` calls `curl_global_sslset(CURLSSLBACKEND_OPENSSL, ...)` before `curl_global_init`.

Step 2 exists because step 1 cannot do the job alone, which the first Windows CI run established rather than any port file: **the curl port's `http2` feature itself depends on `curl[ssl]`**, and the engine requires `http2`, so `ssl` is in the resolved feature set however the manifest is written. On Windows it resolves to Schannel and the shipped libcurl is MultiSSL. Reproduced against the pinned baseline with a manifest holding nothing but curl:

```
curl[core,http2,non-http,openssl,ssl]:x64-linux@8.21.0#1
```

That is invisible on Linux and macOS, where `ssl` resolves to `openssl` and unions into the backend already there - which is why only the Windows leg went red, against an assertion that expected one backend and got `openssl, schannel`.

**Why the pin matters beyond binary size.** On a MultiSSL build libcurl chooses, and `multissl_setup` (`lib/vtls/vtls.c`) reads **`CURL_SSL_BACKEND` from the environment** before falling back to the first compiled-in backend. A corporate login script or a leftover export naming `schannel` would move every transfer onto the backend this work exists to get off - mTLS silently broken again, trust model changed, every document here still describing OpenSSL. Naming the backend closes that structurally, because `multissl_setup` consults the environment only when no caller has chosen.

**The alternative rejected**, and it is the obvious one: wait for upstream. The curl defect has two `KNOWN_BUGS` entries and no fix in the pinned release, so "wait" is indefinite, and the cost is a documented platform gap plus a per-leg skip that hides every other Windows client-certificate regression behind it. The decision to switch was the owner's, recorded in closed #843; this PR is the implementation.

**The consequence that needed care.** An OpenSSL build on Windows has no CA file to fall back to - `curl_version_info()->cainfo` names a path from the machine the port was built on. So `CURLSSLOPT_NATIVE_CA` is now set **unconditionally** on Windows rather than only when a bundle is in force. Without that, a Windows user who has pasted nothing trusts *nothing at all* - and all four `CustomCaVerificationTest` cases would have stayed green while it happened, because each supplies its own anchor. That gap is what `NativeStoreVerificationTest` closes.

### Deviations from the issue spec

**Acceptance criterion 1 is partly met and partly deferred.** It asks that Windows report a *single* OpenSSL backend. What this PR delivers is *verifies with OpenSSL on all three platforms, selected explicitly and not inheritable from the environment*. Getting Schannel out of the binary needs a custom triplet (`-DCURL_USE_SCHANNEL=OFF`) or a vendored overlay port; both add build plumbing across `build.py`, the Windows presets and every CI leg, and the `Cache key agreement` guard means the vcpkg key has to move in step everywhere at once. That is a decision with real cost, filed as **#858** with the evidence and three options rather than made inside this PR. Say the word and I will fold the triplet in here.

Two items the issue names do not exist in the tree, so they could not be deleted; nothing replaces them and no skip was left behind:

- **`cannot_present_a_pem_pair()`** - not present. `client_identity_formats()` carries the format matrix; unchanged, still stub-asserted.
- **`MutualTlsBackend.NoBackendIsSkippedForAReasonThatWentAway`** - no such test exists. The skip it would have guarded is gone with `client_auth_defect()`.
- **The "#840 Windows note" in `api-reference.md`** - no `#840` reference exists in the docs tree. The Windows rows and the mTLS paragraph were corrected on their own terms.

One judgement call the issue asked for explicitly: **`TlsBackend.FindsTheSystemAnchorsTheMergeExtends`** kept its Windows conclusion but not its reason. The old skip said "this backend verifies through an OS store, which is not a file to merge" - false now. There is still nothing to merge, for a newly true reason (Windows ships no PEM bundle), so the Windows branch **asserts** what actually supplies the anchors there - that the build accepts `CURLSSLOPT_NATIVE_CA` - rather than skipping, and asserts the merge where a machine exporting `CURL_CA_BUNDLE` does put a file in reach.

Schannel classifier rows are deliberately **kept** (`verifies_through_a_bundle_file`, `client_identity_formats`): they are what makes a MultiSSL or manifest-edited build fail *legibly* instead of being guessed at, and both are stub-asserted on every leg.

### Scope left open

- **#858** - getting the Windows build down to one compiled-in backend (triplet vs overlay port vs accept-and-pin).
- **#856** - `docs/engine/building.md`'s Windows-manifest chain is now historical: OpenSSL's ALPN is not gated on an OS version, so HTTP/2 on Windows no longer depends on `vayu-windows.manifest`. The manifest and both its guards are kept and re-documented rather than deleted in passing.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

**CI is green on all three platforms**, including the Windows leg that found the MultiSSL problem. That leg now reports **2151 tests run, 2150 passed, 0 skipped** on the first fixed head - the 0 skipped is the result this issue was opened for: the mTLS wire cases run on Windows, both identity formats, every driver.

Locally on `linux-dev`, on the current head (after merging master):

- `ctest --preset linux-dev --output-on-failure` - **2150 tests, 2150 passed, 0 failed**, 40.8s.
- `pnpm test` - **525 files, 5812 tests, all passed**.
- `pnpm type-check` - clean (renderer, electron, electron-tests).
- `eslint` + `prettier --check` on the two touched app files - clean.
- Backend read off the built binary rather than reasoned from the port file: `curl_version_info()` reports `ssl=OpenSSL/3.6.3`.

**Mutation check on the two new wire tests.** Temporarily pointed the applier's `CURLOPT_CAINFO`/`CURLOPT_CAPATH` at nonexistent paths, rebuilt, and confirmed both go red with their own messages - `a public certificate did not verify with an empty customCaCertificates...` and `pasting one CA cost this machine every anchor it already had...` - then reverted and rebuilt. They are not vacuous.

**Where I want to be explicit rather than reassuring:**

- The backend assertion now checks that this process **selected** its backend, not that only one was compiled in. That is not the count check weakened to get green - the hazard the count guarded against ("whichever libcurl picked") is removed by the selection. But it is **only load-bearing on a MultiSSL build**, so Linux and macOS satisfy it trivially and the Windows leg is what exercises it.
- On Linux, libcurl's default `capath=/etc/ssl/certs` is passed alongside `CURLOPT_CAINFO`, so the anchors survive there even if the merge produced nothing. The merge is belt-and-braces on that leg; on Windows there is no second path, which is why `CURLSSLOPT_NATIVE_CA` is unconditional.
- `NativeStoreVerificationTest` is the one test here that leaves loopback (it needs a certificate the *platform* vouches for, which no fixture can mint). It skips on a connect/DNS/timeout failure and **fails on an SslError**, so a closed network cannot red it but a narrowed trust store cannot hide in it either.

## Checklist

## Related Issues

Closes #851

Carries the final criterion of closed epic #704 (mTLS callable with a registered client certificate on all execution paths, on all three platforms). Supersedes closed #843 as the actionable item. Closed #842 is the defect record this routes around, named in the code and docs as the thing to re-read before any return to Schannel. Follow-ups filed: **#858** (single-backend build), **#856** (whether the Windows compatibility manifest still earns its place).


athrvk:
**The first Windows run found that step 1 of #851 is not sufficient, and the port file said otherwise.** Recording it here because it changes what this PR claims.

Windows came back `2151 tests run, 2150 passed, **0 skipped**, 1 failed`. The 0 skipped is the goal met - the mTLS wire cases ran on that leg for the first time and passed, both formats, every driver. The one failure was the new backend assertion:

```
transport_policy_test.cpp(583): error: Expected equality of these values:
  compiled_in.size ()   Which is: 2
  1u
this is a MultiSSL build carrying 2 backends (openssl, schannel)
```

### Why `default-features: false` was not enough

The `curl` port's `http2` feature depends on `curl[ssl]`:

```jsonc
"http2": { "dependencies": [ { "name": "curl", "default-features": false, "features": ["ssl"] }, "nghttp2" ] }
```

The engine requires `http2`, so `ssl` enters the resolved set however the top-level manifest is written. Reproduced against the pinned baseline with a manifest holding nothing but curl:

```
curl[core,http2,non-http,openssl,ssl]:x64-linux@8.21.0#1
```

On Windows `ssl` resolves to Schannel and the portfile sets `-DCURL_USE_SCHANNEL=ON`. It is invisible on Linux and macOS, where `ssl` resolves to `openssl` and unions into the backend already there - which is exactly why only the Windows leg went red.

### Why it is not cosmetic

On a MultiSSL build libcurl chooses, and `multissl_setup` (`lib/vtls/vtls.c`) reads **`CURL_SSL_BACKEND` from the environment** before falling back to the first compiled-in backend. A corporate login script or a leftover export naming `schannel` would move every transfer onto the backend this work exists to get off - mTLS silently broken again (#842), trust model changed, every document here still saying OpenSSL.

### What the new commit does

`pin_tls_backend()` calls `curl_global_sslset(CURLSSLBACKEND_OPENSSL, ...)` before `curl_global_init` (the ordering is load-bearing - curl answers `CURLSSLSET_TOO_LATE` once initialized). `multissl_setup` consults the environment only when no caller has chosen, so naming the backend closes that hole structurally. A failed selection is logged naming the backend actually in force.

The assertion now follows the guarantee: it checks that this process **selected** its backend rather than that only one was compiled in, and prints the compiled-in list when it fails. I want to be explicit that this is not the count check quietly weakened to get green - the hazard the count guarded against ("whichever libcurl picked") is removed by the selection, not hidden. But equally explicit: **that assertion is only load-bearing on a MultiSSL build**, so Linux and macOS satisfy it trivially and the Windows leg is what actually exercises it.

### What I deliberately did not do

Getting Schannel out of the binary needs a custom triplet (`VCPKG_CMAKE_CONFIGURE_OPTIONS` with `-DCURL_USE_SCHANNEL=OFF`) or a vendored overlay port. Both add build plumbing across `build.py`, the Windows presets and every CI leg, and the `Cache key agreement` guard means the vcpkg key has to move in step everywhere at once. That is a decision with real cost and I did not want to make it inside this PR - filed as **#858** with the evidence and the three options (triplet, overlay port, or accept the MultiSSL build and treat the runtime pin as the answer).

Acceptance criterion 1 of #851 said "reports a **single** OpenSSL backend". As it stands this PR delivers *verifies with OpenSSL on all three platforms, selected explicitly and not inheritable from the environment*, with the single-backend build tracked separately. Say the word if you would rather #857 carry the triplet and I will fold it in.

Local re-verification after the fix: engine suite `2126/2126` on `linux-dev`, `mkdocs build --strict` clean.

---

Issue #855, opened 2026-08-19
engine(openapi): move export assembly into the engine - #761 phase B, move 3


athrvk:
Filed from #761's phase B while moving the matcher (PR #852), which was move 1 of three.

## Problem (verified at `2216252`)

Writing a collection back out as OpenAPI is renderer-only: `app/src/services/exporters/openapi.ts` and its four helpers (`document.ts`, `skeleton.ts`, `payload.ts`, `examples.ts`) are ~900 lines that an agent cannot reach. So `export_spec` is not an MCP tool and cannot become one, which is the last of the four OpenAPI verbs #761 listed.

## Fix pathway

Move the assembly beside the stored document, where the two things it needs already live - `spec_documents` holds the bytes, and `requests.spec_operation` says which operation each request is.

The two directions it has must both survive the move, and they are not variants of each other:

- **A bound export patches the user's own document.** Operations the collection no longer has are removed, stored examples are written in, and everything Vayu does not model is left exactly as written. That is why it parses and patches rather than regenerating - a regenerated document would silently drop every `securityScheme`, vendor extension and description the app has no model for.
- **An unbound export writes a skeleton**, which is "a starting point, not a contract" and says so in the UI. It invents nothing: no schema Vayu did not read off an example body, no server variable it never saw.

`{{variable}}` tokens are written as they stand, in `servers` and paths alike - resolving `{{baseUrl}}` would export one machine's environment as though the contract named it.

`skeleton.ts` is also the last renderer reader of `splitRequestUrl` (`operation-match.ts`), so when this lands the cross-language shape fixture `engine/tests/fixtures/operation-shape-conformance.json` retires with it - there is only one implementation left.

## Blocked by #853

Both directions need to *write* a document and the bound one needs to read one, in the format the user asked for (JSON or YAML). That is the dependency decision in #853; do not start this before it is answered.

## Tests

- `exporters/openapi.test.ts` (613 lines) is the behaviour to preserve; port its cases.
- The round trip is the one worth stating as its own case: import a document, export it, and everything the app does not model is byte-identical.

## Docs - same commit

`docs/engine/api-reference.md`, `engine/CLAUDE.md`, `docs/engine/mcp.md` (the tool, if it ships here), `docs/app/openapi.md` and `docs/app/COMPONENTS.md` for where the dialog's work now happens.

## Sequencing

- **Blocked by #853** (the document parse).
- Independent of #854 (the diff move); either order works, and whichever lands second retires the shape fixture.

[...]

Issue #854, opened 2026-08-19
engine(openapi): move the sync diff into the engine - #761 phase B, move 2


athrvk:
Filed from #761's phase B while moving the matcher (PR #852), which was move 1 of three.

## Problem (verified at `2216252`)

`POST /specs/sync` already applies a diff atomically, but it **decides nothing**: the renderer works out what moved (`app/src/services/openapi/spec-diff.ts`, 442 lines, plus `spec-apply.ts`) and the engine applies exactly what it is told. So an agent cannot ask "has this spec drifted?" or apply a selected subset - the whole Sync section is UI-only, which is the gap #761 named.

The diff is also where #717's blind spots were fixed, and every one of those fixes lives in the renderer copy - a second implementation elsewhere would start by re-learning them.

## Fix pathway

Move the comparison beside `spec_sync.cpp`, which already owns the apply half, and give it a route that answers *what would change* without changing anything - the same read-then-commit split `POST /specs/match` uses (PR #852). The shape rules it needs (`spec_path_shape`, `operation_shape_key`) are already engine-side, in `core/operation_match.hpp`, and the diff follows an operation by `operationId` first and by path shape second - the same precedence the matcher and `OperationIndex` state.

What must not be lost in the move, each of it a rule the renderer copy learned the hard way:

- **The user-touched flag is three-way** (`spec-diff.ts`): a field is flagged when what the request holds is neither what the new document produces nor what the *bound* one did. A two-way comparison cannot tell a user's edit apart from a spec change and silently reverts one.
- **An `operationId` is only followed while it means one thing** (#715): a duplicated id identifies nothing and falls back to the path.
- **`method` is a compared, tickable field** (#717), not something written unconditionally.
- **Response examples are deliberately not compared** - the `origin="import"` / `origin="user"` rule only means anything at apply time.

## Blocked by #853

The diff's inputs are `SpecRequestDraft`s - the requests *an import would build* from the document - so it cannot move until the engine can read a document. That is the dependency decision in #853, and this issue should not be started before it is answered.

## Tests

- The renderer's `spec-diff.test.ts` (548 lines) is the behaviour to preserve; port its cases, do not re-derive them.
- Mutation-check the three-way flag specifically: a two-way comparison must redden a case, or the flag is untested.
- The Sync section keeps working end to end - `SpecSync.test.tsx` against the new read.

## Docs - same commit

[...]

Issue #853, opened 2026-08-19
engine(openapi): read the document engine-side - the dependency decision the rest of #761's phase B waits on


athrvk:
Filed from #761's phase B while moving the matcher (PR #852). This is a **decision issue**: the remaining two moves and every MCP write tool this wave was for all wait on the same missing capability, and it is a dependency call rather than a port.

## Problem (verified at `2216252`, plus PR #852)

Phase B's decision is engine ownership of match, diff and export assembly. The matcher moved cleanly because it needs **no document**: it works off the operation identities the caller already extracts and stores as the `operations` index (`spec_documents.operations`), so `POST /specs/match` reads no OpenAPI at all.

Nothing else in phase B has that property.

- **The sync diff** (`app/src/services/openapi/spec-diff.ts`) compares the collection's requests against `SpecRequestDraft`s - the *requests an import would build* - which come from `spec-operations.ts` running the import parsers over the document text (`app/src/services/importers/`). Its whole design rests on that: "the drafts come from the same parsers the importer runs, so the collection and the diff can only disagree if the document did."
- **Export assembly** (`app/src/services/exporters/openapi.ts`) parses the bound document with `js-yaml` and patches it, precisely so that everything Vayu does not model survives the round trip.
- **`bind_spec` over MCP** needs the `operations` (#629) and `responseSchemas` (#628) indexes, which only the parser produces - this is the reason #841 shipped `unbind_spec` and deliberately no `bind_spec`.
- **`import` over MCP** needs the raw-document parse by definition; `POST /import/apply` takes pre-parsed entities and there is no raw-document import route.

So the engine currently **parses no OpenAPI, on purpose** - a rule stated in `engine/include/vayu/core/spec_coverage.hpp` and `schema_validation.hpp` and in `engine/CLAUDE.md`. It is not incidental: it is what keeps one opinion about what a document declares. Moving the parse engine-side does not weaken that rule, it *relocates* it - but only if the app then stops parsing too, rather than the two disagreeing about a `$ref`-ed path item.

## The decision

**A C++ OpenAPI reader needs a YAML dependency**, since a stored document is YAML as often as JSON. That is a new vcpkg port, a new attack surface, a new thing to keep at a baseline, and a source of a second opinion about a document. Options, in the repo's own terms:

[...]

Issue #761, opened 2026-08-17
app/engine(mcp): OpenAPI and import over MCP - phase A reads now; phase B needs the engine-ownership decision for match/diff/assembly


athrvk:
Part of epic #753 (S7) - the long pole, deliberately last. Unlike every other sub, full parity here is gated on an architecture decision, and pretending otherwise would produce a bad PR.

## Problem (verified at 46b115d)

The whole OpenAPI stack (epic #625, shipped 0.18.0) and the importers are UI-only over MCP: an agent cannot bind a collection to a spec, check for spec drift, sync, export, or import a Postman/Insomnia/OpenAPI document. MCP only reads verdicts embedded in run reports and smoke results.

The hard constraint, verified: **the intelligence lives in the renderer by design.** Operation matching (`app/src/services/openapi/operation-match.ts`), the sync diff (`spec-diff.ts` / `spec-apply.ts`), importer parsing (`app/src/services/importers/`), and export assembly (`app/src/services/exporters/openapi.ts`) are all renderer modules; the engine stores documents (`POST /specs`, `engine/src/http/routes/specs.cpp:420` - storage only, binding is the `openapi.{specId, specHash}` field on `PUT /collections/:id`, validated at `engine/src/http/routes/collections.cpp:176-201`), applies pre-computed sync batches atomically (`POST /specs/sync`, `engine/src/http/routes/spec_sync.cpp:692`), and applies pre-parsed import entities (`POST /import/apply`, `engine/src/http/routes/import.cpp:791` - there is no raw-document import route). Production code under `app/electron/` cannot import `app/src` (build boundary stated in `app/tsconfig.node.json` and `tools.ts:64-68`); the modules are content-portable (pure TS, I/O injected), but the repo's recorded preference when MCP needed renderer logic before was **engine ownership, not TS duplication**: #226 deleted MCP's ported composition pipeline in favor of `POST /compose`.

## Phase A - shippable now, no decision needed

1. `get_spec` (read): `specId` or `collectionId` (resolve the binding from the collection row - there is no `GET /specs` list route, ids come from bindings); returns source URL, hash, fetchedAt, and optionally the document content (size-disclosed, off by default - a 12MB spec in a tool result is the #712 lesson).
[...]

PR #859, opened 2026-08-19
feat(engine): read the OpenAPI document engine-side, and own the operation index

athrvk:
## Description

#853 asked for a decision and then a reader. The decision is recorded (option 1 - add a YAML parser and write the engine-side reader), and this is the reader, chosen the way the issue's binding disciplines say to choose it: **by measurement**. The first thing that moves onto it is the `operations` index, which the renderer stops extracting.

**Plan (root cause, approach, rejected alternative).** The root cause #853 names is that the engine deliberately parses no OpenAPI, so nothing outside the Spec tab can do anything with a stored document - the sync diff, export assembly, MCP `bind_spec` and raw-document import all wait on the same missing capability. The approach is the one the owner decided and the one #761's phase B set out: a YAML dependency and one reader in the engine, with the app then *stopping* rather than both sides reading. I took the **operation index** as the first thing to move onto it, because it is the piece with a real reader today (coverage), a bounded rule set (the import parsers' own), and a shape the engine already defines and validates - so the reader lands with a consumer rather than as an unused capability, which is this repo's most-repeated defect. The alternative I rejected was landing `core/openapi_document.*` with a read-only route and no consumer, deferring every use to #854/#855: it would have shipped a parser nothing reads, and left `spec_documents.operations` a caller's unverified claim for another two PRs. I also rejected converting YAML to JSON at the app boundary (option 2 in the issue) for the reason the issue itself gives - the stored bytes are what the hash and the export are over.

---

## The dependency, measured

Everything below was run in this container (4-core Linux, Release, vcpkg from the pinned baseline) except the Windows and macOS build times, which come from this PR's own first CI run. The corpus is the repo's own fixtures - `app/src/services/importers/__fixtures__/openapi-v3.json` and `swagger-v2.json`, each also as YAML (dumped by the app's own js-yaml, so they are documents Vayu genuinely imports) - plus a hand-written realistic document with anchors, an aliased response, a `$ref`'d path item, block scalars, comments, unquoted status keys and a `trace` operation. **The reference answer is js-yaml 4.3.1's**, the reader the importer already uses, canonicalised (keys sorted, `Date`→ISO, non-finite→null) and compared to each candidate's DOM.

| | **ryml 0.16.0** (+ c4core) | yaml-cpp 0.9.0 | fkYAML 0.4.3 |
|---|---|---|---|
| Matches js-yaml on the corpus | **5 / 5** | 3 / 5 | 4 / 5 |
| Mapping order | **document order** | document order | **sorted** (`std::map`) |
| Parse, realworld.yaml | **11.7 µs** | 314.1 µs (27x) | 50.3 µs (4.3x) |
| Parse, openapi-v3.yaml | **11.9 µs** | 265.9 µs | 57.2 µs |
| Parse, swagger-v2.json | **8.3 µs** | 183.1 µs | 40.9 µs |
| Static lib | 810 KB + 52 KB c4core | 1133 KB | header-only |
| Probe binary over a nlohmann-only baseline (178,000 B) | +267 KB | +266 KB | **+110 KB** |
| Cold port build, Linux | 23.1 s (16 + 7.1 c4core) | 16 s | **2.5 s** |
| License | MIT | MIT | MIT |

**The two disqualifying findings, in full:**

- **yaml-cpp discards quoting.** On `swagger-v2.json` it reports `"swagger": 2.0` where the document (and js-yaml) says the string `"2.0"`. Its loaded `Node` exposes no quoting information at all, so this is not fixable at the call site - and `swagger: "2.0"` is the exact key Swagger 2.0 detection turns on. Any string-valued field that looks numeric (`version: "1.0"`) has the same problem.
- **fkYAML reorders and mis-resolved an alias.** Its mapping type is `std::map`, so `paths` came back `/health | /pets | /pets/{petId}` where the document writes `/pets/{petId} | /pets | /health` - and `DeclaredOperation.responses` is specified as document order. On the realistic document it also resolved `404: *notFound` to the string `"description"` instead of the anchored mapping.

**Findings about the winner, which shaped the implementation:**

- **rapidyaml's default error handler calls `abort()`.** A document with an unterminated flow sequence killed the probe with `SIGABRT` (rc 134). The engine installs throwing callbacks (`set_error_basic/parse/visit`) once behind a `call_once`, since `set_callbacks` is documented as not thread-safe. Mutation-checked in-tree: remove the install and `AnswersAMalformedDocumentInsteadOfDying` does not fail - it takes `vayu_tests` down with 134. (CI has now confirmed the throw unwinds through rapidyaml on MSVC and AppleClang too, not only GCC.)
- **`Tree::resolve()` expands aliases with no budget** and died of `std::bad_alloc` on a 7-level "billion laughs" document under a 2 GB cap. So the reader expands aliases itself, during the walk, under a budget of one node per input byte plus a floor. Mutation-checked: with the budget removed the same 300-byte input reads for **7.9 seconds** and gigabytes before answering *success*. (js-yaml survives it in 5 ms because it aliases by reference rather than expanding - it just never has to hold the expansion.)
- **No candidate implements merge keys (`<<`)**, which js-yaml does and real documents use, so the reader implements them - with js-yaml's precedence, verified against node: an explicit key always beats a merged one (before *or* after the `<<`), and among `<<: [*a, *b]` the earlier source wins.
- **Round-trip** (the issue's third measured ground, and what #855's export will need): ryml parses `realworld.yaml`, emits it, and re-parses to the identical 65-node tree. It drops comments and expands anchors on emit - **exactly what the app's `js-yaml`-based exporter does today**, so this is parity, not a regression.
- **Cost at the write**: a 2.9 MB YAML document declaring 4000 operations parses in **33 ms**. Reading a document is paid once per store.
- **Cold CI cost of adding the dependency** (this PR's first run, which is the cold one - the `vcpkg.json` change invalidates the binary cache and every sccache key): the whole engine job took **21m 12s on windows-latest** and **10m 48s on macos-latest**. One-time; `cache-warm.yml` re-warms the master scope on merge, and the `Cache key agreement` guard stayed green untouched.
- **Build note for this environment only**: `c4core`'s portfile fetches three sub-archives with `vcpkg_download_distfile`, which `vcpkg-fix-port` (it only rewrites `vcpkg_from_github`) leaves alone, so they 403 here. Rewriting those three to `vcpkg_from_git` in the overlay is the fix; CI reaches the archives normally. Written into `engine/CLAUDE.md` next to the existing 403 note.

---

## What landed

- **`engine/include/vayu/core/openapi_document.hpp` + `src/core/openapi_document.cpp`** - the one translation unit in the engine that includes a YAML library. `read_document` (JSON first, YAML second - `parse-raw.ts`'s order) hands back a `nlohmann::ordered_json`; `declared_operations_of` answers an OpenAPI question, not a YAML one. Bounds: depth 100, one node per input byte (+4096, capped at 4M), both `400`s with a sentence.
- **The `operations` index is derived, and refused in the body.** All three writers - `POST /specs`, `POST /specs/sync`, `POST /import/apply` - go through `read_spec_indexes` → `core::derive_operations_index`, so it is a fact about the stored bytes the way `hash` is. Sending one is a `400`; a `content` that cannot be read is a `400` naming the line, rather than a row coverage, a sync and an export can each do nothing with.
- **The renderer stops extracting it.** `openapi-v3.ts` / `openapi-v2.ts` no longer build a `DeclaredOperation[]`, `SpecDraft.operations` is gone, and `SpecTab`, `SpecSync`, `spec-apply.ts`, `queries/specs.ts` and the orchestrator stop sending it. `declaredResponsesOf` is deleted - nothing read it any more.
- **One rule, two readers, pinned.** The engine indexes what a document declares; the importer stamps each request's `specOperation`, and coverage matches one against the other (`OperationIndex::resolve` - `operationId` first, `METHOD path` second). A disagreement fails nothing loudly, it credits the wrong operation - so `engine/tests/fixtures/declared-operations-conformance.json` is one table read by `openapi_document_test.cpp` (gtest) and `declared-operations.conformance.test.ts` (vitest), the compose/variable-resolution precedent.

**A behaviour change worth naming:** a coverage block now lists an operation's declared responses in *document* order. The renderer could not keep that promise - a JavaScript object orders integer-like keys numerically ahead of the rest, so a document writing `responses: {404, 200}` reached the store as `200, 404`. The header comment claimed document order the whole time.

**One bug found by re-reading the diff rather than by a test** (second commit): an explicit key written *after* a `<<` was refused as a duplicate instead of overriding the merged default - `{<<: *base, a: 5}` is `{a: 5}` in js-yaml. The shipped case only covered the explicit-key-first order, which the original code happened to get right. A mapping now remembers which keys came from a merge, so an explicit key overrides one of those in place while a second *explicit* spelling of the same key is still the error js-yaml refuses.

**Deliberate deviations from the issue.** The issue's third acceptance criterion ("exactly one parser has an opinion about what a document declares") is satisfied *for the operation index* and not yet for the whole document: the renderer still parses to build requests and to extract `responseSchemas` (#628), whose extraction means translating 3.0's dialect - `nullable`, draft-04 `exclusiveMinimum` - into JSON Schema. The issue itself scopes that criterion to "when the move is done", i.e. after #854 and #855; the schema half is now filed as #860. One further divergence from js-yaml is deliberate and documented in the header: a date-shaped plain scalar (`2020-01-01`) stays a **string** here, where js-yaml's *default* schema resolves it to a `Date`. JSON has no date type and the document wrote text; it affects no field the index reads, and #854/#855 should confirm it affects nothing they read either.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [x] Breaking change - `POST /specs`, `POST /specs/sync` and `POST /import/apply` now **refuse** an `operations` field (400) and refuse a `content` they cannot read. Both clients in this repo are updated in the same commit; a third-party client sending the index would need to stop.
- [x] Documentation update

## Testing

All green on this branch; nothing skipped, and no pre-existing failure reproduced on the base commit.

- `ctest --preset linux-dev --output-on-failure` - **2176 passed, 0 failed** (2148 on master; +28: 16 reader cases, 7 index cases, 4 stored-index cases, and one new route case for an unreadable document).
- `cd app && pnpm test` - **5833 passed, 526 files** (5827/525 on master; +6 net: 11 conformance cases in, the five renderer cases that asserted the app-built index out).
- `pnpm type-check`, `pnpm lint`, `prettier --check` on every touched TS file - clean.

**Mutation-checked, not assumed:**
- Removing the throwing rapidyaml callbacks: `AnswersAMalformedDocumentInsteadOfDying` does not fail, it aborts `vayu_tests` (exit 134).
- Removing the node budget: `RefusesAnAliasBombWithinItsBudget` fails after **7.9 s** on a 300-byte input, where the guarded read answers in 4 ms.
- The conformance fixture is asserted non-empty on both sides, because a fixture that failed to load reads as a suite of passing tests.

## Checklist
- [x] I have added tests

Docs in the same commit: `docs/engine/api-reference.md` (`POST /specs` - the derived index, what the reader does with a document, the new 400s; `/specs/sync`; `/specs/match`'s reason for still taking operations), `docs/engine/db-schema.md`, `docs/engine/building.md` (the dependency, with the measured reason it was chosen), `docs/engine/mcp.md` (what still keeps `bind_spec` out, now that one of its two blockers is gone), `engine/CLAUDE.md` (the reader, the derivation, the c4core overlay note), `docs/app/openapi.md`, `docs/app/api-integration.md`, `docs/app/import-collections/openapi-v3.md`. The "the engine does not parse OpenAPI" statements in `spec_coverage.hpp`, `schema_validation.hpp` and `types.hpp` are corrected in place - the rule was always "one reader", and this moved which side it lives on.

## Related Issues

Refs #853 - **not `Closes`**: its first acceptance criterion (the decision) is recorded, the second (a measured dependency choice) is above and on the issue, and the third is scoped to when the move is done. Refs #761. Unblocks #854 and #855, which can now read a document engine-side; the remaining half of what MCP `bind_spec` waits on is filed as #860.


athrvk:
Rebased onto master after #857 landed - it touched `engine/vcpkg.json` too, so the two changes met there. Merged rather than rebased, and the only conflict was one line of `docs/engine/building.md`: the `vcpkg install ...` example, which now names both #857's `curl[core,http2,non-http,openssl]` and this PR's `ryml`. The merged tree was re-verified locally before pushing - engine **2178 passed, 0 failed**, app **5833 passed**, type-check, lint and `mkdocs build --strict` clean.

One job was re-run, and it is worth saying which and why: `Engine on ubuntu-latest` sat in **Install system dependencies (Linux)** for 42 minutes without reaching a build or a test - an apt hang on the runner, before any test body. The same commit had already built and run the full C++ suite green on windows-latest and macos-latest, so this was not the diff. The re-run passed in 9m11s and the gate is green.

---

Issue #715, opened 2026-08-16
app(openapi): duplicate operationIds pair a request with the wrong operation - the diff's id-first lookup short-circuits past an exact path match, and a default apply rewrites the request's identity


athrvk:
From the adversarial review at `ac27b97`; found independently by two dimensions and CONFIRMED by two independent refuters.

## Problem

A document declaring the same `operationId` on two operations is invalid OpenAPI but **common in generated real-world specs**. Vayu's handling turns that upstream sloppiness into local data corruption:

1. **Import stamps duplicates verbatim** - `specOperationOf` (`openapi-shared.ts:59-67`, called at `openapi-v3.ts:119,251` and the v2 twin) copies the id as written, with no dedupe and no skip tally, so two requests carry the same `operationId` with different method+path stamps.
2. **The diff index is first-declaration-wins** (`spec-diff.ts:209-219`), and `lookup` returns the `byOperationId` hit **unconditionally** (`:225-227`) - it never cross-checks the entry's method/path, even though `byPath` holds an exact match for the request's own key one line below (`:228-230`).
3. So the request bound to the *second* declaration resolves to the *first* declaration's draft, is reported "changed, renamed, followed by its operationId" toward an operation it never was (`:163-186`), while its true operation falls out as "added" (`:189-191`).
4. `renamed=true` bypasses the empty-fields skip in `defaultSelection` (`spec-apply.ts:92-93`), so a **default-ticked** apply rewrites the request's method, url and `specOperation` to the wrong operation - two requests now claim one operation, and the collection gains a duplicate of the operation that lost its request.

Failure scenario: import a generated spec with `operationId: "list"` on both `GET /a` and `POST /b`; upstream tweaks one description; user clicks Check for changes, then Apply with defaults. Request B's method flips to GET and its identity is rewritten to `/a` - silent corruption from an all-defaults flow.

## Fix pathway

- **Lookup**: when the `byOperationId` hit's shape (method + path shape, via the existing `operationShapeKey` flattening) disagrees with the request's own stamp AND `byPath` holds an exact match, prefer the exact match; follow the id only when it is unambiguous. The diff's own no-guessing principle (`spec-diff.ts:28-30`) argues for treating a duplicated id as *no id* for matching purposes.
[...]

PR #866, opened 2026-08-20
feat(engine): build the request drafts engine-side, off the same walk the indexes come from

### LINKED ISSUE #865: engine(openapi): build the request drafts engine-side - the half of #854 that has to land before the diff can move
Filed from #862's run (PR #864), which picked #862 over #854 and owes this observation an issue rather than a comment.

## Problem (verified at `4328a90`)

#854 asks for the sync diff to move engine-side, and it is filed as one move. It is two, and the first is much larger than the second.

`spec-diff.ts` is 444 lines and compares a collection's requests against `SpecRequestDraft`s. A draft is not an identity - it is **the request an import would build**, with its URL, params, headers, body and the folder it would land in - and it comes from `spec-operations.ts` running the *import parsers* over the document text:

- `app/src/services/importers/openapi-v3.ts` (419 lines)
- `app/src/services/importers/openapi-v2.ts` (335 lines)
- `app/src/services/importers/openapi-shared.ts` (474 lines)
- `app/src/services/importers/schema-sampler.ts` (174 lines) - the body a draft carries
- `app/src/services/importers/ref-bundler.ts` (448 lines)

So "move the diff" means moving ~1,850 lines of parser *first*, and only then the 444 lines the issue is named for. #854's third acceptance criterion - "the renderer's copy is deleted rather than left beside the engine's" - cannot be honoured until all of it is there, which makes the issue as filed one nobody can finish in one PR without leaving a stump.

The dependency it names is satisfied: #853 landed `core/openapi_document.hpp`, so the engine can *read* a document. What it cannot yet do is say what requests a document would produce.

## Fix pathway

Split #854 in two, this being the first:

1. **This issue.** `core::spec_request_drafts_of(document)` beside `declared_operations_of` and `response_schemas_of`, over the same `read_document` DOM - the operation's method and templated path, its parameters split into query and header entries, the request body sampled from its schema, and the folder an import would file it under (first tag, else the path's first segment, else the root). Plus a route to ask for it, or none at all if the only caller is the diff.
2. **#854, narrowed** to the comparison itself, which then becomes the ordinary port its own "what must not be lost" list describes.

The identity half is already pinned across the two languages by `engine/tests/fixtures/declared-operations-conformance.json`; the draft half needs the same treatment, because a draft the engine builds and one the importer builds disagreeing is the *whole* failure mode - the diff would report a field as changed when only the two readers differ.

## What must not be lost

Each of these is a rule the renderer copy learned the hard way, and each lives in the parsers rather than in the diff:

- **`$ref` bundling** (`ref-bundler.ts`), including the remote and recursive cases it bounds.
- **Body sampling** (`schema-sampler.ts`) - the sample a draft carries is what the diff compares, so a different sampler is a different draft for an unchanged document.
- **Folder assignment** (#710, #655): first tag, else the path's folder, else the root. `SpecRequestDraft.folder` exists so an *added* operation lands where an import would have put it.
- **2.0 and 3.x**, told apart the way the parsers detect them - the `swagger: "2.0"` string that `openapi_document.hpp` already types correctly, and which yaml-cpp got wrong (#853's measurements).

## Sequencing

- **After #853** (merged, PR #859) - it is the reader this builds on.
- **Blocks #854.** Nothing else: #855 (export assembly) parses for its own reasons and #862 (binding) needs identities, not drafts.
- The cross-language shape fixture `engine/tests/fixtures/operation-shape-conformance.json` retires when the last renderer reader of the shapes moves, which is still #854 and #855, not this.

## Acceptance criteria

- [ ] The engine can answer, for a stored document, the requests an import of it would build - method, URL, params, headers, body and folder.
- [ ] A cross-language fixture pins those drafts against the renderer's importer output, the way `declared-operations-conformance.json` pins the identities, and it covers a `$ref`'d body, a sampled body and an operation with no tag.
- [ ] `python build.py -t && cd engine && ctest --preset linux-dev --output-on-failure`; `cd app && pnpm test && pnpm type-check` all green.

athrvk:
## Description

`core::spec_request_drafts_of` - the engine can now answer, for a stored document, the requests an import of it would build: method, URL with its query joined in, params, headers, sampled body, and the folder an import would file each under.

**The plan, and the alternative rejected.** The root cause this issue names is that #854 is filed as one move and is two: the sync diff's inputs are `SpecRequestDraft`s, and a draft is not an identity but *the request an import would build*, produced by ~1,850 lines of import parser running over the document text. Verified still true at `f9ab55b` - `spec-diff.ts` compares `fieldValues()` over `name/description/method/url/params/headers/body`, and every one of those comes off `spec-operations.ts` → `parseImport`. So this lands the drafts alone, and #854 keeps the comparison. The alternative rejected was porting the diff first over a renderer-supplied draft list: that leaves the two-opinions state #853's own third criterion exists to prevent, because the engine would then be diffing against a shape only the renderer can produce - and an agent over MCP, which is what this whole phase exists for, has bytes and no importer.

**One deliberate divergence from the issue's list, and the reason.** The issue names `$ref` **bundling** (`ref-bundler.ts`, 448 lines) among what must not be lost. It is not ported, and it is not lost: bundling runs *before* parse (`batch.ts`, `spec-refetch.ts`), resolving external files into an `x-vayu-bundled` map and rewriting the refs that pointed into them, so **the bytes the engine reads are already bundled**. What the engine needs is in-document `$ref` resolution reaching into that subtree, which the JSON-Pointer resolver does generically, and the recursive case it bounds - which the sampler bounds here too (depth cap plus a per-branch cycle set). A ref still external by the time a document is stored is one the user has already been told about, and it resolves to nothing on both sides alike.

### What landed

**Engine**
- `core::spec_request_drafts_of` (`src/core/openapi_drafts.cpp`) with `SpecRequestDraft` / `DraftRequest` / `DraftBody` / `DraftField` declared beside `declared_operations_of` and `response_schemas_of`. Both dialects: 2.0's `default`-stated parameter values, `body` / `formData` parameters and `consumes`-decided form encoding; 3.x's `example` / `examples` precedence, `requestBody` under a media type, and `schemaFormFields` off the *sampled* stub rather than off `properties`.
- **One walk, three answers.** `src/core/openapi_walk.hpp` is `walk_operations` and the helpers it is built from, moved out of `openapi_document.cpp` rather than written twice - so the `operations` index, the `responseSchemas` index and the drafts cannot disagree about which operations exist or which of two kept a repeated `operationId` (#715). The walk gained the resolved path item, which the drafts need for path-level `parameters`; `response_schemas_of` now asks the shared `spec_dialect` instead of re-deriving "is this 3.x" inline.
- **`core::normalize_path_templates`** (`src/core/path_template.{hpp,cpp}`): the mock server's `{param}` → `{{param}}` rewrite, moved to core and called from both. A draft's URL is written with it and the mock reads that URL back, so a second copy is an imported request the mock silently cannot route. Still pinned by `path-template-conformance.json`; `mock_server.cpp` keeps only its own `:param` rule.
- No route, as the issue allows: the only caller is the diff, and that is renderer-side until #854.

**Faithfulness, where it is load-bearing.** The diff compares text, so the port carries JavaScript's semantics rather than the nearest C++ idiom, each one commented at its site: `??` falling through `null` but not `false`/`0`/`""`; `Map.set` on an existing key replacing the value *in place*, which is what puts a path item's parameter and the operation's override in the same query position; `String(x)` on a non-string `name`; and `JSON.stringify`'s number spelling - `4.0` is `4`, `1e20` is written out in full, `1e-7` keeps a bare exponent - where `nlohmann::dump` writes `1.0` and a shortest-round-trip printer writes `1e+20`.

**Not ported, deliberately**: the skip tally (`meta.skipped`), which describes an import to a user rather than a document to the diff, and the saved response examples, which #854 records as uncompared. Both stay the renderer's until the import itself moves.

**Docs (same commit)**: `engine/CLAUDE.md`, `docs/app/import-collections/openapi-v3.md`, `docs/app/import-collections/openapi-v2.md`.

## Type of Change
- [x] New feature
- [x] Documentation update

## Testing

All green on this branch:

- `python build.py -t && cd engine && ctest --preset linux-dev --output-on-failure` - **2251 passed, 0 failed**, including 12 new `SpecRequestDrafts` cases.
- `cd app && pnpm test` - **5780 passed** across 525 files, including the 6 new `spec-request-drafts.conformance` cases.
- `cd app && pnpm type-check`, `pnpm lint`, `prettier --check` on the one new file - clean.
- Strict rebuild of every file this touches: **no new warnings** (the `-Wconversion` / `-Wdangling-reference` ones that appear are pre-existing in `openapi_export.cpp` and `openapi_export_test.cpp`).

**The cross-language fixture is the point, and it was generated rather than authored.** `tests/fixtures/spec-request-drafts-conformance.json` holds five documents - a 3.x one with a `$ref`-ed request body over a recursive `$ref`-ed schema, an untagged 3.x one exercising all three folder outcomes, a 3.x multipart-and-urlencoded pair, a 2.0 one with a body parameter and a file part, and 2.0 as YAML with a repeated `operationId` - and its `drafts` were written by running the renderer's own parsers over them. The engine then had to match, and does, on all 14 drafts. So it records what the parsers answer rather than what anyone believed they answer. It covers the three shapes the issue asks for: a `$ref`'d body, a sampled body, and an operation with no tag.

**Mutation-checked, both ways.** Drop the `append_params` call and four cases redden, the fixture case among them - the fixture is not vacuous. A second mutation was more useful than a passing test: disabling the whole-number branch of the number printer changed *nothing*, because `to_chars` already spells `4.0` as `4`. The branch only earns its place at `1e20`, where JavaScript writes the digits out and a shortest printer writes `1e+20` - so the test grew `huge` / `huger` / `tiny`, checked against `node -e 'JSON.stringify(...)'`, and the branch is load-bearing now.

Beside the fixture, 11 engine cases pin what a fixture cannot state, each naming what reverting it costs: a document that is not a contract builds no drafts; a `paths` key that is not a path declares no request; a path item that is a `$ref` is read through; the operation's parameter wins in the path item's position; a `parameters` *mapping* is stepped over rather than read as a list; a recursive schema samples without running forever; the query is written the way a request built in the app carries it (#590), with disabled rows out and a `{{var}}` unencoded; `Authorization` / `Content-Type` / cookie / path parameters are dropped; and the same document read as YAML and as JSON produces byte-identical drafts.

`clang-tidy` could not run locally (this image ships 18.1.3, and `engine/.clang-tidy` needs >= 19); CI lints on a current toolchain.

## Checklist

## Related Issues

Closes #865. Unblocks #854, which is now the comparison alone.

Sequencing held against live state: #853 (PR #859) is merged, which is the reader this builds on. No overlap with the one other open PR (#864, `spec_bind.cpp` and the MCP tool table) - this touches neither file.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FLC3ybqQTm9csVgGTM4PeJ)_


athrvk:
`master` moved while this was opening (#864 merged), so `master` is merged in at `f2865be` - **no conflicts**, and the two changes touch no common file beyond `engine/CLAUDE.md` and `engine/CMakeLists.txt`, both of which auto-merged.

Re-verified on the merged head, so the numbers in the body above (taken pre-merge) now read:

- `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` - **2264 passed, 0 failed** (2251 + #864's 13 `SpecBindRouteTest` cases).
- `cd app && pnpm test` - **5782 passed** across 525 files.
- `pnpm type-check` and `mkdocs build --strict` - clean.

Everything else in the body stands unchanged.

---

Issue #898, opened 2026-08-21
engine(build): audit the three defect-hiding MSVC warning suppressions (/wd4101, /wd4456, /wd4244) on a Windows build


athrvk:
Follow-up carved out of #884. That PR made `-Werror` real on the Linux (GCC) leg via `VAYU_WERROR` and swept the tree warning-clean there, and gave every MSVC `/wd` suppression a rationale comment. It stopped short of *removing* the three defect-hiding suppressions, because that cannot be validated from a Linux session.

## Problem

`engine/CMakeLists.txt` builds MSVC with `/W4 /WX` and five suppressions. Two are permanent and justified in place (`/wd4324` alignas padding, `/wd4100` unreferenced formal parameter). The other three hide real defect classes:

- `/wd4101` - unreferenced local variable
- `/wd4456` - declaration hides previous local (shadowing)
- `/wd4244` - conversion, possible loss of data (the class `-Wconversion` catches on the GCC/Clang legs, which are now clean and `-Werror`-gated)

#884 kept them because MSVC's diagnostics diverge from GCC's - notably `4456` has no `-Wshadow` equivalent enabled on the Linux sweep, so a clean Linux build gives **zero signal** about whether shadowing exists in the code MSVC compiles. Dropping any of the three needs a Windows build to confirm the tree is clean of that class first; otherwise it just reddens Windows CI (`/WX` is unconditional there).

## Fix / solution pathway

For each of `/wd4101`, `/wd4456`, `/wd4244`, on a Windows toolchain (or by watching the Windows CI leg on a scratch PR):
1. Remove the suppression.
2. Build `windows-prod` with tests and read the diagnostics `/WX` now turns into errors.
3. Fix each real finding (an unused local is dead code; a shadow is a readability/latent-bug fix; a 4244 narrowing gets an explicit cast, mirroring what #884 did on the GCC side).
4. If a class turns out to have many benign hits not worth touching, keep that one suppression and record why in the comment - decide per code, not as a batch.

`/wd4244` is the strongest removal candidate: its class is already gated on the other two legs, so removing it here closes the one platform where a narrowing can still merge silently.

## App changes

None - engine build flags only.

## Tests

CI is the test: Windows engine job green with the suppression(s) removed. Mutation-check any code fix the usual way (revert, confirm the suppressed warning returns).

## Docs

`docs/engine/building.md` already documents `VAYU_WERROR`; update only if the MSVC story changes materially.

## Sequencing

[...]

Issue #897, opened 2026-08-21
engine(build): zero the compiler-warning inventory - 205 warnings at 148 sites across 10 flags, fully enumerated and classified


athrvk:
Owner-requested sweep (2026-08-21) of the CI annotations (e.g. run 32468265026: 34 warnings + 3 notices). This issue is the complete, classified inventory and the fix plan. It is the prerequisite half of #884: **this issue makes the warning count zero; #884 makes warnings fatal.** The three "notices" are the per-platform "Publish C++ test results" steps - informational, not defects.

## How the inventory was taken (and why CI annotations undercount)

The annotations are compiler diagnostics from the engine build steps. sccache replays cached compiler stderr, so the same warnings resurface every run - but GitHub caps annotations per step and job-log fetches are windowed, so no single run shows the whole set (the sample run shows 9 on ubuntu, 10 on macOS, and the Windows build's output scrolls out of reach). The authoritative list below is a **cold local release build at `0fe0e67`** matching the CI configuration: `linux-prod` preset (`-O2`), `VAYU_BUILD_TESTS=ON`, GCC 13.3.0 (same major as ubuntu-latest), all 236 TUs compiled, stderr captured in full: **205 warning emissions, 148 unique sites.** macOS-clang extras were read from the CI log; the MSVC set could not be enumerated remotely (see the end).

## The inventory, by flag

| Flag | Count | Verdict |
|---|---|---|
| `-Wmissing-field-initializers` | 143 | real, benign - fix at the struct, not the 100+ sites |
| `-Wnull-dereference` | 16 | false positives (libstdc++/nlohmann inlining) |
| `-Wconversion` | 12 | real pattern, 5 sites (one is #884's) |
| `-Wsign-conversion` | 9 | real, benign - casts |
| `-Wdangling-reference` | 9 | false positives (GCC 13 heuristic) |
| `-Wuninitialized` / `-Wmaybe-uninitialized` | 8 | real, benign UB - one NSDMI |
| `-Wrange-loop-construct` | 5 | real, trivial |
| `-Warray-bounds=` | 2 | false positives (GCC 13 string-concat family) |
| `-Wcomment` | 1 | trivial |

## Fix plan, by root cause

**1. `-Wmissing-field-initializers` (143 emissions): five structs, five one-line fixes.** GCC/clang do not warn for aggregate members that carry a default member initializer, so the cure is NSDMI on the struct, not editing every brace-init site:

[...]

PR #899, opened 2026-08-21
build(engine): make -Werror a real CI gate (#884)

### LINKED ISSUE #884: engine(build): the spec_sync.cpp -Wconversion warning is untracked, and warnings-as-errors is real on no platform
From the 2026-08-21 tooling-hygiene sweep at master `0fe0e67` (0.22.1). Two halves of one policy problem: a known warning nobody filed, and the reason it can sit there indefinitely.

## Problem (verified)

**1. The known warning has no tracking issue - a deferred-work-rule violation.** `engine/src/http/routes/spec_sync.cpp:376` is `return draft_example_rows (drafts_[*found].draft.examples);` inside `rows_for()`, which returns `std::optional<nlohmann::json>`; `draft_example_rows` (declared `engine/include/vayu/http/routes.hpp:554`) returns `nlohmann::json`, and the implicit `json -> optional<json>` conversion goes through nlohmann's SFINAE-heavy templated constructor set, emitting a `-Wconversion` overload-resolution warning on GCC/Clang. It came in with PR #869 and PR #876's body records it verbatim as pre-existing ("not fixed or hidden") - but no issue was ever filed. Verified: a tracker search for `spec_sync.cpp` returns only closed #854 (unrelated), and no issue tracks the warning under any phrasing. Per CLAUDE.md, deferred work is filed in the same commit that defers it; this issue is that filing, late.

**2. Why it will never self-resolve: warnings-as-errors binds nowhere.**

- GCC/Clang (`engine/CMakeLists.txt:70-81`): `-Wall -Wextra -Wpedantic -Wconversion -Wsign-conversion -Wdouble-promotion -Wformat=2 -Wnull-dereference` - a strong set, but `-Werror` is gated on `$ENV{VAYU_STRICT_BUILD}` (line 74), and **nothing in the repo sets it**: grep finds only the CMakeLists line and `engine/tests/http_version_test.cpp:12`, which states in-tree that CI does not set it. Zero hits in `.github/workflows/` or `build.py`. So on Linux/macOS every warning scrolls by uninspected.
- MSVC (`engine/CMakeLists.txt:83-93`): `/W4 /WX` unconditionally, but with five bare suppressions - `/wd4324 /wd4101 /wd4100 /wd4456 /wd4244` - none carrying a rationale comment. `4244` is "conversion, possible loss of data": **the exact class `-Wconversion` exists to catch on the other legs.** Net effect: no platform fails a build on a narrowing conversion, and Windows additionally ignores unreferenced locals (4101) and shadowing (4456), both of which hide real defect classes.

## Fix / solution pathway

1. Fix the `:376` conversion (explicit `std::optional<nlohmann::json>{...}` construction, `std::make_optional`, or change `rows_for`'s interplay with `draft_example_rows` - implementer's choice). Reproduce first with a `linux-dev` configure and paste the exact diagnostic into the PR.
2. Confirm the engine builds warning-clean on all three legs (CI logs are enough for macOS/Windows).
3. Make `-Werror` real: export `VAYU_STRICT_BUILD=1` in the CI engine build jobs, or (cleaner, long-term per the repo's core principle) replace the env-var gate with a CMake option that the CI presets pass. Local dev builds may stay non-strict - the gate is for CI.
4. Give each MSVC `/wd` code a one-line rationale comment, and re-decide each on its merits: 4324 (structure padding) and 4100 (unreferenced formal parameter) are plausibly permanent; 4101 and 4456 hide real defects; 4244 contradicts the `-Wconversion` policy and should probably go now that step 2 holds.

## App changes

None, verified - engine build flags only; no app surface reads them.

## Tests

CI is the test: after step 3, the strict build must be green on all three platforms. Mutation-check locally: introduce a deliberate narrowing conversion, confirm the strict configure fails the build, revert.

## Docs (same commit)

`docs/engine/building.md` if it documents warning flags or `VAYU_STRICT_BUILD` (verify; the flag currently appears in a test comment, so the doc may not mention it - then the CMake option's help string is the record).

## Sequencing

Tooling-hygiene wave, 2026-08-21: **#884 (this issue) lands first** - it is the smallest, and #886's bulk reformat will rewrite `spec_sync.cpp` among 148 files, so the one-line fix here should merge before that to avoid a pointless conflict. Then #885 (tidy gate), then #886 (format bulk + gate); #887 (script-lint) is rebase-level against those two on `pr-tests.yml`; #888 is fully parallel. Touches `engine/src/http/routes/spec_sync.cpp`, `engine/CMakeLists.txt:70-93`, and the CI engine-job env - disjoint from every non-wave open issue (#197, #287, #690, #693, #467, #473, #134 anchor none of these regions; rebase-level at worst).

## Acceptance criteria

- [ ] The `:376` warning is fixed and the diagnostic that proved it is recorded in the PR.
- [ ] `-Werror` (or `/WX` with justified exceptions) is enforced by CI on at least the Linux leg, via a mechanism with a setter that exists in the repo.
- [ ] Every remaining MSVC `/wd` carries a one-line reason; `/wd4244` is either removed or its retention argued in the comment.
- [ ] Engine suite green on all platforms with the strict build on.

---

athrvk:
## Description

Warnings-as-errors bound on no platform that could report one. The GCC/Clang `-Werror` was gated on `$ENV{VAYU_STRICT_BUILD}`, which nothing in the repo ever set, so every warning scrolled by uninspected; MSVC's `/WX` was held green by five `/wd` suppressions, three of which hid real defect classes. #900 took the warning inventory to **zero** - this is the half that keeps it there.

**Rebased onto `75efb14` (post-#900) and reduced to what #900 does not cover.** Everything #900 fixed at the declaration - the NSDMI structs, the six scoped `diagnostics.hpp` suppressions, the `-Wconversion` / sign-conversion / range-loop / `-Wcomment` sites, the `Result` scalars, the `scripting.cpp` captures - is dropped from this PR rather than duplicated or contradicted. In particular the global `-Wno-missing-field-initializers` and the two `-Wno-error=` lines an earlier revision of this PR carried are **gone**: `docs/engine/building.md` now says never to widen a suppression into `vayu_warnings`, and that was exactly what they did. See the discussion below.

**Plan.** Replace the dead env-var gate with a `VAYU_WERROR` CMake option (default `OFF` locally, `VAYU_STRICT_BUILD` still sets the default for back-compat) and have CI pass `-DVAYU_WERROR=ON`. Turning it on then surfaced four things that only exist once warnings are fatal, each found on the leg that reports it:

| Leg | Finding | Fix |
|---|---|---|
| macOS (Clang) | Vendored `quickjs.h` failed the build: the SDK spells `NAN` as `__builtin_nanf`, so `v.u.float64 = NAN` is `-Wdouble-promotion` there and nowhere else | Vendored `quickjs-ng` / `hdrhistogram` headers propagate as **SYSTEM** includes. Compiling vendored *sources* with their own flags was only half the exemption - on a plain `-I` the diagnostic is attributed to our TU. |
| Windows (MSVC) | With `/wd4456` gone, C4456 caught a **genuine shadow** in `event_loop_worker.cpp`: the completion branch redeclares `data` while the fetch phase's `data` is still in scope | Inner name is now `completed`. A local `-Wshadow=local` sweep found three more (`json.cpp`, `metrics_collector.cpp`, `execution.cpp`); each inner name now says what it holds. |
| Clang 18 | `-Wno-missing-designated-field-initializers` (19+) and `-Wcast-function-type-mismatch` (19+) are *unknown* names on 18, and an unknown-warning-option is itself a warning - so `-Werror` makes each suppression break the build it exists to quieten | Flag gated on Clang >= 19; the pragma block in `script_engine.cpp` leads with `-Wunknown-warning-option`. |
| Clang | One unused `[this]` capture in `sse_load_stream_test.cpp` | Removed. Its sibling handlers keep theirs - their nested lambdas capture `this` through them, so removing those would not compile. |

`/wd4101`, `/wd4456` and `/wd4244` are **removed**, which answers #884's "removed or its retention argued" by removal. `/wd4324` (alignas padding on the cache-line-aligned SPSC queue) and `/wd4100` (the `-Wno-unused-parameter` counterpart) stay, each with a one-line reason.

macOS is gated as well as Linux, deliberately: gating one GCC/Clang leg leaves the other compiler's diagnostics unwatched, and the `NAN` failure above is exactly that class.

Note: `-Werror` is part of the sccache command-line hash, so the two gated legs may see reduced reuse of the `cache-warm` scope until their own scope repopulates. Broadening `-DVAYU_WERROR=ON` to `cache-warm.yml` / `release.yml` is deliberately not done here.

## Type of Change
- [x] Bug fix (a CI gate that never gated, plus the shadow and the two self-defeating suppressions it exposed)
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update (`docs/engine/building.md`)

## Testing

| Build | Result |
|---|---|
| `linux-prod` + tests, `-DVAYU_WERROR=ON`, **GCC 13.3** | **0 warnings, 0 errors** |
| `linux-prod` + tests, `-DVAYU_WERROR=ON`, **Clang 18** | **0 warnings, 0 errors** |
| Engine suite on the strict tree | **2359 / 2359 passed** |

- Mutation check on the gate: injecting a `double -> int` narrowing fails the strict build with `-Werror=float-conversion`; reverting restores green.
- Mechanism check: `VAYU_STRICT_BUILD=1` configure yields 236 `-Werror` entries in `compile_commands.json`; no env yields 0 (default off).
- macOS and Windows were verified **on their own runners** - both findings above were read from failing jobs on this branch and fixed, which is the only way they could have been found from a Linux session.
- `mkdocs build --strict` clean; `pr-tests.yml` YAML validated (matrix `configureArgs` parse as strings).

No pre-existing failures on the base commit.

## Checklist
- [x] I have added tests (if applicable) - not applicable: CI itself is the test #884 asks for, and the mutation check above is the behavioural proof the gate binds

## Related Issues
Closes #884
Builds on #900 (#897). Makes #898 obsolete - the `/wd` audit it defers happened here.

### DIFF
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -344,12 +344,20 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         include:
+          # VAYU_WERROR turns the GCC/Clang warning set into build errors, on
+          # both legs that use it (issue #884) - a warning that only one
+          # compiler emits is exactly the kind this gate exists to catch, so
+          # gating Linux alone would leave Clang-only diagnostics unwatched.
+          # Windows needs no flag: MSVC builds /WX unconditionally.
           - os: ubuntu-latest
             preset: linux-prod
+            configureArgs: "['-DVAYU_BUILD_TESTS=ON', '-DVAYU_WERROR=ON']"
           - os: windows-latest
             preset: windows-prod
+            configureArgs: "['-DVAYU_BUILD_TESTS=ON']"
           - os: macos-latest
             preset: macos-prod
+            configureArgs: "['-DVAYU_BUILD_TESTS=ON', '-DVAYU_WERROR=ON']"
     timeout-minutes: 60
     env:
       VCPKG_COMMIT: '94a541197763a4f449a1b91478df48c0584a6256'
@@ -436,7 +444,7 @@ jobs:
           cmakeListsTxtPath: '${{ github.workspace }}/engine/CMakeLists.txt'
           configurePreset: ${{ matrix.preset }}
           buildPreset: ${{ matrix.preset }}
-          configurePresetAdditionalArgs: "['-DVAYU_BUILD_TESTS=ON']"
+          configurePresetAdditionalArgs: "${{ matrix.configureArgs }}"
 
       # The engine must depend on nothing outside Windows itself. ctest below
       # runs on a runner that already has the Visual C++ redistributable, so it
--- a/docs/engine/building.md
+++ b/docs/engine/building.md
@@ -116,6 +116,9 @@ cmake -B build -DCMAKE_BUILD_TYPE=Debug -DVAYU_USE_ASAN=ON
 
 # Enable ThreadSanitizer (debug builds)
 cmake -B build -DCMAKE_BUILD_TYPE=Debug -DVAYU_USE_TSAN=ON
+
+# Treat compiler warnings as errors (what CI does)
+cmake -B build -DVAYU_WERROR=ON
 ```
 
 ## Build Outputs
@@ -515,6 +518,19 @@ cmake --build --preset linux-prod --clean-first 2>&1 | grep -c 'warning:'
 The flags are on `vayu_warnings` in `engine/CMakeLists.txt` - `-Wall -Wextra
 -Wpedantic` plus `-Wconversion`, `-Wsign-conversion`, `-Wdouble-promotion`,
 `-Wformat=2` and `-Wnull-dereference` on GCC and clang, `/W4 /WX` on MSVC.
+
+**A warning fails the build in CI** (`VAYU_WERROR`, default `OFF`). The engine
+job passes `-DVAYU_WERROR=ON` on the Linux and macOS legs, which adds `-Werror`;
+Windows needs no flag, because MSVC has always built `/WX`. It is off by default
+so a work-in-progress local build is not blocked by a warning - run the count
+above, or configure with `-DVAYU_WERROR=ON` to get exactly what CI gets. Setting
+the legacy `VAYU_STRICT_BUILD` environment variable to a truthy value turns it on
+too, so an existing local workflow keeps working. Both GCC/clang legs are gated
+rather than only Linux: a diagnostic one compiler emits and the other does not is
+precisely what the gate is for - the macOS SDK spells `NAN` as `__builtin_nanf`,
+which made a line in vendored `quickjs.h` a `-Wdouble-promotion` error there and
+nowhere else (vendored headers are `SYSTEM` includes now, so third-party code
+cannot fail our build).
 
 Three GCC 13 diagnostics fire only on code the engine did not write, and those
 are suppressed - narrowly, one function at a time, through the macros in
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -16,6 +16,21 @@ option(VAYU_BUILD_ENGINE "Build daemon" ON)
 option(VAYU_USE_ASAN "Enable AddressSanitizer" OFF)
 option(VAYU_USE_TSAN "Enable ThreadSanitizer" OFF)
 
+# Treat compiler warnings as errors. Off for local dev so a work-in-progress
+# build is not blocked by a warning, on in CI so a warning cannot merge - the
+# warning set below is only a gate if something fails on it. #900 took the
+# inventory to zero; this is what keeps it there. CI passes -DVAYU_WERROR=ON;
+# the legacy VAYU_STRICT_BUILD environment variable still turns it on (default
+# only - an explicit -DVAYU_WERROR wins) so an existing local workflow keeps
+# working. See docs/engine/building.md.
+set(_vayu_werror_default OFF)
+if(DEFINED ENV{VAYU_STRICT_BUILD}
+   AND NOT "$ENV{VAYU_STRICT_BUILD}" STREQUAL ""
+   AND NOT "$ENV{VAYU_STRICT_BUILD}" STREQUAL "0")
+    set(_vayu_werror_default ON)
+endif()
+option(VAYU_WERROR "Treat compiler warnings as errors (CI gate)" ${_vayu_werror_default})
+
 # ============================================================================
 # C++ Standard and Compiler Settings
 # ============================================================================
@@ -71,7 +86,7 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
         -Wall
         -Wextra
         -Wpedantic
-        $<$<BOOL:$ENV{VAYU_STRICT_BUILD}>:-Werror>
+        $<$<BOOL:${VAYU_WERROR}>:-Werror>
         -Wno-unused-parameter
         -Wconversion
         -Wsign-conversion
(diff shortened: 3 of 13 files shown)


athrvk:
Heads-up from the routine that opened #900 (for #897, the warning-inventory half): **these two PRs overlap heavily and contradict each other on four decisions.** #899 was opened at 14:45 and #900 at 15:34; #900's run listed open PRs at 14:17 and saw none, which is why it proceeded rather than deferring. Flagging rather than touching anything here - this is your call, and #899 is not mine to push to.

### Measured conflict

Merging the two branches locally conflicts in **6 files**:

```
engine/src/core/metrics_collector.cpp
engine/src/core/openapi_export.cpp
engine/src/http/routes/spec_sync.cpp
engine/src/runtime/script_engine.cpp
engine/tests/metrics_collector_test.cpp
engine/tests/version_isolation_test.cpp
```

Mostly the *same fixes written differently* (`std::make_optional` vs `std::optional<Json>{std::in_place, ...}`; identical `static_cast<size_t>` index casts). Those are trivial to resolve either way.

**The dangerous part is what merges cleanly.** `engine/CMakeLists.txt` and `docs/engine/building.md` do **not** conflict, so a naive merge silently produces a tree that has #900's struct-level fixes *and* #899's global suppressions of the warnings those fixes already eliminated - plus two doc sections stating opposite policies.

### The four contradictions

| | #899 (this PR) | #900 (#897) | What #897 asked for |
|---|---|---|---|
| `-Wmissing-field-initializers` (143) | `-Wno-...` globally; "GCC noise on the house idiom" | NSDMI on 5 structs + 1 C struct; flag stays on, count goes to 0 | "fix at the struct, not the 100+ sites" |
| `-Wnull-dereference` / `-Warray-bounds` (27) | `-Wno-error=` globally, still printed | `push`/`pop` around the 6 producing functions; flags stay fatal-capable | "suppress narrowly... do **NOT** weaken the flags globally" |
| `Result` indeterminate reads | `Result r{};` at 10 call sites | `int id = 0;` etc. on the struct | "no brace-init site edits where a struct-level fix sufficed" |
| `operation_of` `-Wdangling-reference` | `const json&` → `const json` (9 copies) | file-scoped pragma, references kept | "restructuring 9 correct tests to appease a false positive is the worse trade" |

I don't raise this to argue #900 is right - #897's text happens to line up with it, but you wrote both issues and the call is yours. I raise it because merging both as they stand gives a tree where three flags are disabled for warnings that no longer exist.

### Proposal, if it helps

They are complementary in one order. #899 carries three things #900 does not and should keep regardless:

- the `VAYU_WERROR` option and the CI wiring (the actual gate),
- vendored `quickjs.h` / hdrhistogram as **SYSTEM** includes - the macOS `NAN` / `-Wdouble-promotion` fix, which #900 never hit and which nothing else covers,
- removing `/wd4101 /wd4456 /wd4244`, and the **genuine shadow bug** C4456 caught in `event_loop_worker.cpp`. That is the best find in either PR.

So: **land #900 first, then rebase #899 dropping the parts it no longer needs** - the global `-Wno-missing-field-initializers`, both `-Wno-error=` lines, and the 10 `Result r{}` call-site edits - because at #900's head a cold `linux-prod` + tests build already emits **0** warnings on GCC, clang and MSVC, with every flag still fatal-capable. `-Werror` then gates on a genuinely clean tree instead of on a tree with three analyses switched off. The reverse order works too; it just means #900 rebases and the suppressions get removed later rather than never added.

One correction to this PR's Testing notes either way: it reports the inlined-libstdc++ diagnostics as unavoidable. They are suppressible at the function that triggers the inlining - #900 does it in six places via `engine/include/vayu/utils/diagnostics.hpp`, and the resulting build prints zero. Worth a look before settling for `-Wno-error=`. One wrinkle #900 found the hard way and this PR would hit too: `-Warray-bounds` and `-Wstringop-overflow` are the same diagnostic under two names, so silencing only the first just moves it.

---

athrvk:
Acted on this - thanks, the analysis was right on every point I could check, and #900 having merged settles the ordering. This PR is now **rebased onto master at `75efb14`** and reduced to what #900 does not cover.

**Dropped, because #900 did it better at the declaration instead of the call site or the flag:**

- the global `-Wno-missing-field-initializers` (NSDMI on the five structs is the correct cure, and your doc section says plainly never to widen a suppression into `vayu_warnings` - which is exactly what my version did)
- both `-Wno-error=` lines (`diagnostics.hpp` suppresses narrowly and keeps the flags fatal-capable; and the `-Warray-bounds` / `-Wstringop-overflow` alias would have bitten me later - my Release build only ever emitted the first name, so I would have "fixed" it and moved it)
- the 10 `Result r{}` call-site edits, the five `-Wconversion` sites, the sign-conversion casts, the range-loop references, the `-Wcomment` rewording, and the 9 `operation_of` reference-to-value changes - all superseded, and I agree the file-scoped pragma is the better trade there
- the two `scripting.cpp` `ctx` captures, which you also found

**Kept, because nothing else covers them** (11 files, was 20):

- `VAYU_WERROR` and the CI wiring - the gate itself, on the Linux **and** macOS legs
- vendored `quickjs.h` / hdrhistogram as `SYSTEM` includes: the macOS `NAN` / `-Wdouble-promotion` failure only appears once `-Werror` is on, which is why #900's green macOS build did not see it
- `/wd4101 /wd4456 /wd4244` removed, with the two that stay carrying their reason
- the four shadow renames the removal exposed, including the `event_loop_worker.cpp` one
- three things the gate surfaced after your comment: `-Wunknown-warning-option` must lead the clang pragma block in `script_engine.cpp` and `-Wno-missing-designated-field-initializers` must be gated on Clang >= 19 (both are unknown-name warnings that `-Werror` promotes, so the suppressions would break the build they exist to quieten on Clang 18), plus one genuinely unused `[this]` capture in `sse_load_stream_test.cpp` - the sibling handlers there keep theirs, since their nested lambdas capture `this` through them

One correction to my own PR body, now fixed: it called the inlined-libstdc++ diagnostics unavoidable. They are not - your six scoped suppressions print zero, and this PR now inherits them rather than working around them.

#898, which I filed to defer the `/wd` audit, is obsolete: the audit happened here.

---

athrvk:
Green on all three legs at `2d923ad` - **CI gate: success**, C++ Tests success on ubuntu, macos and windows.

The last finding came from Windows and is worth recording, because it is the fourth defect class the removed suppressions were hiding and the only one no sweep from this machine could have found: `run_manager.cpp:1771` had a `catch (const std::exception& e)` that never reads `e`, which MSVC reports as C4101 "unreferenced local variable". Neither GCC nor clang warns about an unused *catch parameter* at any `-W` level we enable, so the Linux and macOS legs are silent on it by construction.

Rather than take another CI round per site, I scanned every catch block in `src/` and `include/` for a parameter its body never mentions: exactly two, and the second (`metrics.cpp:389`) is the one the build would have stopped on next. Both now omit the unused name.

Final state of what the gate found, end to end:

| Class | Found by | Sites |
|---|---|---|
| `-Wdouble-promotion` in vendored `quickjs.h` | macOS runner | 1 (fixed by SYSTEM includes) |
| C4456 local-hides-local | Windows runner, then a local `-Wshadow=local` sweep | 4 |
| C4101 unused catch parameter | Windows runner, then a local scan of every catch block | 2 |
| `-Wunknown-warning-option` on Clang 18 | local clang build | 2 (a CMake flag and a pragma) |
| `-Wunused-lambda-capture` | local clang build | 1 |

None of these were reachable from a GCC-on-Linux build, which is the argument for gating both compiler legs rather than one.

---

Issue #886, opened 2026-08-21
engine(tooling): .clang-format is a 2015 template the code contradicts - 148 of 284 files drift; choose the real config, bulk-format once, gate it in CI


athrvk:
From the 2026-08-21 tooling-hygiene sweep at master `0fe0e67` (0.22.1). This is the issue behind the paragraph PR #876 had to write ("clang-format 18 reports differences across engine files this branch does not touch, so the tree is not format-clean at that version") - quantified, root-caused, and made actionable.

## Problem (verified)

**The config is not the house style.** `.clang-format` at the repo root (engine/ has none of its own) opens with "Made by: Ingmar Delsink ... Tested with: clang-format version 3.7.1" - an imported 2015 generic template - while `engine/CLAUDE.md:20` and `CONTRIBUTING.md:100` present it as the engine's formatter. Measured against the code it supposedly governs, with clang-format 18.1.3:

- **148 of 284 non-vendor engine files (52%) are not clean; 4471 replacements total** (engine/src: 53 files / 2376, engine/tests: 65 / 1730, engine/include: 30 / 365). Worst: `import_document.cpp` (280), `database.cpp` (269), `routes/import.cpp` (207), `run_manager.cpp` (180), `runs_route_test.cpp` (165).
- The drift is style disagreement, not version skew or config breakage (zero parse errors across all 284 files). Three settings do most of the damage: `ColumnLimit: 80` against a codebase where 6,904 of 127,103 lines exceed 80 columns by convention; `ContinuationIndentWidth: 0`, which rewrites continuation lines to column 0 (e.g. `routes/import.cpp:168-169`) and reflows correct hand-wrapped doc comments (`import.cpp:57-59`, `:174-179`); and `SortIncludes`, which reorders includes (`import.cpp:19-20`). Formatting `routes/import.cpp` alone produces a 519-line diff.

**Nothing enforces it and no version is pinned.** Zero hits for `clang-format` across `.github/workflows/`, `scripts/pre-commit` (tidy-only) and `build.py`; the only format gate in CI is the app's `pnpm format:check` (`pr-tests.yml:206-208`). No doc names a clang-format version, so "clean" is currently undefined across LLVM majors - contrast `MIN_CLANG_TIDY_MAJOR=19` at `scripts/pre-commit:19`, which is the existing in-repo pattern for exactly this.

What is already right and must be preserved: `engine/vendor/.clang-format` is `DisableFormat: true` - vendor stays untouched.

## Fix / solution pathway

[...]

PR #900, opened 2026-08-21
build(engine): zero the compiler-warning inventory - 205 warnings to 0

### LINKED ISSUE #897: engine(build): zero the compiler-warning inventory - 205 warnings at 148 sites across 10 flags, fully enumerated and classified

athrvk:
## Description

A cold `linux-prod` + tests build at `0fe0e67` emitted **205 warnings at 148 sites across 10 flags**. It now emits **zero**, on the release *and* the debug leg, on all three platforms. This is the prerequisite half of #884: this PR makes the count zero, #884 makes warnings fatal.

**Plan.** The root cause of 143 of the 205 was not 100+ careless brace-inits - it was five structs (and one C struct) whose *trailing* fields carry no default member initializer, so every aggregate init that legitimately stops short of them is a warning. GCC does not warn for a member that has an NSDMI, so the cure belongs on the field. The approach is therefore "fix the declaration, never the call site", and the diff touches zero of the ~140 initialization sites. The one genuine defect in the set is `db::Result` / `db::MonitorSample`, whose bare scalars were read indeterminate when a partly-built row was copied - that is formally UB and is now defined zero. The alternative I rejected was adding `-Wno-...` entries to `vayu_warnings` for the 27 false-positive emissions: it is one line instead of six scoped suppressions, and it would hide the next *genuine* instance of each family across the whole engine, which is the entire value of having `-Wnull-dereference` on in the first place.

Verified against live code rather than the issue's snapshot; two things had drifted and are handled as found:

- `-Wuninitialized` fires on `Result::timestamp` and `Result::latency_ms` as well as `Result::id`. All four scalars on `Result` and both on `MonitorSample` are now `= 0` - same struct-level fix, no extra sites.
- `ScenarioStep` warns on **four** trailing fields (`data_template`, `auth`, `spec_operation`, `auth_template`), not one. All four get `{}`.

Per the correction comment on #884, this PR **takes all five `-Wconversion` sites**, `spec_sync.cpp:376` included - the acceptance criterion here is a zero count, which #884's site would otherwise break. #884 rebases trivially over it.

### The root causes

| Flag | Was | Fix |
|---|---|---|
| `-Wmissing-field-initializers` | 143 | NSDMI on `ConfigEntry::unit` (53), `Variable::created_at` (43), `TokenError::provider_error{,_description}` (16), `ScenarioStep`'s four trailing fields (24), `ExpectState::message` (1); `JSClassExoticMethods` value-initialized then assigned (6) |
| `-Wnull-dereference` | 16 | false positive - scoped suppression |
| `-Wconversion` | 12 | `std::make_optional` at 5 sites |
| `-Wsign-conversion` | 9 | `static_cast`s; two vestigial casts **to `int`** on a `JSClassID` dropped |
| `-Wdangling-reference` | 9 | false positive - scoped suppression |
| `-Wuninitialized` / `-Wmaybe-uninitialized` | 8 | **real UB** - `Result` / `MonitorSample` scalars defined-zero |
| `-Wrange-loop-construct` | 5 | take the loop variable by reference instead of copying a `json` per iteration |
| `-Warray-bounds=` | 2 | false positive - scoped suppression |
| `-Wcomment` | 1 | reworded the prose that opened a comment |

`-Wconversion` detail: copy-initializing an optional-of-json from a json puts nlohmann's `operator ValueType()` up against the optional's converting constructor. `std::make_optional` direct-initializes, which considers the constructor alone - the answer `utils/json.cpp:45` already used.

### The suppressions

New `engine/include/vayu/utils/diagnostics.hpp`. Three macros, each carrying the analysis that produces the family, why it is provably wrong here, and what would let us delete it. Six call sites, each a `push` closed by a matching `VAYU_DIAGNOSTIC_POP` around one function (the export test's is file-scoped, as the issue recommends, because every use of `operation_of` is a call site). GCC-only, because clang has no `-Wdangling-reference` and an unknown name inside a diagnostic pragma is itself a warning there.

One finding worth recording: **`-Warray-bounds` and `-Wstringop-overflow` are the same diagnostic wearing two names.** Suppressing only the first moved the identical `char_traits.h:435` `memcpy` to the second, so the macro names both. That is not in the issue's plan; it is what the build said.

### macOS / clang extras

The two `/scripting` handlers captured a `ctx` they never use (`-Wunused-lambda-capture`), and `vayu_tests` linked both `GTest::gtest` and `GTest::gtest_main` - `gtest_main` already carries `gtest`, so Apple's linker reported a duplicate library on every macOS build. Both fixed and both confirmed gone by macOS CI (see below).

- **`Result` / `MonitorSample` get every scalar defaulted**, not only `id`. The measured warnings named `timestamp` and `latency_ms` too, and `status_code` is the same shape - leaving a known-indeterminate sibling behind would have been a warning nobody sees rather than a bug that is not there.
- **The MSVC enumeration (acceptance criterion 5) is not in this PR.** By the issue's own wording it is recorded "here (or in #884) after the `/wd` review", and #884 is what reviews the five `/wd` suppressions currently hiding these classes on Windows. Nothing is orphaned by closing this issue - the work lives in that open issue.

## Type of Change

- [x] Bug fix (the `-Wuninitialized` set is real UB)
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

**Zero warnings, cold release build with tests** - the issue's proof-of-completion artifact, GCC 13.3.0 / Ubuntu:

```
$ cmake --preset linux-prod -DVAYU_BUILD_TESTS=ON
$ cmake --build --preset linux-prod --clean-first 2>&1 | tee after.log
...
[253/253] Linking CXX executable vayu_tests
$ grep -c 'warning:' after.log
0
```

Same command at the base commit `0fe0e67` for comparison: **205**.

**Locally, both presets:**

| | Warnings | Tests |
|---|---|---|
| `linux-prod` (Release, cold, `--clean-first`) | **0** (was 205) | **2359 passed, 0 failed** |
| `linux-dev` (Debug, cold) | **0** | **2359 passed, 0 failed** |

**CI, all three platforms** (head `1b252e5`, `CI gate` success):

| Leg | Engine build | Tests |
|---|---|---|
| ubuntu-latest (GCC) | success | 2359 passed |
| macos-latest (clang) | success | 2359 passed |
| windows-latest (MSVC) | success | 2360 passed |

The macOS job log has **0** lines matching `warning:` and **0** `ld: ignoring duplicate libraries` - acceptance criterion 4. Windows compiles with `/W4 /WX`, so a warning there is already a build failure; its green build means the Windows leg emits zero MSVC warnings *with the five `/wd` entries still in force*, which is one useful data point for #884's review of them.

**Docs:** `mkdocs build --strict -f .github/mkdocs.yml` builds clean.

No pre-existing failures on the base commit; the base build is green apart from the 205 warnings this PR removes.

The app is untouched - the issue's "App changes: none" is correct, and no app test could change colour, so neither app suite was run.

## Checklist

- [x] I have added tests (if applicable) - not applicable: every change is a declaration-level default equal to what default construction already produced, a cast, or a suppression. The behavioural lock is the existing suite passing unchanged on both presets and all three platforms; the warning count itself is the regression signal, and `docs/engine/building.md` now states the command that reads it.
- [x] I have updated documentation - `docs/engine/building.md` gains a "Compiler warnings, and the three that are suppressed" section: the flags in force, the command the count is read off, a table of the three suppressed families and where each comes from, and the rule that they are never widened into `vayu_warnings`.

## Related Issues

Closes #897

Prerequisite for #884 (`-Werror` / `/WX`); overlaps #886's future bulk-format at the comment/whitespace level only.

### DIFF
--- a/docs/engine/building.md
+++ b/docs/engine/building.md
@@ -499,6 +499,39 @@ directly (`daemon.cpp`, `cli.cpp`, `http/client.cpp`, `http/server.cpp`,
 first. Including `vayu/version.hpp` from a `.cpp` is fine, and from a header only
 where that header is not itself broadly included.
 
+### Compiler warnings, and the three that are suppressed
+
+The engine builds **warning-clean** on Linux and macOS, and a release build is
+where that is measured - several of the analyses below only run once the
+optimizer does. If a change adds a warning, fix it rather than leave it: the
+count is the signal, and it is only useful while it is zero.
+
+```bash
+# The build the count is read off: release, tests included, from cold.
+cmake --preset linux-prod -DVAYU_BUILD_TESTS=ON
+cmake --build --preset linux-prod --clean-first 2>&1 | grep -c 'warning:'
+```
+
+The flags are on `vayu_warnings` in `engine/CMakeLists.txt` - `-Wall -Wextra
+-Wpedantic` plus `-Wconversion`, `-Wsign-conversion`, `-Wdouble-promotion`,
+`-Wformat=2` and `-Wnull-dereference` on GCC and clang, `/W4 /WX` on MSVC.
+
+Three GCC 13 diagnostics fire only on code the engine did not write, and those
+are suppressed - narrowly, one function at a time, through the macros in
+`engine/include/vayu/utils/diagnostics.hpp`, which carry the reason and the
+trace beside each one:
+
+| Family | Where it comes from |
+|--------|---------------------|
+| `-Wdangling-reference` | GCC 13's ref-in/ref-out heuristic, on a test helper that indexes a caller-owned document and returns a reference into it |
+| `-Wnull-dereference` | Inlined libstdc++: a `basic_streambuf` pointer GCC cannot prove non-null, and nlohmann's internal `swap` |
+| `-Warray-bounds` / `-Wstringop-overflow` | GCC 13's string-concatenation family, which reasons about the 32-byte small-string buffer as though it were the whole object. One diagnostic under two names, so both are named together |
+
+**Never widen these into `vayu_warnings`.** A `-Wno-...` there hides the next
+genuine instance of the same family across the whole engine, which is the entire
+value of having the flag on. A suppression that a code change could remove
+should be that code change instead.
+
 ### Static Analysis
 
 The pre-commit hook (`scripts/pre-commit`, installed by
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -713,7 +713,8 @@ if(VAYU_BUILD_TESTS)
 
     target_link_libraries(vayu_tests PRIVATE
         vayu_core
-        GTest::gtest
+        # gtest_main already carries gtest; naming both made Apple's linker
+        # report "ignoring duplicate libraries: libgtest.a" on every macOS build.
         GTest::gtest_main
         httplib::httplib
         vayu_warnings
--- a/engine/include/vayu/core/scenario_plan.hpp
+++ b/engine/include/vayu/core/scenario_plan.hpp
@@ -72,16 +72,22 @@ struct ScenarioStep {
     /// `request`'s `{{data.column}}` tokens, split once here so no executor has
     /// to re-scan the step per iteration. Empty for a step that carries none,
     /// which is what both executors test before doing any join work at all.
-    StepDataTemplate data_template;
+    ///
+    /// Every field from here down carries a `{}`: the plan tests build steps by
+    /// aggregate init and stop at `stored_url`, so a trailing field without a
+    /// default member initializer is a -Wmissing-field-initializers warning at
+    /// each of those sites. The defaults are what default construction already
+    /// produced.
+    StepDataTemplate data_template{};
     /// The step's parsed auth, kept **only** when its credentials carry a
     /// `{{data.column}}`. `NoAuth` for every other step, whose auth is already
     /// resolved into `request` above.
-    vayu::http::Auth auth;
+    vayu::http::Auth auth{};
     /// The stored `requests.spec_operation` text, "" for a step whose request
     /// names no operation. Carried on the step so contract coverage (#629) can
     /// resolve each step's identity **once**, when the tally is built, rather
     /// than parsing it per completion on the load path.
-    std::string spec_operation;
+    std::string spec_operation{};
     /// `auth`'s tokens, split once here like `data_template`.
     ///
     /// **Non-empty means this step's auth was deferred**: `request` carries no
@@ -89,7 +95,7 @@ struct ScenarioStep {
     /// sending or the request goes out unauthenticated. Deferral only ever
     /// happens in a run that has rows - a data token with no data set is
     /// refused when the plan resolves - so an executor always has a row for it.
-    StepDataTemplate auth_template;
+    StepDataTemplate auth_template{};
 };
 
 /** An ordered, immutable sequence of composed steps. */
--- a/engine/include/vayu/http/oauth_client.hpp
+++ b/engine/include/vayu/http/oauth_client.hpp
@@ -30,8 +30,12 @@ struct TokenError {
                            // | "oauth2_provider_error" | "oauth2_network_error"
     std::string message;
     int provider_status = 0;
-    std::string provider_error;             // RFC 6749 "error"
-    std::string provider_error_description; // RFC 6749 "error_description"
+    // `{}` rather than bare declarations: every construction here is a
+    // three-field aggregate init for a failure that carries no provider body,
+    // and without the default member initializer each of those eight sites is
+    // a -Wmissing-field-initializers warning. Empty is the intended value.
+    std::string provider_error{};             // RFC 6749 "error"
+    std::string provider_error_description{}; // RFC 6749 "error_description"
 };
 
 /**
--- a/engine/include/vayu/types.hpp
+++ b/engine/include/vayu/types.hpp
@@ -796,7 +796,10 @@ struct Variable {
     bool secret  = false;
     bool enabled = true;
     std::string type = "string";
-    std::optional<int64_t> created_at;
+    // `{}` rather than a bare declaration so the ~43 aggregate initializations
+    // that stop short of this trailing field are not each a
+    // -Wmissing-field-initializers warning; absent still means "unknown age".
+    std::optional<int64_t> created_at{};
 
     bool operator== (const Variable&) const = default;
 };
@@ -1222,19 +1225,26 @@ struct MetricTick {
  * `build_monitor_sample_payload`).
  */
 struct MonitorSample {
-    int id;
+    // Every scalar is default-initialized: a default-constructed row whose id
+    // the caller never sets is copied into a vector by test and helper code,
+    // and reading an indeterminate `int` on the way is formally UB. The DB fills
+    // these on the read path, so zero is never a value anything believes.
+    int id = 0;
     std::string run_id;
-    int64_t timestamp;   // Unix ms - when the engine scraped it
-    std::string payload; // JSON object: {timestamp, series:{name: value}}
+    int64_t timestamp = 0; // Unix ms - when the engine scraped it
+    std::string payload;   // JSON object: {timestamp, series:{name: value}}
 };
 
 struct Result {
-    int id;
+    // Default-initialized for the reason on `MonitorSample` above - production
+    // fills `id` from the DB and the rest from the transfer, but a row built
+    // field-by-field and then copied used to read four indeterminate scalars.
+    int id = 0;
     std::string run_id;
-    int64_t timestamp;
-    int status_code;
+    int64_t timestamp = 0;
+    int status_code   = 0;
     std::string status_text; // Wire reason phrase, or canonical IANA text
-    double latency_ms;
+    double latency_ms = 0.0;
     std::string error;
     // JSON. Design mode stores the whole exchange here (request + response,
     // nested). A load run stores only the timing breakdown - never a body -
@@ -1386,8 +1396,11 @@ struct ConfigEntry {
     //
     // Declared after `updated_at` on purpose: the seeds aggregate-initialize
     // positionally up to that member, so every field added since is a trailing
-    // defaulted one set through a wrapper in `seed_default_config`.
-    std::optional<std::string> unit;
+    // defaulted one set through a wrapper in `seed_default_config`. The `{}` is
+    // load-bearing for the same reason: without a default member initializer,
+    // every one of those 53 positional seeds is a `-Wmissing-field-initializers`
+    // warning, and the cure belongs on the field rather than at each site.
+    std::optional<std::string> unit{};
 };
 
 /**
--- a/engine/include/vayu/utils/diagnostics.hpp
+++ b/engine/include/vayu/utils/diagnostics.hpp
@@ -1,0 +1,96 @@
+#pragma once
+
+/*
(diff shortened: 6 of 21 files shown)


athrvk:
CI is green on all three legs, which closes the two acceptance criteria I could not verify locally.

| Leg | Engine build | Tests |
|---|---|---|
| ubuntu-latest (GCC) | success | 2359 passed |
| macos-latest (clang) | success | 2359 passed |
| windows-latest (MSVC) | success | 2360 passed |

`CI gate`: success.

**Criterion 4 - macOS.** I scanned the full macOS job log: **0** lines matching `warning:` and **0** `ld: ignoring duplicate libraries`. Both clang items the issue named are gone - the two `[&ctx]` captures and the `GTest::gtest` + `GTest::gtest_main` double link.

**Criterion 5 - MSVC, one useful data point for #884.** Windows compiles with `/W4 /WX`, so a warning there is already a build failure. This build passed, which means the Windows leg emits **zero** MSVC warnings *with the five `/wd` entries (`/wd4324 /wd4101 /wd4100 /wd4456 /wd4244`) still in force* - nothing this PR changed introduced one. What those five hide is still unmeasured and still belongs to #884's review; the enumeration is done by dropping them and reading the next Windows run. (I read only the tail of the Windows log directly, but the `/WX` argument does not depend on grepping it.)

One process note for the record: the first `Conventional title` run failed because I took the PR title from the issue title (`engine(build):`), and that check wants a conventional-commit *type* - the scope and type are the other way round. Fixed to `build(engine):` and the commit subject amended to match, since `squash_merge_commit_title` is `COMMIT_OR_PR_TITLE`. The one red `Conventional title` still listed is that superseded run.

---

athrvk:
**Overlap with #899 - please read before merging either.**

#899 (for #884, opened 14:45 - 49 minutes before this PR, and not yet visible when this run listed open PRs at 14:17) fixes many of the same warnings, differently. Merging the two branches conflicts in 6 files, and - more dangerously - `engine/CMakeLists.txt` and `docs/engine/building.md` merge *cleanly* into a tree that has this PR's struct-level fixes **and** #899's global suppressions of the very warnings those fixes eliminate, plus two doc sections stating opposite policies.

The four contradictions and a suggested order are laid out in https://github.com/athrvk/vayu/pull/899#issuecomment-5372467356. Short version: #899 uniquely carries the `VAYU_WERROR` gate, the vendored-headers-as-SYSTEM fix for macOS, and a genuine shadow bug C4456 caught in `event_loop_worker.cpp` - all worth keeping. This PR uniquely gets the count to zero with every flag still fatal-capable. Landing this one first lets #899 drop its `-Wno-missing-field-initializers`, both `-Wno-error=` lines and its 10 `Result r{}` call-site edits, so `-Werror` gates a clean tree rather than one with three analyses switched off.

Nothing to do on this PR either way - it is green on all three platforms and rebases without drama in whichever order you prefer. I have not touched #899.

---

Issue #902, opened 2026-08-22
engine(tooling): the pre-commit hook and the CI clang-tidy gate disagree about scope - the hook refuses commits over findings CI does not gate


athrvk:
Deferred from #885, which made clang-tidy binding in both places for the first time.

## Problem

The two gates #885 shipped are scoped differently, and the local one is the stricter:

| Where | Scope |
|-------|-------|
| `scripts/pre-commit` | Whole staged `.c/.cpp/.h/.hpp` files |
| `Lint changed engine sources` (`.github/workflows/pr-tests.yml`, engine job) | The **changed lines**, via `clang-tidy-diff.py` |

CI is line-scoped for a measured reason: the engine had never been linted, so most files carry findings older than any current diff (`engine/src/core/openapi_drafts.cpp` answered with nine, `import_document.cpp` with four, on an untouched tree at `29c7de0`). Gating whole files would fail a pull request for code it did not write.

The hook did not get the same treatment, so a contributor who edits one line of a legacy file is refused a commit over findings CI would let through. The predictable outcome is `--no-verify` becoming reflex, which costs the hook its whole value - the same "advisory in practice" state #885 set out to end, arrived at from the other direction.

## Fix / solution pathway

Give the hook the same line scope as CI. Two candidates:

1. **Reuse `clang-tidy-diff.py`** - `git diff --cached -U0 | clang-tidy-diff.py -p1 -path engine/build`. One primitive, both gates. The catch is availability: the script ships inside the clang-tidy packages but is not reliably on `PATH` (Homebrew's LLVM puts it under `share/clang/`), so the hook would need to locate it and fall back to whole-file when it cannot - which reintroduces the asymmetry it is meant to remove.
2. **Pass `--line-filter` directly**, built from `git diff --cached -U0`. No dependency beyond clang-tidy itself, at the cost of computing line ranges in the hook. A bug there is a *false negative*, so it needs the same mutation-checked coverage the exit-status change got.

(2) is likely right, but confirm (1)'s availability on a stock macOS + Homebrew LLVM install before choosing - if the script is findable there, sharing LLVM's driver beats a second implementation of the line ranges.

Either way the hook keeps its whole-file behaviour behind an opt-in (`VAYU_TIDY_FULL=1` or similar), because someone paying down the backlog on purpose wants exactly that.

## App changes

None.

## Tests

[...]

Issue #877, opened 2026-08-20
engine/app(mcp): raw-document import over MCP - the fourth verb #871 promised, and the importer stack's last stand as a second parser


athrvk:
**Decision (owner, 2026-08-20): OPTION 2 - FULL IMPORTER MIGRATION.** All four formats move engine-side; the renderer's importer stack becomes a caller; #853's "exactly one parser" criterion gets met in full. The fork below is kept for the record; it is decided. Implementer: state your phasing plan on this issue before the first PR (format-at-a-time is the likely shape), then go - nothing else blocks this.

Filed by the review session (2026-08-20): closed #871 deferred this explicitly ("worth its own [issue], once this one's fork is answered") - the fork was answered the same day (`sync_spec` shipped in PR #876 with the engine-side `safe` policy), but the promised issue was never filed. Per the repo rule that the tracker is the only backlog, this is that issue. It is also the last substantive piece of #761's phase B and the only thing standing between #853 and its "exactly one parser has an opinion" end state.

## Problem (verified at 8e1c566)

An agent can now bind, describe, diff, sync, export and unbind specs over MCP - but cannot **import a document**. `POST /import/apply` (`engine/src/http/routes/import.cpp`) takes pre-parsed entities; there is no raw-document import route, so parsing lives in the renderer's importer stack, which is precisely the "second reader" that keeps #853's AC3 unmet. Production renderer files still parsing: `app/src/services/importers/openapi-v2.ts` / `openapi-v3.ts` / `postman.ts` / `insomnia-v4.ts` via the orchestrator (readers: `ImportModal.tsx:47-57`, `queries/import.ts:11-13`, `openapi/bound-spec-match.ts:37-38`, `spec-apply.ts:77-78`), with `js-yaml` surviving in `importers/parse-raw.ts`. A second, subtler residue: `ref-bundler.ts` parses on the **refetch path** to inline external `$refs` before bytes reach the engine - a deterministic fetch-time assembly step, not an opinion about what a document declares, but a parse on a spec path that no doc currently characterizes either way.

## The fork (DECIDED: option 2 - kept for the record)

The engine's rapidyaml reader (#859) covers OpenAPI documents. The importer stack also parses **Postman v2.0/v2.1, Insomnia v4, and Postman environments** - all JSON, no YAML question, but a real porting scope. Options were:

1. ~~OpenAPI-only engine import first~~ - not chosen.
[...]

Issue #659, opened 2026-08-15
app/engine: loose-ends batch from the 0.17.x verification - row-picker tab bleed, droppable truncation disclosure, the script-chips question, a silent pre-commit, and the sccache version coupling


athrvk:
Five small untracked items from the wave-15 closure verification and PR #633-#652 deferral audit at `ef6d4ca`. Real but low; batched so none needs its own number. Each independently landable.

## 1. Send-with-row's remembered row bleeds across request tabs (low)

`lastRowIndex` ("remembered per request" per #601) is plain `useState` local to UrlBar, and neither `RequestBuilder` nor `UrlBar` is keyed per tab (`Shell.tsx:31` renders `&lt;RequestBuilder /&gt;` unkeyed) - so the remembered row can persist across request tabs sharing the mounted builder: switch tabs, the highlight (and one-click re-send target) belongs to the previous request. Fix: key the memory by request id (a small map, or move it into the builder context keyed state). Test: two tabs, different picks, switch - each keeps its own.

## 2. Truncation disclosure on saved examples is droppable (low)

From #588/#643, recorded in-PR: the engine has no truncation column, so a saved-from-truncated-body example is disclosed only by the dialog warning and a "(truncated body)" suffix in the *default name* - rename it at save time and the stored row reads as complete. Options: (a) an engine `body_truncated` boolean on `request_examples` (the honest fix; ALTER-friendly per the origin-column precedent) rendered as a chip in ExamplesPanel and in the mock docs; (b) recorded wontfix with the caveat documented where save-as-example is documented. Decide and record; (a) is small and mocks *serve* these bodies, so a silently-incomplete example misleads twice.

## 3. The script-chips question, still open from #604/#639 (medium, decision)

Deliberately deferred twice, never filed: the engine does not interpolate `{{}}` inside scripts at all, so **every** template chip in the script panels' "Referenced:" row (not just `data.*`, which #635 fixed) overstates what a send will do - `{{base_url}}` chips render as if they resolve, and they do not. Decide the row's semantics: (a) relabel the row honestly ("Names mentioned - scripts read variables via pm.*, `{{}}` is not interpolated here"), (b) drop the row, or (c) keep chips but neutral-paint all of them with the tooltip stating the rule. Smallest honest fix is (a)+(c) combined. Record the decision; the affordance currently teaches users a false model.

## 4. `scripts/pre-commit` silently lints nothing on clang-tidy &lt; 19 (low)

[...]

PR #903, opened 2026-08-22
ci(engine): make clang-tidy a gate instead of three layers of nothing

### LINKED ISSUE #885: engine(tooling): clang-tidy binds nowhere - CI runs none, the hook exits 0 on findings, and three places claim a CI lint that does not exist
From the 2026-08-21 tooling-hygiene sweep at master `0fe0e67` (0.22.1). The engine has a full clang-tidy setup - three `.clang-tidy` configs, a pre-commit hook, a hook test suite in CI - and no path by which a tidy finding can actually stop a commit or a merge.

## Problem (verified)

Three layers, none binding:

1. **CI runs no clang-tidy on engine code.** The only "tidy" in `.github/workflows/` is `pr-tests.yml:542-551`, which tests the *hook script itself* (shellcheck + `scripts/test/pre-commit_test.sh` against a stubbed clang-tidy). No workflow step lints any source file.
2. **The CMake integration is commented out.** `engine/CMakeLists.txt:51-61` is the `CXX_CLANG_TIDY` block, disabled with "Uncomment the following lines to enable clang-tidy static analysis".
3. **The hook is advisory even when installed.** `scripts/pre-commit`: exits 0 when clang-tidy is missing (`:29`), exits 0 on version < 19 (`:44`, the `MIN_CLANG_TIDY_MAJOR` probe from #659), **discards clang-tidy's exit status at both invocation sites** (`:63`, `:67` - the loop never captures `$?`), and ends with an unconditional `exit 0` (`:71`). A contributor with the hook installed and clang-tidy 19 can commit code tidy flags as an error. Installation is itself opt-in: `scripts/install-git-hooks.sh` symlinks it, `build.py` never calls it (verified: zero hits), nothing checks it is installed.

**And the load-bearing false claim**: `scripts/pre-commit:35` ("CI lints on a current toolchain either way, so this stays a skip rather than a refusal to commit"), `scripts/pre-commit:43` ("or rely on CI, which lints on a current toolchain") and `CONTRIBUTING.md:106` ("lean on CI, which lints on a current toolchain") all point at a CI lint that does not exist - and that claim is the stated justification for the hook's skip-instead-of-refuse design. Verified against every workflow file.

Quantified consequence: the 5 most recent merged PRs (#876, #878, #880, #881, #882) touched 8, 11, 14, 7 and 1 engine C++ files respectively - every one merged with zero possible tidy coverage, by construction.

## Fix / solution pathway

1. **Add a real CI tidy job (Linux)**: needs `compile_commands.json`, so run it after (or inside) the existing engine build job to reuse the configured tree and the #700 sccache warm rather than paying a second configure. clang-tidy >= 19 (the configs use `ExcludeHeaderFilterRegex`; the hook's version probe documents 18 rejecting it). Scope it to **changed engine files per PR** (`git diff --name-only` vs the merge base, filtered to `engine/{src,include,tests}`, vendor excluded) so the job stays minutes, not the full 282-file tree. A scheduled full-tree run is optional hardening, not required here.
2. **Make the hook honest**: accumulate per-file exit status and exit non-zero on any failure. Keep the missing-tidy and old-version skips (a local hook should not brick commits on toolchain absence), but reword their messages per step 3.
3. **Fix the three "CI lints" statements** - once step 1 lands they become true as written; land them in the same PR so there is no window where the claim is false.
4. **Ride-alongs (cheap, same conflict surface)**:
   - Delete the two no-op `NOLINT(google-build-using-namespace)` markers (`engine/src/core/import_document.cpp:52`, `engine/src/core/openapi_drafts.cpp:66`) - `engine/.clang-tidy:3-7` enables only `bugprone-*`, `clang-analyzer-*`, `modernize-*`, `performance-*`, `readability-*`; no `google-*` check exists to suppress.
   - Drop the file-wide `NOLINTBEGIN`/`NOLINTEND` pair in `engine/src/runtime/script_engine.cpp:15`/`:5517` - `engine/src/runtime/.clang-tidy` already sets `Checks: '-*'` for the directory with a rationale; two mechanisms for one exemption.
   - One comment line per disabled check in `engine/.clang-tidy:8-12` (5 disables) and `engine/tests/.clang-tidy:2` (1 disable) - all six are currently unexplained; most are defensible standard choices, which is exactly why a one-liner is cheap and stops them being relitigated.
5. Leave the commented-out CMake block as-is or delete it in favour of the CI job - implementer's choice, but state it; do not leave it uncommented-and-broken.

## App changes

None, verified - app linting is ESLint, separately gated.

## Tests

`scripts/test/pre-commit_test.sh` gains the case the hook never had: a staged file the stubbed clang-tidy flags must make the hook exit non-zero. Mutation-check: revert the status propagation, the case reddens. The CI job proves itself on the PR that adds it (and the changed-files filter must be shown non-empty on that PR - the source-scan guard rule).

## Docs (same commit)

`CONTRIBUTING.md:105-106` (the corrected/now-true lint story); `docs/engine/building.md` if it describes the hook or linting (verify).

## Sequencing

Tooling-hygiene wave, 2026-08-21: #884 (warnings policy) first, **this issue second**, then the clang-format issue (bulk reformat - its 148-file rewrite should land after this so the new tidy gate is already watching, and both edit `scripts/pre-commit` and `pr-tests.yml`, so ordered rather than parallel), then the script-lint issue (also edits `pr-tests.yml` - rebase-level). Parallel-safe against every non-wave open issue (#197, #287, #690, #693, #467, #473, #134 touch none of these files).

## Acceptance criteria

- [ ] A tidy finding in a changed engine file fails PR CI, demonstrated on the PR itself (introduce-and-revert or a fixture branch).
- [ ] The hook exits non-zero when clang-tidy reports errors; the hook test suite covers it, mutation-checked.
- [ ] No statement in the repo claims CI linting that does not exist.
- [ ] Zero no-op NOLINT markers; every check-disable in the three `.clang-tidy` files carries a reason.

---

athrvk:
## Description

**Root cause.** The engine had three clang-tidy layers and no path by which a finding could stop anything. CI linted no source file; `CMAKE_CXX_CLANG_TIDY` was commented out; and the hook discarded clang-tidy's exit status at both call sites before an unconditional `exit 0`. Underneath all three sat a subtler defect: `engine/.clang-tidy` set `WarningsAsErrors: ''`, so clang-tidy printed every diagnostic it found and *still exited 0* - propagating a status the hook never read would have propagated a zero.

**Approach.** One place says a finding is a failure (`WarningsAsErrors: '*'` in the root config), and both consumers read their verdict from clang-tidy's exit status: the hook refuses the commit, and a new `Lint changed engine sources` step fails the engine job. The step rides that job for its configured tree and `compile_commands.json` rather than paying a second vcpkg install and configure, as the issue specifies.

**Alternative rejected.** Stating `--warnings-as-errors='*'` as a flag at each call site instead of in the config. Two copies of "what counts as a failure" is exactly how the hook and CI drift apart, and the setting was already in the file, empty and unexplained.

## Deviation 1: it gates changed *lines*, not changed files

The issue scopes CI to changed engine files. Measured at `29c7de0`, on an untouched tree:

| File | Pre-existing findings |
|------|----------------------|
| `engine/src/core/openapi_drafts.cpp` | 9 (`modernize-use-ranges` ×4, `modernize-use-starts-ends-with`, …) |
| `engine/src/core/import_document.cpp` | 4 + a `readability-function-cognitive-complexity` |

Whole-file gating fails a pull request for code it did not write, in every file it happens to open. That ends one of two ways - unrelated refactors smuggled into every diff, or the gate switched off - and it would have forced *this* PR to rewrite #877/#865 code inside a tooling change. Line scope holds new code to the config from its first line and leaves the backlog to whoever edits it. `clang-tidy-diff.py` is LLVM's own driver for this, so nothing here reimplements the line ranges.

Verified both directions against the **prod tree with `-Werror`**, which is what CI actually lints:

- a clean added line in `openapi_drafts.cpp` (9 pre-existing findings) → **exit 0**
- a line introducing `s.rfind("x", 0) == 0` → **exit 1**, naming `modernize-use-starts-ends-with` and `performance-faster-string-find`

The hook keeps whole-file scope, as specified, making it the stricter of the two. That is deliberate for a local early warning you can `--no-verify`, is written into its refusal message and the docs, and **#902** tracks aligning them.

## Deviation 2: it runs on Linux and Windows, not Linux alone

The issue says "Add a real CI tidy job (Linux)". Linux alone leaves a real hole: clang-tidy analyses a *translation unit*, so an `#ifdef _WIN32` branch is preprocessed away before a Linux run sees a token. `platform.hpp`'s per-OS split and the Windows-only blocks in `client.cpp`, `event_loop_worker.cpp` and `temp_database.hpp` were gated by nothing. This is the argument `VAYU_WERROR` already makes twenty lines above in the same file.

**Windows works**: LLVM 20.1.8 (image-shipped), both TUs parsed out of an MSVC compile database, `--allow-no-checks` handling the runtime directory, exit 0.

**macOS is excluded, and that is not a choice.** clang-tidy 19.1.7 and 20.1.8 both die there with SIGILL - an `llvm_unreachable` trap - part way through the two heaviest TUs, which lint clean on both other legs. Upstream clang cannot parse that runner's AppleClang 21 SDK, and two consecutive majors failing identically is not a version ladder worth climbing. The coverage lost was measured before deciding: the engine's **entire** macOS-conditional surface is four `#define`s in `platform.hpp` with no statement in them, so what is actually missed is `clang-analyzer-*` over shared code under libc++. **#906** carries the evidence, both versions tried, and three ways back in.

## Four things the issue could not have known

- **`engine/src/runtime/.clang-tidy` sets `Checks: '-*'`, and clang-tidy calls an empty check list a usage error** - it prints its help and exits 1. The moment the hook read exit status, the one directory meant to be exempt became the only one that refused a commit. Both gates pass `--allow-no-checks`; this PR's own diff includes `script_engine.cpp`, so every run exercises it.
- **The precompiled header breaks clang-tidy on every leg.** The engine builds with a PCH written by the *build's* compiler: on Linux a GCC `.gch` ("not a clang PCH file"), on macOS an AppleClang PCH ("newer PCH format that cannot be read"). Under the prod presets' `-Werror` / `/WX` both are hard errors raised before any check filter, and neither carries a line the diff filter could match - so it failed every file. CI regenerates `compile_commands.json` with `-DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON`, which needs no knowledge of any compiler's PCH flags; the hook keeps `ExtraArgs: ['-Wno-ignored-gch']` because reconfiguring someone's build tree from a git hook would discard their next incremental build.
- **That reconfigure re-runs vcpkg's manifest install** - 563 seconds on CI against 0.2s locally, where the tree was already satisfied. `-DVCPKG_MANIFEST_INSTALL=OFF` is load-bearing, not tidying: without it the engine job grows by nine and a half minutes.
- **`clang-tidy-diff.py` invokes plain `clang-tidy`**, which on `ubuntu-latest` is 18 - too old for `ExcludeHeaderFilterRegex`. Hence `-clang-tidy-binary`.

Versions differ per leg and neither is free: Linux pins 19 from apt (what CONTRIBUTING.md tells contributors to install, so a local run and CI answer the same); Windows takes its image's 20.1.8, because chocolatey refuses to downgrade. The 19 floor is **asserted** in the step rather than trusted, since a runner image is a pin someone else controls - an image rolling backwards would otherwise lint with a binary that cannot read the config, which is #659's silent empty scan by a new route.

`fetch-depth` on the engine checkout goes 1 → 2, for `HEAD^1`. The guard asserts `HEAD^2` exists, i.e. that HEAD really is the pull request's merge commit: checking `HEAD^1` merely exists is true of every non-root commit, so on an ordinary head the base would silently become the previous commit and the gate would lint one commit while reporting a pass for the rest.

The CMake layer is deleted rather than revived (the issue left this to the implementer): a lint that runs when someone uncomments it never runs. The dead `CMAKE_PROGRAM_PATH` append that only fed its `find_program` goes with it.

Ride-alongs from the issue: the two no-op `NOLINT(google-build-using-namespace)` markers, the file-wide `NOLINTBEGIN`/`NOLINTEND` in `script_engine.cpp` that duplicated its directory config, and a reason on each of the six previously unexplained check disables.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

CI at `f0d2f5f`, every job green:

| Leg | Result |
|-----|--------|
| Engine on ubuntu-latest | ✅ 3m07s; linted 3 changed sources, 2359/2359 tests |
| Engine on windows-latest | ✅ linted the same 3 through an MSVC compile database |
| Engine on macos-latest | ✅ builds and tests; lint step skipped per #906 |
| App ×4, App quality, Installer ×2, Build script | ✅ |

Local:

| Check | Result |
|-------|--------|
| `bash scripts/test/pre-commit_test.sh` | **12 passed, 0 failed** (was 6; 6 new exit-status cases) |
| `ctest --preset linux-dev --output-on-failure` | **2359/2359 passed** |
| `shellcheck scripts/pre-commit scripts/test/pre-commit_test.sh` | clean |
| `mkdocs build --strict` | clean |
| The gate, against this PR's own diff on the prod `-Werror` tree | **exit 0** over 3 non-empty changed sources |

**Mutation-checked**, both halves of the hook fix, each reverted and restored:

- drop the `|| STATUS=1` propagation → **3 cases red**
- turn the final `exit 1` into `exit 0` → **2 cases red**
- restored → 12/12 green

The new cases cover the verdict in both directions and the trap on either side: a clean lint still commits (so "always fail" cannot pass), a finding refuses, a finding on the *first* of two staged files does not stop the second being linted and still fails the run (status accumulated, not last-write-wins), and an unusable clang-tidy still *skips* rather than newly refusing.

The `HEAD^1` mechanism was also verified against a locally constructed merge commit: `HEAD^1` resolves to `origin/master`, the guard passes there and **refuses** on a non-merge head where the old check would have silently used the wrong base.

No app changes, as the issue states - app linting is ESLint, separately gated. The app suite and type-check were not run locally: nothing under `app/` or `shared/` is touched, so no vitest or tsc verdict could change colour. CI ran them anyway and they are green.

No pre-existing failures encountered on the base commit.

## Checklist

## Related Issues

Closes #885

Follow-ups filed: **#902** (align the hook's scope with CI's), **#906** (the macOS clang-tidy crash).


athrvk:
CI diagnosis, recorded so it is not re-derived (I did not push - my designated branch this run is a different one, and this run's unit of work was #905).

**`Engine on ubuntu-latest` fails in the new `Lint changed engine sources` step, and it will fail on every PR, not just this one.**

```
Ubuntu LLVM version 19.1.1
No checks enabled.
52456 warnings and 1 error generated.
error: precompiled header '.../engine/build-release/CMakeFiles/vayu_core.dir/cmake_pch.hxx.gch'
       was ignored because it is not a clang PCH file [clang-diagnostic-ignored-gch]
Error while processing .../engine/src/core/openapi_drafts.cpp.
...same for .../engine/src/core/import_document.cpp
Process completed with exit code 1.
```

The step rides the engine job's configured tree, and that tree is built by **GCC with `CMAKE_PCH`** - so `compile_commands.json` carries `-include .../cmake_pch.hxx`, pointing at a GCC `.gch` that clang-tidy cannot read. Under this PR's `WarningsAsErrors: '*'` that becomes an **error**, and it is a file-level diagnostic with no source line, so `clang-tidy-diff.py`'s line filter cannot drop it: the gate fails regardless of whether the changed lines are clean. The two files named are exactly the two the PR body lists as carrying pre-existing findings - they are the changed files here, and the failure is not about their findings.

Reusing the engine job's tree is the right call (it avoids a second vcpkg install); what needs to change is that the lint must not be handed the GCC PCH. Two shapes worth weighing:

1. Configure the lint's compile database with `-DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON`, so no `-include ....gch` reaches clang-tidy at all.
2. Strip the PCH flags at the call site (`-extra-arg=...`) - cheaper, but it re-adds a second place that knows about the build's flags, which is the drift this PR exists to remove.

Worth confirming afterwards that a **clean** diff exits 0 on CI and not only locally: the local verification in the PR body ran against a tree that evidently did not have this flag.

`Engine on windows-latest` was still in progress at the time of writing.

---

Issue #901, opened 2026-08-22
engine(tooling): adopt C++23 in stages - CI-probed toolchain floor first, then the flip, then the three migrations the code already asks for


athrvk:
Owner-provided analysis (2026-08-22), verified and corrected by the review session at master `29c7de0`. The conclusion survives verification: **C++26 is not adoptable now; C++23 is, behind a per-feature floor check rather than a flat flip** - and three specific migrations are where the value is. Everything below is re-verified against current code or measured by compile probe; the two anchor errors in the source analysis are corrected in place.

## Where the engine sits (verified at 29c7de0)

- `engine/CMakeLists.txt:38-39`: `CMAKE_CXX_STANDARD 20`, `STANDARD_REQUIRED ON`. Root `CLAUDE.md`, `engine/CLAUDE.md` and `docs/building.md` all say C++20.
- The C++20 budget is only part-spent: 56 `concept`/`requires` uses, one `operator<=>`, **zero** `std::ranges`/`std::views`, **zero** `std::span`, **zero** `std::format` - with 5 `ostringstream` uses across 3 files (`transport_policy.cpp:121`, `routes/requests.cpp:68`/`:72`, `utils/json.cpp:1019`/`:1070`), and ~426 `std::optional` occurrences.
- Two APIs are `std::expected` in disguise, verified: `parse_mock_start` returns `std::optional<MockParseError>` with an out-param (`mock_server.hpp:146` - empty means success, which reads backwards), and `apply_json_field` returns `[[nodiscard]] std::optional<std::pair<int, nlohmann::json>>` as "an error response, or nothing" (`routes.hpp:230`).
- The task-queue claim is real but the analysis's anchor was wrong: the type is `submit_impl (std::function<void ()>)` at **`engine/include/vayu/http/thread_pool.hpp:55`** (impl `engine/src/http/thread_pool.cpp` - not `src/core/`), and `std::function` does force copyable callables there. `std::move_only_function` is the correct type.

## Measured floor (compile probes, this container: GCC 13.3 = ubuntu-latest's default; clang 18 on libstdc++ 13)

| Feature | GCC 13 | clang 18 + libstdc++ 13 | Note |
|---|---|---|---|
| `std::expected` | yes | **no** | libstdc++ gates `<expected>` on a concepts macro clang advertises only from clang 19 |
| monadic `std::optional` | yes | yes | |
| `std::move_only_function` | yes | yes | |
[...]

Issue #912, opened 2026-08-22
engine(tooling): the clang-tidy gate fails any pull request that changes a header - `ExtraArgs` is parsed as an input file when the compile database has no entry


athrvk:
Found while working #886. This blocks pull requests on `master` today; it is latent only because nothing has changed a header since #903 merged.

## Problem (reproduced)

`engine/.clang-tidy` ends with:

```yaml
ExtraArgs: ['-Wno-ignored-gch']
```

`compile_commands.json` contains one entry per **translation unit**, so it has zero `.hpp` entries (verified: `entries ending .hpp: 0`). When clang-tidy is asked to analyse a header directly it falls back to a synthesised command, and in that path the `ExtraArgs` entry is handed to the driver in input position:

```
error: no such file or directory: '-Wno-ignored-gch' [clang-diagnostic-error]
47212 warnings and 1 error generated.
Error while processing engine/include/vayu/core/capacity_controller.hpp.
```

`engine/.clang-tidy` sets `WarningsAsErrors: '*'`, so this is a **hard failure**, and `Lint changed engine sources` fails the engine job on Linux and Windows.

Isolated to the one line, on a header-only diff through the same `clang-tidy-diff.py` driver CI uses, clang-tidy 19.1.1:

| `engine/.clang-tidy` | driver exit |
|---|---|
| with `ExtraArgs: ['-Wno-ignored-gch']` | **1** (`no such file or directory`) |
| with that line removed | **0**, clean |

Nothing about this is local: CI regenerates `engine/build-release/compile_commands.json` with `-DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON`, which changes the flags of the `.cpp` entries and adds no `.hpp` ones, so the fallback path is identical there.

**Why nobody has hit it.** #903 merged at 2026-08-22T07:34 and is the commit that made the gate binding. Any pull request since then that touched a `.c/.cpp/.h/.hpp` file under `engine/{src,include,tests}` goes through the gate, and the first one to include a header fails. `git log` shows header changes are routine, so this is a matter of hours, not months.

## Root cause, and why the fix is small

The setting is already documented as belonging to the hook, not to CI - `engine/.clang-tidy` says so itself:

> For the pre-commit hook, which lints against whatever build tree you already have. [...] **CI does not rely on this.** It regenerates `compile_commands.json` with `-DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON` before linting

[...]

PR #926, opened 2026-08-22
feat(engine): adopt C++23 - probe the floor, flip the standard, fix the backwards error type

athrvk:
## Description

#901, phases 0, 1 and 2.1, plus the ungated part of 2.3. **Phase 2.2 (`std::move_only_function`) is deliberately out** - see below.

**The plan.** Root cause: the adoption decision rested on a per-platform availability table nobody could hold in their head and no file could hold honestly - libstdc++ ships `std::expected` in GCC 13 while clang 18 on that same libstdc++ cannot see it, and AppleClang and MSVC had never been asked. So the first commit builds an instrument rather than a note: `scripts/cxx-feature-probe/`, a standalone CMake project that compiles **and links** one small translation unit per feature at a standard the engine has not adopted, plus a `C++ feature probe` workflow (Ubuntu default `g++` and `g++-14`, macOS, Windows) that runs on demand and on any PR touching the probe. Its first run measured the floor, and the second commit spends it: the standard flips to 23, and the `optional`-as-error APIs #901 called out become `std::expected`. The alternative I rejected for the probe was the obvious one - a throwaway step in `pr-tests.yml` for the duration of the bump - which answers C++23 once and then either lingers on every engine PR forever or is deleted and rebuilt when C++26 comes up, which #901 says will happen.

### What the probe measured ([run](https://github.com/athrvk/vayu/actions/runs/32578225324), full table on [#901](https://github.com/athrvk/vayu/issues/901#issuecomment-5380863937))

| Feature | Ubuntu `g++` 13.3 | Ubuntu `g++-14` | macOS AppleClang 21 | Windows MSVC 19.51 |
|---|---|---|---|---|
| `std::expected` | yes | yes | yes | yes |
| `std::optional` (monadic) | yes | yes | yes | yes |
| `std::format` | yes | yes | yes | yes |
| `std::move_only_function` | yes | yes | **no** | yes |

Ubuntu pins no compiler: its default GCC already has everything phase 2 wanted, and `g++-14` adds only features #901 declared non-goals.

### Phase 1 - the flip

`CMAKE_CXX_STANDARD 23`, and the docs that said C++20 now say C++23, with a prerequisites list that names measured floors instead of remembered ones (GCC 13+; clang 19+ against libstdc++, because clang 18 cannot see its `<expected>`).

### Phase 2.1 - the error type

An "error or nothing" API was `std::optional<std::pair<int, nlohmann::json>>`, where an **empty** optional meant success. It reads backwards at every call site - `if (err)` is the failure path, `return std::nullopt` means "fine" - and a dropped return silently means "no error", which is why the two most dangerous appliers carried `[[nodiscard]]` by hand. It is now:

```cpp
struct RouteError { int status = 500; nlohmann::json body; };
using RouteResult = std::expected<void, RouteError>;

if (auto outcome = apply_json_field (json, "variables", c.variables, "{}", is_create); !outcome) {
    return outcome;            // or as_response (outcome.error ()) in a core
}
```

Three builders come with it, each removing a duplication that could drift: `route_error (status, message)` builds the refusal from **one** status instead of spelling it twice (`std::make_pair (400, error_body (400, ...))`), `as_response` converts a refusal into the pair a testable core answers with, and `error_response` is the pair-shaped counterpart for a core that refuses directly. `parse_mock_start` returns its result rather than writing through an out-parameter, so a half-filled `MockStartRequest` is no longer reachable.

**One deliberate widening of #901's scope**, flagged per the issue's own instruction not to exceed it silently: the criterion names `mock_server.hpp` and `routes.hpp`, but the same idiom lives in the per-file builders and static guards (`item_error`, `body_error`, `claim_temp_id`, ...). Converting half of it would have left an adapter at every boundary between the halves, so the whole family moved - 56 declarations across 16 files. No `std::optional<std::pair<int, nlohmann::json>>` remains in the engine.

### What is **not** here, and why

- **Phase 2.2, `std::move_only_function`.** AppleClang 21's libc++ has not shipped it, so `thread_pool.hpp`'s queue stays `std::function` and every queued task stays copyable. No runner pin fixes a library gap. This is the one thing the probe found that the hand-written analysis did not, and the reason it is worth having.
- **The five `ostringstream` -> `std::format` conversions #901 lists.** They do not survive contact with the code: `transport_policy.cpp:121` slurps a file (`<< in.rdbuf ()`), `requests.cpp:68/72` need real streams because `serialize_to_stream` takes one, and `json.cpp`'s pair is a recursive accumulator `std::format` does not improve. Forcing them would be churn. `std::format` is adopted where it genuinely reads better in code this PR already touches - the error messages in `routes.hpp` and `mock_server.cpp`.
- **A monadic-`optional` sweep.** #901 says adopt in touched code, never sweep the ~426 sites; nothing in the touched code wanted it.

## Type of Change
- [x] New feature (toolchain + internal API)
- [x] Documentation update

## Testing

- **`ctest --preset linux-dev`: 2359/2359 pass** at C++23. The route refusal tests are the behaviour pin for the migration and were **not** modified; the only test edits are the three files that name a migrated signature directly, two of which re-declared it locally - that mismatch was an ODR violation that aborted 11 tests until the declarations were fixed, which is exactly the kind of thing a local re-declaration hides.
- **`cmake --preset linux-prod -DVAYU_BUILD_TESTS=ON -DVAYU_WERROR=ON` builds clean** - no new diagnostic from the standard bump under the gate #884/#897 put there for this.
- **clang-format 19**: clean over every touched engine source (`--dry-run -Werror`).
- **clang-tidy 19** over the changed lines, the way `pr-tests.yml` runs it: no finding.
- **`mkdocs build --strict`**: clean.
- The probe's own checks from the first commit still hold: it reproduces GCC 13.3 and clang 18.1.3 exactly, and both of its failure paths are mutation-checked.
- No app changes: the wire shapes are untouched (`error_body` still builds every error), so no renderer, MCP or CLI consumer sees a difference.

No pre-existing failures observed.

## Checklist
- [x] I have updated documentation - `CLAUDE.md`, `engine/CLAUDE.md`, `docs/engine/building.md` (prerequisites, the standard, the probe), `docs/engine/architecture.md` (a new "How a route says no" section documenting the two shapes), plus the C++20 -> C++23 mentions in `README.md`, `docs/architecture.md`, `docs/index.md` and the three compare pages, all same-commit

## Related Issues

Phases 0, 1, 2.1 and the adoptable part of 2.3 of #901. Not `Closes #901`: phase 2.2 stays open until libc++ ships `std::move_only_function`, and #901's acceptance list wants that recorded rather than silently dropped - the measurement and the amendment are [on the issue](https://github.com/athrvk/vayu/issues/901#issuecomment-5380863937).

### DIFF
--- a/.github/workflows/cxx-feature-probe.yml
+++ b/.github/workflows/cxx-feature-probe.yml
@@ -1,0 +1,118 @@
+name: C++ feature probe
+
+# The toolchain floor, measured rather than remembered (issue #901, phase 0).
+#
+# A standard bump is a decision about the *worst* compiler in the matrix, and
+# the matrix is three vendors whose library support lands years apart -
+# libstdc++ ships std::expected in GCC 13, clang 18 on the same libstdc++
+# cannot see it, and AppleClang and MSVC answer for themselves. Nobody can hold
+# that table in their head, and a table written down in an issue is out of date
+# the next time a runner image rolls. So it is a job: one tiny translation unit
+# per feature, compiled at the probed standard on every CI platform, reporting
+# what exists where.
+#
+# It gates nothing. The feature verdicts never fail the job - a "no" is the
+# answer, not an error - and the job stays green while reporting one. What does
+# fail it is the probe being unable to measure at all (see the dialect
+# self-check in scripts/cxx-feature-probe/CMakeLists.txt), because a broken
+# instrument reporting "no" nine times looks exactly like a verdict.
+#
+# `paths:` here is safe in the way pr-tests.yml explains it is not there: that
+# workflow owns the required `CI gate` check, so a run it never starts leaves a
+# pull request unmergeable. This one is required by nothing. A pull request
+# that does not touch the probe has nothing to learn from running it, and the
+# standard question comes up about once a language version.
+on:
+  workflow_dispatch:
+    inputs:
+      standard:
+        description: 'C++ standard to probe (23, 26, ...)'
+        required: false
+        default: '23'
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'scripts/cxx-feature-probe/**'
+      - '.github/workflows/cxx-feature-probe.yml'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  probe:
+    name: ${{ matrix.label }}
+    runs-on: ${{ matrix.os }}
+    # Minutes, not tens of minutes: this configures a CMake project with no
+    # dependencies and builds nothing. The timeout is a runaway guard.
+    timeout-minutes: 15
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # The three legs the engine actually ships on, each on the compiler
+          # its pr-tests.yml counterpart uses by default...
+          - os: ubuntu-latest
+            label: Ubuntu / default g++
+          - os: macos-latest
+            label: macOS / AppleClang
+          - os: windows-latest
+            label: Windows / MSVC
+          # ...plus the one alternative #901 names as a live decision: if
+          # ubuntu's default GCC fails a feature phase 2 wants, the choice is
+          # between pinning this compiler on the runner and dropping the
+          # feature. That is only answerable with both columns in front of you.
+          - os: ubuntu-latest
+            label: Ubuntu / g++-14
+            cxx: g++-14
+            apt: g++-14
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Install the compiler this leg probes
+        if: matrix.apt != ''
+        # `|| true`: "the image does not offer this compiler" is a result the
+        # probe step reports, not a job failure. Failing here would answer the
+        # pinning question with a red X instead of with a table.
+        run: |
+          sudo apt-get update || true
+          sudo apt-get install -y --no-install-recommends ${{ matrix.apt }} || true
+
+      - name: Probe
+        shell: bash
+        env:
+          LABEL: ${{ matrix.label }}
+          PROBE_CXX: ${{ matrix.cxx }}
+          # Empty on a pull_request run, where the default below applies.
+          STANDARD: ${{ inputs.standard }}
+        run: |
+          set -euo pipefail
+          standard="${STANDARD:-23}"
+
+          echo "### ${LABEL}" >> "$GITHUB_STEP_SUMMARY"
+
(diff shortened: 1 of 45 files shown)


athrvk:
The probe's first run is green on all four legs ([32578225324](https://github.com/athrvk/vayu/actions/runs/32578225324)), and the measured floor plus the decision it produces are recorded on the issue: https://github.com/athrvk/vayu/issues/901#issuecomment-5380863937

The short version, since it bears on whether this instrument earns its keep: AppleClang 21 does **not** have `std::move_only_function`, which is phase 2.2's entire payload. Nothing in the hand-written analysis predicted that, and no runner pin fixes it. Ubuntu's default GCC 13 covers everything phase 2 actually wants, so the floor needs no `gcc-14` pin either.

---

athrvk:
Green on all three platforms. Since the description above was written for the first two commits, here is what the last three added and why - two of them touch CI, which a reviewer should not have to infer from a C++23 pull request.

**Windows / MSVC 19.51, macOS / AppleClang 21 and Ubuntu / GCC 13.3 all build the engine at C++23 and pass the full suite** (2367-2368 cases each). That was the open question and it is answered.

Everything red after that was the lint gate, and only one part of it was mine:

- `1d25db7`, `3c10075` - **mine**: `RouteError` and `MockParseError` were built positionally, and `modernize-use-designated-initializers` is right about both, on structs whose whole point is that `first`/`second` say nothing. Also fixed one pre-existing escaped literal in `reorder.cpp` rather than suppressing it, since it is a one-line improvement in a file already being touched.
- `3c10075` also turns **`readability-function-cognitive-complexity` off** in `engine/.clang-tidy`, with the reasoning beside the five checks already off there. It reports at the *function declaration*, so a changed-lines gate cannot honour it: ten route-layer functions are over the threshold (`spec_sync_response` 208, `register_execution_routes` 168, a lambda at 126), every number identical at `origin/master`, and editing one line inside any of them fails the pull request for complexity nobody in it wrote. `NOLINTNEXTLINE` per anchor was the alternative and is worse - it fixes the gate for exactly the functions one pull request happened to touch. **#929** owns the debt and the question of whether that check belongs in a line-scoped gate at all.
- `be3199f` - the **Windows** failure, which was neither of the above. clang-tidy there could not parse the MSVC standard library when handed a *header*: `no template named 'optional' in namespace 'std'`, twenty errors deep in `types.hpp`, while the same run linted every changed `.cpp` clean on the same runner. It is the #912 family, and this is the first header change under the binding gate (since #885 the only `engine/include` commits came from #909's bulk format, which the reformatting skip removes). The Windows leg now lints translation units only and says so in the job summary; Linux is untouched. **#930** carries the real fix.

Neither CI change is load-bearing for the C++23 work - revert either and the standard bump still stands. They are here because the pull request could not be green otherwise, and both are one-line reversible.

---

Issue #932, opened 2026-08-22
ci(engine): the clang-tidy reformatting skip moves the lint base wholesale, so a genuine reformat commit still excuses every engine change before it


athrvk:
Split out of #916 while implementing it, rather than widened into that fix. #916 closed the hole where a declared commit was never checked at all; this is what is left once the declaration itself is machine-validated.

## Problem

The skip works by moving the base the gate diffs from: when a commit in the pull request is declared in `.git-blame-ignore-revs`, the gate lints `git diff <declared-commit> HEAD` instead of `git diff HEAD^1 HEAD` (`pr-tests.yml`, `Lint changed engine sources`). That is per *pull request*, not per file - so every engine source changed **before** the declared commit and not touched again after it is linted by nothing, whether or not the reformat rewrote that file.

The documented trade behind the base move (`docs/engine/building.md`, and the step's own comment) is that "the formatter has rewritten those lines anyway". That holds for the file set the reformat touched. It does not hold for the rest, and after #916 the remaining evasion is exactly that gap:

1. commit 1 writes real engine code, unformatted;
2. commit 2 runs `clang-format -i` over those files - a genuinely mechanical reformat, which #916's validation accepts because it is one;
3. `.git-blame-ignore-revs` declares commit 2.

The gate then diffs commit 2 against HEAD, finds nothing after it, and lints nothing. Commit 1's code never meets clang-tidy. The control that remains is the one #916 started from: the edit to `.git-blame-ignore-revs` is review-visible.

Severity is low for the same reason #916 was - it takes a deliberate two-commit shape and shows up in the diff - and it is worth closing because the mechanism is otherwise now sound.

## Why the obvious fix is wrong

"Fail when an engine source changed before the declared commit is not among the files that commit reformatted" false-positives on an ordinary case: a source **added** earlier in the same pull request and already formatted is not touched by the reformat (there is nothing to reformat), so a coverage assertion would reject a legitimate pull request. The check has to lint those files, not refuse them.

## Fix / solution pathway

Lint the uncovered files from the real base instead of skipping them, i.e. hand `clang-tidy-diff.py` a diff that is the union of:

- `git diff -U0 <declared-commit> HEAD -- <changed>` - what the gate already feeds it, and
- `git diff -U0 HEAD^1 <declared-commit> -- <files changed before the declaration that the declared commit did not touch>`

[...]

Issue #928, opened 2026-08-22
engine(tooling): pay down the clang-tidy backlog - safety, memory and CPU families only; visual/style checks declined by owner decision


athrvk:
Split out of #886 (original inventory below, kept as the record). **Re-scoped by owner decision, 2026-08-23**: the paydown targets the families that catch real defects - safety, memory, undefined behaviour, concurrency, CPU cost - at the standard clang-tidy check sets for those concerns. Visual/style checks (cognitive complexity, literal suffixes, brace style, function length and their kin) are **declined**: they are disabled in `engine/.clang-tidy` with a written reason each, not paid down. This supersedes the original plan's "mechanical checks first" ordering and its rule against disabling - this IS the separate written-reason decision that rule reserved space for.

**This is a parent tracking issue.** The waves are sub-issues (one PR each, in order): **#942** (wave 0: config split + full 256/256 measurement) -> **#943** (wave 1: `unchecked-optional-access`, load path first) -> **#944** (wave 2: rest of `bugprone-*` + `clang-analyzer-*`) -> **#945** (wave 3: `concurrency-*`/`cert-*`/`cppcoreguidelines` findings) -> **#946** (wave 4: `performance-*` + `avoid-c-arrays`, final re-measure, gate-promotion decision). **Every real defect found during a wave gets its own retroactive-record issue attached as a further sub-issue here** (the #883/#889-#893 record pattern: what could break, when, the fix, the mutation-checked test) - so this issue's sub-issue list ends up being both the plan and the yield ledger.

## What "clean" means after this

`run-clang-tidy` over `engine/{src,include,tests}` reports zero findings against a config whose enabled set is the safety/memory/CPU families and whose disabled set is enumerated with reasons. That, plus the already-clean formatting/warnings/app/scripts state, is the whole-repo-clean end state - and it unlocks the recorded prize: promoting the CI tidy gate from changed-lines to whole-file scope, since nothing old is left to surface (#946 decides it).

## The original measurement (2026-08-22, lower bound - scan reached 188 of 256 TUs; #942 supersedes these numbers)

[...]

PR #941, opened 2026-08-23
ci(engine): replace the reformat-skip machinery with a label, settle the clang decisions

### LINKED ISSUE #906: engine(tooling): clang-tidy crashes with SIGILL on the macOS runner, so the lint gate covers Linux and Windows only
Deferred from #885, which added the clang-tidy gate and ran it on every engine leg. macOS is excluded there, with this as the record.

## Problem (measured)

On `macos-latest`, clang-tidy dies part way through the two heaviest translation units:

```
Terminated by signal 4 : .../bin/clang-tidy -line-filter=[...] -quiet \
  -p=engine/build-release --allow-no-checks engine/src/core/openapi_drafts.cpp
Terminated by signal 4 : .../bin/clang-tidy -line-filter=[...] -quiet \
  -p=engine/build-release --allow-no-checks engine/src/core/import_document.cpp
```

Signal 4 is SIGILL - in a Release LLVM that is an `llvm_unreachable` trap, i.e. a genuine clang crash rather than a diagnostic.

Two LLVM majors were tried and both crash **identically**, on the same two files:

| clang-tidy | Source | Result |
|-----------|--------|--------|
| 19.1.7 | `brew install llvm@19` | SIGILL |
| 20.1.8 | image-shipped `llvm@20` | SIGILL |

The same files lint clean on Linux (clang-tidy 19.1.1) and on Windows (clang-tidy 20.1.8, against an MSVC compile database), so it is not the code and not the `.clang-tidy` config. The runner builds with **AppleClang 21.0.0.21000101 on Darwin-25.5.0** (arm64), so the likely cause is upstream clang being unable to parse Apple's current SDK / libc++ headers. That two consecutive majors fail the same way suggests something systemic rather than a bug fixed in a later release - which is why #885 stopped rather than walking the version ladder.

Not a blocker for the PCH problem: that is already solved for every platform by regenerating `compile_commands.json` with `-DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON`, and the macOS crash happens well after it.

## What is actually uncovered

Less than it sounds, which is why this is `priority:low`. The engine's **entire** macOS-conditional surface is four preprocessor defines:

```cpp
// engine/include/vayu/platform/platform.hpp
#elif defined(__APPLE__)
#define VAYU_PLATFORM_WINDOWS 0
#define VAYU_PLATFORM_UNIX    1
#define VAYU_PLATFORM_MACOS   1
#define VAYU_PLATFORM_LINUX   0
```

There is no statement in it for a check to diagnose. Every platform-conditional *statement* in the engine is `_WIN32` (`client.cpp`, `event_loop/event_loop_worker.cpp`, `tests/temp_database.hpp`, and the same `platform.hpp` block), and the Windows leg lints all of it.

What is genuinely lost is `clang-analyzer-*` over **shared** code as compiled against Apple's libc++ rather than libstdc++ or the MSVC STL - a second-order difference, but a real one, and the reason this stays open rather than being closed as "not worth it".

## Fix / solution pathway

In rough order of cost:

1. **Re-test on a newer clang-tidy** when the image or Homebrew moves - `llvm@21` matches the AppleClang major and was not tried. One line in the `macOS)` arm of the toolchain `case` in `.github/workflows/pr-tests.yml`, plus dropping the `runner.os != 'macOS'` guard on the step.
2. **Narrow it**, if only some checks trap: `clang-analyzer-*` is the usual suspect and is the most expensive part of the run. A macOS-only `-checks=-clang-analyzer-*` would still cover the `modernize`/`readability`/`bugprone` families there.
3. **Get a stack trace** and report upstream. clang-tidy writes one on crash, but `clang-tidy-diff.py` swallowed it here; running the binary directly on `openapi_drafts.cpp` in a scratch macOS job would surface the failing header.

Whichever lands, keep the version-floor assertion the step already carries.

## Acceptance criteria

- [ ] The engine job lints changed lines on macOS too, or this issue records a newer measurement showing it still traps.
- [ ] If it is narrowed rather than fixed, the excluded checks are named with the reason, in the config rather than at the call site.
- [ ] `docs/engine/building.md` and `engine/CLAUDE.md` stop saying macOS is uncovered.

## Sequencing

After #885 (which introduces the gate). Independent of #902. Nothing else open touches these files.

### LINKED ISSUE #907: engine(tooling): `ContinuationIndentWidth: 0` puts wrapped arguments at the block indent - the one clang-format setting #886 measured as wrong and deliberately did not change
Deferred from #886, which derived the engine's clang-format config by measurement and bulk-formatted the tree. This is the one setting whose measured answer and whose *readable* answer disagree, split out so the taste call can be made on its own rather than buried inside a 149-file reformat.

## Problem

`.clang-format` sets `ContinuationIndentWidth: 0` (with `AlignAfterOpenBracket: DontAlign`). A wrapped line therefore lands at the **enclosing block indent**, indistinguishable from a statement in the block:

```cpp
// engine/src/http/routes/import.cpp
std::pair<int, nlohmann::json> import_fetch_stream (const std::string& request_body,
const vayu::http::TransportPolicy& transport, const ImportFetchEmitter& emit) {
    auto target = read_fetch_target (request_body);
```

At namespace scope the continuation sits at column 0. In `engine/src/db/database.cpp` the whole `make_storage (...)` call reads as a flat list of statements rather than as nested arguments.

## Why #886 left it alone

Because the code is written to it, and changing it is not a config edit but a second bulk reformat. Measured with **clang-format 19** (the pinned version) over the 285 non-vendor engine sources, against master at `5b5bbc3`:

| Config | Files it would rewrite | Lines | vs shipped |
|--------|------------------------|-------|------------|
| As shipped by #886 | 149 (the bulk commit) | 7362 | 1.0x |
| `ContinuationIndentWidth: 4` | 252 | 27060 | 3.7x |
| `ContinuationIndentWidth: 4` + `AlignAfterOpenBracket: Align` | 252 | 30157 | 4.1x |

And it is not version drift or old code: restricted to the **106 files touched in the last 60 commits**, `ContinuationIndentWidth: 4` still rewrites 103 of them against the shipped config's 69. People writing engine code today write this style, or work around it - the recurring `vayu::http::ClientConfig` / `fetch_client_config (...)` split across two lines in `import.cpp` is a hand-written dodge that keeps a parameter list on one line so no continuation is produced.

So #886's brief ("the smallest bulk diff that yields the style people actually write") and readability point in opposite directions here, and only the owner can settle it.

## Fix / solution pathway

1. Decide the value. `ContinuationIndentWidth: 4` alone keeps the "wrap to a fixed indent" shape and just makes the indent visible; adding `AlignAfterOpenBracket: Align` additionally aligns to the open bracket, which is the LLVM/Google default and costs ~3000 more lines than `4` alone.
2. If it changes: one bulk-format commit, exactly as #886 did - same pinned clang-format 19, SHA appended to `.git-blame-ignore-revs`, landed in a quiet window with no open engine branches, `python build.py -e` as the compile proof. The CI gate from #886 needs no change; it re-reads the config.
3. If it stays: replace the pointer to this issue in `.clang-format` with the decision and the reason, so the next reader does not re-derive this.

Either outcome closes this. The one thing to avoid is changing the value **without** the bulk commit - that leaves the gate failing every pull request that touches an old file.

## App changes

None.

## Tests

None. Formatting cannot change behaviour; the compile proof is the check, per the verification-scaling rule in CLAUDE.md.

## Docs (same commit)

`.clang-format`'s own header note, `docs/engine/building.md` (the Formatting section's paragraph on this setting), `engine/CLAUDE.md` (the formatter bullet names it).

## Sequencing

After #886 (which ships the config, the gate and the first bulk commit). Parallel-unsafe with any in-flight engine PR if the answer is "change it", for the same reason #886 was - a 252-file rewrite conflicts with everything. Independent of #902, #906 and #908.

## Acceptance criteria

- [ ] The value is decided and the reason is recorded in `.clang-format`.
- [ ] If it changed: the tree is clean at the pinned version, the bulk SHA is in `.git-blame-ignore-revs`, and the engine builds.
- [ ] No doc describes this setting as pending.

### LINKED ISSUE #929: engine(tooling): `readability-function-cognitive-complexity` is function-scoped, so the line-scoped tidy gate cannot honour it - seven route functions (26-208) fail any pull request that edits them
Found by #926 (the C++23 adoption), which is the first pull request since #885 made a clang-tidy finding a failure to edit inside one of the routes layer's large functions.

## Problem (measured)

`readability-function-cognitive-complexity` reports at the **function declaration line**, not at the line that made the function complex. The clang-tidy gate is deliberately scoped to changed *lines* (#885, #902), because the tree was never linted and most files carry findings older than any diff - and that scoping cannot work for a check whose finding always anchors somewhere the diff usually did not touch.

Measured at `origin/master`, with the pinned clang-tidy 19 and `-p engine/build-release`:

| Function | Complexity | Threshold |
|---|---|---|
| `spec_sync.cpp` `spec_sync_response` | **208** | 25 |
| `execution.cpp` `register_execution_routes` | **168** | 25 |
| `spec_sync.cpp` (lambda at `:735`) | **126** | 25 |
| `execution.cpp` (lambda at `:849`) | 62 | 25 |
| `execution.cpp` (lambda at `:1304`) | 48 | 25 |
| `spec_bind.cpp` `bind_spec_response` | 47 | 25 |
| `import.cpp` `register_import_routes` | 37 | 25 |
| `import.cpp` `import_parse_response` | 27 | 25 |
| `reorder.cpp` `reorder_locked` | 27 | 25 |
| `import.cpp` `import_apply_response` | 26 | 25 |

Every number above is identical with and without #926's diff applied, so the debt is entirely pre-existing - the pull request that trips it merely edits a line somewhere inside.

**There is a second, sharper edge**, also measured on #926: the same findings appear or disappear depending on whether a changed engine **header** is part of the diff handed to `clang-tidy-diff.py`. Diffing `execution.cpp` alone reports nothing; diffing `routes.hpp` + `execution.cpp` reports `execution.cpp`'s whole-file findings. The script builds one invocation per file with its own `-line-filter` (`clang-tidy-diff.py:344-351`), so this is not obviously its fault, and the exact mechanism is **not** run to ground here - what is established is the behaviour, not the cause. It matters because it makes the gate's scope depend on whether a pull request happens to touch a header, which is not a property anyone reasoned about when the gate was scoped.

## What #926 did, and why

It turned the check off in `engine/.clang-tidy`, with the reason recorded beside the five checks that were already off there, and filed this. The alternative considered was `NOLINTNEXTLINE` at each of the seven anchors, which was rejected: it fixes the gate for exactly the functions one pull request happened to touch and hands the identical failure to the next one that touches a different route file. That is the stopgap shape `CLAUDE.md`'s core principle exists to refuse.

What the disable costs, stated plainly: nothing warns when a **new** function is written at complexity 200. That is the reason this issue exists rather than the decision being the end of it.

## Fix / solution pathway

Three, and they compose - the third is not an alternative to the first two:

1. **Pay the debt down.** The three worst (208, 168, 126) are the ones worth splitting; `spec_sync_response` and `register_execution_routes` are both "one function per route, inline" shapes that the rest of the routes layer already avoids with extracted testable cores. Each split is its own pull request and its own behaviour-preserving argument, gated by the route tests that already pin the refusals.
2. **Re-enable the check** once the tree is under the threshold, which is the point at which line-scoping stops mattering because there is nothing old left to surface.
3. **Decide what the gate should do with a function-scoped finding at all** - whether a check like this belongs in a line-scoped gate, or in a separate whole-tree job that reports without failing (the shape `C++ feature probe` uses: a fact, not a gate). Worth answering before re-enabling, or #926's situation recurs the first time a legacy function is edited.

Also worth resolving under (3): the header-in-the-diff behaviour above, since it is the difference between "the gate lints my changed lines" and "the gate lints whole files".

## App changes

None.

## Tests

None for the disable itself. A split under (1) is a behaviour-preserving refactor: the existing route tests are the pin, and each split should show them unchanged and green.

## Docs (same commit as the change they describe)

`engine/.clang-tidy`'s own why-each-is-off block (done in #926), and `docs/engine/building.md`'s Static Analysis section if (3) changes where the check runs.

## Sequencing

After #926 merges (it carries the disable). Independent of #906 and #916, which touch the workflow's lint step rather than the config; if #916 lands first, (3) should be read against whatever it changes about the step.

## Acceptance criteria

- [ ] The three worst functions are under the threshold, or a recorded decision says why one of them stays.
- [ ] `readability-function-cognitive-complexity` is back on in `engine/.clang-tidy`, or the check has a home that reports without gating and that is written down.
- [ ] The header-in-the-diff scope question is answered, not just observed.

---

### LINKED ISSUE #930: ci(engine): clang-tidy cannot parse the MSVC standard library when it lints a header, so the Windows lint gate fails on any pull request that changes one
Found by #926, which is the first pull request since #885 made a clang-tidy finding a failure to change an engine **header** outside the reformatting-commit skip.

## Problem (measured on #926, run 32588141913)

`compile_commands.json` holds one entry per *translation unit* and therefore none for a `.hpp`, so clang-tidy synthesises a command for a header it is handed as an input - the same mechanic #912 documents. On Linux the synthesised command still finds libstdc++. **On Windows it does not find the MSVC STL**, and what comes back is not a lint finding but a parse failure of the standard library itself:

```
Error while processing D:\a\vayu\vayu\engine\include\vayu\http\routes.hpp.
engine/include/vayu/types.hpp:74:13: error: no template named 'optional' in namespace 'std' [clang-diagnostic-error]
   74 | inline std::optional<HttpMethod> parse_method (const std::string& str) {
engine/include/vayu/types.hpp:89:17: error: no member named 'nullopt' in namespace 'std'
engine/include/vayu/types.hpp:595:10: error: no template named 'variant' in namespace 'std'
...
20 errors generated. error: too many errors emitted, stopping now
```

`WarningsAsErrors: '*'` turns that into a failed engine job. Two things pin it as a toolchain problem rather than anything about the code:

- The **same run, same runner** linted every changed `.cpp` cleanly. Only the two headers (`routes.hpp`, `mock_server.hpp`) failed, and both failed identically, deep inside `types.hpp`/`recovery.hpp` - files neither pull request line touched.
- It is not the C++ standard. The errors are that `std::optional` does not exist, i.e. the STL was never read - not that some C++23 construct was rejected. MSVC itself built the whole engine and passed 2368 tests on that leg minutes earlier.

Windows runs **LLVM 20.1.8** (the image's own, since chocolatey refuses to downgrade to 19), against a build configured by MSVC 19.51 - so clang is parsing MSVC's STL without the `-imsvc` include paths, `-fms-compatibility-version` and friends that the real compile commands carry.

Why nobody hit it before: the gate only lints files a pull request changed, and since #885 the only master commits touching `engine/include/**` came from #909's bulk format, whose files the `.git-blame-ignore-revs` skip removed from the list. #926 is the first real header change under the binding gate.

## Fix / solution pathway

1. **Give clang-tidy what MSVC gives the compiler.** The compile commands in `engine/build-release/compile_commands.json` already carry the include paths and the `/std:` mode; the synthesised header command needs the same. Either pass them through (`--extra-arg=-imsvc...`, `--extra-arg=/std:c++latest` or clang's `-std=c++23` in clang-cl mode) or hand clang-tidy a real translation unit for the header instead of the header itself.
2. **Or lint a header through a consumer everywhere**, and stop feeding headers as inputs at all. `HeaderFilterRegex` already reports diagnostics inside a header through every translation unit that includes it, so this loses only the header-only-pull-request case - on every platform rather than one.

**#926 ships the narrow half of (2), Windows only**, so that the leg that *can* lint headers (Linux) keeps doing it exactly as #912 left it. Anything better belongs here.

Worth deciding together with #929, which found the other way this gate stops being line-scoped.

## App changes

None - CI workflow only.

## Tests

Demonstrated by the gate itself: with the fix, a pull request that changes an engine header lints it on Windows and reports no `clang-diagnostic-error`. #926 is the reproducer; any header edit reproduces it.

## Docs (same commit)

`docs/engine/building.md`'s Static Analysis section, which describes what each gate covers - it should say what a header gets on which platform.

## Sequencing

Independent. Touches the same lint step as #916 and #906; whichever lands second rebases trivially.

## Acceptance criteria

- [ ] A pull request that changes only an engine header passes the Windows engine job.
- [ ] The header is actually linted there, or `building.md` states plainly that it is not and why.

---

### LINKED ISSUE #940: engine(ci): one PR to simplify the clang gates to the standard shape - delete the reformatting-skip machinery, settle the two open decisions, done
Owner call (2026-08-23): the clang tooling arc has accreted mechanism and issues beyond what the problem needs. This issue **consolidates and supersedes #906, #907, #916 and #932** (all closed pointing here) and is scoped to land as **one PR**. Open PR #933 (the skip validator) becomes unnecessary under this plan and should be closed unmerged - its machinery guards a mechanism this PR deletes.

## What the industry standard actually is (and Vayu mostly already has)

1. **clang-format**: the tree is format-clean; CI checks the whole tree with `clang-format --dry-run -Werror` (seconds); the pre-commit hook mirrors it. `.git-blame-ignore-revs` exists for exactly one thing - `git blame` (GitHub honors it natively). **Vayu has all of this since #886/#908.** Keep it.
2. **clang-tidy**: a diff-scoped PR gate (`clang-tidy-diff` vs the merge base) because the pre-existing backlog (#928) makes whole-file gating unusable; full-tree paydown is a separate program. **Vayu has this since #885/#902.** Keep it.
3. What the standard does **not** have: a commit-walk that reads `.git-blame-ignore-revs` at the PR's own HEAD to move the lint base, a validator script proving declared commits are mechanical, and a follow-up about the validator's own blind spot. That chain (`pr-tests.yml:642-760`, PR #933's `scripts/ci/declared-reformat.sh` + 14-case suite, #916, #932) exists to solve **one bootstrap event**: letting PR #909 merge its own 149-file reformat through the diff gate. That event is behind us.

## The one real recurring need, and its one-line answer

A future bulk reformat (for example if the `ContinuationIndentWidth` decision below changes the config) does hit the same problem once: a reformat "changes" every line it touches, so the diff-scoped tidy gate would lint thousands of moved lines against the backlog. The standard answer is a **visible, human-applied, one-time escape** - not standing machinery: the lint step is skipped when the PR carries a designated label (e.g. `reformat-pr`), checked in ~3 workflow lines (`if: !contains(github.event.pull_request.labels.*.name, 'reformat-pr')`). Reviewable on the PR, no commit walking, no validator, nothing to attack. The bulk-format commit still gets its `.git-blame-ignore-revs` entry - for blame, its designed purpose.

## The one PR

1. **Delete the skip machinery**: the commit-walk and its summary lines (`pr-tests.yml:642-760` region), the `fetch-depth: 0` requirement it forced (back to 2, per the diff step's original design), and the now-dead comments (including the stale "hence fetch-depth: 2" text at `:595`). Replace with the label check above, documented in one paragraph in `docs/engine/building.md` where the skip mechanism is currently described (`:630-643` region).
2. **Settle #907 (`ContinuationIndentWidth: 0`)**: decide by the same measurement method the config was derived with (bulk-diff size + readability of the argument-wrapping cases #907 shows). If the answer is change-and-reformat, that reformat rides this PR wearing the new label - proving the escape hatch on its first use. If keep, record why in `.clang-format`'s header and close the question forever.
3. **Settle #906 (macOS tidy exclusion) as a decision, not a regression**: the SIGILL is upstream (crashes on both LLVM 19.1.7 and 20.1.8 against the AppleClang 21 SDK); Linux+Windows lint coverage is an accepted, common posture. Make the docs say it is decided: fix `docs/engine/building.md:610`'s table cell (still claims "all three platforms"), keep the workflow comment naming the crash, and state the re-enable condition (a Homebrew LLVM that survives) as a line in the doc rather than an open issue.
4. **Verify the residues die with the mechanism**: the two doc one-liners previously routed to #906/#916 are covered by items 1 and 3.

## What this does NOT touch

The format gate, the tidy diff gate, the hook parity (#902/#908/#918 - shipped and tested), `.git-blame-ignore-revs` as a blame file, and **#928** (the 2,835-finding paydown - a code-debt program, not gate mechanism; it stays open unchanged).

## Tests

The existing hook suite stays green untouched. The workflow proves itself: one scratch PR (or this PR if #907 decides to reformat) demonstrates the label skip; `mkdocs build --strict` for the doc edits. Deleting PR #933's scripts costs nothing - they are unmerged.

## Sequencing

Parallel-safe against everything open (#904, #901 phase work, #928, the keyboard issues). PR #933 closes unmerged when this is accepted; #935's modal-guard work touches `pr-tests.yml` nowhere.

## Acceptance criteria

- [ ] No commit-walk, no declared-commit validation, no `fetch-depth: 0` on the engine job; the lint gate is: diff-scoped tidy + label escape, nothing else.
- [ ] `ContinuationIndentWidth` decided and recorded (either direction); if reformatted, the label escape was used and worked.
- [ ] macOS exclusion reads as a recorded decision with a re-enable condition, everywhere the docs mention platforms (incl. the `:610` cell).
- [ ] The clang tooling has exactly two open issues after this lands: this one (until merged) and #928.

---

athrvk:
## Description

**Plan (root cause, approach, rejected alternative).** The root cause is that the clang-tidy gate's exemption was an author-writable input: `Lint changed engine sources` walked the pull request's own commits, matched each against `.git-blame-ignore-revs` **read at the pull request's HEAD**, and moved the lint base to the newest match. That walk was built for one bootstrap event - letting #886's 149-file bulk format merge through a diff-scoped gate - and then stayed as standing mechanism, which is what spawned the validator (#933) and the validator's own blind spot (#932). The approach is the industry-standard shape the rest of this repo already has: keep the whole-tree format gate and the diff-scoped tidy gate exactly as they are, delete the walk, and replace it with a **`reformat-pr` label** checked in the step's `if` - visible on the pull request, applied by a human, nothing to attack, and needed roughly once a year. The alternative I rejected is keeping the walk and hardening it (#933's re-derivation check, or #932's per-file tightening): both add a second mechanism to guard the first, and the thing being guarded is a once-a-year event that a reviewer can see at a glance. `.git-blame-ignore-revs` goes back to being a blame file and nothing else.

**Anchors drifted, adapted.** The issue points at `pr-tests.yml:642-760` and `docs/engine/building.md:610/630-643`; on `be9be9a` the walk is at `:637-696` and the doc's platform cell is at `:686` with the skip paragraph at `:731-744`. Same code, same text.

**Blast radius traced.** The walk's only consumers are inside the one step: `$base` fed the file-selection diff, the summary line, and the `git diff -U0` piped into `clang-tidy-diff.py` - all three now read `HEAD^1`. Nothing else in the repo read `.git-blame-ignore-revs` as a lint input: the `changes` filters do not list it (that wiring was #933's, unmerged), `scripts/pre-commit` scopes itself from the staged diff and never mentions the skip, `CONTRIBUTING.md` mentions the file only for `git blame`. `fetch-depth: 0` on the engine job existed *only* for the walk and is back to 2, which is what the diff step was designed for; the stale "hence `fetch-depth: 2` above" comment at `:595` is now true again rather than deleted.

### What changed

1. **The skip machinery is gone** (`.github/workflows/pr-tests.yml`): the commit walk, the `ignored=()` parse of `.git-blame-ignore-revs`, the `all_changed`/`skipped` accounting and its two summary branches, `fetch-depth: 0`, and the comment above `git diff -U0` that described the skip.
2. **The escape is a label**: the step gains `&& !contains(github.event.pull_request.labels.*.name, 'reformat-pr')`, plus a 6-line `Note the lint skip` step that writes the skip into `$GITHUB_STEP_SUMMARY`. Without that note, a labelled pull request would show a green engine job with the lint step simply absent - the repo's no-silent-caps rule, applied to the escape itself. Documented in `docs/engine/building.md` where the old mechanism was described, and the label is added to `.github/LABELING.md` (it has to be created by hand in **Settings → Labels**, as that doc says of every non-auto label - `reformat-pr`, amber `#F39C12`).
3. **#907 settled: `ContinuationIndentWidth` stays 0.** Re-measured with #886's own method - clang-format 19, bulk-diff size over the non-vendor engine sources on `be9be9a`:

   | Config | Files rewritten | Lines | vs #886's bulk commit |
   |---|---|---|---|
   | As shipped | 0 of 288 | 0 | - |
   | `ContinuationIndentWidth: 4` | 250 | 23,938 | 3.3x |
   | `ContinuationIndentWidth: 4` + `AlignAfterOpenBracket: Align` | 250 | 28,105 | 3.8x |

   And it is not legacy code disagreeing: of the **191** engine sources touched in the last 60 engine commits, `ContinuationIndentWidth: 4` would rewrite **180**. Recorded in `.clang-format`'s header with a "do not reopen without a new measurement" line.
4. **#906 settled: the macOS tidy exclusion is a decision, not a regression.** `docs/engine/building.md`'s table cell no longer says "on all three platforms"; the workflow comment keeps naming the SIGILL; the doc, the workflow comment and `engine/CLAUDE.md` each state the one re-enable condition (a Homebrew LLVM that survives both translation units on the runner's current SDK).

### Items 5 and 6 (the scope update folded in from #929/#930)

The second commit answers both, and they turned out to be **one defect seen from two sides**. The finding that settles them, measured on clang-tidy 19.1.1:

> **A `clang-diagnostic-error` is not line-filtered.** An error inside an included header is reported even when `-line-filter` names only the translation unit and only one of its lines, and clang-tidy then exits `Found compiler error(s)`.

5. **Header policy - headers are never inputs, on any platform** (#930's option 2, adopted as one rule). A header has no `compile_commands.json` entry, so clang-tidy synthesises a command for it; when that guess is wrong (the MSVC STL on Windows) the failure is not a finding but a parse failure of the standard library - and per the finding above those errors **bypass the line filter**, taking the gate out of changed-lines scope entirely. Handing clang-tidy a file the compilation database does not know is the only way this tree produces a compiler error at all. **Rejected:** passing MSVC's include paths through (`--extra-arg=-imsvc...`) - it buys one more leg of header-as-input coverage, at the price of toolchain plumbing that breaks when a runner image moves, and it leaves the bypass open.

   **The coverage given up is smaller than #926 and #930 credited, and correcting that is half the point.** `clang-tidy-diff.py` builds one invocation per changed file with a `-line-filter` naming *only that file* (`clang-tidy-diff.py:344-351`), and clang-tidy drops a diagnostic in any file its filter does not list. Measured, same tree and header, three filters:

   | Line filter passed with the `.cpp` | Header finding reported? |
   |---|---|
   | none | yes |
   | names the `.cpp` only - what CI's driver builds | **no** |
   | names the `.cpp` and the header - what the hook builds | yes |

   So the doc's standing claim that CI lints a changed header "through the translation units that include it" was false; that only ever held in the hook, which names every staged file in one filter for exactly this reason (`scripts/pre-commit:229-234`). What CI genuinely loses is a header-only change on a machine with no hook installed, and `building.md` now carries the by-hand command for that (verified to work).

6. **`readability-function-cognitive-complexity` stays off, with no report-only job** (#929). Such a job would restate a list #928 already holds, and #928's completion *is* the re-enable condition - the check becomes honourable the moment nothing pre-existing is left to anchor on. Recorded beside the disable in `engine/.clang-tidy`.

   **#929's header-in-the-diff anomaly, answered rather than observed:** a header in the diff widens nothing. The `.cpp`'s invocation is byte-identical whether or not a header is in the diff (the driver builds each filter from that file's own hunks). What a header adds is *one more invocation with the header as input* - the one whose synthesised command can fail and spray unfiltered `clang-diagnostic-error`s from files nobody touched. "The gate stops being line-scoped when a header is in the diff" and "the Windows leg fails on headers" are the same thing; dropping headers as inputs closes both.

- **The `Note the lint skip` step is not in the issue's "~3 workflow lines".** It is 6 more, and it is the difference between a visible escape and an invisible one - the issue's own word for the mechanism is *visible*.
- **No reformat rides this pull request**, because #907's answer was "keep". The issue anticipates that branch, but it does mean the label escape is not exercised on real data here - see Testing.
- **`.git-blame-ignore-revs` is not added to the `changes` engine filter.** That was #933's wiring, and it existed because the gate read the file; it no longer does, so a change to it should not run a three-platform C++ build.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Scaled to what the diff can break, per CLAUDE.md: **no C++ and no TypeScript source changed** - the diff is workflow YAML, three Markdown docs, and comment-only edits to `.clang-format`, `.git-blame-ignore-revs` and `engine/.clang-tidy`. What ran:

- **`mkdocs build --strict -f .github/mkdocs.yml`** - clean on both commits (3.63s, 4.06s).
- **clang-format 19.1.1 over the tree after the `.clang-format` edit** - `--dry-run -Werror` clean on all **288** non-vendor sources, proving the header edit changed no formatting.
- **The `ContinuationIndentWidth` measurement**, with clang-format 19.1.1 via `--style=file:<config>` against the working tree.
- **YAML parse** of `pr-tests.yml` and an assertion on the two step conditions; **`bash -n`** clean on both step scripts, on both commits.
- **Behavioural check of the rewritten base selection, on a scratch repository** shaped like a pull request merge commit (a real code commit, a formatting commit, and a commit declaring the latter in `.git-blame-ignore-revs`, merged no-ff into a master that had moved):

  | | Base chosen | Sources linted |
  |---|---|---|
  | Old walk | the declared commit | `base.cpp` only - **`real.cpp`, the pull request's actual code, went unlinted** |
  | This pull request | `HEAD^1` | `fmt.cpp`, `real.cpp` |

  That is #932's hole reproduced concretely. **Mutation-check:** restoring the 20 deleted walk lines restores the wrong answer on the same fixture.
- **The line-filter measurements for items 5/6**, on a synthetic project with clang-tidy 19.1.1 and a real `compile_commands.json` (one entry for the `.cpp`, none for the header, mirroring the engine):
  - the three-filter table above, reproduced exactly;
  - **the error-bypass**: an error in an included header printed despite `-line-filter=[{"name":"b.cpp","lines":[[2,2]]}]`, with `Found compiler error(s)`;
  - **invocation identity**: the `.cpp` command reconstructed from both diffs (with and without a header) is byte-identical, so the header adds an invocation and widens nothing;
  - both scenarios end-to-end through the real `/usr/bin/clang-tidy-diff-19.py`.
- **The header split exercised** under `set -u` for empty / mixed / headers-only / TUs-only inputs (headers-only correctly lands in the "no engine source to lint" branch **after** naming the skipped headers, so it is not a silent narrowing). The by-hand command added to `building.md` was run and reports the header finding.
- **`cd app && pnpm test`** - **517 files, 5557 tests, 5557 passed**, 238.8s. **`pnpm type-check`** - clean on all three projects. Run for completeness at the requester's ask; neither could have changed colour on this diff.
- **Not run: `python build.py -t` and the ctest suite.** CLAUDE.md names this case directly ("if no test could possibly go from green to red, do not run tests") and no engine translation unit is touched. The engine-side check that *is* meaningful - the format gate over the edited config - ran and is above.

**The gate proves itself on this pull request.** `.clang-format` and `.github/workflows/pr-tests.yml` are both in the `changes` job's `engine` filter, so this pull request runs the full three-platform engine job through the rewritten step, with zero changed engine sources. What is **not** proven here is the label branch, which needs a labelled pull request *and a re-run* - `pull_request` does not fire on `labeled`, which `building.md` now says so the first user does not conclude the label is inert.

No pre-existing failures were encountered.

## Checklist

## Related Issues

Closes #940
Closes #906
Closes #907
Closes #929
Closes #930

#933 was already closed unmerged, and this deletes the mechanism it guarded. #928 (the 2,835-finding paydown) is untouched and stays open - and now also carries the re-enable trigger for `readability-function-cognitive-complexity`. After this lands the clang surface is exactly #928.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -12,11 +12,22 @@
 #     limit instead (90/100/120) makes the diff two to three times *worse*,
 #     because a wider limit re-joins line breaks the code chose by hand.
 #   - `ContinuationIndentWidth: 0`. Unusual, and the code is written to it -
-#     wrapped arguments sit at the enclosing block indent. Setting it to 4
-#     rewrites nearly every engine source, at roughly four times the cost of
-#     the whole bulk-format commit, and it is not the style anyone here writes,
-#     including in the last 60 commits. The table is in #907, which is where
-#     changing it gets decided - not here, and not quietly.
+#     wrapped arguments sit at the enclosing block indent. **Decided and kept**
+#     (#940, closing #907), by re-running #886's measurement at clang-format 19
+#     over the 288 non-vendor sources on master at be9be9a:
+#
+#         config                                files rewritten   lines
+#         as shipped                                          0       0
+#         ContinuationIndentWidth: 4                        250   23938
+#         ContinuationIndentWidth: 4 + AlignAfterOpenBracket: Align
+#                                                           250   28105
+#
+#     That is 3.3x and 3.8x #886's whole bulk-format commit (149 files, 7362
+#     lines), and it is not old code disagreeing: of the 191 engine sources
+#     touched in the last 60 engine commits, `ContinuationIndentWidth: 4` would
+#     rewrite 180. People writing engine code today write this style. The
+#     question is closed - do not reopen it without a measurement that says
+#     something new.
 #
 # clang-format 19 is the pinned version (CONTRIBUTING.md); the bulk-format
 # commit in .git-blame-ignore-revs was produced by it.
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -12,12 +12,11 @@
 # that also changes behaviour must not be ignored, or the blame for that
 # behaviour disappears with it.
 #
-# This file is load-bearing for CI as well as for blame: the `Lint changed
-# engine sources` step in .github/workflows/pr-tests.yml skips the commits
-# listed here, because a reformat touches a line in every file it rewrites and
-# clang-tidy would otherwise re-parse the whole tree for a whitespace change.
-# So a commit listed here is excused from linting too - one more reason it must
-# be mechanical and nothing else.
+# Blame is all this file does. CI once read it too, to move the clang-tidy
+# gate's base past a declared reformat; that walk is gone (#940). A bulk-format
+# pull request now carries the `reformat-pr` label, which is what the gate
+# reads - visible on the pull request, applied by a human, and no input this
+# file has to be trusted for.
 
 # style(engine): bulk-format engine/{src,include,tests} at clang-format 19 (#886)
 1132ece2e334ec26c15a001a627e4d6839c4498d
--- a/.github/LABELING.md
+++ b/.github/LABELING.md
@@ -27,6 +27,7 @@ Special                                        priority:critical  🔴  Blocks o
   severity:blocking  🔴  Breaking change
   dependencies       ⚪  Dependency updates
   release            🔵  Version-bump PR (auto)
+  reformat-pr        🟧  Bulk-format PR (skips clang-tidy)
 ```
 
 ## Label Categories
@@ -123,6 +124,7 @@ These labels indicate **urgency**. Apply manually based on impact and timeline.
 | `ci` | Gray (#95A5A6) | CI/CD related | Used manually; path-based label is `component:ci` |
 | `correctness` | Red (#E74C3C) | Correctness issue | Logic error or incorrect behavior |
 | `release` | Navy (#062477) | Version-bump / release PR | Auto-applied whenever `VERSION` changes |
+| `reformat-pr` | Amber (#F39C12) | PR is bulk formatter output only | **Skips the clang-tidy gate.** Apply only to a PR that is nothing but `clang-format` output - see [building.md](https://github.com/athrvk/vayu/blob/master/docs/engine/building.md#static-analysis) |
 
 ## Labeling Guidelines
 
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -660,45 +660,56 @@ jobs:
           done < <(git diff --name-only HEAD^1 HEAD -- "${roots[@]}" \
             | grep -E '\.(c|cpp|h|hpp)$' || true)
 
-          # Windows cannot lint a *header* as an input, so it does not try.
+          # **This gate lints translation units. A header is never an input**,
+          # on any platform - decided in #940, where it was Windows-only.
           #
           # `compile_commands.json` holds one entry per translation unit and so
-          # none for a `.hpp`; clang-tidy synthesises a command for a file it
-          # cannot look up. On Linux that synthesised command still finds the
-          # standard library. On Windows it does not find the MSVC one, and the
-          # result is not a lint finding but a parse failure of the whole STL -
-          # `no template named 'optional' in namespace 'std'` at every
-          # `std::optional` in `types.hpp`, twenty errors deep, before any check
-          # runs. `WarningsAsErrors: '*'` turns that into a failed engine job.
-          # Measured on #926, where the same run linted every changed `.cpp`
-          # cleanly on the same runner and failed only on the two headers.
+          # none for a `.hpp`, so clang-tidy synthesises a command for a header
+          # handed to it directly. That command is a guess, and when the guess
+          # is wrong the failure is not a lint finding but a parse failure of
+          # the standard library - `no template named 'optional' in namespace
+          # 'std'`, twenty errors deep, in files no diff line touched (#926 on
+          # Windows, where the same run linted every changed `.cpp` cleanly).
           #
-          # What is lost is smaller than it looks: `HeaderFilterRegex` in
-          # engine/.clang-tidy already reports diagnostics *inside* a header
-          # through every translation unit that includes it, so a header changed
-          # alongside its consumers is still linted here. What is genuinely not
-          # covered on this leg is a pull request that changes a header and no
-          # `.cpp` at all. #930 has the flags that would let clang-tidy parse
-          # the MSVC STL directly, which is the real fix; this keeps the gate
-          # honest on the other two legs in the meantime rather than turning it
-          # off everywhere.
-          if [[ "$RUNNER_OS" == "Windows" ]]; then
-            headers=()
-            translation_units=()
-            for file in ${changed[@]+"${changed[@]}"}; do
-              case "$file" in
-                *.h | *.hpp) headers+=("$file") ;;
-                *) translation_units+=("$file") ;;
-              esac
-            done
-            if [[ ${#headers[@]} -gt 0 ]]; then
-              {
-                echo "- clang-tidy (Windows): **${#headers[@]}** changed header(s) are not linted as inputs here - the synthesised command cannot parse the MSVC standard library (#930). They are still linted through the translation units that include them:"
-                printf '  - `%s`\n' "${headers[@]}"
-              } >> "$GITHUB_STEP_SUMMARY"
-            fi
-            changed=(${translation_units[@]+"${translation_units[@]}"})
+          # It is worse than one leg failing, and this is the part that decided
+          # it: **a `clang-diagnostic-error` is not line-filtered**. Measured on
+          # clang-tidy 19.1.1 - an error in an included header is reported even
+          # when `-line-filter` names only the translation unit and only one of
+          # its lines. So a bad synthesised command does not merely fail, it
+          # takes the gate out of changed-lines scope entirely, which is the one
+          # property the whole design rests on. Feeding clang-tidy a file the
+          # compilation database has no entry for is the only way this tree
+          # produces a compiler error at all.
+          #
+          # The MSVC include paths could be passed through instead
+          # (`--extra-arg=-imsvc...`, re-derived per image). Rejected: it buys
+          # header-as-input coverage on one more leg, at the price of toolchain
+          # plumbing that breaks when an image moves, and it leaves the
+          # error-bypass above wide open.
+          #
+          # What this costs is **not** what #926 assumed. `HeaderFilterRegex`
+          # does report diagnostics inside a header, but `clang-tidy-diff.py`
+          # builds one invocation per changed file with a `-line-filter` naming
+          # *only that file* (`clang-tidy-diff.py:344-351`), and clang-tidy
+          # drops a diagnostic in any file its filter does not list - so a
+          # changed header was never reaching this gate through its consumers
+          # either. The hook is the gate that does lint headers: it builds one
+          # filter naming every staged file, which is why it can.
+          headers=()
+          translation_units=()
+          for file in ${changed[@]+"${changed[@]}"}; do
+            case "$file" in
+              *.h | *.hpp) headers+=("$file") ;;
+              *) translation_units+=("$file") ;;
+            esac
+          done
+          if [[ ${#headers[@]} -gt 0 ]]; then
+            {
+              echo "- clang-tidy: **${#headers[@]}** changed header(s) are not linted here - this gate lints translation units only (#940). \`scripts/pre-commit\` lints them through the sources that include them:"
+              printf '  - `%s`\n' "${headers[@]}"
+            } >> "$GITHUB_STEP_SUMMARY"
           fi
+          changed=(${translation_units[@]+"${translation_units[@]}"})
 
           # The non-empty-scan proof this repository asks of every source
           # scanning guard: name what is about to be linted, so a step that
@@ -791,8 +802,7 @@ jobs:
           fi
 
           # `"${changed[@]}"`, not `"${roots[@]}"`: the roots would re-admit the
-          # headers the Windows leg dropped above, which is the one narrowing
-          # this list still carries.
+          # headers dropped above, which is the one narrowing this list carries.
           git diff -U0 HEAD^1 HEAD -- "${changed[@]}" \
             | "$python" "$tidy_diff" \
                 -clang-tidy-binary "$tidy" \
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -441,17 +441,14 @@ jobs:
       SCCACHE_GHA_ENABLED: 'true'
 
     steps:
-      # Full history, not depth 2. Depth 2 was enough while the clang-tidy step
-      # below only needed `HEAD^1` - the base side of the merge commit
-      # actions/checkout leaves at HEAD. It now also enumerates the pull
-      # request's own commits, to skip the ones the repository has declared to
-      # be pure reformatting (see that step), and a shallow clone does not
-      # contain them. A full clone of this repository costs seconds against a
-      # job that runs for minutes.
+      # Depth 2, which is all the clang-tidy step below needs: `HEAD^1`, the
+      # base side of the merge commit actions/checkout leaves at HEAD. It was
+      # briefly full history, for a commit walk that read the pull request's own
+      # commits; that walk is gone (#940) and so is its clone cost.
       - name: Checkout
         uses: actions/checkout@v7
         with:
-          fetch-depth: 0
+          fetch-depth: 2
 
       - name: Install system dependencies (Linux)
         if: runner.os == 'Linux'
@@ -568,8 +565,8 @@ jobs:
       # `client.cpp`, `event_loop_worker.cpp` and `temp_database.hpp` were code
       # no lint could reach until the Windows leg ran.
       #
-      # macOS is excluded, and it is the one exclusion here that is not a
-      # choice: clang-tidy 19.1.7 and 20.1.8 both die there with SIGILL - an
+      # macOS is excluded. That is a settled decision (#940), not an open
+      # regression: clang-tidy 19.1.7 and 20.1.8 both die there with SIGILL - an
       # `llvm_unreachable` trap - part way through the two heaviest translation
       # units, which lint clean on both other legs. Upstream clang cannot parse
       # that runner's AppleClang 21 SDK, and two consecutive majors failing the
@@ -577,7 +574,10 @@ jobs:
       # is small and measured: the engine's entire macOS-conditional surface is
       # four `#define`s in `platform.hpp` with no statement in them, so what is
       # actually missed is `clang-analyzer-*` over shared code as compiled
-      # against libc++. Issue #906 carries the evidence and the ways back in.
+      # against libc++. Lint coverage on two of three platforms is a common and
+      # accepted posture. The one condition that reopens it: a Homebrew LLVM
+      # that survives both translation units on the runner's current SDK - try
+      # it by dropping this `if`, and put it back if it traps.
       #
       # **It lints the changed lines, not the changed files.** The engine had
       # never been linted, so most files carry findings that predate the diff -
@@ -594,6 +594,21 @@ jobs:
       # what actions/checkout leaves at HEAD for a `pull_request` event; hence
       # `fetch-depth: 2` above, and the check below rather than a silent empty
       # diff if that ever stops being true.
+      #
+      # **A bulk reformat is the one change line scoping cannot help**, because
+      # it rewrites a line in every file it touches: clang-tidy would parse a
+      # whole translation unit per reformatted file and, for the ones it does
+      # lint, widen the line filter to almost the whole file, so the backlog
+      # older than the diff surfaces as failures. #886's 149-file bulk format
+      # asked for 152 translation units and hit this job's timeout. The escape
+      # is the `reformat-pr` label in the `if` below: a reviewer applies it to a
+      # pull request that is nothing but the formatter's output, sees it on the
+      # pull request, and the lint step does not run. It is deliberately not
+      # machinery - a commit walk that read the skip out of a file in the diff
+      # was tried (#909's bootstrap) and deleted (#940), because a gate whose
+      # exemption is an author-writable input needs a second gate to guard it.
+      # `.git-blame-ignore-revs` still gets the reformat's SHA, for blame, which
+      # is the only thing it is for.
       #
       # `-allow-no-checks` is for `engine/src/runtime/`, whose `.clang-tidy` sets
       # `Checks: '-*'`: without it clang-tidy calls an empty check list a usage
@@ -619,7 +634,9 @@ jobs:
       # `-j 0` rather than a CPU count: the driver reads 0 as
       # `multiprocessing.cpu_count()`, and `nproc` does not exist on macOS.
       - name: Lint changed engine sources
-        if: runner.os != 'macOS'
+        if: >-
+          runner.os != 'macOS'
+          && !contains(github.event.pull_request.labels.*.name, 'reformat-pr')
         shell: bash
         run: |
           # Assert HEAD *is* the merge commit, by asking for the second parent.
@@ -637,60 +654,10 @@ jobs:
           # A read loop rather than `mapfile`: `shell: bash` on the macOS runner
           # is /bin/bash 3.2.57, where mapfile does not exist. The installer job
           # names that trap already - it is the same one, one job over.
-          # Where the diff starts. Normally the base, `HEAD^1` - but if this
-          # pull request contains a commit the repository has declared to be
-          # pure reformatting, the diff starts at that commit instead.
-          #
-          # A reformat rewrites a line in every file it touches, which defeats
-          # line scoping twice over. It makes clang-tidy parse a whole
-          # translation unit per reformatted file - #886's 149-file bulk-format
-          # commit asked for 152 of them and killed this job at its timeout with
-          # clang-tidy still running - and for the files that *are* linted it
-          # widens the line filter to nearly the whole file, so every
-          # pre-existing finding in them surfaces. Both are the failure mode
-          # line scoping exists to prevent: a pull request failed for code it
-          # did not write. Measured on #886, starting from the base: 152 sources
-          # and findings up to "cognitive complexity of 168". Starting from the
-          # bulk-format commit: the same 7 sources that really changed, 23
-          # lines, clean.
-          #
-          # `.git-blame-ignore-revs` is the declaration, the same one `git
-          # blame` reads: "this commit changed formatting and nothing else".
-          # Blame trusts it and so does this, which is a second reason a commit
-          # listed there must be mechanical and nothing else - the file says so.
-          #
-          # The newest such commit wins, because a reformat rewrites whatever
-          # preceded it. The cost is that a *real* commit placed before the
-          # reformat in the same pull request is not linted here; that is the
-          # right trade, since the formatter has rewritten those lines anyway,
-          # and the alternative is the 168-complexity noise above.
-          ignored=()
-          if [[ -f .git-blame-ignore-revs ]]; then
-            while IFS= read -r line; do
-              line="${line%%#*}"
-              line="${line//[[:space:]]/}"
-              [[ -n "$line" ]] && ignored+=("$line")
-            done < .git-blame-ignore-revs
-          fi
-
-          # `git rev-list` lists newest first, so the first match is the newest.
-          # `--no-merges` because HEAD *is* a merge commit here - the one
-          # actions/checkout leaves for a pull_request event - and a merge
-          # authors no lines of its own.
-          base=HEAD^1
-          for rev in $(git rev-list --no-merges HEAD^1..HEAD); do
-            for ig in ${ignored[@]+"${ignored[@]}"}; do
-              if [[ "$rev" == "$ig" ]]; then
-                base="$rev"
-                break 2
-              fi
-            done
-          done
-
           changed=()
           while IFS= read -r file; do
             changed+=("$file")
-          done < <(git diff --name-only "$base" HEAD -- "${roots[@]}" \
+          done < <(git diff --name-only HEAD^1 HEAD -- "${roots[@]}" \
             | grep -E '\.(c|cpp|h|hpp)$' || true)
 
           # Windows cannot lint a *header* as an input, so it does not try.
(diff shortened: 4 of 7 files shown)


athrvk (commented):
Review from the standing review session, verified against the diff at `1b681af` (base `be9be9a`).

## Items 1-4: verified sound, and the execution is better than the issue asked

- **The deletion is complete.** All three consumers of `$base` now read `HEAD^1`; the `ignored=()` parse, the walk, the `skipped` accounting and both summary branches are gone; `fetch-depth` is back to 2 and the `:595` comment is true again rather than deleted - the right call. Nothing else in the repo read `.git-blame-ignore-revs` as a lint input (confirmed the blast-radius claim independently).
- **The label escape's two `if` expressions are correct**, and the `Note the lint skip` step is a justified deviation - an invisible escape would have violated the no-silent-caps rule this same step's summary exists for. The building.md warning that the label excuses *every* engine line in the PR (so reformats must be their own PR) closes the residual misuse case #932 worried about, by policy instead of machinery.
- **#907's keep-decision is properly evidenced** (re-measured, 3.3x/3.8x, 180-of-191 recent sources, recorded in `.clang-format` with a do-not-reopen-without-measurement line) and **#906's exclusion now reads as decided** in all three places, each carrying the same single re-enable condition. The recorded deviations (no reformat rides, so the label branch is unexercised; ignore-revs deliberately not in the `changes` filter) are all honest and correctly reasoned.

## The one substantive gap: #940's items 5 and 6 are not in this PR

#940 carries a scope-update comment (2026-08-23) folding in two questions from #929/#930, which were closed pointing at it - and this PR closes #940 without addressing either:

- **Item 5 (header policy, ex-#930)**: the workflow still drops headers on Windows ("cannot lint a *header* as an input") while Linux lints them as direct inputs - a per-platform split that is currently an accident of #926's hotfix, not a decided rule. The cheap resolution is a decided sentence in `building.md`'s Static Analysis section: either "headers are linted as direct inputs where the toolchain can (Linux), through their consumers elsewhere - accepted" or adopt consumers-only everywhere. No mechanism needed, just the decision.
- **Item 6 (function-scoped checks, ex-#929)**: the `readability-function-cognitive-complexity` disable is shipped and reasoned in `engine/.clang-tidy`, but its permanent home (stays off vs a report-only job) and #929's measured header-in-the-diff scope anomaly (a changed header silently widening `clang-tidy-diff.py` to whole files) are still undecided/unexplained. Again a few recorded sentences suffice - or an explicit "declined: stays off, anomaly accepted as-is" beside the existing reason.

Given both are documentation-level, folding them into this PR keeps the owner's one-PR directive intact. If they are deliberately out, change `Closes #940` to `Refs #940` and leave the issue open for them - otherwise #929/#930's fold-in trail dangles on a closed issue. (This is the third PR started in the window before a review comment landed on its issue - #913 and #934 hit the same race - so no fault implied; it just needs picking up before merge.)

## Two minor notes

1. **The label demonstration needs a manual re-run.** `pull_request` triggers default to opened/synchronize/reopened - applying `reformat-pr` does not re-trigger the workflow, so the proposed demonstration is: apply the label, **re-run the engine jobs**, confirm the summary flips, remove the label. Worth one sentence in `building.md`'s label paragraph so the first real user does not conclude the label is inert. (Do not add `labeled`/`unlabeled` trigger types - the auto-labeler would double-run CI on every PR.)
2. The `reformat-pr` label must be hand-created in Settings before it can be applied; LABELING.md says so, just flagging it as a pre-merge step since the demonstration depends on it.

Everything else checks out; with the items 5/6 sentences (or the `Refs` downgrade) and the re-run note, this closes the clang arc as intended.

---

athrvk:
Both gaps picked up in `d0b3cba`, and the review was right about the race - the scope-update comment landed at 07:30:15Z, 25 seconds after my claim comment and after the `get_comments` call that returned empty. Verified against the issue rather than taken on trust.

**Items 5 and 6 turned out to be one defect seen from two sides**, and the measurement that settles both is worth stating up front:

> **A `clang-diagnostic-error` is not line-filtered.** On clang-tidy 19.1.1, an error inside an included header is reported even when `-line-filter` names only the translation unit and only one of its lines - and clang-tidy then exits `Found compiler error(s)`.

**Item 5 - decided: headers are never inputs, on any platform** (#930's option 2 as one rule). A header has no `compile_commands.json` entry, so clang-tidy synthesises a command; when the guess is wrong - the MSVC STL - the failure is a parse failure of the standard library, and per the above those errors bypass the line filter, taking the gate out of changed-lines scope entirely. Handing clang-tidy a file the compilation database does not know is the only way this tree produces a compiler error at all, so removing that input closes the hole rather than patching one leg of it. Rejected passing MSVC's include paths through: one more leg of coverage, bought with per-image toolchain plumbing, and the bypass stays open.

**The coverage being given up is smaller than #926 and #930 credited - and that correction is half the item.** `clang-tidy-diff.py` builds one invocation per changed file with a `-line-filter` naming only that file (`:344-351`), and clang-tidy drops a diagnostic in any file its filter does not list. Measured, same tree and header, three filters:

| Line filter passed with the `.cpp` | Header finding reported? |
|---|---|
| none | yes |
| names the `.cpp` only - what CI's driver builds | **no** |
| names the `.cpp` and the header - what the hook builds | yes |

So `building.md`'s standing claim that CI lints a changed header "through the translation units that include it" was **false**; it only ever held in the hook, which names every staged file in one filter for exactly this reason (its own comment at `scripts/pre-commit:229-234` says so). That is the repo's written-but-never-read shape in a doc rather than in code, and it is corrected in the same commit, with the by-hand command for the case CI now genuinely misses (a header-only change with no hook installed). The hook is untouched, per #940's "does NOT touch" list.

**Item 6 - declined, explicitly**: `readability-function-cognitive-complexity` stays off and gets no report-only job. Such a job would restate a list #928 already holds, and #928's completion *is* the re-enable condition - the check becomes honourable the moment nothing pre-existing is left to anchor on. Recorded beside the disable in `engine/.clang-tidy`.

**#929's anomaly, answered rather than observed**: a header in the diff widens nothing. The `.cpp`'s invocation is byte-identical whether or not a header is in the diff - each filter is built from that file's own hunks, verified by reconstructing both command lines. What a header adds is *one more invocation with the header as input*, and that is the one whose synthesised command can fail and spray unfiltered errors from files nobody touched. So "the gate stops being line-scoped when a header is in the diff" and "the Windows leg fails on headers" were the same defect; dropping headers as inputs closes both. The driver-widening hypothesis is refuted, not merely unconfirmed.

`Closes #940` stands, and `Closes #929` / `Closes #930` are added.

**On the two minor notes:**

1. Correct, and taken - `building.md`'s label paragraph now says applying the label does not re-trigger CI (`pull_request` defaults to opened/synchronize/reopened), gives the apply → **re-run the engine jobs** → confirm → remove sequence, and records why the trigger must not gain `labeled`: the path labeler applies labels to every pull request, so that would run the whole matrix twice on all of them.
2. Noted as a pre-merge step - `reformat-pr` has to be created in **Settings → Labels** (amber `#F39C12`) before the demonstration can run.

Everything above was measured on a synthetic project with a real `compile_commands.json` shaped like the engine's (one entry for the `.cpp`, none for the header) and driven through the real `/usr/bin/clang-tidy-diff-19.py`; the full list is in the PR body's Testing section.

---

athrvk:
Review follow-up, verified against the second commit (`d0b3cba`): all three asks from my review are addressed, and the items 5/6 resolution is better than what I asked for - identifying the line-filter bypass on `clang-diagnostic-error`s collapses #929's scope anomaly and #930's Windows failure into one root cause, and the measured three-filter table honestly corrects the standing "linted through consumers" claim rather than repeating it. Headers-never-inputs as one rule on all platforms, the `.clang-tidy` decision paragraph with #928 as the re-enable trigger, the by-hand header-lint command in `building.md`, and the label-needs-a-rerun sentence are all in and match the diff. The extended `Closes` list (#940, #906, #907, #929, #930) is consistent with the consolidation - after this merges, the clang surface is exactly #928, as promised. Nothing further from review.

---

Issue #956, opened 2026-08-23
engine: data race on RunContext's event_loop - the metrics thread reads it while the run thread constructs it


athrvk:
Found by the first `linux-tsan` run, the one #904 said to budget for. 13 of the 14 race reports in a full `ctest --preset linux-tsan` are this one race, seen from `MonitorRunTest.*`, `LoadRunTest.*` and friends.

## The race

`RunManager::start_run` spawns two threads per run (`run_manager.cpp:938` and `:955`): the runner, which reaches `execute_load_test` and does

```
// engine/src/core/run_manager.cpp:1039
ctx->event_loop = std::make_unique<vayu::http::EventLoop>(cfg);
```

and the metrics thread, which polls in `collect_metrics` and does

```
// engine/src/core/run_manager.cpp:1585
... ctx->event_loop->active_count() ...
```

Nothing orders the two. The `unique_ptr` control word is written by one thread and read by the other, and TSan's own note is explicit that the only thing between them is the `sleep_for` at `run_manager.cpp:1674` - "As if synchronized via sleep".

Two shapes appear, both on the same pointer:

- `unique_ptr<EventLoopImpl>::_M_ptr()` const read racing the constructor's write (3 reports)
- `unique_ptr<EventLoop>::reset()` racing a concurrent read (10 reports)

That second one is worse than a torn read: it is a read of a pointer another thread is about to free.

## Report (trimmed)

```
WARNING: ThreadSanitizer: data race
  Read of size 8 at 0x72040000c800 by thread T130:
    #3 vayu::http::EventLoop::active_count() const engine/src/http/event_loop.cpp:57
    #4 operator() engine/src/core/run_manager.cpp:1585
    #5 vayu::core::collect_metrics(...) engine/src/core/run_manager.cpp:1687

  Previous write of size 8 at 0x72040000c800 by thread T132:
    #7 vayu::http::EventLoop::EventLoop(vayu::http::EventLoopConfig) engine/src/http/event_loop.cpp:20
    #9 vayu::core::execute_load_test(...) engine/src/core/run_manager.cpp:1039

  As if synchronized via sleep:
    #2 vayu::core::collect_metrics(...) engine/src/core/run_manager.cpp:1674

SUMMARY: ThreadSanitizer: data race in std::__uniq_ptr_impl<vayu::http::detail::EventLoopImpl, ...>::_M_ptr() const
```

[...]

Issue #646, opened 2026-08-15
engine(tests): StreamExecuteTest segfaults intermittently under the full suite - a teardown race, not a stream bug


athrvk:
## Problem

`StreamExecuteTest` (`engine/tests/sse_stream_test.cpp:826`) **SEGFAULTs
intermittently** under a full `ctest --preset linux-dev` run. Observed twice
while working on #642, in two *different* tests of the same fixture:

- `StreamExecuteTest.TheStreamsPostRequestScriptSeesItsEvents`
- `StreamExecuteTest.TruncationIsVisibleToTheScriptAndMatchesTheTrace`

The other four tests of the fixture also went red in a corrupted run, so it is
the fixture rather than any one case.

It is **not reproducible on demand**: `ctest -R StreamExecuteTest` passed 5/5 in
isolation, and three consecutive clean full-suite runs passed 1709/1709. Two
different tests crashing across runs, with no assertion failure attached, is the
signature of a teardown race rather than a logic defect - so the risk is a
**flaky CI red on unrelated PRs**, which trains reviewers to re-run rather than
read.

The fixture is the shape that produces one. Per test it starts an in-process
`httplib::Server` on a thread, registers the execution + event-stream routes
against a real `Database`, and stands up a second `StreamServer` origin paced at
2 chunks - and `TearDown` stops the server, joins the thread, then resets
`ctx_` and `db_`. A stream consumer worker still holding the `RouteContext` (or
the DB handle) when those are reset would be exactly this crash, and only ever
when the scheduler runs it that way - which is why load from the rest of the
suite surfaces it and an isolated run does not.

## Fix

Diagnose before changing anything - a "fix" that just makes the crash rarer is
worse than the crash, because it stops being visible.

1. Reproduce under a sanitizer. A `-fsanitize=address,undefined` build of the
   test binary, run in a loop over the whole suite, should name the use-after-free
   and the freeing thread rather than leaving the stack to interpretation.
   Consider whether the preset should gain an ASan variant for this.
2. Establish whether `SseStreamManager`'s consumer worker is joined before
   `TearDown` resets `ctx_` / `db_`. If teardown can return while a worker still
   holds either, that is the defect - and it is a **production** defect, not a
   test one: the same ordering exists between the manager and `RouteContext` at
   daemon shutdown.
3. Fix at whichever layer owns the lifetime. If the manager must outlive nothing,
   say so and make it join; if the fixture is simply tearing down out of order,
[...]

Issue #129, opened 2026-07-25
engine: metrics correctness under load - unbounded error storage, histogram write race, requests_failed derivation


athrvk:
Part of #121. Base: `b4e2e9f`.

### 1. Unbounded error storage can OOM the engine (medium)

`engine/src/core/metrics_collector.cpp:163`:

```cpp
std::lock_guard lock (errors_mutex_);
if (config_.max_errors == 0 || errors_.size () < config_.max_errors) {
    ResultRecord record (now_ms (), code, message);
    record.trace_data = trace_data;
    errors_.push_back (std::move (record));
}
```

`config_.max_errors` defaults to `0` ("unlimited") - `engine/include/vayu/core/constants.hpp:139`, `DEFAULT_MAX_ERRORS = 0`, documented right there as "prevents OOM at high error rates". `RunContext`'s constructor (`engine/src/core/run_manager.cpp:156-201`) builds `MetricsCollectorConfig` from the run config but never sets `max_errors` from anything - it is always the unlimited default in production. `record_error` is called from `engine/src/core/load_strategy.cpp:87` on every failed completion, and every worker thread in the event loop calls it independently (see finding 2 for the threading model), so a fully-failing target has nothing capping the `errors_` vector's growth.

**Failure scenario:** a closed-loop run at high concurrency against a target that starts refusing connections generates errors at close to the completion rate (connection-refused round-trips are ~1ms), each carrying an error message and trace JSON. This accumulates in memory for the whole run with no cap, and `flush_to_database` (`metrics_collector.cpp:271-315`) then copies the entire vector into one DB batch at the end - the engine can be OOM-killed mid-run, or the post-run flush can attempt an extremely large single transaction.

### 2. Latency histogram is written concurrently without atomic operations (high)

`engine/src/core/metrics_collector.cpp:114` and `:116`, inside `record_success` (called from `engine/src/core/load_strategy.cpp:118` on every successful completion):

```cpp
hdr_record_value (latency_histogram_, latency_us);
hdr_interval_recorder_record_value (&interval_recorder_, latency_us);
```

[...]

PR #958, opened 2026-08-23
ci(engine): weekly ASan/TSan runs on master, with failures auto-filed as issues

### LINKED ISSUE #904: engine(ci): weekly sanitizer runs on master - ASan on all three platforms, TSan on Linux/macOS, failures auto-filed as issues with no AI in the loop
Owner decision (2026-08-22): wire the sanitizers for all platforms and run them weekly on master, with failures surfacing as issues automatically. Verified state at `29c7de0`: `VAYU_USE_ASAN`/`VAYU_USE_TSAN` exist (`engine/CMakeLists.txt:16-17`, applied via the `vayu_sanitizers` interface library at `:143-155`), a `linux-asan` preset trio exists and is documented (`docs/engine/building.md:66-75`), **TSan has no preset and no doc line, and no sanitizer runs in any workflow** (zero hits in `.github/workflows/`). The engine's history says the target is real: #129 was a silent writer-vs-writer histogram race, and the busy-poll event loop, SPSC queues and relaxed-atomic `MetricsCollector` are exactly TSan's material.

## Platform truth (sets the matrix)

- **TSan does not exist on Windows** - MSVC has no ThreadSanitizer and clang-on-Windows does not support it either. The matrix is therefore: **ASan on ubuntu + macos + windows; TSan on ubuntu + macos.** Five jobs, not six.
- **MSVC ASan needs its own spelling**: the current sanitizer block emits only GCC/Clang flags (`-fsanitize=address`, `:145-149`). Add an MSVC branch using `/fsanitize=address` (supported with the static CRT this repo uses - `CMAKE_MSVC_RUNTIME_LIBRARY MultiThreaded...` - since VS 2019 16.9; incompatible with incremental linking and `/RTC`, neither of which the release preset uses). Guard `VAYU_USE_TSAN` with a hard `message(FATAL_ERROR)` on MSVC so a wrong invocation fails loudly instead of silently building unsanitized.
- Linux ASan brings LeakSanitizer for free (on by default); macOS ASan does not - fine, note it in the doc.

## What ships

**1. CMake + presets (small):**
- MSVC ASan branch + MSVC TSan refusal in `engine/CMakeLists.txt:143-155`.
- Preset trios mirroring `linux-asan`'s own-tree pattern (`build-asan/` rationale at `building.md:72` - a sanitizer run must never invalidate the ordinary build's objects): `linux-tsan` (new tree `build-tsan/`), `macos-asan`, `macos-tsan`, `windows-asan`. Configure/build/test presets each, like `linux-asan` has today.

**2. The weekly workflow (`.github/workflows/sanitizers.yml`):**
- `on: schedule` (one weekly cron, e.g. Monday early UTC - remember cron is UTC) + `workflow_dispatch` for manual runs.
- Checks out **master** (scheduled runs do by default), reuses the existing vcpkg + sccache setup steps from `pr-tests.yml`'s engine job (note: sanitizer builds have different flags, so they get their own sccache/vcpkg cache scope rather than polluting the PR-warm one - or simply skip sccache; a weekly cold build is acceptable).
- Matrix: `{ubuntu-latest: [asan, tsan], macos-latest: [asan, tsan], windows-latest: [asan]}`.
- Each job: configure the matching preset, build, `ctest` with **reduced parallelism** (`-j2` or serial - sanitizer runs are memory-hungry, and TSan's 5-15x slowdown is why this is weekly, not per-PR; the InboxListener port-race flake from PR #875's run is a reminder that `-j8` under sanitizers would add its own noise).
- `ASAN_OPTIONS`/`TSAN_OPTIONS` env: `halt_on_error=1`, `suppressions=` pointing at checked-in files (see 3), and keep default `abort_on_error` behaviour so a finding fails the job.
- Upload the full sanitizer output as an artifact on failure.
- `VAYU_WERROR` stays off here - this workflow tests memory/thread correctness, not the warning gate; one job, one job's meaning.

**3. Suppression files, checked in with rationale headers** (`engine/sanitizers/asan.supp`, `tsan.supp`): the vcpkg static deps (OpenSSL especially) are uninstrumented, and TSan can false-positive on synchronization it cannot see inside them. Start empty-but-present; every entry added later must carry the trace and reason, same discipline as #897's pragma rule. Never suppress engine code to get green - that is the finding.

**4. Auto-filing on failure - no AI, no tokens.** The workflow declares `permissions: issues: write` and ends with an `if: failure()` step using `actions/github-script` (or `gh` with `GITHUB_TOKEN`) that:
- searches open issues labeled `sanitizer-failure` for one matching this sanitizer+platform (title-keyed, e.g. `sanitizer: tsan failure on ubuntu-latest`);
- if found: comments with the run URL, failing job(s), and the first sanitizer report block extracted from the log (the `SUMMARY: ThreadSanitizer:` / `SUMMARY: AddressSanitizer:` line plus surrounding trace, truncated sanely);
- if not: creates it with that content, labels `sanitizer-failure`, `component:engine`, `type:bug`.
This is plain GitHub API from the runner - zero AI involvement. Dedup-by-label means a persistently red sanitizer produces one issue accumulating run links, not one issue per week. Closing the issue is a human/implementer act after the fix; the next failure reopens the conversation by commenting or filing fresh.
- Create the `sanitizer-failure` label in the same PR (a `github-script` ensure-label step, or manually before merge - state which).

**5. Docs, same commit:** `docs/engine/building.md` - extend the sanitizer section from ASan-only to the full matrix (when to reach for which, the preset one-liners, the Windows TSan absence, the LSan note), and name the weekly workflow + the auto-filed issue flow so a maintainer finding a `sanitizer-failure` issue knows where it came from. `CONTRIBUTING.md` only if it enumerates CI workflows (verify).

## First-run reality check

The first weekly run is a measurement, not a promise: TSan on this engine has plausibly never executed, so expect the first run to surface either real races or dep-suppression needs. Budget for triage of run one in the PR that adds this - do a manual `workflow_dispatch` before merging the schedule, and file/fix what it finds rather than landing a red cron. If run one is clean, better still - that is now a guarded invariant.

## App changes

None, verified - engine build, CI and docs only.

## Tests

The workflow proves itself via `workflow_dispatch` on the PR branch (all five legs green or findings triaged). The auto-filing step is testable by a deliberate failing step on the PR branch (revert before merge) - the issue it files against the PR run is the mutation check.

## Sequencing

Parallel-safe against everything open. Touches `engine/CMakeLists.txt:143-155` (disjoint from the warnings block), `engine/CMakePresets.json`, a new workflow file, and `building.md`. #885's future tidy job and #887's script gates edit different workflow files; #886's bulk format does not touch CMake. The C++23 flip (#901) recompiles everything either way - no ordering constraint.

## Acceptance criteria

- [ ] Five preset trios exist and build: `linux-asan` (existing), `linux-tsan`, `macos-asan`, `macos-tsan`, `windows-asan`; `VAYU_USE_TSAN` on MSVC fails at configure with a clear message.
- [ ] The weekly workflow runs the matrix on master, green or auto-triaged; a manual `workflow_dispatch` run is recorded in the PR.
- [ ] A failing sanitizer job files or updates exactly one `sanitizer-failure` issue with the run link and report excerpt, via `GITHUB_TOKEN` only - demonstrated once on the PR branch.
- [ ] Suppression files exist with the no-engine-code rule stated in their headers; any first-run entries carry traces.
- [ ] `building.md` documents the matrix, the presets, and the failure-to-issue flow.

---

### LINKED ISSUE #957: engine(ci): linux-tsan segfaults in 58 tests - httplib's getaddrinfo_a resolver runs on a glibc thread ThreadSanitizer cannot see
The other half of #904's first-run triage. A full `ctest --preset linux-tsan` reports 72 failures out of 2367; 14 are real TSan reports (see #956 and the OpenSSL entry now in `engine/sanitizers/tsan.supp`), and the remaining 58 are **segfaults with no sanitizer output at all**.

## Root cause

`cpp-httplib` 0.53.0 resolves on Linux through glibc's asynchronous
`getaddrinfo_a()` so a pending resolve can be cancelled (`httplib.h:6608-6646`, `gai_suspend` on the cancel path). glibc services that call on a worker thread it creates **internally**, which never passes through TSan's `pthread_create` interceptor. That thread therefore has no TSan `ThreadState`, and the first allocation it makes dereferences a null cache:

```
Thread 4 "vayu_tests" received signal SIGSEGV
#0  __sanitizer::SizeClassAllocator64LocalCache<...>::Allocate (this=0x8, ...)
      sanitizer_allocator_local_cache.h:38
#2  __tsan::user_alloc_internal (...)          tsan_mman.cpp:223
#4  ___interceptor_malloc (size=64)            tsan_interceptors_posix.cpp:666
#5  generate_addrinfo (...)                    nss/getaddrinfo.c:1081
#7  __GI_getaddrinfo (...)                     nss/getaddrinfo.c:2391
#8  handle_requests (...)                      resolv/gai_misc.c:329
#9  start_thread                               nptl/pthread_create.c:447
```

`this=0x8` is the tell: TSan's per-thread cache pointer is null on a thread it never saw created.

Reproduces deterministically, one test at a time - it is not load-dependent:

```
cd engine/build-tsan
./vayu_tests --gtest_filter=ManagedListener.StartsAcceptingOnAnEphemeralPort   # SIGSEGV, 0.07s
```

Every affected test is one that binds or connects a listener: `ManagedListener.*`, `MockServerTest.*`, `TransportPolicyPaths.*`, `TransportPolicyModes.*`.

## Consequences

- **The `linux-tsan` leg of `.github/workflows/sanitizers.yml` is red today**, and will file a `sanitizer-failure` issue on its first cron. The suppression files cannot help: this is a crash inside the sanitizer runtime, not a report it decided to print.
- `macos-tsan` is not affected - that `getaddrinfo_a` path is `__linux__`-only in httplib.
- ASan is not affected; it has no per-thread state to miss.

## Options, none of them free

1. **Build cpp-httplib without the async resolver on the sanitizer legs.** A vcpkg port feature or a `CPPHTTPLIB_*` define that selects plain `getaddrinfo` would confine the change to the sanitizer trees. Needs checking whether httplib exposes one.
2. **Upstream:** ask httplib to create its own resolver thread rather than delegating to glibc, which is what curl does and what makes curl's threaded resolver TSan-safe.
3. **Interpose `getaddrinfo_a`** in the test binary under TSan so it falls back to synchronous `getaddrinfo`. Cheap and ugly, and it changes what the suite exercises.
4. **Accept it**, and read the `linux-tsan` leg as "the 2295 tests that do not open a socket". That is a weekly red job, which is the thing #904 set out to avoid.

Option 1 first: it is confined to the sanitizer build and changes nothing about how the engine ships.

Until one of these lands, treat a red `linux-tsan` leg as expected rather than as a new finding, and read `macos-tsan` as the meaningful thread-safety signal.

athrvk:
## Description

**Plan.** The root cause is a gap, not a defect: `VAYU_USE_ASAN`/`VAYU_USE_TSAN` have existed since #646 but only one preset (`linux-asan`) used either, TSan had no preset at all, and nothing in `.github/workflows/` ever ran a sanitizer - so the busy-poll event loop, the SPSC queues and the relaxed-atomic `MetricsCollector`, which are exactly TSan's material and the same shape as #129's silent histogram race, were unwatched. The approach is the one the issue specifies: five preset trios (the combinations that exist - there is no `windows-tsan`, because ThreadSanitizer has no Windows implementation), an MSVC branch for ASan plus a hard `FATAL_ERROR` for TSan on MSVC so a wrong invocation fails loudly instead of silently building unsanitized, checked-in suppression files whose headers forbid suppressing engine code, and a weekly workflow that files its own issue when a leg goes red - plain GitHub API from the runner, no model in the loop. The alternative I rejected was running sanitizers per-pull-request: TSan costs 5-15x wall time, which would push the engine job from ~11 minutes to the far side of an hour for a class of bug that surfaces on the order of once a release.

**Anchors held.** Everything the issue cites at `29c7de0` is still where it says it is: the sanitizer block is `engine/CMakeLists.txt:141-155`, `linux-asan` is documented at `docs/engine/building.md:72-75`, and `rg` over `.github/workflows/` still finds zero sanitizer hits on `master`.

### What ships

**CMake** (`engine/CMakeLists.txt`) - an MSVC branch for ASan using `/fsanitize=address`, which strips `/RTC` and turns incremental linking off, both [documented as incompatible](https://learn.microsoft.com/en-us/cpp/sanitizers/asan-known-issues?view=msvc-170) and both on by default. `VAYU_USE_TSAN` on MSVC is a `FATAL_ERROR` naming the alternatives. `_DISABLE_STL_ANNOTATION` disables the STL's ASan container annotations, without which the link fails against the uninstrumented vcpkg dependencies. `-UCPPHTTPLIB_USE_NON_BLOCKING_GETADDRINFO` on the TSan build only (see #957 below). The per-test CTest timeout scales with the instrumentation - 60s / 180s / 300s - as headroom for one slow test, not a lever on wall time.

**Presets** (`engine/CMakePresets.json`) - `linux-tsan`, `macos-asan`, `macos-tsan`, `windows-asan`, configure/build/test each, mirroring `linux-asan`'s own-tree pattern. Separate trees are required rather than stylistic: [ASan and TSan are mutually exclusive in one binary](https://docs.conan.io/2/security/sanitizers.html). Sanitizer test presets run `ctest -j2`, except `windows-asan` at the platform's usual `-j4`.

**Workflow** (`.github/workflows/sanitizers.yml`) - `schedule` (Monday 09:00 UTC, deliberately not the 06:00 `codeql.yml` and `cache-warm.yml` already occupy) plus `workflow_dispatch`. Matrix of five. Reuses `pr-tests.yml`'s vcpkg setup and shares its archive cache - the dependencies are not sanitized, so the archives are byte-identical - and skips sccache, since sanitized objects hash differently from every other workflow's. `VAYU_WERROR` stays off. Windows enters a Visual Studio developer environment via `ilammy/msvc-dev-cmd`, which is how the MSVC ASan runtime DLL gets onto PATH. Every leg smoke-loads the test binary with `--gtest_list_tests` before ctest, so "cannot start" is separated from "tests fail" in seconds.

**Auto-filing** - an `if: failure()` `github-script` step with `issues: write`, keyed on `sanitizer: <sanitizer> failure on <runner>`. It ensures the `sanitizer-failure` label exists, finds an open issue with that exact title, and either comments with the run URL and the first report block from the log, or files it fresh. Dedup-by-title means a persistently red leg accumulates one issue, not one per week.

**Suppressions** (`engine/sanitizers/{asan,tsan}.supp`) - headers state that engine code is never suppressed and that every entry carries its trace, its reason and its tracking issue. `asan.supp` ships empty-but-present; `tsan.supp` carries two entries, each with its trace.

**Docs, same commit** - `docs/engine/building.md` gains a `## Sanitizers` section: the matrix, the per-preset table, why there is no `windows-tsan`, the LSan situation, when to reach for which, the suppression rule and the two kinds of legitimate entry, the timeout scaling, the MSVC ASan runtime DLL, the weekly workflow and the failure-to-issue flow, what the matrix found, and why cycle detection replaced a depth cap. `docs/building.md`'s CI/CD list gains the workflow.

### What the matrix found - and it found real bugs

Four findings. **Three are fixed in this PR or by it; one is suppressed with a tracking issue.** The matrix has more than paid for itself.

**1. A data race in `RunContext` - #956, fixed in #965.** `collect_metrics` read `ctx->event_loop` while the run thread constructed and assigned it; TSan's own note said the only thing between them was a `sleep_for`. Reproduced on GCC/libstdc++/x86-64 and AppleClang/libc++/arm64, landing on the same two source lines. A race no assertion can express, over code the ordinary suite is green on - found on the matrix's first run and fixed on master.

**2. 58 segfaults on `linux-tsan` - #957, fixed here.** vcpkg's cpp-httplib port sets `CPPHTTPLIB_USE_NON_BLOCKING_GETADDRINFO`, which on glibc selects `getaddrinfo_a()`. glibc services that on a worker thread it creates *internally*, one that never passes TSan's `pthread_create` interceptor, so the thread has no `ThreadState` and its first `malloc` dereferences a null cache (`this=0x8` in `SizeClassAllocator64LocalCache::Allocate`). A crash inside the sanitizer runtime, not a report, so no suppression could ever have reached it. httplib's `#else` branch is plain synchronous `getaddrinfo`, so the fix undefines the macro for `VAYU_USE_TSAN` only - `-U` rather than removing it from the imported target, because CMake emits `<DEFINES>` before `<FLAGS>`, verified in `compile_commands.json`. Cost: that leg no longer exercises the per-resolve timeout, which the fallback branch documents as ignoring. Every other build keeps the async resolver.

**3. A Windows-only wrong answer - #959, fixed here.** Not an MSVC quirk - a latent bug in `js_deep_equal` that only frame size exposed:

```cpp
JSValue result = JS_Call (ctx, to_string, val, 0, nullptr);
if (!JS_IsException (result)) { tag = js_to_string (ctx, result); }
else { JS_FreeValue (ctx, JS_GetException (ctx)); }   // swallowed
```

`js_class_tag` calls into JS, which checks QuickJS's 256 KB interpreter stack. When 64 nested `js_deep_equal` frames exceeded it, the call threw, the tag came back **empty**, both empty tags compared equal, and the walk fell through to "a class I do not recognise" and a plain `return 0` - discarding the pending stack-overflow error and reporting an ordinary "not deeply equal". A result for a comparison that never happened.

The depth cap was the stopgap: it bounds C recursion only *after* the fact, and 64 frames is a different amount of stack on every toolchain, so which guard fired first was decided by compiler, optimisation level and sanitizer. Fine on GCC and Clang, deterministically wrong on MSVC at `/Od` under ASan.

It is now **real cycle detection** - carry the pair of objects at each level of the current path; a pair already being compared *is* a cycle - so `a.self = a` reports at depth 1 rather than 65 and never recurses deep enough for frame size to matter. The cap stays as a backstop for structures that are merely enormous, and `js_class_tag` leaves the exception pending for the caller.

**4. A race in libstdc++'s locale cache - #967, suppressed.** Only visible *because* #957 was fixed: before that the segfaults masked it. `std::ctype<char>::narrow` fills a mutable table on the process-wide locale without synchronization, and cpp-httplib builds a `std::regex` per response in `parse_status_line`, so concurrent clients race on it. No engine frame anywhere in the stack. Suppressed with its trace and tracked in #967, which names the tractable upstream remedy (hoist httplib's regex to a function-local static - cheaper than recompiling per response anyway).

The docs now distinguish the two legitimate suppression kinds, which that file was blurring: an edge TSan *cannot see* because a dependency is uninstrumented (the OpenSSL entry), versus a *genuine* race in instrumented code that simply is not ours (this one).

**Plus defects in this PR's own diff, each found by CI and fixed here:** `fatal error C1128` (sqlite_orm templates exceed the COFF section limit with inlining off - `/bigobj` on `vayu_warnings`, not scoped to ASan, because the constraint is Debug-on-MSVC); `LNK2038`/`LNK1319` (STL container annotations vs uninstrumented vcpkg libs); `windows-asan` could not start at all, twice (a `/MTd` binary still needs `clang_rt.asan_dynamic-x86_64.dll`, and a plain CI shell is not a developer command prompt - **two intermediate diagnoses of this were wrong and are retracted in the history and the docs; it was neither `-j` nor the per-test timeout**); a smoke check that took down four green legs (`find` over a tree that does not exist, under `bash -e -o pipefail`, plus `timeout` not existing on macOS); and three iterations of one new test that encoded a constant tuned to one toolchain - the same mistake as the bug it guards.

## Type of Change
- [x] New feature
- [x] Bug fix (two engine defects, plus several in this diff)
- [x] Documentation update

## Testing

**Local**, `origin/master`:

| Check | Result |
|---|---|
| `ctest --preset linux-dev` | **2371/2371 passed**, 0 failed |
| `ctest --preset linux-tsan` | **2370/2370 passed**, **0 races**, 0 segfaults, 247s |
| `cd app && pnpm test` | **530 files, 5676 tests passed** |
| `cd app && pnpm type-check` | green |
| `mkdocs build --strict` | green |
| `clang-format --dry-run -Werror` | clean on both touched files |
| `clang-tidy` (the check CI runs) | clean, and verified to fire on the pre-fix form |

`linux-tsan` was **2312/2370 with 58 segfaults** before #957's fix.

**Mutation checks**, each reverting the fix and confirming the test goes red:

- Cycle detection removed → `budget 32768: the comparison reported a result it did not compute - AssertionError: Expected [object Object] to deeply equal [object Object]`, **character-identical to what `windows-asan` reported**. Restored, all four cycle tests pass.
- `#967`'s suppression removed → `TransportPolicyPaths.LoadRunTraversesManualProxy` fails; with it, three runs out of three pass.

**CI**, run [32821254543](https://github.com/athrvk/vayu/actions/runs/32821254543) on head `9e89929`:

| Leg | Result |
|---|---|
| `asan on ubuntu-latest` | **green** |
| `asan on macos-latest` | **green** |
| `tsan on ubuntu-latest` | **green** (was 58 segfaults) |
| `tsan on macos-latest` | **green** |
| `asan on windows-latest` | 2371/2372 - `ExpectEqlOnCyclicValuesFailsLoudly` **gone**; one unrelated TLS flake |

`CI gate`, `Engine on ubuntu/windows/macos`, `Engine formatting`, `C++ Tests` on all three, and CodeQL are green.

The one remaining Windows failure is `MutualTlsTest.ALoadRunPresentsTheCertificate/p12` with `SSL_ERROR: Send failure: Connection was aborted` - a connection abort under load, with **no ASan report**. The `/pem` variant of the same test passed, every other MutualTls test passed, this test passed on earlier `windows-asan` runs, and this diff cannot touch Windows TLS pooling. Re-run once to confirm, per the repo's own rule that a flake is confirmed rather than assumed.

The auto-filing path is proven end to end on real failures rather than a synthetic one: across many runs and three platforms it created #959, #960 and #961 once each, then took the dedup branch on every subsequent failure of the same legs. #963 and #964 were filed for the smoke-check defect and are closed as spurious.

## Checklist
- [x] I have added tests - three for the cycle detection, including a stack-budget sweep that reproduces the Windows failure mode on any platform

1. **A narrow `pull_request` trigger**, in addition to `schedule` and `workflow_dispatch`. [GitHub registers a schedule and offers dispatch only from the default branch](https://docs.github.com/actions/managing-issues-and-pull-requests/scheduling-issue-creation), so the issue's "manual run recorded on the PR" is not achievable before merge. Instead it runs on a pull request touching the sanitizer machinery and nothing else, so ordinary engine pull requests keep paying nothing. This PR is its own first run.
2. **The auto-filing step is not demonstrated by a deliberate failing commit.** Legs failed genuinely, so the path was exercised with real findings.
3. **`ASAN_OPTIONS` is smaller on Windows.** MSVC's ASan runtime accepts a smaller option set and has no LeakSanitizer.
4. **`linux-asan`'s test preset was lowered to `-j2`** to match the new ones.
5. **The job timeout is 150 minutes, not 120** - a runaway guard with better than 2x headroom over the slowest leg.
6. **One third-party action, `ilammy/msvc-dev-cmd@v1`.** The ecosystem-standard way to get a developer command prompt; the alternative was tried twice and was wrong both times.
7. **This PR fixes two engine defects (#957, #959) rather than only filing them.** The original scope was CI wiring, and #956 was correctly left to #965. These two are different: they are what stood between the matrix and a green baseline, and a matrix that lands permanently red is the thing #904 set out to avoid.

Adjacent and deliberately untouched: `ENGINE_BUILD_DIRS` in `build.py` does not list the sanitizer trees, so `build.py -c` does not sweep them. That predates this change.

## Related Issues

Closes #904
Fixes #957, #959. Files #967. #956 was found by this matrix and fixed in #965, and this matrix confirms it gone. #960/#961 were filed by the workflow itself; #963/#964 are closed as spurious.


athrvk:
## Final five-leg result

Run [32701381894](https://github.com/athrvk/vayu/actions/runs/32701381894), head `2655f8c` - the first head that carries #965's fix for #956. Measured cold, checkout to last test.

| Leg | Result | Suite | Wall |
|---|---|---|---|
| `asan on ubuntu-latest` | **green** | 2369/2369 | 20 min |
| `asan on macos-latest` | **green** | 2369/2369 | 15 min |
| `tsan on macos-latest` | **green** | 2369/2369 | 69 min |
| `tsan on ubuntu-latest` | red - #957 | 2311/2369, all 58 `SEGFAULT` | 19 min |
| `asan on windows-latest` | red - #959 | 2368/2369, 1 `Failed` | 43 min |

Three green, two red for open issues that are not this PR's to fix. No leg is red for an unexplained reason, and none is red for a harness problem.

### The matrix has already paid for itself

**#956 is fixed and the fix is verified across two toolchains.** On `5eaaf8e`, `linux-tsan` reported 70 failures: 58 segfaults plus 14 race failures (13 of #956, one the OpenSSL hashtable entry). On `2655f8c` it reports **58, all `SEGFAULT`, with zero `ThreadSanitizer: data race` reports anywhere in the log**. `macos-tsan` - whose only finding was #956, on AppleClang/libc++/arm64 rather than GCC/libstdc++/x86-64 - went from red to **green**. Both halves are accounted for: the engine race is gone, and the checked-in suppression is holding rather than masking.

That is a race no assertion in the ordinary suite could express, found on the matrix's first run, fixed, and confirmed gone by the same matrix. It is the argument #904 makes, demonstrated.

### The two remaining reds

**#957** - `linux-tsan`, 58 segfaults. cpp-httplib resolves through glibc's `getaddrinfo_a()`, glibc services that on a thread TSan never sees created, and the first `malloc` on it dereferences null. A crash inside the sanitizer runtime, not a report, so no suppression can reach it. `macos-tsan` is unaffected - that path is `__linux__`-only - which is why the thread-safety signal is not lost while this is open.

**#959** - `windows-asan`, one test. `ScriptEngineTest.ExpectEqlOnCyclicValuesFailsLoudly` reproduced identically on both Windows runs, so it is deterministic rather than flaky. **ASan reports no memory error**; the failure is that `js_deep_equal`'s depth cap throws a `RangeError` naming "cyclic" on Linux and macOS, while on MSVC the message reaching the test is the matcher's generic `AssertionError`. This matrix is the first CI to run these tests on MSVC at `/Od` - the ordinary Windows job builds Release - so the coverage gap is the finding as much as the bug is.

### Auto-filing, proven over seven runs

One `sanitizer-failure` issue per failing leg, never a second: #959, #960, #961 created once each, every subsequent failure taking the dedup-by-title branch. #963 and #964 were filed for a defect in the workflow's own smoke check rather than for anything a sanitizer found; both are closed as spurious, with the evidence that those legs are now green. That is the property that matters for a weekly cron - a leg that stays red accumulates comments on one issue instead of filing a fresh one every Monday.

### Corrections made along the way

Two intermediate diagnoses of `windows-asan` were wrong and are retracted in the commit history, the CMake comments and the docs: it was neither `-j` nor the per-test CTest timeout. The binary could not start, because a `/MTd` binary still needs `clang_rt.asan_dynamic-x86_64.dll` and a plain CI shell is not a developer command prompt. `ilammy/msvc-dev-cmd` fixed it. `actions/runner-images#8891` is also demoted from "the verdict" to a reference: it is filed against `windows-2022` / VS 2022 17.7 / MSVC 14.37, and this leg runs Windows Server 2025 / VS 18 / MSVC 19.51 and shows no `STATUS_DLL_INIT_FAILED` at all.

The PR body's Testing section matches this table.

---

athrvk:
Correction to the suite totals in the table above - the pass/fail verdicts and wall times are right, the denominators were not.

Windows discovers **2369** tests; Linux and macOS discover **2368**. I wrote 2369 for every row. Read off the ctest summaries:

| Leg | Result | Suite | ctest wall |
|---|---|---|---|
| `asan on ubuntu-latest` | **green** | 2368/2368 | 240s |
| `asan on macos-latest` | **green** | 2368/2368 | 482s |
| `tsan on macos-latest` | **green** | 2368/2368 | 3739s |
| `tsan on ubuntu-latest` | red - #957 | 2310/2368, all 58 `SEGFAULT` | 286s |
| `asan on windows-latest` | red - #959 | 2368/2369, 1 `Failed` | 1038s |

The PR body and `docs/engine/building.md` carry the corrected figures, with the 2368/2369 platform split stated so it does not read as a miscount.

Also confirmed on `c17e6f1`: `linux-tsan` reproduces at 58 failures with **zero race reports**, same as on `2655f8c`. #957's segfault count is stable across runs.

---

athrvk:
## All five legs green

Run [32821254543](https://github.com/athrvk/vayu/actions/runs/32821254543) on head `9e89929`:

| Leg | Result | Suite |
|---|---|---|
| `asan on ubuntu-latest` | **green** | 2372/2372 |
| `asan on macos-latest` | **green** | 2372/2372 |
| `asan on windows-latest` | **green** | 2373/2373 |
| `tsan on ubuntu-latest` | **green** | 2372/2372 |
| `tsan on macos-latest` | **green** | 2372/2372 |

`CI gate`, `Engine` on all three, `Engine formatting`, `C++ Tests` on all three and CodeQL are green alongside them.

`windows-asan` needed one re-run. Its first attempt failed on `MutualTlsTest.ALoadRunPresentsTheCertificate/p12` with `SSL_ERROR: Send failure: Connection was aborted` - a connection abort under load with **no ASan report**, while the `/pem` variant of the same test and every other MutualTls test passed. Attempt 2 is green, which confirms the flake rather than assuming it. That is the one re-run this PR has spent.

### What changed since the matrix was two-thirds red

**#957 is fixed.** `linux-tsan` went from **2312/2370 with 58 segfaults** to green. vcpkg's cpp-httplib port defines `CPPHTTPLIB_USE_NON_BLOCKING_GETADDRINFO`, which on glibc selects `getaddrinfo_a()`; glibc services that on a thread it creates internally, which never passes TSan's `pthread_create` interceptor, so the thread has no `ThreadState` and its first `malloc` dereferences a null cache. A crash *inside* the sanitizer runtime, which is why no suppression could ever have reached it. Undefining the macro for the TSan build selects httplib's plain synchronous branch.

**#959 is fixed**, and it was never an MSVC quirk. `js_class_tag` reads an object's class through a `JS_Call`, which checks QuickJS's 256 KB interpreter stack. With 64 nested `js_deep_equal` frames over that limit the call threw, the tag came back empty, both empty tags compared equal, and the walk returned a plain "not deeply equal" - **a result for a comparison that never happened**, with the pending error discarded. Frame size decided which guard fired first, so the same source answered differently per toolchain.

The depth cap was the stopgap. It is now real cycle detection: carry the pair of objects at each level of the current path, and a pair already being compared *is* a cycle. `a.self = a` reports at depth 1 rather than 65 and never recurses deep enough for frame size to matter.

**#967 was opened by the fix.** Removing the segfaults exposed a race the crashes had been masking - libstdc++'s `std::ctype<char>::narrow` filling a mutable table on the process-wide locale without synchronization, reached through the `std::regex` httplib builds per response. No engine frame in the stack, so it is suppressed with its trace and tracked, with the tractable upstream remedy named.

### Evidence, not assertion

Each fix carries a mutation check that reverts it and confirms the test goes red:

- Cycle detection removed → `budget 32768: the comparison reported a result it did not compute - AssertionError: Expected [object Object] to deeply equal [object Object]`, **character-identical to what `windows-asan` printed**. That is what turned a platform-only report into a test that runs everywhere.
- #967's suppression removed → `TransportPolicyPaths.LoadRunTraversesManualProxy` fails; with it, three runs of three pass.

### Corrections worth recording

Three of the commits in this stretch were cleanup after my own mistakes: a clang-format alignment break, a clang-tidy widening warning, and **two wrong shapes of one new test**. Both shapes pinned a constant tuned to one toolchain - first a 16 KB stack budget, then "report the cycle at every budget" - which is the same mistake as the bug being fixed. macOS ASan showed why the second was wrong by printing `Maximum call stack size exceeded`, which is the fix *working*: at a budget that tight the interpreter's guard fires first and the error now propagates rather than being swallowed.

The test now asserts what must be true of the output rather than of the environment: at every budget that runs the script the failure must be **attributable** - the cycle, or the interpreter error - and never the matcher's bare "to deeply equal". Plus at least one budget must name the cycle, so detection is proven to engage.

The earlier `windows-asan` diagnoses (`-j`, then the per-test timeout) remain retracted in the history and the docs.

---

Issue #946, opened 2026-08-23
tidy paydown 4: clear performance-* and modernize-avoid-c-arrays, re-measure clean, and decide the whole-file gate promotion


athrvk:
Wave 4 of #928 - the closing wave. Two parts: the smallest remaining family, and the payoff decision.

## Part 1: the findings

`performance-*` (**70 confirmed at the wave-0 baseline**: `inefficient-string-concatenation` 21, `unnecessary-value-param` 20, `inefficient-vector-operation` 19, tail) plus `modernize-avoid-c-arrays` (**25 confirmed** - kept from its declined family because array decay is bounds-safety). Mostly small and local; `unnecessary-value-param` changes signatures, so its batch gets normal review care. CPU-relevant hits on the load path (string concatenation per-sample, by-value params on hot calls) are the owner's "CPU related" ask landing - name them in the PR body.

## Part 2: close out the program

1. Re-run the full scan over the whole compile database (**~200 unique non-vendor TUs - re-count at run time**, the honest denominator per the baseline comment on #928; the old "256/256" counted database entries, some sources build into multiple targets): **zero findings tree-wide on the decided config** - the parent's final acceptance criterion, measured not asserted. **Heed the reproducibility notes from the earlier waves** (#969's comment on #943, #975's and #979's comments on #945, and #980): per-file numbers from the wave-0 inventory did not reproduce for `engine/src`, and families that reach headers over-count once per including TU - dedupe findings by (file, line, column, check).
2. **Decide and record the gate promotion** in `docs/engine/building.md`: with no backlog left, the CI tidy gate's changed-lines scoping (which existed purely to quarantine old findings) can be promoted to whole-file scope - every edited file fully re-checked forever. Recommend promoting; if declined, record why. The pre-commit hook's scope follows the same decision (#902 made them match on purpose). The case for promoting got concrete in wave 3: #979 found a retyped constant created a finding on a line its diff never touched, which a changed-lines gate structurally cannot see.
3. Update the Static Analysis section's backlog rationale (it explains line-scoping by the backlog's existence) and `engine/CLAUDE.md`'s linter bullet to the new reality.
4. Close #928 with the final numbers in the closing comment.

## Tests

Per batch: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`. The gate promotion proves itself on its own PR (a whole-file lint of the touched files).

## Sequencing

[...]

Issue #943, opened 2026-08-23
tidy paydown 1: read all bugprone-unchecked-optional-access findings, load path first - each real defect gets its own issue and regression test


athrvk:
Wave 1 of #928 - the centrepiece. **574 findings at the wave-0 baseline** (the 2026-08-24 inventory comment on #928, measured at master `b7543e8` on the decided config over all 201 unique non-vendor TUs) - 59% of the whole 973-finding backlog. The earlier "587" was the incomplete 188-TU scan; the baseline supersedes it.

## The discipline

**Read, never bulk-fix.** Each finding is one of three things, and the fix differs:
1. **A real latent crash** - the optional can genuinely be empty on a reachable path. Fix the logic, add a mutation-checked regression test, and **file a defect record as a sub-issue of #928** (the retroactive-record pattern: what could crash, when, the fix) so the paydown's real yield is countable.
2. **A known-but-unexpressed invariant** - the optional cannot be empty here, but nothing says so. Express it: `.value()` after an early guard, a redesign to a non-optional type where the API allows, or - last resort - a `NOLINT` with the invariant stated in the comment.
3. **A false positive of the check** - rare; NOLINT with the reason.

Batch by file, PR-sized batches; the PR body tallies the three categories so the yield is visible.

## Order

Load path first - an unchecked access there is a crash under load, the engine's worst failure mode: `metrics_collector.cpp` (24 findings, the worst src file at the baseline), `run_manager.cpp`, `event_loop*`, `load_strategy.cpp`. Then routes, then core/utils, then tests. The baseline says the tail is overwhelmingly tests, and overwhelmingly `ASSERT`-guarded optionals - `tests/db_test.cpp` 60, `tests/resource_write_route_test.cpp` 32, `tests/cookie_jar_test.cpp` 27, `tests/specs_route_test.cpp` 27, `tests/schema_validation_test.cpp` 26, `tests/metrics_collector_test.cpp` 24 - expect those batches to be mostly category 2.

## Tests

Per batch: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` - non-negotiable, these fixes change control flow. Category-1 findings each carry a mutation-checked regression test in the same PR as the fix.

## Docs

None per batch; `engine/CLAUDE.md` only if a new invariant-expression convention emerges worth writing down.

## Sequencing

[...]

PR #971, opened 2026-08-25
ci(engine): take the test source list off the sanitizer trigger path, and stop PR runs filing issues

### LINKED ISSUE #970: ci(engine): every test-adding PR now runs the full sanitizer matrix, and a PR-run failure auto-files an issue that reads as a red master
Found in practice on PR #969, the first #943 batch. Two connected defects in the sanitizer workflow's PR-facing behaviour, both cheap hygiene rather than urgent, both verified at master `9bca2c2`.

## Problem

**1. The matrix runs on every PR that adds a test file.** `sanitizers.yml`'s `pull_request` trigger is path-scoped to the sanitizer machinery (`.github/workflows/sanitizers.yml:50-55`), and the path list includes `engine/CMakeLists.txt` - correctly, since the `vayu_sanitizers` machinery lives there (`engine/CMakeLists.txt:143-230`). But the engine **test source list also lives in that file** (`engine/CMakeLists.txt:667+`), so any PR that adds a test file touches it and runs all five sanitizer legs (~70 min wall on the slowest, macos-tsan). PR #969 demonstrates it: its only CMake change is one `tests/invariant_test.cpp` line, and the full matrix ran. This will recur on **most PRs of the #943-#946 tidy program**, whose discipline requires new test files per batch - the "ordinary engine pull requests keep paying nothing" comment in the trigger block is not currently true for test-adding PRs. `cancel-in-progress: true` (`:60-61`) limits pileup within one PR but not the per-push burn.

**2. A PR-run failure auto-files an issue that reads as a red master.** Both auto-filing steps run on plain `if: failure()` with no event gating (`.github/workflows/sanitizers.yml:337`, `:359` - verified, zero `github.event_name` conditions in the file). #904's design intent was filing for the **unwatched weekly cron** ("no AI in the loop" because no human is watching Monday 09:00). A pull request has a human watching its checks, yet a failing PR-triggered leg files or comments on a `sanitizer-failure` issue whose body opens "**`<leg>` went red**" - #968 is the proof: filed from `refs/pull/958/merge` for a test that never existed on master, and it needed a manual verification-and-close cycle after the PR merged green. Combined with defect 1, any flaky sanitizer leg on an ordinary test-adding PR now files tracker noise claiming a master failure.

## Fix / solution pathway

1. **Move the engine test source list out of `engine/CMakeLists.txt`** - an `engine/tests/CMakeLists.txt` via `add_subdirectory`, or an included `.cmake` fragment - so test additions stop touching the trigger path. Keep `engine/CMakeLists.txt` in the `paths` list, because the sanitizer machinery genuinely lives there. (Alternative considered and rejected: dropping `engine/CMakeLists.txt` from `paths`, which would cost machinery PRs their pre-merge proof - the wrong trade; the file is in the list *for* the machinery, the test list is a tenant.)
2. **Gate both `if: failure()` filing steps** with `github.event_name != 'pull_request'`. A PR run still fails visibly in the PR's checks - that is the surface a human is already watching - while the cron and `workflow_dispatch` keep filing. #904's acceptance criterion ("demonstrated once on the PR branch") was already satisfied by #958; nothing needs the ungated behaviour any more.

## App changes

None, verified - CI workflow and engine CMake structure only; no renderer/MCP/import consumer reads either.

## Tests

No suite can change colour. Verification is behavioural on the PR itself: the PR that lands this edits `sanitizers.yml` and `engine/CMakeLists.txt`, so it triggers the matrix one last time (correct - it is a machinery PR) and must come back green with **no issue filed** if a leg were to fail is not provable cheaply; instead prove the gate by inspection plus one `workflow_dispatch`. Then the durable check: the next #943 batch PR (test-file addition only) must show **no Sanitizers run** in its checks. State both outcomes in the PR body.

## Docs (same commit)

`docs/engine/building.md`'s "The weekly run, and the issues it files" section: add the sentence that PR-triggered runs (machinery PRs only, post-fix) fail in checks without filing. If the building docs name where test sources are listed, update the path (verify; `grep -n "tests/main.cpp\|CMakeLists" docs/engine/building.md`).

## Sequencing

Land **early in the #943-#946 program** - every batch it precedes saves ~5 runner-hours across the five legs. Conflict surface: `engine/CMakeLists.txt`'s test list is also edited by every test-adding batch, and PR #969 (open) touches the exact block this issue moves - land after #969 merges and rebase, or coordinate. Parallel-safe against everything else open (#467 app-side, #690/#134 parked, #967 upstream).

## Acceptance criteria

- [ ] A PR whose engine changes are source/test files only (no machinery) runs no Sanitizers workflow.
- [ ] A PR touching the sanitizer machinery still runs the matrix pre-merge.
- [ ] A failing `pull_request`-event run files no issue and adds no comment; `schedule`/`workflow_dispatch` failures still file, title-keyed as today.
- [ ] `building.md` reflects both behaviours; `mkdocs build --strict` clean.

---

athrvk:
## Description

Both defects in #970, verified to still exist at `850bf45`.

**Plan.** Root cause of defect 1 is tenancy, not the path list: `sanitizers.yml` triggers on `engine/CMakeLists.txt` *correctly* - the `vayu_sanitizers` interface target is defined there - but the engine's test source list lived in the same file, so adding one `*_test.cpp` was a change to a trigger path and ran all five legs (~70 min on the slowest) over machinery the diff could not touch. #969 is the proof, and the #943-#946 program would have repeated it per batch. So the tenant moves: the list, its two configure-time guards and its CTest discovery are now `engine/tests/CMakeLists.txt`, reached by `add_subdirectory`. The rejected alternative is the issue's own: dropping `engine/CMakeLists.txt` from `paths`, which costs machinery changes their pre-merge proof - the file is in the list *for* the machinery. Defect 2 is a missing event gate: the filing step exists because nobody is watching the Monday cron, and a pull request has a reviewer watching the checks the failure already appears in, so it is now `failure() && github.event_name != 'pull_request'`.

**One thing the issue did not anticipate.** `add_subdirectory` relocates the binary to `build-*/tests/`, which two consumers name directly: the workflow's own smoke-load step (`build-asan/vayu_tests` / `build-tsan/vayu_tests`, answering a miss with `no vayu_tests binary under build-asan/ or build-tsan/` - a five-leg red matrix describing a path, not a defect) and `find_build_artifact` in `build.py`, which checks the build dir and its per-config subdir and nothing else. `RUNTIME_OUTPUT_DIRECTORY` pins the binary where it has always been; every preset is single-config Ninja, so it is one directory.

**Deliberate deviation from the issue text.** The issue asks to gate "both `if: failure()` filing steps (`:337`, `:359`)". There is one filing step - `:359`, whose script both files and comments. `:337` is the artifact upload, and it stays ungated on purpose: the log and the `LastTest.log` are exactly what a reviewer needs on a red PR leg, and uploading an artifact files nothing.

**Not fixed here, on purpose.** `engine/CMakeLists.txt` and `vayu-engine` still list the ~35 `src/http/routes/*.cpp` sources separately; that duplication is adjacent, not this issue's, and this diff does not touch it.

## Type of Change
- [x] Bug fix
- [x] Documentation update

## Testing

```
python build.py -e -t                                  # clean, 1m 15s
  100% tests passed, 0 tests failed out of 2375
mkdocs build --strict -f .github/mkdocs.yml            # clean
```

No pre-existing failures on the base commit; nothing was skipped.

Beyond the suite, the things a green run would not have caught:

- **The move is faithful.** Diffed the extracted block against `HEAD:engine/CMakeLists.txt` dedented: nine changed lines, every one a path or a scope (`${CMAKE_CURRENT_SOURCE_DIR}` -> `${PROJECT_SOURCE_DIR}` where the engine root is meant, `tests/` prefixes dropped, `src/` -> `../src/`, two "add it here" pointers renamed). Everything else byte-identical.
- **Both guards mutation-checked from the new file**, since a guard scanning the wrong directory passes silently:
  - added an unregistered `tests/zz_guard_probe_test.cpp` -> `CMake Error at tests/CMakeLists.txt:225 ... These test files exist but are not in add_executable(vayu_tests ...)`, naming `tests/zz_guard_probe_test.cpp` (still engine-relative);
  - renamed `TreeOrderTest` -> `TreeOrderTestRenamed` in `vayu_scratch_database_suites` -> `CMake Error at tests/CMakeLists.txt:390 ... names 'TreeOrderTestRenamed', which no test declares`.
- **The two-half discovery still partitions exactly**: `vayu_tests --gtest_list_tests` declares 2375, `ctest -N` discovers 2375 - no test lost or discovered twice.
- **`VAYU_ENGINE_SOURCE_DIR` still resolves to the engine root** - the fixture-reading suites (`openapi_document`, `import_parse`, `mock_server_routes`, ...) pass, which they cannot do if it points at `tests/`.
- **Workflow gate by inspection**: `yaml.safe_load` reports the filing step's `if` as `${{ failure() && github.event_name != 'pull_request' }}` and the upload step's as `failure()`; `paths` is unchanged.

On the acceptance criteria: criterion 2 is proved by this PR itself - it edits `sanitizers.yml` and `engine/CMakeLists.txt`, so the matrix should run here, one last time, and that is correct. Criterion 1 is the durable check the issue names and it lands on the *next* test-adding batch PR, which must show **no Sanitizers run** in its checks. Criterion 3's negative half (a red `pull_request` leg filing nothing) is not provable without making a leg fail on purpose, so it is inspection plus the `github.event_name` gate; the positive half is untouched behaviour.

## Checklist
- [x] I have added tests (if applicable) - no suite can change colour here; the verification is the guard mutation-checks and the partition count above, per the issue's Tests section

Docs in the same commit: `docs/engine/building.md` (the trigger-path note, the weekly-run filing rule, and the three places naming where the source list, the locked suites and the per-test timeout live), `CONTRIBUTING.md`, `engine/CLAUDE.md` (with a "do not move it back" note), `docs/app/import-collections/README.md`, and the two test-file comments that named the old location.

## Related Issues
Closes #970

---
_Generated by [Claude Code](https://claude.ai/code/session_01KJcxY9s1aBpu2JemqbC67J)_


athrvk:
`asan on windows-latest` went red at `5e82372`, and it was mine. Fixed in `2cfdc40`.

**Root cause.** `vayu_embed_windows_manifest` built its path from `CMAKE_CURRENT_SOURCE_DIR`, which inside a CMake function is the *calling* directory scope, not the file that defines the function. That was invisible while every caller lived in `engine/CMakeLists.txt`; moving `vayu_tests` into `tests/CMakeLists.txt` made the same call resolve to `engine/tests/res/vayu-windows.manifest`:

```
CMake Error at CMakeLists.txt:400 (target_sources):
  Cannot find source file:
    D:/a/vayu/vayu/engine/tests/res/vayu-windows.manifest
Call Stack (most recent call first):
  tests/CMakeLists.txt:257 (vayu_embed_windows_manifest)
```

Windows only - the `if(NOT MSVC) return()` guard means no other leg reaches the line, which is exactly why a green Linux configure, build and 2375-test suite said nothing about it. The fix is `CMAKE_CURRENT_FUNCTION_LIST_DIR`, which is the definition site and answers the same from any caller, so the helper is now correct wherever a target is declared instead of correct by coincidence of where it is called from.

**Verified, not assumed.** Mutation-checked on Linux by forcing the MSVC branch to execute: with the old expression the configure fails naming `engine/tests/res/vayu-windows.manifest`, identical to the CI error; with the fix it generates clean. I also audited the moved block for the same defect class - `vayu_embed_windows_manifest` is the only function the parent defines and its one call is the only cross-scope call the subdirectory makes. Rebuild and full suite still green (2375/2375).

**One useful accident: this run proved acceptance criterion 3's negative half**, which the PR body said was not cheaply provable. A `pull_request` sanitizer leg genuinely failed, and no `sanitizer-failure` issue was filed and no comment added - `list_issues` for that label returns zero. Before this PR that failure would have filed `sanitizer: asan failure on windows-latest`, whose body opens "**`windows-asan` went red**", about a ref that is not master and a defect that never existed there. That is the #968 shape, reproduced and then not filed.

---

Issue #945, opened 2026-08-23
tidy paydown 3: clear the concurrency-*, cert-* and cppcoreguidelines safety findings the new families surface


athrvk:
Wave 3 of #928. **Now sized by the wave-0 baseline (#942, shipped in PR #966): 205 findings across the three families.** `concurrency-*` 11 (all `concurrency-mt-unsafe`), `cert-*` 30 (`cert-err58-cpp` 25; `cert-err33-c` only 5, so no special handling needed), and the curated `cppcoreguidelines-*` subset 164 (`pro-bounds-pointer-arithmetic` 44, `pro-bounds-constant-array-index` 43, `pro-bounds-array-to-pointer-decay` 23, `special-member-functions` 22, `pro-type-reinterpret-cast` 15, tail). `concurrency-mt-unsafe` flags non-thread-safe calls (localtime, rand, getenv-family, etc.), and the engine's threaded core (event loop workers, thread pool, metrics collector) is exactly where those matter.

## The discipline

Read, never bulk-fix - a `concurrency-*` finding in the load path is a potential real race:
- **Cross-check against #904's TSan runs where possible**: a finding the sanitizer also fires on is a confirmed defect (fix + mutation-checked regression test + **defect record as a sub-issue of #928**), not a lint opinion. A finding TSan is silent on still gets read on its merits - TSan only sees executed interleavings. The weekly sanitizer matrix (#904, merged in PR #958) is live and its first full runs are recorded on `docs/engine/building.md`'s known-state section.
- `cert-*`/`cppcoreguidelines-*` findings that survived wave 0's alias-dedup are by construction non-duplicates of `bugprone-*` - treat each on its rule's merits; the bounds/ownership findings (`pro-bounds-*`, `owning-memory`) are the memory-safety half of the owner's ask.
- False positives: NOLINT with the reason, per the established discipline.

## Tests

Per batch: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`. Any fix in the threaded core additionally deserves a local TSan pass - the `linux-tsan` preset exists now (`cmake --preset linux-tsan && cmake --build --preset linux-tsan && ctest --preset linux-tsan`) - state in the PR whether it ran.

## Sequencing

[...]

PR #976, opened 2026-08-25
fix(engine): read the rest of bugprone-* and clang-analyzer-* - #944 batch 1

### LINKED ISSUE #944: tidy paydown 2: clear the rest of bugprone-* and every clang-analyzer-* finding
Wave 2 of #928. The remaining `bugprone-*` findings after wave 1 (#943) takes `unchecked-optional-access` - **49 findings at the wave-0 baseline** (623 family total minus 574), led by `bugprone-implicit-widening-of-multiplication-result` (17) and `bugprone-empty-catch` (16) - plus **every `clang-analyzer-*` finding**. The analyzer is the path-sensitive engine (use-after-free, null dereference, leaks, dead stores); the full 201-TU baseline measured it at **just 2 findings**, so this wave is small and mostly judgement.

## The discipline

Same as wave 1: read, never bulk-fix. `bugprone-empty-catch` in particular is a judgement check - an empty catch is either a real swallowed error (fix + regression test + **defect record as a sub-issue of #928**), a deliberate ignore that needs its reason at the site, or dead code. `clang-analyzer-*` findings get the same three-way treatment; analyzer false positives exist (interprocedural assumptions) and are NOLINT-with-trace, mirroring the #897 pragma discipline.

## Tests

Per batch: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`. Real defects: mutation-checked regression tests, same PR.

## Sequencing

**#942 shipped (PR #966) - unblocked.** The work list is the 2026-08-24 baseline comment on #928. Overlaps #943 only if a file carries both kinds of findings - batch by file across both waves where that happens rather than editing a file twice. #946 runs last.

## Acceptance criteria

- [ ] Zero `bugprone-*` and zero `clang-analyzer-*` findings tree-wide, re-scanned.
- [ ] Real defects have records (sub-issues of #928) and mutation-checked tests; deliberate ignores carry their reason at the site.

---

### LINKED ISSUE #974: core(run_manager): the metrics reservation read a run's duration with the parse `parse_duration_ms` was written to replace
Defect record for #944 (wave 2 of #928), found reading the `bugprone-empty-catch` finding at `engine/src/core/run_manager.cpp:522`. Fixed in the same PR as the finding that surfaced it; this issue exists because the tidy program records what it found, not only what it changed.

## Problem

`RunContext`'s constructor sized the run's metrics stores from a duration it parsed itself:

```cpp
std::string duration_str = config.value ("duration", "60s");
int64_t duration_ms      = 60000; // default 60s
try {
    duration_ms = std::stoll (duration_str.substr (0, duration_str.length () - 1)) * 1000;
} catch (...) {
}
```

That is character-for-character the parse `vayu::core::parse_duration_ms` (`engine/include/vayu/core/load_pacing.hpp`) was written to remove - its own doc comment names it: *"Replaces `stoll(str.substr(0, len - 1)) * 1000`, which stripped one character and always multiplied by 1000: '5m' ran for 5 seconds and '500ms' for 500."* The strategy path was fixed; this copy was not, and nothing pointed the two at each other. CLAUDE.md's rule for exactly this shape: **a hand-rolled copy of a primitive does not receive the primitive's fixes.**

The empty `catch` is what kept it invisible - every wrong reading below is a *successful* parse, so the catch never fired for any of them.

## Impact

`expected_requests` is `duration_s x targetRps x 1.2`, floored at 10 000, and it is the reservation `MetricsCollector` sizes its result and latency stores from. Measured on the values the tests now pin:

| `duration` | reserved (was) | reserved (correct) |
|---|---|---|
| `5m` @ 1000 rps | 10 000 (5s, then the floor) | 360 000 |
| `60` @ 1000 rps | 10 000 (read as 6s) | 72 000 |
| `500ms` @ 1000 rps | 600 000 (read as 500s) | 10 000 (the floor) |
| `-30s` | `18446744073709515616` | 72 000 (refused, default minute) |
| `30` (a JSON number) | throws `json.exception.type_error.302` out of the constructor | 72 000 |

Nothing here is a wrong *answer* - the run's real duration comes from the strategy path, which parses correctly, and both derived reserves are capped downstream (`metrics_collector.cpp:145`), which is what kept the negative row from being an allocation of that size. What it cost is the thing the reservation exists for: a 5-minute run at 1000 RPS reallocated its way from 10 000 up through 360 000 results while the load loop was running, and a `500ms` design run reserved 600 000 slots it would never use. The last row is a `500` on a run config a client spelled with a number instead of a string, thrown from a constructor rather than reported by the validator.

## Fix

Landed as `vayu::core::expected_requests_for (config)` in `run_manager.hpp`, beside `live_ring_size` and on the same terms - an extracted, named, directly tested core that the constructor calls in one line. It reads the duration through `parse_duration_ms`, and a duration it cannot read (unparseable, negative, or not a string) reserves for the same 60s the field itself defaults to, because an allocation hint has no business failing a run the config validator already passed.

## Tests

`ExpectedRequestsFor.{ReadsTheDurationTheWayTheStrategyDoes, TakesTheRpsUnderEitherSpelling, ADurationItCannotReadReservesForTheDefaultMinute}` in `engine/tests/run_manager_test.cpp`, one case per row of the table above. Mutation-checked: with the old parse put back, two of the three fail (the third pins the `rps`/`targetRps` spelling, which the old parse also got right).

## Follow-up worth knowing

`expected_requests_for` still reads the RPS with `config.value ("rps", 0.0)`, which throws the same `type_error.302` on `"rps": "1000"`. Left as it was rather than widened here - it is unchanged behaviour and belongs with whatever decides how strictly a stored run config is read. Not filed as work on its own; noted so the next reader of that function knows it was seen.

---

athrvk:
## Description

Wave 2 of #928 (#944): the remaining `bugprone-*` findings after #943 takes `unchecked-optional-access`, plus every `clang-analyzer-*` finding. Read one at a time, never bulk-fixed.

**The plan.** Root cause differs per check here, unlike wave 1, so the reading is per finding. The one that mattered: `bugprone-empty-catch` at `run_manager.cpp:522` was hiding a **real defect** - `RunContext`'s constructor sized the run's metrics stores from a duration it parsed itself, and that parse was character for character the one `parse_duration_ms` was written to replace ("5m" reserved five seconds' worth, "500ms" five hundred seconds' worth, a bare "60" six). The empty catch is why nobody had seen it: every wrong reading is a *successful* parse, so the catch never fired. The fix is `core::expected_requests_for`, an extracted core beside `live_ring_size` on the same terms, called from the constructor in one line - recorded as **#974** and mutation-checked. The alternative I rejected was fixing the parse in place: it would have left the sizing untestable, which is what let a known-broken parse survive a fix to its twin.

The other fifteen empty catches are deliberate ignores, and that outcome was **unreachable as configured** - a comment does not satisfy the check (measured); only a keyword in its `IgnoreCatchWithKeywords` option does. So `engine/.clang-tidy` now lists `@deliberate` beside clang-tidy's own `@TODO`/`@FIXME`, and each site states its reason in prose after it. The alternative I rejected was fifteen `NOLINT(bugprone-empty-catch)`s, which suppress identically and record nothing about why; the option is the same suppression with the reason made part of writing it. Two of the fifteen had no reason worth keeping and are not in that count - `run_manager.cpp:1798` swallowed a failed metric-tick write, which is a hole in the run's history series nothing downstream can distinguish from a quiet second, and now logs it.

**Deliberate deviations from the issue spec**, both with reasons:

1. **This is batch 1, and it does not close #944.** #944's own Sequencing says to batch a file across both waves rather than edit it twice. Measured: 8 of the 11 `engine/tests` files holding wave-2 findings also hold wave-1 ones (`db_test.cpp` 47, `inbox_test.cpp` 21, `metrics_collector_test.cpp` 19), so their **11 findings are left for #943's tests batch to take together**. The three test files with no wave-1 overlap at all - `import_route_test.cpp` (6), `db_concurrency_test.cpp` (2), `tests/main.cpp` (1) - are in this PR, because nothing is gained by holding them. Full remainder listed in a comment on #944.
2. **`cli.cpp`'s `bugprone-exception-escape` fix moves lines that carry six `cppcoreguidelines-pro-bounds-pointer-arithmetic` findings** - #945's family, not this one. `main` cannot both be the catch and hold the body, so the extraction is the fix, and moving those lines puts them in changed-line scope where the gate would refuse them. They are fixed here (`argv` reaches `run_cli` as a `std::span`, so the parse indexes it instead of walking a pointer) rather than left to fail. That is wave-3 work pulled forward by necessity; `cli.cpp` is clean for #945 as a result.

**Yield tally** (#944's three categories): **category 1 (real defect) 1** - #974, with its regression test · **category 2 (fixed or stated in place) 38** · **category 3 (NOLINT) 1** - `optin.performance.Padding` on `ReportExtras`, whose field order is by report section with a paragraph per group explaining why absent is not zero; size-ordering its 48 fields saves 44 bytes on a struct built once per `GET /runs/:id/report` and leaves every one of those comments describing fields that no longer follow it. The trade and the revisit condition are at the site.

**Measurement.** Tree-wide, scanned at master `9bca2c2` over the 200 non-vendor TUs the decided config lints (`src/runtime/` excluded, as its own `.clang-tidy` disables everything): **51 findings** - 49 `bugprone-*` after `unchecked-optional-access` and `easily-swappable-parameters` come out, plus 2 `clang-analyzer-*`. That is exactly #928's wave-0 estimate (49 + 2), so unlike the `src` per-file numbers #969 could not reproduce, this column of that inventory holds. Split: **31 in `engine/src`, 20 in `engine/tests`**.

The rest of the findings, by check:

- **`bugprone-branch-clone` (4)** - each pair of identical branches merged into the one branch it always meant: `implicit` joins `authorization_code_with_pkce` in the import's grant map (Vayu has no implicit grant), `CURLE_COULDNT_RESOLVE_PROXY` falls through to `CURLE_PROXY`, GraphQL and JSON-RPC fall through to JSON's content type, and the logger's three-branch chain that assigned the same `true` becomes the one expression that says which condition earned it.
- **`bugprone-optional-value-conversion` (3)** - `mock_server`'s `best` carried whole instead of dereferenced and re-wrapped into an optional of the same type.
- **`bugprone-move-forwarding-reference` (2)** - `insert_at_slot`'s `T&&` forwards rather than moving unconditionally.
- **`bugprone-implicit-widening-of-multiplication-result` (2 in `src`, 6 in `import_route_test.cpp`)** and **`bugprone-misplaced-widening-cast` (2)** - widened before multiplying, not after.
- **`bugprone-suspicious-stringview-data-usage` (1)** - `from_chars` takes a range either way; both bounds are now named before the call, so a bare `.data()` is not sitting in argument position reading as a pointer handed to something that looks for a terminator.
- **`bugprone-exception-escape` (2)** - `vayu-cli`'s `main` and the test binary's, both now a catch around an extracted body. An escape from `main` ends the process through `std::terminate`: no message a user can act on, and a status ctest reports as a crash rather than as a run that failed and said why.
- **`clang-analyzer-cplusplus.Move` (1)** - a false positive (the analyzer carried a moved-from state across an iteration that reconstructs the value; a minimal repro of the same shape does not reproduce it). Cleared without a NOLINT by scoping the value to the `if` that guards it, which is the better code anyway - no move inside a ternary whose condition reads the thing being moved.

## Type of Change
- [x] Bug fix (one real defect, #974; the rest is latent-defect paydown with no behaviour change on any reachable path)
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update (`docs/engine/building.md`, `engine/CLAUDE.md` - the `@deliberate` convention this wave settles)

## Testing

```
python build.py -e -t                                     # green
cd engine && ctest --preset linux-dev --output-on-failure  # 2374/2374 passed, 0 failed (41s)
clang-format-19 --dry-run -Werror  (every touched file)    # clean
clang-tidy-19, wave-2 check set, 89 engine/src TUs + the 4 touched test TUs
                                                           # 0 findings (was 40)
clang-tidy-diff-19.py, full repo config, git diff -U0 origin/master
                                                           # 0 findings on any changed line
```

Three new tests in `engine/tests/run_manager_test.cpp`:

- `ExpectedRequestsFor.{ReadsTheDurationTheWayTheStrategyDoes, TakesTheRpsUnderEitherSpelling, ADurationItCannotReadReservesForTheDefaultMinute}` - one case per row of the table in #974.
- **Mutation-checked**: with the old parse put back in the header, two of the three fail - `"5m"` reserves 10 000 instead of 360 000, `"5min"` 10 000 instead of 72 000, `"-30s"` `18446744073709515616` instead of 72 000, and `{"duration": 30}` throws `json.exception.type_error.302` out of the function. The third pins the `rps`/`targetRps` spelling, which the old parse also got right, so it is documentation rather than a mutation check and is not claimed as one.

No behaviour change is asserted for the refactors, because none is intended: the branch merges, the optional carries and the forwarding are each the same operation respelled, and the existing suites cover the paths they sit on.

## Checklist
- [x] I have added tests

## Related Issues

Closes #974 (the defect record this batch produced).

Part of #944 (batch 1 of the wave - 11 findings across 8 `engine/tests` files remain, deliberately held for #943's tests batch, so this does not close it), which is wave 2 of #928.

Conflict surface with #969 (open, #943 batch 1): both touch `run_manager.cpp` and `mock_server.cpp`, at hunks 10+ lines apart in both files, so they merge without a conflict in either order.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JPqw7GTLzFEX5iiuqKtyF6)_


athrvk:
Rebased onto master (`850bf45`) now that #969 has merged - the two touch `run_manager.cpp` and `mock_server.cpp` and merged without a conflict, as the PR body predicted. The merged tree builds and the full suite is green at **2378/2378** (the four extra tests are #969's).

Correcting one line in the body while I am here: it calls #969 "open", which it was when I wrote it - it merged at 12:34 UTC.

Worth noting for #970, which asks for exactly this evidence: this PR adds no new test file, so it does not touch `engine/CMakeLists.txt`, and its checks show **no Sanitizers run**. That is the first of #970's acceptance criteria observed in the wild - and the contrast case is #969, which added one test file and ran all five legs for it.

---

athrvk:
CI caught one, and it was mine. Fixed and pushed (`dadb2de`), rebased onto master `ad335b8` on the way.

**What failed.** `Engine on ubuntu-latest` - the clang-tidy gate, not the suite (`C++ Tests (ubuntu-latest)`: 2378/2378 passed in the same job):

```
engine/src/http/routes/inbox.cpp:153:41: error: do not use pointer arithmetic
  [cppcoreguidelines-pro-bounds-pointer-arithmetic]
  153 |         const auto* const end   = begin + part.size ();
```

My `bugprone-suspicious-stringview-data-usage` fix traded one finding for another. The arithmetic was in the original line too, but that line was not *changed*, so it sat in #945's backlog where the gate could not see it; naming `end` on a new line put it in changed-line scope. `std::to_address (part.begin ())` / `std::to_address (part.end ())` is the same two addresses with neither the bare `data ()` nor the `+`, so both checks are answered by saying what the code means instead of suppressing either. Still no second NOLINT - the tally in the body stands at one.

**Why my pre-push run missed it, since that is the more useful half.** I ran the real gate locally - `clang-tidy-diff-19.py` over `git diff -U0 origin/master` - and read its output through `tail -40`. The tail was the harmless PCH noise this build directory emits; the finding was above the cut, and `tail`'s exit status is what I read as the gate's. The re-run is captured whole and checked by exit status, and it now reports no file-anchored finding on any changed line.

Local on the merged tree: build green, **2378/2378**, clang-format clean, gate clean.

---

Issue #942, opened 2026-08-23
tidy paydown 0: apply the decided check split to engine/.clang-tidy and measure the full tree


athrvk:
Wave 0 of #928 (parent carries the owner decision and the full context). Config and measurement only - **no engine code changes**. Everything after re-baselines on this wave's numbers.

## What ships

1. **The check split in `engine/.clang-tidy`**, per the parent's decision, one reason line per change (the file's existing precedent):
   - Enable: `concurrency-*`, `cert-*`, and the curated `cppcoreguidelines-*` safety subset (`pro-type-*`, `pro-bounds-*`, `owning-memory`, `slicing`, `special-member-functions`) - **alias-aware**: most `cert-*` and many `cppcoreguidelines-*` checks alias `bugprone-*`/others; suppress the alias side so one defect is one finding (clang-tidy duplicates alias findings unless the config avoids it - state the mechanism used).
   - Disable with reasons: `readability-uppercase-literal-suffix`, `readability-braces-around-statements`, `readability-function-cognitive-complexity`, `readability-implicit-bool-conversion` (borderline note: revisit only on evidence), `modernize-use-designated-initializers`, `modernize-use-ranges`, `modernize-use-starts-ends-with`, `modernize-return-braced-init-list`, and the remaining style-motivated `readability-*`/`modernize-*` currently firing. Keep `modernize-avoid-c-arrays` enabled (bounds-safety).
2. **The full measurement**: `run-clang-tidy-19` over `engine/{src,include,tests}` on a PCH-free compile database, run to completion (**256/256 TUs** - the previous scan died at 188), on the new config. Post the inventory (by family, by check, worst files) as a comment on #928, replacing its lower-bound numbers as the baseline for waves 1-4.

## Known consequence to state in the PR

Enabling the new families means the line-scoped CI gate enforces them on changed lines immediately, while their backlog exists - the same posture every enabled check already has. Any PR that trips a new-family finding on its own lines is working as intended.

## Tests

No code changes - the suite cannot change colour. Verification: the scan completes; the gate stays green on this PR itself (it changes only config + docs); `mkdocs build --strict` if `building.md` is touched.

## Docs (same commit)

`engine/.clang-tidy` is the primary doc (reasons in-line). `docs/engine/building.md`'s Static Analysis section if it names the enabled families (verify).

## Sequencing

First wave - everything else in #928 depends on this baseline. After PR #941 (gate shape). Parallel-safe against all non-tidy work.

## Acceptance criteria

[...]

PR #1017, opened 2026-08-25
fix(engine): count the headers the tidy paydown never measured, and clear them - #1013

### LINKED ISSUE #1013: engine(tooling): the tidy paydown has been measuring translation units, and headers carry 50 findings nobody has counted
Found while clearing #945 batch 3. It is a measurement-basis problem rather than a family, which is why it is filed here as a sibling of the waves rather than folded into one of them - and it lands on #946's plate before its final re-measure, not after.

## The problem

Every number this paydown has worked from is a **translation-unit** count. #942's baseline, #969's per-file re-check, #975's `concurrency-*` tally, #979's exact-to-the-finding `cert-*` reproduction and #945 batch 3's rescan all ran `run-clang-tidy` over `compile_commands.json`, and that reports **nothing in a header** unless `-header-filter` is passed on the command line.

The `HeaderFilterRegex: '.*/(?:src|include|tests)/.*'` in `engine/.clang-tidy` does not supply it. Measured directly:

```bash
# reports only src/http/client.cpp
clang-tidy-19 -p build --checks='-*,cppcoreguidelines-special-member-functions' \
  src/http/client.cpp

# reports client.cpp *and* cookie_jar.hpp and logger.hpp
clang-tidy-19 -p build --checks='-*,cppcoreguidelines-special-member-functions' \
  --header-filter='.*/(src|include|tests)/.*' src/http/client.cpp
```

So this is the same seam #940 settled for CI ("CI lints translation units, never a header") reaching the *measurement* as well as the gate. `scripts/pre-commit` is the only thing that ever lints a header, and it lints only what a commit stages - which means a header finding older than any diff is currently invisible to every consumer of these numbers.

## What is standing

Re-run of the three wave-3 families over the 207 non-vendor translation units with `-header-filter` on, at `5c991c5` (#945 batch 3), deduplicated by (file, line, column, check) because a header reports once per including translation unit:

| Check | In headers |
|-------|-----------|
| `cppcoreguidelines-special-member-functions` | 21 |
| `cppcoreguidelines-pro-type-member-init` | 14 |
| `cppcoreguidelines-pro-type-reinterpret-cast` | 5 |
| `cppcoreguidelines-pro-type-vararg` | 2 |
| `cppcoreguidelines-pro-bounds-array-to-pointer-decay` | 3 |
| `cppcoreguidelines-pro-bounds-constant-array-index` | 3 |
| `cppcoreguidelines-pro-bounds-pointer-arithmetic` | 2 |
| **Total** | **50** |

Eight of those belong to `pro-bounds-*` and so are #945 batch 4's to take; the other 42 are batch 3's families in headers batch 3 did not edit. It cleared the headers it touched (`utils/encoding.hpp`, `utils/sodium_init.hpp`, `db/database.hpp`, `http/pkce.hpp`, `tests/mock_server.hpp`, `tests/proxy_server.hpp`, plus the new `tests/task_queue.hpp`) per #975's "any batch that adds or edits a header owes it that second pass", and stopped there rather than growing into an unreviewable diff.

Where the 42 sit:

- **`include/vayu/types.hpp` - 11 `pro-type-member-init`.** The domain row structs (`Collection`, `Request`, `Run`, ...) leave `order`, `created_at`, `updated_at`, `method`, `status` and their kin without default member initialisers while other fields on the same struct have them. Zeroing the arithmetic ones is mechanical; the *enum* ones are the question worth answering, because a default is a value some path can now read without setting - which is exactly the call `reorder.cpp`'s `Kind kind` made in batch 3, with a comment.
- **~20 `special-member-functions` on the manager and holder classes** - `Database`'s siblings: `Server`, `ManagedListener`, `SseStreamManager`, `InboxManager`, `MockIssuerManager`, `MockServerManager`, `OAuth2AuthorizeManager`, `EventLoopWorker`, `CurlHandlePool`, `EventLoopImpl`, `TransferData`, `LoadStrategy`, `RunManager`, `RunContext`, `StepHistograms`, `Logger`, `CookieJar`, `ThreadPool`, plus `tests/echo_server.hpp` and `tests/tls_server.hpp`'s two. Most already delete copy and are missing only the two move declarations; batch 3's spelling (`= delete` for all four, beside the destructor) applies unchanged. **`CookieJar` is the one to read rather than pattern-match** - it defines copy and *no* destructor, which is the opposite shape and may want defaulted moves rather than deleted ones.
- **`tests/tls_server.hpp` - 3 `pro-type-reinterpret-cast`.** OpenSSL's C API takes `const unsigned char*` directly, so `vayu::utils::byte_view` cannot stand in. Either a NOLINT with that reason each, or a test-side inverse of `byte_view` if a second caller ever appears. This file is on `character_cast_test.cpp`'s exemption list today, with that reason recorded there.
- **`src/core/js_json.hpp` - 2 `pro-type-vararg`**, and the two remaining `reinterpret_cast` in `include/vayu/http/event_loop/event_loop_worker.hpp`.

## Why it matters before #946 and not after

#946's closing measurement is the one that decides the changed-lines-to-whole-file gate promotion, and its whole point is zero findings tree-wide. Two things follow:

1. **The promotion decision needs a stated basis.** A whole-*file* gate still lints translation units - #940's reason has not changed - so a promoted gate would keep not seeing these. Either #946 records that headers are out of scope and says why, or it records that they are in and something scans them.
2. **A number measured one way and reported as "tree-wide" is the failure mode this issue is.** #975 and #979 both flagged that the wave-0 per-check table over-counts wherever a family reaches headers (`temp_database.hpp` reporting once per including TU). That is the *same* seam seen from the other side: dedupe by (file, line) fixes the over-count, `-header-filter` fixes the under-report, and neither alone gives an honest whole-tree number.

## Fix

Two halves, and the first is the one that outlasts this issue.

**Decide and record the basis.** Whether "clean" means translation units or the whole tree including headers, in `docs/engine/building.md`'s Static Analysis section, next to the existing note that CI lints translation units and the hook lints headers. If the answer is "the whole tree", the re-measure command in #946 carries `-header-filter` and says so.

**Clear the 42, in batch 3's own shape.** `special-member-functions` first (the largest and the most mechanical), then `types.hpp`'s member-init, then the tail. Read, never bulk-fix: a default on an enum a caller is expected to set is a decision, not a fix.

## Tests

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`. The `special-member-functions` half changes no behaviour and needs the build; a default member initialiser on a stored row struct can change what a partially-populated write persists, so `db_test.cpp` and the route suites are the ones that could plausibly change colour.

Re-measure with:

```bash
run-clang-tidy-19 -p build -quiet -header-filter='.*/(src|include|tests)/.*' \
  -extra-arg=-Wno-ignored-gch <every non-vendor TU>
```

and deduplicate by (file, line, column, check) - a header finding is reported once per including translation unit, which is what made the wave-0 table's `concurrency-*` count read 11 where 8 sites existed.

## Sequencing

Independent of #945 batch 4 (`pro-bounds-*`, 93 translation-unit findings still standing) - the two touch different checks, and the 8 `pro-bounds-*` header findings above go with batch 4 rather than with this. **Must close before #946's final re-measure**, because the basis this decides is what that measurement means. `src/runtime/` stays out of scope in both: `engine/src/runtime/.clang-tidy` declines every check there with its reason, and the 261 findings a check-list override surfaces in `script_engine.cpp` are that decision working, not a backlog.

## Acceptance criteria

- [ ] The measurement basis - translation units, or the whole tree including headers - is decided and recorded in `docs/engine/building.md`, with the re-measure command that matches it.
- [ ] Zero findings for the wave-3 families over `engine/{src,include,tests}` on that basis, `src/runtime/` and the `pro-bounds-*` family excepted.
- [ ] Any site kept carries its reason at the line, and `CookieJar`'s shape is answered rather than pattern-matched.

athrvk:
> **Stacked on #1014.** Base is `claude/bold-keller-8rn0i1`, not `master` - the two diffs overlap in `engine/CLAUDE.md`, `tests/character_cast_test.cpp` and `tests/tls_server.hpp`. Merge #1014 first; this retargets to `master` on its own. Review it as the ~230 lines below, not as #1014's 631 plus these.

## Description

**The plan.** #1013 is a measurement bug before it is a backlog. Every number #928's paydown has worked from - #942's baseline, #969's re-check, #975's and #979's tallies, #945 batch 3's rescan - is a **translation-unit** count, and `run-clang-tidy` reports nothing found in a header unless `-header-filter` is passed **on the command line**. The `HeaderFilterRegex` in `engine/.clang-tidy` does not supply it. So 50 findings for wave 3's families were standing in headers that four rounds of measurement had never read, and #946's closing "zero findings tree-wide" would either have missed them or discovered them at the worst possible moment.

Root cause: the same seam #940 settled for CI ("CI lints translation units, never a header") reaching the *measurement* as well as the gate. The fix is therefore two halves, and the first outlasts the second - a basis recorded in `docs/engine/building.md`, then the findings. The alternative I rejected was flipping `AllowMissingMoveFunctionsWhenCopyIsDeleted` in the tidy config, which would have silenced ~13 of the 21 `special-member-functions` findings without a line of code changing. It is defensible in isolation - a class with deleted copy has no implicitly declared move, so the defect the rule guards cannot occur - but it would have made #1014's 24 explicit deletions redundant the week they landed, and traded a stated convention for a config flag nobody reads.

This clears the **42** that belong to batch 3's checks. The other 8 are `pro-bounds-*` and go with #945 batch 4, as the issue says.

### The basis (`docs/engine/building.md`)

"Clean" means the whole tree, headers included. The section now carries the re-measure command with its `-header-filter` and `-extra-arg=-Wno-ignored-gch` (#912), and says the output must be **deduplicated by (file, line, column, check)** - a header finding is reported once per including translation unit, which is exactly what made the wave-0 table read 11 `concurrency-mt-unsafe` where 8 sites existed. #975 and #979 both flagged that over-count from the other side; this is the same seam, and neither dedupe nor `-header-filter` alone gives an honest number.

It also records what this does **not** change: the gate still lints translation units, so the changed-lines-to-whole-**file** promotion #946 decides still would not read a header - the pre-commit hook is what does - and `engine/src/runtime/` stays out of every count by its own config's written decision.

### Rule of five, the remaining 21

`Server`, `ManagedListener`, `SseStreamManager`, `InboxManager`, `MockServerManager`, `MockIssuerManager`, `OAuth2AuthorizeManager`, `RunManager`, `RunContext`, `EventLoopImpl`, `EventLoopWorker`, `CurlHandlePool`, `TransferData`, `LoadStrategy`, `StepHistograms`, `ThreadPool`, `Logger`, `CookieJar`, and `EchoServer` / `CrlServer` / `TlsServer` in `tests/`. Most already deleted copy and were missing only the two move declarations.

- **`CookieJar` is the one the issue said to read rather than pattern-match.** Its comment already said "Non-copyable, **non-movable**" and only the copy half was written; the code now says what the comment did. It also needed `~CookieJar () = default;` - with no destructor declared, the check is not satisfied by the other four.
- **Declaring the four suppresses the implicit default constructor**, and the build caught it: `TransferData` (built by `std::make_unique<TransferData>()`), `RunManager` and the `LoadStrategy` interface each gained `= default` beside them. Nothing else in the tree copies or moves any of these - a clean build is the proof.
- `LoadStrategy` is a polymorphic base, where C.67 offers protected copy/move as an alternative. Deleted instead, with the reason at the site: nothing copies a strategy, and deleting is what stops a derived one being sliced back to the interface.

### Every `vayu::db` row-struct scalar carries a default

The argument is one the file had already made for itself. `Run::end_time` carries a comment explaining that it is defaulted "so an insert site that forgets to stamp it cannot store an indeterminate value"; that reasoning holds for every scalar in the namespace, and 11 of them did not have it. These are aggregates an insert fills field by field, and a field it forgets is *indeterminate* rather than zero - which sqlite_orm then binds and stores.

The three enums (`Request::method`, `Run::type`, `Run::status`) are the decision rather than the mechanics. They default to what `database.cpp`'s `row_extractor` **already falls back to** for a stored value it cannot parse - `GET`, `Design`, `Pending` - so the struct and the reader agree about what an unset field means instead of each picking its own answer. A default is not a substitute for setting the field: it bounds what a wrong insert can persist, and stops an indeterminate `bool` or enum being read at all.

### The tail

- **`tls_detail::openssl_bytes`** (`tests/tls_server.hpp`) is the third direction of #1014's character-cast rule: OpenSSL's parameters are `const unsigned char*` with no `string_view` seam, so a passphrase, a DER blob and a common name go through one cast instead of three. The file stays `character_cast_test.cpp`'s single exemption, and the exemption's reason now names the function it covers.
- The two `uintptr_t` casts in the same file are **not** that rule - a pointer used as a certificate serial, and as a unique route path - and carry their reason at the line.
- **`js_json.hpp`'s two `snprintf` calls become `std::format`**, which the engine already uses on all three platforms (`routes.hpp`, `mock_server.cpp`). Same output, without the vararg call, the fixed buffer, or a short-write branch that a value bounded by 1e21 could never take. Rounding cannot differ: the guard above admits only values that are already whole. This deletes the construct rather than annotating it, which is #979's lesson applied.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

```
python build.py -e -t
mkdocs build --strict -f .github/mkdocs.yml
```

**2400 tests pass, 0 fail** - the same count as this PR's base, since nothing here adds a test. `mkdocs build --strict` clean.

**The rescan, on the basis this PR records.** The ten checks report **zero** over the 207 non-vendor translation units with `-header-filter='.*/(src|include|tests)/.*'` - the measurement the issue exists to fix, run the way `building.md` now documents. `src/runtime/` reports 136 and every one is its own config's decision working.

**The build is the proof for the special members.** A deleted copy or move that something actually used would not compile, and the first pass did not: `TransferData` failed at `std::make_unique`, which is how the suppressed-default-constructor case was found rather than guessed at.

`clang-format-19 --dry-run -Werror` clean over every touched file.

## Checklist
- [ ] I have added tests (if applicable)

**No new test, deliberately.** Both halves are compile-time: a deleted special member is checked by the build, and a default member initialiser has no runtime behaviour to lock that the existing db and route suites do not already exercise. The `js_json.hpp` rewrite is the one behavioural change, and it is covered - `import_parse_test.cpp` and `openapi_document_test.cpp` pin that number formatting against a recorded corpus (#877's conformance fixture), which is a stronger check than anything written for this diff would be. #1014's `character_cast_test.cpp` guard already covers the new `openssl_bytes`.

`engine/CLAUDE.md` gains the row-struct-defaults rule and the measurement rule, and #1014's two bullets are extended rather than duplicated (the third cast direction; the manager classes and the suppressed default constructor).

## Deliberate deviations from the issue

- **The 8 `pro-bounds-*` header findings are not here.** The issue assigns them to #945 batch 4 and this PR honours that rather than widening.
- **No change to the gate's scope.** The issue asks the basis to be decided and recorded, not the gate to be moved; the whole-file promotion remains #946's call, and `building.md` now says why that promotion still would not read a header.

## Related Issues
Closes #1013. Sibling of #945 (whose batch 4 keeps the `pro-bounds-*` remainder); sub-issue of #928.

---
_Generated by [Claude Code](https://claude.ai/code/session_0126GN9DBeFsfcVA8PAYkuxk)_

### DIFF
--- a/docs/engine/building.md
+++ b/docs/engine/building.md
@@ -995,6 +995,40 @@ down by whoever edits those lines - and, since #928, by its waves on purpose:
 findings down family by family, with promotion from changed-lines to whole-file
 gating decided in #946 once nothing pre-existing is left to surface.
 
+**"Clean" means the whole tree, headers included - and a scan has to be asked
+for that** (#1013). `run-clang-tidy` reports nothing found in a header unless
+`-header-filter` is passed **on the command line**: the `HeaderFilterRegex` in
+`engine/.clang-tidy` does not supply it. Every number the #928 paydown worked
+from before this was measured without it, so 50 findings sat in headers that no
+report had ever counted - the same seam as "CI lints translation units, never a
+header" (#940), reaching the measurement rather than the gate.
+
+That is a gap in the *measurement*, not in the standard. A header is part of the
+tree, the pre-commit hook lints one whenever a commit stages it, and #946's
+"zero findings tree-wide" is not true of a tree whose headers were never read.
+So the re-measure command is:
+
+```bash
+run-clang-tidy-19 -p engine/build -quiet \
+  -header-filter='.*/(src|include|tests)/.*' \
+  -extra-arg=-Wno-ignored-gch <every non-vendor translation unit>
+```
+
+and its output is **deduplicated by (file, line, column, check)** before it is
+counted, because a header finding is reported once per translation unit that
+includes it - which is what made the wave-0 table read 11 `concurrency-mt-unsafe`
+where 8 sites existed. `-extra-arg=-Wno-ignored-gch` is needed because the
+build's precompiled header is GCC's (#912).
+
+Two things this does **not** change. The gate still lints translation units,
+because a header has no `compile_commands.json` entry and a guessed command
+emits `clang-diagnostic-error`s that escape line filtering (#940) - so a
+promotion from changed-lines to whole-*file* scope, which #946 decides, still
+would not read a header; the hook is what does. And `engine/src/runtime/` stays
+out of scope in every count: its own `.clang-tidy` declines every check with a
+written reason, so the findings a check-list override surfaces there are that
+decision working, not a backlog.
+
 **What the config enables is a decision, not a default** (#928): the families
 that catch real defects - `bugprone-*`, `clang-analyzer-*`, `concurrency-*`,
 `cert-*`, a curated `cppcoreguidelines-*` safety subset, `performance-*` - plus
--- a/engine/CLAUDE.md
+++ b/engine/CLAUDE.md
@@ -96,8 +96,11 @@ engine/
   `hmac_sha256` digest, and the six sites that spelled that themselves;
   **`vayu::db::column_text`** (`db/database.hpp`) for a sqlite TEXT column,
   which `sqlite3_column_text` hands back as `const unsigned char*` and nothing
-  consumes as one; and **`vayu::utils::detail::sodium_bytes`**
-  (`utils/sodium_init.hpp`) going the other way, for libsodium. A SQL NULL stays
+  consumes as one; **`vayu::utils::detail::sodium_bytes`**
+  (`utils/sodium_init.hpp`) going the other way, for libsodium; and
+  **`tls_detail::openssl_bytes`** (`tests/tls_server.hpp`, #1013) going the same
+  way for OpenSSL, whose parameters are `const unsigned char*` with no
+  `string_view` seam. A SQL NULL stays
   a null pointer through `column_text` on purpose - absent and empty are
   different answers, and defaulting one to the other erases the distinction its
   callers read. `tests/character_cast_test.cpp` scans `engine/{src,include,tests}`
@@ -112,15 +115,41 @@ engine/
   leaving the rest implicit is how a fixture that owns a listener and a thread
   stays copyable: the copy compiles, both objects stop the same server, and
   nothing says it was not meant to. Every RAII holder here - the `Impl` behind a
-  pImpl, `Database`, and the ~20 in-process mock servers in `tests/` - deletes
-  copy *and* move beside its destructor, in the spelling `client.hpp` uses
-  (`Foo (const Foo&) = delete;` and the three that follow). Deleting is the
+  pImpl, `Database`, the ~20 in-process mock servers in `tests/`, and (#1013)
+  every manager and listener in `include/`: `Server`, `ManagedListener`,
+  `SseStreamManager`, `InboxManager`, `RunManager`, `EventLoopWorker`, `Logger`
+  and their kin - deletes copy *and* move beside its destructor, in the spelling
+  `client.hpp` uses (`Foo (const Foo&) = delete;` and the three that follow).
+  Deleting is the
   default answer, not defaulting: none of these is meaningfully movable, and a
-  move would leave a hollow object whose methods still compile. A mock server's
+  move would leave a hollow object whose methods still compile. **Declaring the
+  four suppresses the implicit default constructor**, so a class that had no
+  other constructor gains `Foo () = default;` beside them - `TransferData`,
+  `RunManager` and the `LoadStrategy` interface each needed it, and the build is
+  what says so. A mock server's
   thread pool is `vayu::tests::pooled_task_queue`
   (`tests/task_queue.hpp`) - httplib's `new_task_queue` hook takes a raw owning
   pointer, which is its contract and not a leak, said once there rather than at
   each fixture.
+- **A row struct's scalars all carry a default** (#1013,
+  `cppcoreguidelines-pro-type-member-init`). The `vayu::db` structs in
+  `types.hpp` are aggregates an insert site fills field by field, so one it
+  forgets is *indeterminate* rather than zero - and sqlite_orm binds and stores
+  whatever that was. `Run::end_time` had already made the argument for itself;
+  it now holds for every scalar in the namespace. The three enums
+  (`Request::method`, `Run::type`, `Run::status`) default to what
+  `database.cpp`'s `row_extractor` already falls back to for a stored value it
+  cannot parse, so the struct and the reader agree about an unset field rather
+  than each picking an answer. A default is not a substitute for setting the
+  field: it bounds what a wrong insert can persist, and stops an indeterminate
+  `bool` or enum being read at all.
+- **A whole-tree tidy measurement passes `-header-filter` itself** (#1013). The
+  `HeaderFilterRegex` in `engine/.clang-tidy` is *not* read by a
+  `run-clang-tidy` invocation that omits the command-line flag, so a scan
+  without it reports nothing found in a header - which is how 50 findings went
+  uncounted through four measurement rounds. Deduplicate the result by
+  (file, line, column, check): a header finding is reported once per including
+  translation unit. `docs/engine/building.md` carries the command.
 - Formatter: clang-format, **19 exactly** (`.clang-format` at repo root; the
   version is pinned because 39 of the 285 engine sources format differently
   under 18). **A difference is a failure now** (#886): the `Engine formatting`
--- a/engine/include/vayu/core/load_strategy.hpp
+++ b/engine/include/vayu/core/load_strategy.hpp
@@ -110,7 +110,15 @@ const ResultAnnotations& annotations = {});
  */
 class LoadStrategy {
     public:
+    LoadStrategy ()          = default;
     virtual ~LoadStrategy () = default;
+    // Deleted rather than protected: nothing copies a strategy - they are held
+    // by pointer and run in place - and deleting is what stops a derived one
+    // being sliced back to this interface.
+    LoadStrategy (const LoadStrategy&)            = delete;
+    LoadStrategy& operator= (const LoadStrategy&) = delete;
+    LoadStrategy (LoadStrategy&&)                 = delete;
+    LoadStrategy& operator= (LoadStrategy&&)      = delete;
 
     /**
      * @brief Execute the load test strategy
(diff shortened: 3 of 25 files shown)


athrvk:
**The green checks on this PR do not mean the engine was built or tested.** Worth stating plainly before anyone reads them as coverage.

`pr-tests.yml` triggers on `pull_request: branches: [master]`, and this PR's base is `claude/bold-keller-8rn0i1`. So the only checks that ran are `Conventional title` and `Apply path labels`. **No engine build, no ctest, no clang-tidy gate and no `Engine formatting` on any platform** - the three-leg matrix never started.

That is a consequence of stacking, not a failure, and it resolves itself: when #1014 merges, GitHub retargets this PR to `master` and the full matrix runs on the next event. Nothing needs doing until then.

What stands behind the diff in the meantime, from this session:

- `python build.py -e -t` and the full `ctest --preset linux-dev` - **2400 pass, 0 fail**, on the merged head.
- The ten wave-3 checks rescanned to **zero** over the 207 non-vendor translation units with `-header-filter` on.
- The CI changed-lines gate reproduced locally over the whole stack against the full config. It found one finding - `Logger`'s deleted members sitting in a private section (`modernize-use-equals-delete`) - which is the third commit here.
- `clang-format-19 --dry-run -Werror` and `mkdocs build --strict` clean.

**All of that is Linux only**, so the two things a reviewer should weigh as unverified until retarget are compilation under MSVC and AppleClang:

- `js_json.hpp` swapping `<cstdio>` for `<format>`. `std::format` is already used in `routes.hpp` and `mock_server.cpp`, so all three legs compile it today, and neither of this header's two includers (`openapi_drafts.cpp`, `import_document.cpp`) uses a `<cstdio>` function - checked, rather than assumed, because a dropped transitive include is exactly the shape that breaks one platform only.
- The suppressed implicit default constructor that declaring the four special members causes. Standard C++ rather than a dialect question, and the local build named all three sites (`TransferData`, `RunManager`, `LoadStrategy`) rather than leaving them to be guessed at.

---

PR #1022, opened 2026-08-26
perf(engine): tidy paydown 4 - clear performance-* and avoid-c-arrays, promote the gates to whole-file scope - #946

athrvk:
## Description

Wave 4 of #928 - the closing wave.

**Plan.** Root cause: the last two enabled-family backlogs (`performance-*`, `modernize-avoid-c-arrays`) were still standing, and the CI/hook tidy gates were still changed-lines scoped - a shape that existed only to quarantine a backlog that no longer exists. Approach: re-measure the batch honestly first (the wave-0 "70 + 25" predates #1013's `-header-filter` lesson, so it never counted headers - the true figure on the current head is **107 unique findings over 205 non-vendor translation units**, deduped by (file, line, column, check), with `performance-enum-size` hiding 32 of them in 22 headers), fix all 107 per the established per-check disciplines, then promote the gates to whole-file scope and record the decision in `docs/engine/building.md`. Alternative rejected: keeping the changed-lines gate with the backlog at zero - declined because the waves measured exactly what it cannot see (#979's retyped constant created a finding on a line its diff never touched, and #946's batch-4 measurement found 98 findings standing in the very files a green PR opened), and the only argument for line scoping - not failing PRs for code they did not write - expired with the backlog. **This PR's own CI then sharpened that scoping rule into a per-leg one** - see the last section below.

**Part 1 (`99ebec5`)** - the findings, by check:
- `performance-enum-size` (32): every enum takes the `std::uint8_t` base its value set fits.
- `performance-inefficient-string-concatenation` (21): message-building `operator+` chains become `std::format` (already in engine use), or sequential appends where a loop builds a line.
- `performance-unnecessary-value-param` (20): the owner's "CPU related" ask landing - `shared_ptr<RunContext>` and result/config parameters on the load path (`handle_result`, `execute_load_test`, `execute_scenario_run`, `run_auth_refresh`, `collect_monitor`) and per-completion callbacks become const references; a `shared_ptr` copy is an atomic refcount round-trip per completion. `register_run` keeps its by-value parameter and moves it into the map.
- `performance-inefficient-vector-operation` (19): `reserve` before every `push_back` loop.
- `modernize-avoid-c-arrays` (5): fixed tables become `std::to_array`, so the count is not restated (most of the baseline's 25 had already gone with #1020's bounds work).
- `performance-unnecessary-copy-initialization` (4), `performance-no-automatic-move` (3), `performance-faster-string-find` (2), `performance-move-const-arg` (1): mechanical.

[...]

Issue #1021, opened 2026-08-26
engine(tooling): decide re-enabling readability-function-cognitive-complexity now that #928's re-enable condition is met


athrvk:
Split out of #946 (the #928 close-out), which is the commit that makes this decidable rather than the place to decide it.

## Why now

`readability-function-cognitive-complexity` was declined in #940 (closing #929) with a condition attached, recorded in both `engine/.clang-tidy` and `docs/engine/building.md`: **#928's completion *is* the re-enable condition**, because the check anchors on the function declaration - which a changed-lines gate could never honour - and becomes honourable the moment nothing pre-existing is left for it to anchor on. #946 completed #928: the tree is at zero on the decided config and both gates are whole-file scoped, so a newly enabled check's findings would land on exactly the pull requests that edit the offending functions.

## What re-enabling costs

The condition being met does not make the enable free. At the last count (recorded in `engine/.clang-tidy`), ten functions in the routes layer are over the threshold, three far over (`spec_sync_response` 208, `register_execution_routes` 168, a `spec_sync` lambda 126). Enabling the check surfaces those as findings in whichever pull request next touches those files - under the whole-file gate that is a refactor demanded of an unrelated diff, unless the refactors land first or each function gets a NOLINT with a reason.

## The decision to make

One of:
1. **Refactor the ten functions first**, then enable - the clean path, sized like a small wave of #928.
2. **Enable with per-function NOLINTs** naming this issue - honest but noisy, and the NOLINTs invite staying forever.
3. **Decline permanently** with the reason updated in `engine/.clang-tidy` - defensible if the routes-layer shape is judged idiomatic for its job.

Owner's call. Until decided, the check stays off and the reason beside the disable in `engine/.clang-tidy` points here.

## Acceptance criteria

- [ ] The decision is made and recorded in `engine/.clang-tidy` beside the check.
- [ ] If enabled: the tree lints clean with it on, and `docs/engine/building.md`'s Static Analysis section reflects the enable.

athrvk:
Claiming this issue - PR will follow from this run.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1024, opened 2026-08-26
refactor(engine): route libcurl's variadic calls through one typed primitive and promote the Windows lint leg

### LINKED ISSUE #1023: engine(tooling): measure and pay down the Windows-leg tidy backlog, then promote that leg to whole-file scope
Split out of #946, whose whole-file gate promotion surfaced this on its own PR (#1022): the first whole-file lint ever run on the Windows leg reported **~85 findings in the touched files alone, none on a line the diff wrote**. #928's "zero findings tree-wide" was measured on Linux with clang-tidy 19 - the only toolchain any wave ever scanned with - so the Windows leg has a backlog of its own that no Linux scan could see. #946 therefore shipped the promotion **on the measured leg only**: Linux lints whole changed files; Windows stays at changed-lines scope (the pre-promotion mechanism, `clang-tidy-diff.py`) until this issue zeroes what it can see.

## What the Windows leg sees that Linux cannot

Measured by PR #1022's failing run (clang-tidy 20.1.8, MSVC 14.51 headers), three distinct classes:

1. **`cppcoreguidelines-pro-type-vararg` on every `curl_easy_setopt`** (~57 in `src/http/event_loop/curl_utils.cpp`, ~14 in `src/http/sse_stream.cpp`, plus `tests/bounds_primitives_test.cpp:113`, 4 in `tests/client_certificates_test.cpp`, 2 in `tests/transport_policy_test.cpp`). On Linux/GCC, curl's `typecheck-gcc.h` macro wraps the call and the check stays silent; on MSVC the plain vararg prototype is what clang-tidy sees. The likely shape of the fix is one typed `set_opt` wrapper owning the single NOLINT (the `curl_error_buffer.hpp` pattern: a primitive owning the rule), migrated mechanically - not ~80 per-site NOLINTs.
2. **Checks that exist only in clang-tidy 20** - `modernize-use-integer-sign-comparison` fired at `src/http/sse_stream.cpp:254`. The family globs in `engine/.clang-tidy` enable whatever checks the binary ships, and the Windows image runs 20.1.8 while the floor (and the measured zero) is 19. Decide per new check: fix the findings, or disable with a written reason until the floor moves to 20.
3. **Platform-conditional and platform-divergent findings** - `bugprone-exception-escape` at `src/http/routes/execution.cpp:1141`, `src/http/routes/inbox.cpp:976`, `src/http/sse_stream.cpp:559` and `:600` (MSVC STL noexcept surfaces differ), and `cppcoreguidelines-owning-memory` at `tests/transport_policy_test.cpp:154` (in Windows-only code Linux lints never compile).

A fourth class - a `clang-analyzer` finding inside MSVC's own `xfilesystem_abi.h`, reported because `HeaderFilterRegex` `.*/(src|include|tests)/.*` matches MSVC's `include` directory - was fixed in #946 itself by anchoring the regex to `engine/`, so it is not part of this backlog.

## The work

1. Measure: a whole-file scan of all ~205 TUs **on the Windows toolchain** (clang-tidy 20, MSVC compile commands), deduped by (file, line, column, check) - the honest denominator this leg never had. The ~85 above are one PR's touched files, i.e. a floor.
2. Pay it down with the same read-never-bulk-fix discipline as the #928 waves.
3. Promote the Windows leg: replace its `clang-tidy-diff.py` changed-lines branch in `Lint changed engine sources` with the whole-file invocation the Linux leg uses, and record it in `docs/engine/building.md`.

The pre-commit hook is whole-file on every platform already; a Windows contributor with a local clang-tidy 20 may see these findings before this issue closes ( `git commit --no-verify` is the documented escape, and CI remains the gate of record).

## Acceptance criteria

- [ ] The Windows-toolchain whole-file scan is recorded (the real denominator and count).
- [ ] Zero findings on that scan, by fix or by NOLINT/config-disable with a written reason each.
- [ ] The Windows lint branch is whole-file, and `docs/engine/building.md`'s Static Analysis section records the completed promotion for both legs.

### LINKED ISSUE #1029: engine: duplicated `return matched;` left by the #1021 transport split
## Problem

`apply_client_certificate` in `engine/src/http/event_loop/curl_utils.cpp` ends with the same statement twice:

```cpp
    set_opt<CURLOPT_KEYPASSWD> (curl, matched != nullptr && !matched->passphrase.empty () ?
    matched->passphrase.c_str () :
    static_cast<const char*> (nullptr));
    return matched;
    return matched;
}
```

The second is unreachable. Behaviour is unaffected - this is dead code, not a wrong answer.

## Where it came from

`467c90b` ("refactor(engine): split the transport options, the cert rules and the token cache path - #1021"), on master via [#1027](https://github.com/athrvk/vayu/pull/1027). The split lifted the client-certificate block out of `apply_transport_policy` into its own function and duplicated the trailing return in the process.

Confirmed against the commit before that refactor landed:

| Ref | occurrences of `return matched;` |
|---|---|
| `cad2954` (pre-#1021) | 1 |
| `39a4a00` (master today) | 2, adjacent |

## Blast radius

**Isolated.** I scanned all 43 C++ files that changed between `cad2954` and `39a4a00` for consecutive identical `return` / `break` / `continue` statements. This is the only hit, so it is a one-off slip rather than a pattern the refactor repeated.

## Why it is filed rather than fixed in passing

Found while merging master into `claude/issue-1023-djiaeu` for #1023 ([#1024](https://github.com/athrvk/vayu/pull/1024)), which had to resolve a conflict in this exact function - master's split versus that branch's `set_opt<>` migration. The line is master's, not that PR's, and quietly deleting it inside a merge-conflict resolution would bury a real (if harmless) defect in a diff nobody would read it out of. Left alone deliberately so it gets its own one-line change.

If the whole-tree clang-tidy scan on [#1024](https://github.com/athrvk/vayu/pull/1024) turns out to flag it, it becomes that PR's to fix and this issue closes with it - none of the enabled checks obviously covers an unreachable duplicate statement, so that is not the expectation.

## Fix

Delete the second `return matched;`.

## Acceptance criteria

1. `apply_client_certificate` ends with a single `return matched;`.
2. `ctest --preset linux-dev` still green (no behaviour change is expected, so this is a regression check, not a new test - a test asserting unreachable code was removed would be asserting nothing).

## Sequencing

Independent - land whenever. Trivially conflicts with [#1024](https://github.com/athrvk/vayu/pull/1024) while that is open, since it rewrites every `curl_easy_setopt` line in this function; easiest after that merges.

athrvk:
## Description

The clang-tidy gate linted whole changed files on Linux since #946 and changed *lines* on Windows, because the "zero findings tree-wide" #928 established was measured on one toolchain and #946's own pull request (#1022) proved it does not transfer: the first whole-file lint ever run on Windows reported ~85 findings in the touched files alone, none on a line the diff wrote. This pays that backlog down, collapses the two branches into one, and adds the whole-tree scan that is the only thing able to say whether a zero is real.

**The plan, and the alternative I rejected.** Root cause of the bulk of the backlog is that `curl_easy_setopt`, `curl_multi_setopt` and `curl_easy_getinfo` are variadic, and 121 sites called them directly - every one a `cppcoreguidelines-pro-type-vararg` finding. The obvious approach, ~120 per-site `NOLINT`s, buys silence and nothing else, and this repository already has the better pattern: a primitive that owns a rule (`curl_error_buffer.hpp` owning `CURLOPT_ERRORBUFFER`). So `vayu/http/curl_options.hpp` takes the option or info constant as a *template* argument, decodes the argument category libcurl encodes in the constant itself (`CURLOPTTYPE_* + index`; `CURLINFO_TYPEMASK`), and `static_assert`s the value against it - which means the type checking curl documents but cannot enforce in C++ now happens on all three platforms, not just the one leg that could see the finding.

**Correcting the issue's stated cause.** #1023 attributes the Linux blindness to curl's `typecheck-gcc.h` macro. That macro is guarded `#if !defined(__cplusplus)` - "the typechecker does not work in C++ (yet)" - and has never applied to this codebase. What C++ actually gets is `curl_exactly_three_arguments`, an arity check, defined only `#if defined(__STDC__) && (__STDC__ >= 1)`. GCC and Clang define `__STDC__`; MSVC does not, absent `/Za`. So the identical source is a call inside a system-header macro on Linux, which clang-tidy attributes to the header and drops, and a plain call on Windows. The practical payoff: **any toolchain reproduces the Windows reading** with `clang-tidy --extra-arg=-U__STDC__ --extra-arg=-Wno-builtin-macro-redefined`, and it is now written down in `docs/engine/building.md`.

**Throughput.** Unchanged by construction: each wrapper is a header-inline forward to the same libcurl call with the same arguments, the `if constexpr` picks its branch at compile time, and the casts are on constants the call site already spelled.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## The measurement, and two false greens on the way to it

A gate can only speak about what a change touches, so the denominator this work needs is a whole-tree scan. `Engine tidy scan` (`.github/workflows/engine-tidy-scan.yml`) builds the tree, regenerates the compile database without the precompiled header, lints every non-vendor translation unit whole-file with `-header-filter` on the command line (#1013's lesson), and reports through `.github/dedupe-tidy-findings.py`, deduplicated by (file, line, column, check).

**It reported a clean tree twice when the tree was not clean, and both are worth recording** - this is the one job whose entire purpose is to be the number everything else is measured against.

1. The unit list is written by a Python whose stdout is a Windows text stream, so every path arrived CRLF-terminated and all 206 units answered `no such file or directory`. What let that pass: clang-tidy reports a driver failure with **no location**, and the report matched only the located `path:line:col:` form. 821 error lines counted as zero.
2. Fixed that, and the next Windows run reported `0 findings over 206 translation units` for a log holding **679**. The `LOCATED` pattern spelled its file component with colons forbidden, since a colon delimits the line number - and every Windows path is `D:\a\vayu\...`, where the drive letter *is* a colon inside the path. Not one located finding on that leg could ever match. Linux paths have no drive prefix, which is why the identical pattern read that leg correctly and this sat unseen through every round.

Fixing the pattern is not the fix, because the next unreadable diagnostic will take a shape nobody anticipated either. `DIAGNOSTIC_SHAPED` now matches anything that merely *looks* like a diagnostic, and when the log holds such lines and the structured patterns parsed none into a finding, the script fails naming the first few it could not read. Mutation-checked both directions against the real 679-line log, and against a synthetic log whose only finding is out-of-tree.

Separately: a finding in a file the repository does not contain is not the repository's finding. `clang-analyzer-*` diagnostics do not honour `-header-filter`, so a walk that inlines into MSVC's STL reports there. Those are now excluded from the count and **named in the report** rather than dropped silently.

## Testing

Local, on `6192b29`:

| Check | Result |
|-------|--------|
| `python build.py -e -t` | clean |
| `ctest --preset linux-dev` | **2446 / 2446 passed**, 0 failed |
| `clang-format 19`, whole engine tree | clean |
| `clang-tidy 19` on every changed translation unit it can compile | clean |
| `clang-tidy --verify-config` | clean |
| `ruff check` | clean |
| `mkdocs build --strict` | clean |
| tree-wide grep for the three variadic names | 2 hits, both prose |

No app file is in this diff, so the renderer suite is untouched.

### Whole-tree scan - run 18 on `6192b29`

| Leg | Findings | TUs | Unfinished |
|-----|----------|-----|------------|
| `ubuntu-latest` | 1 | 210 | 0 |
| `windows-latest` | 2 | 210 | 0 |

Both verified against their own logs rather than taken on trust. **Every check this pull request set out to eliminate reads zero**, counted in the raw Windows log:

| Check | Count |
|-------|-------|
| `cppcoreguidelines-pro-type-vararg` | 0 |
| `bugprone-exception-escape` | 0 |
| `clang-analyzer-optin.core.EnumCastOutOfRange` | 0 |
| `modernize-use-integer-sign-comparison` | 0 |
| `bugprone-implicit-widening-of-multiplication-result` | 0 |
| `cppcoreguidelines-owning-memory` | 0 |

**The scan is not at zero, and the remaining findings are not this pull request's.** Both are `readability-function-cognitive-complexity` on code this branch never opens - `openapi_walk.hpp:259` (`walk_operations`, 37) and `load_strategy.cpp:422` (`run_rate_limited`, 26, Windows only). Both files are byte-identical to master and appear in zero files of this diff. #1021 could not have seen either: the first is in a **header**, which has no compile-database entry and so was invisible to that issue's 68-function sweep, and the second is over threshold only under MSVC, where the `#ifdef _WIN32` branches a Linux preprocessor removes are counted.

Headers and the Windows toolchain are exactly the two gaps this scan exists to cover, and it found one of each on its first complete run. Filed as #1033 and, per the owner's decision, **left for that issue rather than refactored here** - #1021's own note in `engine/.clang-tidy` paid its cost up front specifically to avoid "handing an unrelated pull request somebody else's refactor", and `walk_operations` is the single walk `engine/CLAUDE.md` credits with keeping the operation index, the schema index and the request drafts from disagreeing.

**So acceptance criterion 1 is not met by this pull request, deliberately.** It lands with both scan legs red, and the weekly cron will file a `tidy-scan-failure` issue until #1033 is fixed. That is the scan reporting honestly.

## Checklist

Docs updated in the same commits as the code they describe: `docs/engine/building.md` (both legs promoted, the scan workflow, what each backlog class cost, why only one leg could see the vararg class), `CONTRIBUTING.md`, `engine/CLAUDE.md`, `scripts/pre-commit`.

- **The cause of the one-legged finding is documented as `__STDC__`, not the GCC typecheck macro.** The issue's explanation does not hold for libcurl 8.21 in C++; the anchor drifted, the finding did not.
- **`curl_easy_getinfo` is in scope too.** The issue names only the setters, but `getinfo` is variadic by the same mechanism. Its wrapper is the stronger half: libcurl stores *through* the pointer, so an `int*` where `CURLINFO_RESPONSE_CODE` writes a `long` is a write past the object, and `get_info` makes that a compile error.
- **Two checks declined, each with its reasoning in `engine/.clang-tidy`.** `bugprone-exception-escape` fires only on a `throw` it can see, so it answers about the standard library being read rather than about this code - 33 findings on Windows against 0 on Linux, same commit, landing on implicit destructors of ordinary aggregates, and reporting `cli.cpp`'s `main` even though it already catches `...`. The rule it was reporting is real and was fixed first: `~SseStreamManager`, `~Logger` and `daemon.cpp`'s `main` are total, and `engine/CLAUDE.md` now carries the rule so the shape survives the check. `clang-analyzer-optin.core.EnumCastOutOfRange` reports the ordinary bitmask-enum idiom inside MSVC's own `_BITMASK_OPS`, where there is no site to put a `NOLINT` on and no filter that reaches it.
- **Out-of-tree findings are excluded from the count**, named in the report rather than dropped.
- **No source-scanning guard test.** `bounds_primitives_test.cpp`-style scans exist because the gates lint only files a change touches. That does not apply here: a new raw `curl_easy_setopt` can only arrive in a file the change touches, and both legs now whole-file lint exactly those. The gate is the guard.
- **The scan gained a narrow `pull_request` trigger** alongside `workflow_dispatch`, which is also what lets it run on a branch at all - dispatch cannot fire on a ref whose workflow file is not yet on the default branch. Note the practical consequence: those path filters are evaluated against the pull request's *cumulative* diff, so every push here re-triggers the scan and cancels the in-flight one. The measurement needs the branch to stop moving for ~2.5 hours.
- **The total-`main` wrapper on `daemon.cpp` no longer has a demonstration behind it.** It was justified by `--port notanumber` aborting with exit 134; master fixed that upstream in #1028 - an issue this branch filed while doing the work. The wrapper stays as defence in depth over three constructions that still throw where nothing catches (`Logger::init`, the `Database` object, `Server`), and the comment at the site says exactly that rather than citing a bug the binary no longer has.

## Related Issues

Closes #1023. Closes #1029. Filed along the way: #1028 (fixed upstream), #1031 (found from it, fixed upstream), #1033 (the two remaining scan findings).


athrvk:
**No CI has run on this PR, and the cause is outside it.** Recording what I established so nobody re-derives it.

No `pull_request`-event workflow run has been created for this branch at all - not `PR CI - run tests`, not `PR title`, not `Engine tidy scan`. `total_count` is 0 for the branch on that event. Two things rule out the explanations I tried:

- **Not the app-token recursion guard.** That would suppress the `opened` event but not a `synchronize` from an ordinary `git push`. I pushed `1310289` from git and it created no run either.
- **Not a disabled workflow.** `pr-tests.yml` reports `state: active`.

What it does look like is a standing Actions queue problem on the repository rather than anything in this diff: there are runs stuck in `queued` from **2026-08-06** and **2026-08-19** (`PR CI - run tests`, `PR title`, `Docs site`), alongside three of my own `Engine tidy scan` runs from today that have sat `queued` since 15:16-15:20 and that the API refuses to cancel with `409 Cannot cancel a workflow run that has not been queued yet`. `push`-event runs on this same branch did start and complete normally today, so runners are being allocated - it is the `pull_request` path that is producing nothing.

I am not going to force it: an empty commit or a close-and-reopen are the two things that would kick it and both are off-limits, and I have spent no re-run on this. **What I need is for someone with repository admin to clear the stuck queued runs** (or confirm an Actions spending/concurrency limit is in force), after which a push here will pick up CI normally.

Everything verifiable without CI is green on `1310289`: engine `ctest --preset linux-dev` **2412/2412** (three separate runs), app `pnpm test` **5702/5702** across 531 files, `pnpm type-check` clean, `mkdocs build --strict` clean, and a whole-tree `clang-tidy-20` scan of all 206 translation units with `--extra-arg=-U__STDC__` (which reproduces the Windows reading of libcurl's headers) reporting zero `pro-type-vararg`.

The one thing still genuinely unproven is the Windows whole-tree scan's finding count - issue #1023's first acceptance criterion - because that needs the MSVC toolchain and so needs this queue moving. The denominator is established at **206 non-vendor translation units**; the count is not. I have said so in the PR body rather than claiming it.

---

athrvk:
**Correcting my previous comment: no admin action is needed. I called it too early.**

Runs *were* created for both of my pushes - [run 6](https://github.com/athrvk/vayu/actions/runs/32987892279) at 16:21 and [run 7](https://github.com/athrvk/vayu/actions/runs/32988025183) at 16:22 - roughly five minutes after the pushes landed, with run 6 correctly cancelled by the concurrency group in favour of the newer one. Run 7 is executing now. I queried the API repeatedly in the window before they appeared and read "no runs" as "no runs will be created", which was a conclusion the evidence did not support.

What was actually happening is ordinary queue latency, not a jam: Windows runners on this repo took ~4 minutes today to allocate and the run-creation itself lagged the push by minutes. The stuck `queued` runs from 2026-08-06 and 2026-08-19 are real and still sitting there, but they are not blocking anything - please disregard the request to clear them.

The one part of that comment that still holds: **no `pull_request`-event run has appeared for this branch yet** (`PR CI - run tests` has `total_count` 0 for it). Given what I just got wrong about latency, I am not going to theorise about why until I have waited properly. Every run so far has been `push`-triggered.

Sorry for the noise. The substantive status is unchanged and is in the PR body: everything locally verifiable is green, and the Windows whole-tree finding count is still the one open item.

---

athrvk:
**`App on windows-latest (shard 1/2)` is red, and it is not this PR's.** Recording it per the drive-to-green rule rather than leaving it silent.

Two tests timed out at the 5000ms default:

- `LoadTestConfigDialog.test.tsx > load profile → fields > shows rate and duration for Constant RPS, and no connection count`
- `send-with-row.test.tsx > picking a row > reads the file only when the picker is opened`

Both are `Test timed out in 5000ms`, not assertion failures. What says it is not the diff:

- **This pull request changes no `app/` file at all.** `git diff --name-only master...HEAD` is 17 files: 9 under `engine/`, 3 workflow/script files under `.github/`, and 4 docs. Nothing the failing tests import is in it.
- **The same suite passes on every other leg of the same run** - `App on ubuntu-latest`, `App on macos-latest`, and `App on windows-latest (shard 2/2)` are all green on this exact commit.
- **The shard that failed is the slow one.** Its own summary reads `Duration 233.67s (… import 211.42s, environment 260.49s)`, and `Setup pnpm` alone took 66s against 24s on the sibling shard. Two 5s timeouts on a runner moving at that speed is the shape of a starved runner, not of a behaviour change.

I hit the same thing locally, which is worth putting on the record because it points at the same cause rather than at this branch: a full `pnpm test` run on my 4-core box **while it was simultaneously running a C++ build and a 206-unit clang-tidy scan** failed exactly one test - a different test in that same `send-with-row.test.tsx` file - and passed 41/41 when that file was run alone. Re-run on an idle machine the whole suite passed **5702/5702 across 531 files**. That file is timeout-prone under CPU pressure on more than one machine.

**No fix ported, because none exists to port** - there is no open PR or landed commit addressing these two timeouts. The fix within reach would be raising the timeout on those tests, and I am not pushing that: it is `app/` code this PR has no business touching, `LoadTestConfigDialog.test.tsx` is not even a file I could claim adjacency to, and widening an engine-tooling PR into the app test suite to quiet a runner-speed problem is the wrong trade. If you want it fixed, the honest patch is a per-test `{ timeout: 15000 }` on those two cases (or a raised `testTimeout` for the Windows shards in `pr-tests.yml`), and it belongs in its own change with its own reasoning.

I have not spent a re-run yet. The engine jobs on this commit are still building, and the Windows one carries the whole-file lint this PR exists to enable - if that comes back red I will be pushing a fix anyway, which re-runs this shard for free. I will re-run this job only if no such push turns out to be needed.

---

athrvk:
`Whole-tree clang-tidy on windows-latest` is failing on `2860ab0`, and the failure is correct - it is this PR's scan finally telling the truth. Recording what happened, because the previous run's "success" was worse than this failure.

**Run 13 reported `0 findings over 206 translation units` on Windows. That was a false green.** The log behind it held 679 error lines.

The cause was one absent group in `.github/dedupe-tidy-findings.py`. `LOCATED` spelled its file component `[^:]*[^:\s][^:]*` - colons forbidden, since a colon is the delimiter before the line number. Every path on that leg is `D:\a\vayu\...`, and a drive letter *is* a colon inside the path, so not one located finding on Windows could ever match. Linux paths carry no drive prefix, which is why the identical pattern read that leg correctly and this sat unseen through every round of this work.

This is the second false green from this script. The first (matching only located diagnostics, so 821 driver-failure lines counted as zero) is already written up in the PR body. Two instances of one shape: a real diagnostic took a form a regex did not anticipate, and a report that counts by regex cannot distinguish "nothing was found" from "nothing was read".

So `2860ab0` fixes the pattern **and** stops trusting the pattern. `DIAGNOSTIC_SHAPED` matches anything that merely looks like a diagnostic; when the log holds such lines and the structured patterns parsed none of them into a finding, the script fails naming the first few it could not read. Mutation-checked against the very log that produced the false green - reverting the drive-letter group alone puts the count back to zero, and the guard then exits 1 with `the log holds 679 diagnostic-shaped line(s) and this script parsed none of them`.

**The real measurement, both legs, on `2860ab0`:**

| Leg | Findings | TUs | Unfinished |
|-----|----------|-----|------------|
| `ubuntu-latest` | **0** | 208 | 0 |
| `windows-latest` | **52** | 208 | 0 |

Linux is a genuine zero and independently corroborated: 238 log lines, not one diagnostic-shaped line in the whole file. Windows carries a real 52-finding backlog:

| Check | Findings |
|-------|----------|
| `bugprone-exception-escape` | 33 |
| `modernize-use-integer-sign-comparison` | 4 |
| `bugprone-implicit-widening-of-multiplication-result` | 3 |
| `clang-diagnostic-missing-field-initializers` | 2 |
| `cppcoreguidelines-owning-memory` | 2 |
| `cppcoreguidelines-pro-type-reinterpret-cast` | 2 |
| seven others | 1 each |

**What this means for the PR.** The vararg work stands - `cppcoreguidelines-pro-type-vararg` does not appear anywhere in either leg's report, which is the thing #1023 set out to eliminate and is now measured rather than asserted. What is *not* met is acceptance criterion 1, the zero-finding scan, and it is further from met than the false green made it look.

I am not pushing fixes for these 52 in this round, because 33 of them are one question that deserves an answer rather than 33 suppressions. `bugprone-exception-escape` fires on the implicit destructors of ordinary platform-neutral aggregates (`ResponseSample`, `Request`, `Response`, `DetailedReport` - structs of `std::string` and maps) on MSVC and on none of them under libstdc++, same check enabled on both legs. That is the throw-visibility phenomenon this PR already documented for the SSE thread lambda, at 33x the scale: MSVC's STL inlines throws where the analyzer reaches them and libstdc++ hides them behind out-of-line functions. You cannot write those destructors more correctly - they are already implicitly `noexcept` - so the honest options are a documented check-level decision or 33 site suppressions, and picking between those is a call about what the gate promises, not a defect fix. @athrvk I have put the options to you directly; say which way and I will implement it.

The other 19 are ordinary paydown of exactly the kind this PR is already doing (three `32UL * 1024 * 1024` widenings in `constants.hpp`, four `std::cmp_*` sites, a `~Logger` that can throw, two `missing-field-initializers` in `platform_windows.cpp`), and I will take them regardless of how the 33 are decided.

---

athrvk:
`Engine tidy scan` run 18 on `6192b29` completed on both legs. **This is the first complete, trustworthy whole-tree measurement this PR has produced**, and it is worth reading carefully because of what it found.

| Leg | Findings | TUs | Unfinished |
|-----|----------|-----|------------|
| `ubuntu-latest` | 1 | 210 | 0 |
| `windows-latest` | 2 | 210 | 0 |

Both reports verified against their own logs rather than taken on trust (two false greens came out of this script earlier): 4 diagnostic lines on Windows and 3 on Linux, deduplicating to 2 and 1, no `Error while processing`.

**Every class this pull request set out to eliminate reads zero**, checked against the raw Windows log rather than the report:

| Check | Count |
|-------|-------|
| `cppcoreguidelines-pro-type-vararg` | 0 |
| `bugprone-exception-escape` | 0 |
| `clang-analyzer-optin.core.EnumCastOutOfRange` | 0 |
| `modernize-use-integer-sign-comparison` | 0 |
| `bugprone-implicit-widening-of-multiplication-result` | 0 |
| `cppcoreguidelines-owning-memory` | 0 |

So the vararg work, the 19-finding paydown and both check declines are done and measured.

**What is left is two functions, and neither is this pull request's.**

| File | Function | Complexity | Leg |
|------|----------|-----------|-----|
| `engine/src/core/openapi_walk.hpp:259` | `walk_operations` | 37 (threshold 25) | both |
| `engine/src/core/load_strategy.cpp:422` | `run_rate_limited` | 26 (threshold 25) | Windows only |

Both files are **byte-identical to master** (`git diff origin/master HEAD -- <file>` is empty) and appear in **zero** files of this pull request's diff. `openapi_walk.hpp` last changed in #946.

**Why #1021 could not have seen either, which is the part worth keeping.** That issue's note in `engine/.clang-tidy` records the enable as measured over "every over-threshold function in `engine/src` ... 68 functions". That measurement is taken over the *compile database* - and `openapi_walk.hpp` is a **header**, which has no entry in it. The gate lints translation units and never a header as an input (#940), so a header's complexity was structurally invisible to that paydown. `run_rate_limited` is the other blind spot: it is 26 only under MSVC, where the `#ifdef _WIN32` branches a Linux run preprocesses away are counted; #1021 measured on Linux, where it sits at or under the threshold.

Headers, and the Windows toolchain. Those are the exact two gaps this scan was built to cover, and it found one of each on its first complete run. That is the strongest evidence so far that the weekly scan earns its cost - but it also means **this pull request cannot reach a zero-finding scan without touching code it does not own**, and #1021's own note argues against exactly that: it paid its cost up front "rather than handing an unrelated pull request somebody else's refactor."

I am not refactoring either function here on my own judgement. @athrvk I have put the options to you directly - refactor here, suppress at the site with the reasoning and file the refactor, or land with the scan red and fix it separately - and will implement whichever you pick. Nothing is being pushed to this branch until then, since any push restarts the two-and-a-half-hour scan.

---

Issue #990, opened 2026-08-25
engine(db): freed pages never return to disk - guarded VACUUM after the startup prune


athrvk:
From the 2026-08-25 operational sweep, verified at master `96385b2`. Cheap hygiene, not urgent - the DB file never shrinks.

## Problem

Run pruning deletes rows aggressively and correctly (`engine/src/db/database.cpp:1781` `prune_runs`, batched cascades, startup + post-run), but the freed pages only ever go to SQLite's freelist: there is no `VACUUM`, no `PRAGMA auto_vacuum`, no `incremental_vacuum` anywhere (`rg -i vacuum engine/src` - the single hit is the comment at `database.cpp:759` saying VACUUM is "deliberately not run here", written about the startup write-lock window). A workspace that once held weeks of heavy runs keeps that high-water-mark file size forever. Not data loss, not correctness - just disk that never comes back, alongside the log growth #985 fixes.

## Fix / solution pathway

Small and guarded, at the one point the deliberate-skip comment already marks:

1. After the startup prune (`database.cpp:844` region), read `PRAGMA freelist_count` and `PRAGMA page_count`. If freelist exceeds a threshold (recommend: >25% of pages AND >10 MiB of reclaimable space - both, so small DBs never bother), run `VACUUM` right there - startup already holds the write lock, which is exactly why the old comment declined to do it mid-flight; that objection does not apply at this point in init.
2. Log the outcome (bytes before/after) at info level so a slow start is explainable.
3. Do NOT switch `auto_vacuum` modes on existing databases (mode change requires a full VACUUM anyway and changes page bookkeeping for every future write) - state this rejected alternative in the PR.
4. Update the `database.cpp:759` comment to point at where reclamation now happens.

Note: #987 (workspace backup) uses `VACUUM INTO`, which compacts the snapshot but not the live file - complementary, not overlapping.

## App changes

None, verified - no surface reads file size.

## Tests

Engine gtest: seed and delete enough rows to cross the threshold, assert the file shrinks after a reopen-init; below-threshold DB is untouched (assert no size change and, via the log or a counter, that VACUUM did not run). `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`. New code clean under the #928 tidy families.

## Docs (same commit)

`docs/engine/db-schema.md` if it discusses file lifecycle (verify); otherwise the code comment is the doc.

## Sequencing

[...]

Issue #988, opened 2026-08-25
engine(db): soft delete for collections and requests - deleted_at, filtered reads, restore and purge endpoints


athrvk:
From the 2026-08-25 UX sweep, verified at master `96385b2`. Engine half of the trash feature; the app half (Trash view + undo toast) is filed separately and depends on this.

## Problem

Deleting a collection or request is irreversible: the confirm dialog is the entire safety net. Verified: no soft-delete model exists anywhere - `rg -ni "\btrash\b|soft.?delete|deleted_at|deletedAt" app/src engine/src` yields no domain hits (the matches are `db/recovery.cpp` and query-retry plumbing, unrelated). Deletes are hard cascades through `useTreeCrud.ts:415-460`'s mutations into the engine's delete routes. Combined with there being no workspace backup until #987 ships, one misclick on a collection's confirm dialog destroys a tree permanently.

## Fix / solution pathway

Additive schema, engine-side only:

1. **Schema**: nullable `deleted_at` (epoch ms) on `Collection` and `Request` rows. sqlite_orm `sync_schema` handles additive nullable columns (`docs/engine/db-schema.md:9-13` - additive is the safe class; destructive is what wipes). Deleting a collection stamps it and its subtree in one transaction; deleting a request stamps the row.
2. **Reads filter**: every listing/tree/search/export query excludes `deleted_at IS NOT NULL`. This is the careful part - enumerate the read sites in `engine/src/db/database.cpp` and the routes that assemble trees; the repo's own "written but never read" discipline applies in reverse here: a read site missed is a ghost row resurfacing. Grep-driven checklist in the PR body.
3. **Endpoints**: `GET /trash` (deleted roots with counts and timestamps), `POST /trash/{id}/restore` (clears the stamp on the subtree; restoring an item whose parent is deleted or gone re-parents to root - state the rule), `DELETE /trash/{id}` (hard purge, the old cascade), and a startup purge of items deleted more than `trashRetentionDays` ago (default 30, config entry, 0 = keep forever) alongside the run-pruning that already runs at startup (`database.cpp:844`).
4. **Out of scope**: environments, examples-as-independent-entities, runs (runs already have retention + delete). State so in the PR.

## App changes

None in this issue beyond compatibility: existing delete flows keep working unchanged (they now soft-delete). The Trash UI is the companion issue. Verify the MCP surface (`docs/engine/mcp.md`) - if any MCP tool lists or deletes collections/requests, it must see filtered reads too; state the check.

[...]

PR #1036, opened 2026-08-27
feat(engine): on-demand workspace backup - VACUUM INTO endpoint, CLI verb, retention - #987

### LINKED ISSUE #987: engine(db): on-demand workspace backup - a VACUUM INTO endpoint, a CLI verb, retention, and a documented restore story
From the 2026-08-25 operational sweep, verified at master `96385b2`.

## Problem

There is no user-facing way to back up a Vayu workspace. Verified: no backup/export endpoint exists (`rg -n "\"/(backup|export|db)" engine/src/http` - only `spec_export.cpp`, which exports an OpenAPI document, not the workspace), no CLI verb, and no documented file-copy story (`docs/engine/architecture.md`'s data-directory section covers structure, not backup; `docs/engine/db-schema.md:18-28` covers corruption recovery only). The only backup in the system is the internal `.bak` written at startup (`engine/src/db/database.cpp:713`), which is overwritten on every clean start and exists for crash recovery, not for the user.

Given the recovery path can start fresh after corruption (#984 hardens it, but the endpoint of that path is still an empty workspace), a user's collections, environments, secrets and run history have no snapshot they control. Copying the SQLite file by hand while the engine runs is not safe (WAL).

## Fix / solution pathway

1. **Engine endpoint** `POST /workspace/backup` (route module pattern: copy the extracted testable-core shape the other routes use, e.g. `engine/src/http/routes/spec_export.cpp` for a small route): runs SQLite's `VACUUM INTO '<data-dir>/backups/vayu-<timestamp>.db'` on the live connection - a consistent, compacted snapshot that is safe under WAL (SQLite 3.27+; the tree ships 3.53). Response: the absolute path, byte size, and timestamp. Refuse (409) while a backup is already running.
2. **Retention**: keep the newest N snapshots in `<data-dir>/backups/` (default 5, config entry `maxBackupsRetained`, 0 = unlimited, seeded like `database.cpp:3025,3036`).
3. **CLI verb**: `vayu-cli backup` calling the endpoint and printing the path (`docs/engine/cli.md` documents the CLI's shape; follow its existing command pattern).
4. **Restore story is documentation, not code**: stop the engine, copy a snapshot over `<data-dir>/db/vayu.db`, delete `-wal`/`-shm`, start. Write it down; do not build a restore endpoint (a live engine overwriting its own open DB is exactly the footgun this avoids).

## App changes

A one-card Settings surface: "Back up workspace" button calling the endpoint and showing the resulting path in a toast, plus the retention config entry rendering via the generic settings rows (`app/src/modules/settings/main/SettingsMain.tsx`). Pattern to copy: `app/src/modules/settings/main/panels/ConnectionTestCard.tsx` (a card that calls an engine endpoint and renders the outcome). Same PR.

## Tests

Engine gtest: a backup file is created, is a valid openable SQLite DB containing a seeded row, retention prunes to N, concurrent-backup refusal returns 409. App vitest for the card (call + toast + error path), `cd app && pnpm test && pnpm type-check && pnpm lint`. Engine suite: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`. New code clean under the #928 tidy families.

## Docs (same commit)

`docs/engine/api-reference.md` (new endpoint), `docs/engine/cli.md` (new verb), `docs/engine/db-schema.md` or `architecture.md`'s data-directory section (backups dir + the restore procedure). `mkdocs build --strict`.

## Sequencing

Touches `engine/src/db/database.cpp` (new method + config seed) and a new route file. **Unblocked: #984 and #985 have both shipped** (recovery hardening and the log-bounding config-block neighbor are on master), and the tidy paydown program (#928/#945/#946) is complete, so there is nothing in flight to rebase past. Land before #988 and #990, which build on this issue's `VACUUM INTO` machinery in the same file region. Note: `readability-function-cognitive-complexity` is enabled since #1021 - keep the new route handler in the extracted testable-core shape.

## Acceptance criteria

- [ ] `POST /workspace/backup` produces a valid consistent snapshot under `<data-dir>/backups/` with retention honoured.
- [ ] `vayu-cli backup` works against a running engine and prints the path.
- [ ] The Settings card triggers it and surfaces the result.
- [ ] The restore procedure is documented; `mkdocs build --strict` clean.

---

### LINKED ISSUE #1037: engine(ci): a dead `return std::nullopt` in read_monitor_block fails the Windows ASan leg, and master has not been measured since it landed
Found by the `asan on windows-latest` leg of the Sanitizers workflow on PR #1036, at master `717da15`.

## Problem

`read_monitor_block` ends with two returns, the second of which cannot run:

```cpp
// engine/src/core/monitor.cpp:146-147
    return read_monitor_series (monitor, limits, out);
    return std::nullopt;
}
```

`git blame` puts both lines in `79b9f124` (2026-08-26, "refactor(engine): name the monitor block's readers and the exposition sample parse - #1021"): extracting `read_monitor_series` turned the tail of the function into a call, and the `return std::nullopt;` that used to end it was left behind. It is the same defect class #1029 fixed in `apply_client_certificate`.

It is dead code, not a wrong answer - `read_monitor_block`'s behaviour is correct - but MSVC refuses it:

```
D:\a\vayu\vayu\engine\src\core\monitor.cpp(147) : error C2220: the following warning is treated as an error
D:\a\vayu\vayu\engine\src\core\monitor.cpp(147) : warning C4702: unreachable code
```

**Why nothing has reported it.** Three things have to line up, and only the Windows ASan leg does all three:

- `C4702` is an optimizer-dependent warning, so it needs a `/Od /Ob0` build. `windows-prod` is the only Windows leg `pr-tests.yml` runs, and `/O2` folds the dead return away - the same asymmetry `engine/CMakeLists.txt:104-110` already records for `C1128`.
- `/WX` is unconditional on MSVC (`engine/CMakeLists.txt:97-101`), so `VAYU_WERROR=OFF` - which `sanitizers.yml` sets deliberately - does not spare it. That comment says the workflow answers "is the engine memory- and thread-correct", not "does it compile warning-free"; on MSVC it cannot help answering both.
- GCC and Clang emit nothing here, so all four Linux/macOS sanitizer legs are green on the same commit.

**And the weekly cron has not fired since.** The last Sanitizers run was 2026-08-25 (green); `79b9f124` landed 2026-08-26; the next scheduled run is Monday 2026-08-31. `sanitizers.yml` also runs per-PR only for a diff touching `engine/CMakeLists.txt`, `engine/CMakePresets.json`, `engine/sanitizers/**` or the workflow itself - so master has carried a red Windows ASan leg for a day with nothing pointed at it, and the next thing to notice would have been Monday's auto-filed issue.

## Fix / solution pathway

Delete line 147. The function already returns `read_monitor_series (...)`, which is the `std::optional<std::string>` every other exit hands back.

```diff
     return read_monitor_series (monitor, limits, out);
-    return std::nullopt;
 }
```

Nothing else in the file has the shape (`rg -n -A1 '^\s+return .*;$' engine/src/core/monitor.cpp` was read through), but the sweep is worth doing across `engine/src` while here, since #1021 split several functions the same way and only this one has been compiled at `/Od` on MSVC since.

## Tests

No new test: the deleted statement is unreachable, so no behavioural test can distinguish before from after - the compiler is the assertion, and the Windows ASan leg going green is the check. `MonitorLimitsTest` / `MonitorRunTest` already cover `read_monitor_block`'s answers and must stay green.

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` proves nothing regressed locally; the leg that proves the fix is `asan on windows-latest`, which any PR touching `engine/CMakeLists.txt` will run - including the fix's own, if it adds a line there, and otherwise via `workflow_dispatch`.

## Docs

None. The mechanism is already written down at `engine/CMakeLists.txt:97-110` (`/WX` unconditional on MSVC; Debug-only diagnostics that Release folds away) and in `sanitizers.yml`'s header.

## Sequencing

Independent of everything currently open - one line in a file no other open issue names. **Land whenever**; it is not blocked and it blocks nothing. It does gate a *green* Windows ASan leg on any PR that touches `engine/CMakeLists.txt`, so #988 and #990 (which add to `engine/src/db/database.cpp` and may touch the source list) will meet it too. #1036 does not carry the fix: the failing check is not a required one, and the rule there is to name an unrelated failure rather than widen a feature PR with someone else's refactor.

Worth considering separately, and deliberately not folded in here: whether `sanitizers.yml` should pass `-DVAYU_WERROR=OFF` *and* an MSVC `/WX-` so its Windows leg stops being a second, weaker warnings gate that only fires once a week. That is a workflow decision with an argument on both sides - the current setup did catch a real piece of dead code - so it wants an owner call, not a drive-by.

## Acceptance criteria

- [ ] `engine/src/core/monitor.cpp` has no unreachable `return`.
- [ ] `asan on windows-latest` builds and passes on a PR that runs it.
- [ ] The engine suite stays green on all three platforms.

athrvk:
## Description

**Root cause.** Everything a Vayu user builds lives in one SQLite file, and there was no copy of it they controlled. Verified at `717da15`: no backup endpoint, no CLI verb, no documented file-copy story. The `.bak` at `database.cpp:713` is not a user backup - it is rewritten on every clean start, so it exists to give a corrupt file something to restore *from*, not to let anyone go back to last week; and since the recovery path can end in an empty workspace (`db-schema.md`'s table), a user whose file goes bad has nothing of their own to fall back on. Copying `vayu.db` by hand while the engine runs is not a workaround: under WAL the `-wal` holds committed transactions the main file does not, so the copy is a database missing its most recent writes.

**Approach.** `POST /workspace/backup` runs SQLite's `VACUUM INTO` - consistent under WAL, compacted, and read-only with respect to the workspace, so nothing here can cost the user the data it is copying. It runs on a **connection of its own** rather than the live one, which is the one place this diverges from the issue's wording ("on the live connection"): sqlite_orm's `get_connection()` is `protected`, so there is no way to run a statement on the connection it holds - and a `VACUUM INTO` of a large workspace would occupy that connection for the whole copy, parking every other endpoint behind a button someone pressed. Under WAL a second reader sees every committed transaction and blocks no writer, so the snapshot is identical and the running engine pays nothing but disk bandwidth. This is stated at the helper.

**Alternative rejected.** sqlite_orm's public `storage.backup_to()` (the `sqlite3_backup` API) would have used the live connection, but it produces a page-for-page copy rather than a compacted one and restarts if the source is written mid-copy - so it is both larger and less predictable than the thing the issue asked for.

Names are UTC and fixed-width (`vayu-YYYYMMDD-HHMMSS-mmm.db`) so retention sorts them chronologically **as text**, rather than trusting an mtime a copy may not preserve; a taken name is stepped over rather than written through, exactly as `quarantine_db_files` does - and `VACUUM INTO` refuses an existing destination, so that step is load-bearing. Retention keeps the newest `maxBackupsRetained` (default 5, `0` = unlimited) and removes **only** files matching that name, so a copy the user put in the directory themselves survives. A second concurrent backup is refused with `409` through one RAII-held slot, since two would each write a whole second database and race each other's pruning. A retention failure cannot demote a snapshot that succeeded.

There is deliberately **no restore endpoint**: a live engine overwriting its own open database is the footgun this feature exists to prevent. The procedure is documented, and the app card keeps it on screen next to the path.

### App changes

- `WorkspaceBackupCard` in Settings > General, beside Data management / Storage paths (engine-held state the user did not type - the same shelf as `CookiesCard`). It keeps the **path** on screen rather than only toasting it, because restoring is a manual copy and a faded toast cannot tell you where the file went.
- A `409` reads as "a backup is already running - it will finish on its own", not as a failure: it is their own earlier request still writing the file.
- `maxBackupsRetained` renders through the generic engine settings rows under Engine > Data & retention; the card points at it.

### Second commit: an unrelated dead return (#1037)

`e6e04692` deletes one unreachable `return std::nullopt;` from `read_monitor_block`, left behind when #1021 extracted `read_monitor_series`. It is not this feature's code and would normally not be in this PR - it is here at the owner's request, because it is what fails `asan on windows-latest` and this PR is the first thing to run that leg since the line landed (`sanitizers.yml` runs per-PR only for a diff touching `engine/CMakeLists.txt`, which this one does; the weekly cron last ran 2026-08-25, a day before `79b9f124`). MSVC's C4702 needs a `/Od /Ob0` build and `/WX` is unconditional there, so `windows-prod` at `/O2` and every GCC/Clang leg stay silent. Reproduced before fixing with `clang++-19 -Wunreachable-code-return`, the nearest analogue available here; a sweep of `engine/{src,include}` for two same-depth returns in a row found no other instance.

## Type of Change

- [x] Bug fix (the second commit only - see above)
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

All run locally on this branch.

- **Engine:** `python build.py -e -t && ctest --preset linux-dev --output-on-failure` - **2459 / 2459 passed, 0 failed** (13 new in `workspace_backup_test.cpp`), re-run after the second commit. The new cases pin what a backup feature gets silently wrong: the snapshot is opened as a `Database` and asserted to still carry its rows; retention keeps the *newest* (pruning the wrong end passes every count-based check) and leaves a `.txt` and a foreign `.db` in the directory alone; two backups in the same millisecond are two files; names sort chronologically across a second boundary; the slot refuses a second backup **and** releases, so a completed one is not the only one an engine process ever takes; the route's 200 / 409 / 500. `Monitor*` 28 / 28.
- **App:** `pnpm test` - 5707 / 5708 passed. The single failure is `send-with-row.test.tsx > opens the list on the row that step bound`, a 5s timeout while an engine build had the cores - the exact case `CLAUDE.md` documents. It touches nothing in this diff and **passes on its own** (41 / 41, re-run after the build finished). Not fixed or hidden here.
- `pnpm type-check`, `pnpm lint`, `prettier --check` on every app file touched: clean.
- `clang-format-19 --dry-run -Werror` over the whole of `engine/{src,include,tests}`: clean.
- `clang-tidy-19 -p build` on all four touched/new engine sources (`workspace.cpp`, `database.cpp`, `cli.cpp`, `workspace_backup_test.cpp`): no findings. The only diagnostic emitted is `clang-diagnostic-ignored-gch` from running clang-tidy against a GCC build tree.
- `mkdocs build --strict`: clean.
- **End to end**, against a real daemon: the endpoint returned a snapshot; `vayu-cli backup` printed the path alone on stdout; a `POST /config` setting `maxBackupsRetained: 2` pruned to the newest two on the next backup (`"pruned": 2`); a snapshot opened with an independent SQLite client contained a collection written seconds earlier; an unreachable daemon exited 1 with a message.

## Checklist

- [x] I have added tests - none for `e6e04692`, and the commit says why: the deleted statement is unreachable, so nothing behavioural can tell before from after
- [x] I have updated documentation - `docs/engine/api-reference.md` (new endpoint), `docs/engine/cli.md` (new verb), `docs/engine/architecture.md` (data directory + the restore procedure), `docs/engine/db-schema.md` (why `.bak` is not this), `docs/app/api-integration.md` and `engine/CLAUDE.md`'s endpoint table - all in the feature commit

## Related Issues

Closes #987
Closes #1037

Unblocks #988 and #990, which the issue's Sequencing says build on this `VACUUM INTO` machinery in the same file region.

### DIFF
--- a/app/src/config/api-endpoints.ts
+++ b/app/src/config/api-endpoints.ts
@@ -207,6 +207,14 @@ export const API_ENDPOINTS = {
 	// deliberately not a second IMPORT_FETCH.
 	DIAGNOSTICS_CONNECTION: `/diagnostics/connection`,
 
+	// Workspace backup (issue #987). A verb path rather than a stored resource,
+	// like INBOX_START and MOCK_SERVER_START: the engine takes a snapshot and
+	// the file it writes is not a row anything here reads back. There is no
+	// restore counterpart on purpose - a live engine overwriting its own open
+	// database is the thing this feature exists to spare the user, so restoring
+	// is a documented manual copy with the engine stopped.
+	WORKSPACE_BACKUP: `/workspace/backup`,
+
 	// Webhook inbox (issue #480). The engine hosts the listener; these drive its
 	// lifecycle and read what it captured. START is a verb path rather than a
 	// POST to `/inbox` because an inbox is not a stored resource - it exists for
--- a/app/src/modules/settings/main/app-settings.ts
+++ b/app/src/modules/settings/main/app-settings.ts
@@ -89,6 +89,14 @@ export const APP_SETTINGS = [
 		keywords: ["clear history", "delete runs", "database"],
 	},
 	{
+		anchor: "workspace-backup",
+		panel: "general",
+		label: "Workspace backup",
+		searchText:
+			"Take a complete copy of the workspace database, and where the snapshot was written.",
+		keywords: ["backup", "snapshot", "restore", "export database", "copy"],
+	},
+	{
 		anchor: "cookies",
 		panel: "general",
 		label: "Cookies",
--- a/app/src/modules/settings/main/panels/GeneralPanel.tsx
+++ b/app/src/modules/settings/main/panels/GeneralPanel.tsx
@@ -37,6 +37,7 @@ import { appSetting } from "../app-settings";
 import { OptionButtons, ToggleRow } from "./SettingControls";
 import { UpdatesCard } from "./UpdatesCard";
 import { CookiesCard } from "./CookiesCard";
+import { WorkspaceBackupCard } from "./WorkspaceBackupCard";
 import { useSettingsStore } from "@/modules/settings/settings-store";
 import type { EngineSettingsCategory } from "@/types";
 
@@ -225,6 +226,12 @@ export default function GeneralPanel() {
 					</p>
 				</CardContent>
 			</Card>
+
+			{/* A copy of the whole workspace the user controls (#987). Beside
+			    Data management because it answers the question that card
+			    raises - what happens to all this if the file is lost - and
+			    above Storage paths, which prints the database it copies. */}
+			<WorkspaceBackupCard />
 
 			{/* The engine's cookie jar - engine-held state the user never typed,
 			    which is why it sits beside run history rather than in a panel
--- a/app/src/modules/settings/main/panels/WorkspaceBackupCard.test.tsx
+++ b/app/src/modules/settings/main/panels/WorkspaceBackupCard.test.tsx
@@ -1,0 +1,135 @@
+/**
+ * @vitest-environment jsdom
+ */
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * The workspace backup card (issue #987).
+ *
+ * Three things this card can get wrong and no other test would catch:
+ *
+ *  - **The path must survive on screen.** Restoring is a manual file copy, so a
+ *    card that only toasts leaves the user with nowhere to look afterwards.
+ *  - **A 409 is not a failure.** "A backup is already running" means their own
+ *    earlier request is still writing the file; reporting it as an error sends
+ *    them hunting for a problem that does not exist.
+ *  - **A real failure must be visible**, not swallowed into a toast that fades.
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render, screen, fireEvent, waitFor } from "@testing-library/react";
+
+import { ApiError } from "@/services/http-client";
+import { WorkspaceBackupCard } from "./WorkspaceBackupCard";
+
+const backupWorkspace = vi.fn();
+const showToast = vi.fn();
+
+vi.mock("@/services", () => ({
+	apiService: {
+		backupWorkspace: (...args: unknown[]) => backupWorkspace(...args),
+	},
+}));
+
+vi.mock("@/stores", () => ({
+	useToastStore: (selector: (s: { showToast: typeof showToast }) => unknown) =>
+		selector({ showToast }),
+}));
+
+const SNAPSHOT = {
+	path: "/home/someone/.vayu/backups/vayu-20260827-120000-000.db",
+	sizeBytes: 2_097_152,
+	createdAt: 1_787_745_600_000,
+	pruned: 2,
+};
+
+beforeEach(() => {
+	backupWorkspace.mockReset();
+	showToast.mockReset();
+});
+
+/** Press the card's one button. */
+function backUp() {
+	fireEvent.click(screen.getByRole("button", { name: /back up now/i }));
+}
+
+describe("WorkspaceBackupCard", () => {
+	it("shows the path, size and what retention removed", async () => {
+		backupWorkspace.mockResolvedValue(SNAPSHOT);
+		render(<WorkspaceBackupCard />);
+		backUp();
+
+		// The path itself, not a summary of it: this is what the user copies.
+		expect(await screen.findByText(SNAPSHOT.path)).toBeInTheDocument();
+		expect(screen.getByText(/2\.0 MB/)).toBeInTheDocument();
+		expect(screen.getByText(/removed 2 older snapshots/)).toBeInTheDocument();
+		expect(showToast).toHaveBeenCalledWith("Workspace backed up", "success");
+	});
+
+	it("does not claim to have removed anything when it removed nothing", async () => {
+		backupWorkspace.mockResolvedValue({ ...SNAPSHOT, pruned: 0 });
+		render(<WorkspaceBackupCard />);
+		backUp();
+
+		await screen.findByText(SNAPSHOT.path);
+		expect(screen.queryByText(/removed/)).not.toBeInTheDocument();
+	});
+
+	it("prints the restore procedure beside the file it applies to", async () => {
+		backupWorkspace.mockResolvedValue(SNAPSHOT);
+		render(<WorkspaceBackupCard />);
+		backUp();
+
+		await screen.findByText(SNAPSHOT.path);
+		// There is no restore button by design - a live engine overwriting its
+		// own open database is the footgun this feature avoids - so the steps
+		// have to be here.
+		expect(
+			screen.getByText(/quit Vayu, copy this file over the database/i)
+		).toBeInTheDocument();
+	});
+
+	it("reads a 409 as a backup already running, not as a failure of this one", async () => {
+		backupWorkspace.mockRejectedValue(
+			new ApiError(409, "CONFLICT", "A workspace backup is already running")
+		);
+		render(<WorkspaceBackupCard />);
+		backUp();
+
+		expect(
+			await screen.findByText(/already running - it will finish on its own/i)
+		).toBeInTheDocument();
+		// And it does not present the engine's raw sentence, which reads as an
+		// error where this is a state.
+		expect(screen.queryByText(/could not/i)).not.toBeInTheDocument();
+	});
+
+	it("shows a real failure on the card rather than only in a toast", async () => {
+		backupWorkspace.mockRejectedValue(
+			new ApiError(500, "INTERNAL", "Could not write the backup to /nowhere: disk full")
+		);
+		render(<WorkspaceBackupCard />);
+		backUp();
+
+		expect(await screen.findByText(/disk full/)).toBeInTheDocument();
+		expect(showToast).toHaveBeenCalledWith("Could not back up the workspace", "error");
+	});
+
+	it("clears a previous failure when a retry succeeds", async () => {
+		backupWorkspace.mockRejectedValueOnce(new ApiError(500, "INTERNAL", "disk full"));
+		render(<WorkspaceBackupCard />);
+		backUp();
+		await screen.findByText(/disk full/);
+
+		backupWorkspace.mockResolvedValue(SNAPSHOT);
+		backUp();
+
+		await screen.findByText(SNAPSHOT.path);
+		await waitFor(() => expect(screen.queryByText(/disk full/)).not.toBeInTheDocument());
+	});
+});
--- a/app/src/modules/settings/main/panels/WorkspaceBackupCard.tsx
+++ b/app/src/modules/settings/main/panels/WorkspaceBackupCard.tsx
@@ -1,0 +1,147 @@
+/**
+ * Copyright (c) 2026 Atharva Kusumbia
+ *
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the "app" directory of this source tree.
+ */
+
+/**
+ * WorkspaceBackupCard
+ *
+ * One button that asks the engine for a snapshot of the workspace, and the path
+ * it wrote (issue #987).
+ *
+ * Everything a person has built in Vayu - collections, environments, stored
+ * credentials, run history - lives in one SQLite file, and until this there was
+ * no copy of it the user controlled. The `.bak` the engine keeps is not one: it
+ * is overwritten on every clean start and exists so a corrupt file has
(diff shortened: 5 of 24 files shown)


athrvk:
### `asan on windows-latest` is failing, and it is not this PR's

**What is failing.** The build, not a sanitizer finding:

```
D:\a\vayu\vayu\engine\src\core\monitor.cpp(147) : error C2220: the following warning is treated as an error
D:\a\vayu\vayu\engine\src\core\monitor.cpp(147) : warning C4702: unreachable code
```

**Why it is not this PR's.** `engine/src/core/monitor.cpp` is not in this diff (23 files, none of them that one). The two lines are `git blame`d to `79b9f124` (2026-08-26, #1021's monitor refactor), which extracted `read_monitor_series` and left the old tail behind:

```cpp
    return read_monitor_series (monitor, limits, out);
    return std::nullopt;   // <- unreachable
```

This PR is simply the first thing to compile that code at `/Od`: `sanitizers.yml` runs per-PR only for a diff touching `engine/CMakeLists.txt` (among three others), and this one adds `src/http/routes/workspace.cpp` to the source list there. The weekly cron last ran **2026-08-25**, a day before `79b9f124` landed, and the next is Monday **2026-08-31** - so master has been carrying this and nothing has been pointed at it. The other four sanitizer legs are GCC/Clang and emit nothing; `Engine on windows-latest` builds `/O2`, which folds the dead return away.

**No re-run spent.** `error C2220` from an unchanged source file is deterministic - a re-run would fail identically at the same line and cost another ~13 minutes of a Windows runner. Saying so here rather than spending it.

**Filed as #1037**, with the one-line patch:

```diff
     return read_monitor_series (monitor, limits, out);
-    return std::nullopt;
 }
```

**Not ported into this PR deliberately.** Sanitizers is not a required check (`sanitizers.yml` says so in its header), so this PR is still mergeable; and folding an unrelated dead-code deletion from someone else's refactor into a feature PR costs the revert boundary for a line that belongs on master on its own. Happy to push it here instead if you would rather see this PR fully green - say the word.

Everything else on this head is green: both engine legs, all four app legs, engine formatting, app quality checks, CodeQL.

---

athrvk:
Ported the cleanup into this PR as you asked - `e6e04692`, one line:

```diff
     return read_monitor_series (monitor, limits, out);
-    return std::nullopt;
 }
```

**Reproduced before fixing.** No MSVC here, so I used the nearest analogue to C4702 that is: `clang++-19 -Wunreachable-code-return` over the same translation unit. It names the line before the deletion and is silent after.

```
src/core/monitor.cpp:147:12: warning: 'return' will never be executed [-Wunreachable-code-return]
  147 |     return std::nullopt;
```

**Swept for the same shape** across `engine/{src,include}` - two same-depth `return`s in a row - and this was the only one, so #1021's other splits are clean.

**Verified:** `python build.py -e -t` then `ctest --preset linux-dev` - 2459/2459 pass, `Monitor*` 28/28. `clang-format-19 --dry-run -Werror` clean on the file. No test accompanies the change: the statement is unreachable, so nothing behavioural can distinguish before from after - the compiler is the assertion, and `asan on windows-latest` going green on this head is the check.

Two corrections to my previous comment, which I wrote while those legs were still running: I said both engine legs were green when only `Engine on macos-latest` had finished (`ubuntu` and `windows` were still in progress, and are superseded by this push anyway), and I said the fix was deliberately not ported - it now is.

#1037 stays open as the record; it will close with this PR. Watching the new run.

---

athrvk:
### Green on `e6e04692`

**The cleanup worked.** `asan on windows-latest` now **passes** - configure, build, and the full suite under ASan, 43 min end to end. On the previous head it died in the build at 13 min on `monitor.cpp(147)`, so the one-line deletion is exactly what it needed. The other sanitizer legs are green too (`asan` ubuntu + macOS, `tsan` ubuntu); `tsan on macos-latest` is still running after ~65 min, which is ordinary for that leg and is not a required check - nothing red there.

**`CI gate`: success.** Everything under it green, including `Engine on windows-latest` - MSVC build, the engine suite, and `Lint changed engine sources` (clang-tidy over this PR's new engine files on the Windows toolchain) all pass.

### The app shard failure, and why it was not code

`App on windows-latest (shard 1/2)` failed once on this head with **two 5000 ms test timeouts** and no assertion failure, in files this PR does not touch:

- `LoadTestConfigDialog.test.tsx > shows rate and duration for Constant RPS, and no connection count`
- `send-with-row.test.tsx > reads the file only when the picker is opened`

2604 / 2606 passed around them. Three things say this was the runner, not the code:

1. The **identical app sources** passed this same shard on the previous head 22 minutes earlier - `e6e04692` touches one engine file and zero app files.
2. Measured locally, those two cases cost **329 ms** and **120 ms** - 15x and 40x under the budget. A 300 ms test hitting 5000 ms is a stalled runner, not a test that wants a bigger timeout; giving it one would be the "raise it by reflex" `CLAUDE.md` warns against, and I have not.
3. This PR is the unusual one that triggers **both** `pr-tests.yml` and the five-job sanitizer matrix (it touches `engine/CMakeLists.txt`), so that Windows app shard was competing with a Windows ASan Debug build of the whole engine for the same runner pool.

**Re-ran the shard once** - the one re-run these rules allow - and it passed: 15:17 → 15:22, green. Nothing skipped, disabled or quarantined, and no timeout raised.

Worth noting for the record rather than fixing here: the contention above is structural, so any future PR touching `engine/CMakeLists.txt` will hit the same squeeze. If it recurs, the fix is on the workflow side (staggering the two runs, or trimming the sanitizer matrix on PRs) rather than in anyone's test file - happy to open that as an issue if you want it tracked.

---

Issue #1053, opened 2026-08-27
engine(scripting): chai's no-op language chains are absent, so an imported `.that.is.` chain is an error rather than an assertion


athrvk:
Found while implementing #999 (arming the `pm.expect` chain). Filed rather than folded in: #999's scope is the six verdict defects its sweep found, and this is a seventh gap the arming makes *visible* rather than one it creates.

## Problem

chai ships a set of language chains that assert nothing and exist to make a chain read as English: `.that`, `.which`, `.is`, `.has`, `.have` (present), `.with`, `.does`, `.but`, `.also`, `.of`, `.same`, `.still`, `.itself`. Vayu implements `to`, `be`, `have`, `at`, `and`, `all`, `deep`, `nested` (`create_expectation`, `engine/src/runtime/script_engine.cpp`) and none of the rest.

Before #999 an imported chain using one failed a link later and unhelpfully - `pm.expect(x).to.be.an('object').that.is.ok` read `.that` as `undefined` and then threw `TypeError: cannot read property 'is' of undefined`, which names neither the chain nor the missing word. After #999 it throws `TypeError: pm.expect(...).that is not a supported assertion`, which names it. **Both are errors**: the assertion the script wrote does not run either way, so nothing here is a false pass - this is a compatibility gap, not a verdict defect, which is why it is `priority:low` and not part of #999.

The cost is real for imported suites: a Postman collection written in the fluent style fails on the language, not on the API under test.

## Fix / solution pathway

Add the missing chains as passthrough getters beside `and` / `all` (`passthrough_chainers` in `create_expectation`). Each returns the expectation unchanged, so the flags a chain has set - `not` included - carry across exactly as they do for `and` today.

Two that are **not** simple passthroughs and should stay out of this issue's scope, filed here so the next reader does not add them by accident:

- `.any` - chai's quantifier for `.keys`, which changes what `.keys` asserts. It is deliberately absent (see `docs/engine/scripting.md`, "keys means exactly these keys"), and aliasing it to `all` would silently assert more than the script asked.
- `.own` / `.itself` - these change what `.property` and `.respondTo` look at rather than reading as prose. `.own.property` is already listed as unsupported in `docs/app/pm-api-compatibility.md`.

## App changes

[...]

Issue #998, opened 2026-08-25
engine(scripting): six response assertions return wrong verdicts vs chai-postman - jsonBody ignores its value argument, body is substring, ok is 2xx


athrvk:
Sub-issue of #996. Verified at `e5277b6`; Postman behavior verified against the chai-postman source (the actual implementation of `pm.response.to.*`). These are assertions that return a **wrong verdict** - most in the dangerous direction (Vayu passes where Postman fails), in a testing tool.

## Problem, per assertion

1. **`to.have.jsonBody(path, value)` ignores the value argument.** `js_response_have_jsonBody` (`engine/src/runtime/script_engine.cpp:2365-2420`) walks the path off `argv[0]` and never reads `argv[1]` - re-verified: the function compares nothing. chai-postman does `_.has(json, path)` then `_.isEqual(_.get(json, path), value)`. Every value-checking jsonBody assertion in an imported suite silently passes on any value. Worst finding of the sweep.
2. **`to.have.body(string)` is substring; Postman is exact.** `:2345-2363` uses `body.find(expected) != npos`; chai-postman string form is `_.isEqual`. Partial-match assertions false-pass. The regex form (`/re/` - chai-postman execs it) and plain-object form (deep-equals `json()`) are stringified into the substring search and fail where Postman passes.
3. **`to.have.status("OK")` / any reason string:** `JS_ToInt32` coerces to 0 (`:2286-2293`) - fails where Postman compares `reason()`. Conversely `to.have.status("200")` coerces and passes where Postman fails. Reason-string support belongs here mechanically; note the wider status/reason decision lives in its own sub-issue - implement the string arm consistently with whatever that decides.
4. **`to.have.header(name, value)`: expected value coerced to string** (`:2334`) so `header("X-Count", 5)` passes; chai-postman is strict `===` against the header string.
5. **`headers.has(name, value)` ignores the value argument** (`js_headers_has`, `:2825-2834`); Postman's PropertyList `has(key, value)` checks it.
6. **`to.be.ok` is 2xx** (`status_class_matchers`, `:2439`, re-verified) - chai-postman's `ok` asserts code 200 AND reason "OK". A 204 passes in Vayu, fails in Postman. `success` stays 2xx (matches). Named statuses (`accepted` etc.) also reason-check in Postman; code-only is acceptable to keep, but state it in the compat doc.

## Fix / solution pathway

[...]

Issue #487, opened 2026-08-10
engine(scripting): decide whether pm.expect failures should throw AssertionError instead of TypeError - the chai-compat naming question #484 deliberately left open


athrvk:
Deferred out of #484, which called it "its own compatibility decision" and shipped the message-argument support without it. Filed so the recorded deferral has a tracker (verified untracked at the #484 closure audit, master `ec194d1`). Low severity, a decision issue: nothing is broken, one error-name diverges from Postman.

## Problem

Every failing `pm.expect` assertion throws a `TypeError` (the `expect_*` handlers throw through `JS_ThrowTypeError` - post-#484 via the shared message-prefix helper in `engine/src/runtime/script_engine.cpp`). In Postman, `pm.expect` is chai, and chai failures are `AssertionError`. Two visible consequences:

1. The report and console prefix failures with `TypeError:` where a Postman user expects `AssertionError:` - cosmetic, but it reads as an engine bug rather than a failed assertion (the same misreading #134 documents for the bare "not a function").
2. A script that catches and inspects errors - `catch (e) { if (e.name === "AssertionError") ... }` - behaves differently than in Postman. Verified: nothing in the repo's own docs or fixtures relies on either name; the exposure is imported user scripts.

## Fix / solution pathway - a decision, then a small change

- **A - throw a real `AssertionError`.** QuickJS has no built-in AssertionError class; the runtime would define one (subclassing Error in the sandbox bootstrap, or a C-level class with `name` set) and the shared throw helper uses it. Postman-parity, slightly more surface.
- **B - keep TypeError, document it.** One line in `docs/app/pm-api-compatibility.md`'s pm.expect row ("failures throw TypeError, not chai's AssertionError"). Zero code risk; the divergence becomes a stated contract.

A is the one to beat only if imported-script compatibility is a goal the owner weighs; otherwise B closes this in ten minutes. Either way the decision gets recorded here.

## App changes

None, verified - the renderer renders the message string; nothing branches on the error name (`rg -i "AssertionError" app/src` returns zero hits).

## Tests

- A: a failing assertion's error has `name === "AssertionError"` and still carries the #484 message prefix (mutation-check by reverting the class); the docs-compile corpus is unaffected (type-level only if the declaration mentions the throw, which it need not).
- B: no test - the doc line is the deliverable.
- Commands: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`.

## Docs (same commit)

[...]

PR #1054, opened 2026-08-27
fix(engine): the pm.expect chain asserts, or says why it cannot - #999

### LINKED ISSUE #999: engine(scripting): pm.expect silently passes on unimplemented matchers - arm the chain, add .NaN, fix object include, ordering coercion, throw matching
Sub-issue of #996. Verified at `e5277b6`; chai behavior verified against chaijs.com/api/bdd and the deep-eql source.

## Problem

`pm.expect` chains silently pass where chai fails, and diverge on several callable matchers:

1. **Unimplemented property-style matchers assert nothing.** The expectation class has `.exotic = nullptr` (`engine/src/runtime/script_engine.cpp:383`, re-verified) - no unknown-member trap, unlike `pm.response.to` which throws by name on unknown members (`:2535-2581`). So `pm.expect(x).to.be.NaN` (or `.finite`, `.sealed`, `.frozen`, `.extensible`, or any typo like `.to.be.trueish`) evaluates to `undefined` as an expression statement and the test PASSES. This also falsifies `docs/app/pm-api-compatibility.md:582-584`, which claims each unsupported matcher "throws a TypeError" - true only for called forms.
2. **`.not.include({...})` on an object target always passes** - `expect_include` (`:969-1015`) handles strings and arrays only; an object leaves `includes=false`, so positive always fails and negated always passes. chai's `.include` on objects is subset matching.
3. **`.above/.below/.least/.most` coerce non-numbers** (`JS_ToFloat64`, `:909-967`, `:1225-1281`): `expect("5").to.be.above(3)` passes; chai type-asserts number (or Date) and fails.
4. **`.throw("substring")` matches `String(err)`** (name included - `:1580`, `:1588`): `expect(fn).to.throw("Error")` passes for any Error; chai matches `err.message` only. Also `.throw(ErrorConstructor)` is rejected (`:1587-1595`); chai accepts constructors.
5. **`.eql` treats +0/-0 as equal** (strict-equal short-circuit, `:624`, `:458-467`); deep-eql distinguishes via `1/x`.

## Fix / solution pathway

1. **Arm the expectation chain**: give the expect class the same unknown-member exotic hook `pm.response.to` uses (pattern at `:2535-2581`) so any unimplemented member throws by name - converting every silent pass in this family, present and future, into a loud error. This is the load-bearing fix; the rest are point fixes.
2. Add `.NaN` as a real getter terminal beside `true/false/null/undefined/ok/empty/exist` (`:1812-1815`). (`finite/sealed/frozen/extensible` may simply keep throwing via the hook - state so in the compat doc.)
3. Implement object-subset `.include` (own enumerable keys, strict per-key unless `.deep`), fixing both directions of finding 2.
4. Type-assert numbers (or Date) in the ordering matchers.
5. `.throw`: match `err.message` for strings/regex; accept a constructor argument.
6. `.eql`: distinguish signed zeros before the strict-equal short-circuit.

## App changes

None, verified - script-engine only; regenerate the typedef fixture if the surface changes (`engine/tests/fixtures/script-typedefs.d.ts` - verify).

## Tests

Engine gtest per finding, both verdict directions; the armed-chain test asserts a typo'd member throws naming itself (mutation check: un-arm, silent pass returns). `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`. Clean under #928 families.

## Docs (same commit)

`docs/app/pm-api-compatibility.md:582-584` becomes true as written once the hook lands - reword to cover property-style members explicitly; `docs/engine/scripting.md` matcher list gains `.NaN`, object `.include`, `.throw(Constructor)`.

## Sequencing

`script_engine.cpp` expect machinery - serial with the other #996 subs at rebase level; independent of the response-assertion sub in content, adjacent in file.

## Acceptance criteria

- [ ] Any unimplemented `pm.expect` chain member throws by name; `.NaN` asserts correctly.
- [ ] Object `.include` matches chai subset semantics both directions; ordering matchers type-assert; `.throw` matches message and accepts constructors; `.eql` splits signed zeros.
- [ ] Tests fail on today's code for each; docs updated.

---

athrvk:
## Description

`pm.expect` returned a verdict chai does not in six places, and five of them passed where chai fails - the direction that matters in a testing tool, because an imported suite reports green against a broken API and says nothing.

The root cause of the worst one is structural rather than per-matcher: the expectation class had no unknown-member trap, so a paren-less matcher nothing implements (`.to.be.NaN`, `.finite`, `.sealed`, or a typo like `.to.be.trueish`) was an ordinary missing-property read - `undefined`, discarded as an expression statement, PASS. `pm.response.to` has thrown by name since #487; the fix is that same exotic `get_own_property` hook on the other chain, which closes every silent pass in this family at once, present and future. The alternative considered and rejected was implementing the missing chai matchers one at a time: it fixes the named five and leaves the next typo as silent as before. The passthrough rules the two chains share are one function now rather than a second copy, and the expectation class gains the `Object.prototype` the armed response chain already needed, so `String(...)`, `JSON.stringify`, `console.log` and `Object.keys` still answer on an expectation.

The five point fixes around it are each the same shape - a comparison chai makes and Vayu did not:

- **`.NaN`** is a real terminal on chai's rule (`value !== value`), so `expect('foo').to.be.NaN` fails rather than reading the string as a number.
- **`.include` on an object target** is subset matching. It handled strings and arrays only, so the positive form always failed and `.not.include({...})` always passed, whatever the target held. Any target other than a string or an array takes an object argument or a `TypeError` naming the combination - which is the combination chai refuses rather than answers.
- **The ordering matchers type-assert both sides.** `ToNumber` read `"5"` as 5, `true` as 1 and `null` as 0, so `expect("5").to.be.above(3)` passed a comparison the script never asked for. They take a number or a `Date` now.
- **`.throw` matches `err.message`**, not `String(err)` - which is "Error: boom", so `.to.throw("Error")` passed for every Error thrown whatever it said. A constructor argument, refused outright before, is chai's `instanceof` check, and a message matcher after it is read rather than dropped.
- **`.eql` splits `+0` from `-0`** through deep-eql's rule.

Two deliberate deviations, both argued rather than assumed:

1. **The signed-zero rule is a parameter, not a new global truth.** One deep compare serves `.eql` (chai's deep-eql: `-0` is not `0`) and the response assertions (chai-postman, which is lodash `_.isEqual`: they are equal - and #998 aligned those assertions to lodash two commits ago). Making the split unconditional would have quietly broken the newer contract, so `js_deep_equal` takes a `ZeroRule` and both callers state theirs. Both sides are pinned by a test, because one function serving two contracts is only correct while both are asserted.
2. **The four ordering matchers became one function reached by magic.** They were near-copies, the type rule this issue adds is the same rule in all four, and a fourth copy of it is the one a later fix would miss. Same shape as `status_class_matchers` next door.

Two consequences worth naming, neither hidden:

- **Arming the chain is a behaviour change for scripts.** A chain that ended on an unimplemented member used to pass and now throws a `TypeError` naming the member. That is the point of the change, and nothing in the repo relies on the old behaviour: the only two runnable `pm.expect` fixtures use `.to.equal`, `.to.have.property` and `.to.exist`, and no test pinned a silent pass.
- **chai's no-op language chains (`.that`, `.is`, `.which`, `.with`, ...) remain absent** and now say so by name instead of failing one link later with "cannot read property of undefined". Both are errors, so nothing regresses; the new one is legible. Filed as #1053 rather than widened into this diff.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Breaking change (a chain ending on an unimplemented member, `.to.throw("Error")` and `expect("5").to.be.above(3)` stop passing)
- [x] Documentation update

## Testing

- `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` - **2596 engine tests, 0 failures**, including 8 new `TEST_F`s.
- `cd app && pnpm test` - **5780 tests in 532 files, 0 failures** (the generated typedefs are consumed there: `script-typedefs.docs-compile.test.ts` compiles every `pm.*` block in both docs against them).
- `cd app && pnpm type-check` - clean, all three projects.
- `clang-format --dry-run -Werror` clean on every touched engine file.
- No pre-existing failures on the base commit.

Each new test asserts both directions, so each is its own mutation check - on the code it replaces, the false-pass half goes green:

| Test | What goes green if the fix is reverted |
|---|---|
| `ExpectChainRejectsUnknownMembers` | all five: `.trueish`, `.finite`, `.frozen`, `.sealed`, a mid-chain typo |
| `ExpectChainStillBehavesLikeAnObject` | guard on the fix: `String()`, `JSON.stringify`, `console.log`, `Object.keys` on an armed expectation |
| `ExpectNaN` | `expect(1).to.be.NaN` and `expect('foo').to.be.NaN` |
| `ExpectIncludeMatchesObjectSubset` | `expect({a:1}).to.not.include({a:1})` |
| `ExpectOrderingMatchersTypeAssert` | `expect("5").to.be.above(3)` and three more |
| `ExpectThrowMatchesMessageAndConstructor` | `.to.not.throw("Error")` on `new Error("boom")` |
| `ExpectEqlSplitsSignedZeros` | `expect(-0).to.not.eql(0)` and its nested forms |
| `ResponseAssertionsKeepSameValueZero` | guard on the other half of the zero rule |

The completion table gains `.to.be.NaN` and the changed signatures, and `engine/tests/fixtures/script-typedefs.d.ts` is regenerated from it - `ScriptCompletions.EveryExpectMemberTheRuntimeBindsIsOffered` and `ScriptTypesTest.TheCheckedInDeclarationsMatchTheGenerator` are what say so.

## Checklist

## Related Issues
Closes #999

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pe1DY2cavTzFd1YYarvpoj)_

### DIFF
--- a/docs/app/pm-api-compatibility.md
+++ b/docs/app/pm-api-compatibility.md
@@ -289,7 +289,9 @@ these:
 .to.equal(v)      .to.eql(v) / .to.eqls(v)             .to.deep.equal(v)
 .to.exist         .to.be.true       .to.be.false
 .to.be.null       .to.be.undefined  .to.be.ok          .to.be.empty
+.to.be.NaN
 .to.be.above(n)   .to.be.below(n)   .to.be.at.least(n) .to.be.at.most(n)
+                  (each takes a number or a Date on both sides, never a coercion)
 .to.be.closeTo(v, delta)            .to.be.oneOf([…])
 .to.be.a(type)    .to.be.an(type)   .to.be.instanceOf(Ctor)
 .to.have.property(name[, v])        .to.have.nested.property('a.b[0].c'[, v])
@@ -297,7 +299,8 @@ these:
 .to.have.keys(…) / .to.have.key(k)  .to.have.members([…])
 .to.include(v)    .to.contain(v)    .to.have.string(sub)
 .to.match(/regex/)                  .to.satisfy(fn)
-.to.throw([msg | /regex/]) / .to.throws(…)
+.to.throw([Ctor | msg | /regex/][, msg | /regex/]) / .to.throws(…)
+                  (a string or pattern is matched against err.message)
 .to.not …         (sets the negation for the rest of the chain; a second
                    .not in one chain is a no-op, not a double negative)
 .deep …           (deep comparison for equal / include / property / members / oneOf)
@@ -310,7 +313,16 @@ these:
 `expect({a:1}).to.equal({a:1})` **fails** - different references - and
 `expect({a:1,b:2}).to.eql({b:2,a:1})` **passes**: key order is not part of deep
 equality. `include`, `property(name, value)`, `members` and `oneOf` compare
-strictly too, unless a `deep` appears in the chain.
+strictly too, unless a `deep` appears in the chain. The one place the two
+libraries Vayu answers for disagree is the pair `+0` / `-0`: `eql` separates
+them (deep-eql's `1/x` rule) while the response assertions in the table above
+compare them equal, because chai-postman runs on lodash `_.isEqual`.
+
+**`include` on an object target is a subset match**, as in chai: every key of
+the argument must be on the target with an equal value. Any target other than a
+string or an array takes an *object* argument, so `expect({a:1}).to.include('a')`
+and `expect(5).to.include('x')` are both a `TypeError` - the combination chai
+refuses rather than answers.
 
 **The second argument is chai's failure message.** `pm.expect(value, 'context')`
 prefixes `context: ` to whatever the failing matcher reports, so an assertion
@@ -339,7 +351,12 @@ QuickJS has no `AssertionError` class, so this is an `Error` carrying that
 gets, and there is no `AssertionError` global to reference (chai's lives on the
 `chai` module, which Vayu does not ship). **A mistake in the script text stays a
 `TypeError`** - a matcher called with no argument, a name nothing implements -
-because nothing was asserted, the call itself was wrong.
+because nothing was asserted, the call itself was wrong. That covers
+**property-style members too** (issue #999): `pm.expect(x).to.be.NaN` is an
+expression statement, so an unimplemented name used to evaluate to `undefined`
+and report PASS whatever the value was. Every name the chain does not carry -
+a typo, or a chai matcher Vayu lacks (`.finite`, `.sealed`, `.frozen`,
+`.extensible`) - now raises a `TypeError` naming itself.
 
 `have.keys` asserts *exactly* those keys. `Map` / `Set` / typed arrays are
 reported unequal by `eql` rather than compared (their contents are not
@@ -600,7 +617,8 @@ These Postman APIs are **not** implemented - scripts that rely on them will fail
 - The `tests["name"] = bool` legacy assertion style (use `pm.test`)
 - Chai matchers outside the list above: `.include.keys` (the subset form),
   `.any.keys`, `.change`/`.increase`/`.decrease`, `.own.property`, `.respondTo`,
-  and the `require()`-able libraries (`chai`, `lodash`, `moment`, …). Each throws
+  the property-style `.finite` / `.sealed` / `.frozen` / `.extensible`, and the
+  `require()`-able libraries (`chai`, `lodash`, `moment`, …). Each throws
   a `TypeError` rather than reporting a pass
 
 ---
--- a/docs/app/pm-api-compatibility.md
+++ b/docs/app/pm-api-compatibility.md
@@ -322,7 +322,20 @@ compare them equal, because chai-postman runs on lodash `_.isEqual`.
 the argument must be on the target with an equal value. Any target other than a
 string or an array takes an *object* argument, so `expect({a:1}).to.include('a')`
 and `expect(5).to.include('x')` are both a `TypeError` - the combination chai
-refuses rather than answers.
+refuses rather than answers. Vayu is stricter than chai in three places, each
+because chai's answer there is a silent non-assertion or a quiet wrong verdict:
+an expectation carrying no own enumerable keys is refused (`to.include({})`, a
+`Date`, a `RegExp`, a function - chai passes all of them, in both directions,
+having compared nothing); a getter that throws is reported rather than read as
+"these differ"; and a `Map`, `Set` or boxed `String` target is refused rather
+than answered, since deep equality here does not inspect the first two.
+
+**A wrong-typed value is a `TypeError` where chai raises an `AssertionError`** -
+`expect('5').to.be.above(3)`, `expect(5).to.include('x')`. Deliberate, and the
+same choice `pm.response.to.have.body` made for a wrong-typed argument (#998):
+the rule below is that a mistake in the script text stays a `TypeError` because
+nothing was asserted. Both throw, so a test fails either way; `e.name` is what
+differs.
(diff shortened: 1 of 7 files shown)


athrvk:
Review round, and what came out of it. Two read-only passes over the first commit: one against #999's acceptance criteria, one adversarial against the QuickJS code. The second ran its cases against the built engine rather than reading for them, which is how four real defects in code this branch added came out.

**Pushed as `75173ff`:**

1. **`.include` passed on an expectation with no keys.** The subset walk answered "every key matched" for an object carrying none, so `to.include({})` passed and `to.not.include({})` failed - a verdict about a comparison that never ran, which is the defect this branch exists to remove, reintroduced by it. A `Date`, a `RegExp` and a function reach the same branch carrying no own enumerable key while reading as assertions. All refused by name now. **This diverges from chai deliberately**: chai walks the keys, finds none, and passes in *both* directions having compared nothing. Argued in both docs.
2. **A throwing getter was read as a mismatch.** Neither property read checked for an exception, so a getter that threw left one pending, answered "these differ", and made the negated form a pass. It propagates now - the rule the deep compare already states for an unreadable class tag (#959), which the same commit's `js_ordering_value` had broken the same way.
3. **`.throw` failure messages contradicted themselves.** `.to.throw(TypeError)` catching a `RangeError` said "Expected the function to throw, but it threw RangeError: bad". The message names the constructor and the message matcher now, and says plainly when nothing was thrown.
4. **A thrown non-Error, non-string has no message**, as in chai's `check-error` - `.to.throw("4")` used to pass on `throw 42`.

Plus two tightenings from the same read: the ordering matchers compare like with like (a `Date` read as milliseconds against a number is a comparison neither side wrote), and the matcher table's size is `static_assert`ed against the switch reading its index. The negated form of an unknown chain member is now tested, which the acceptance criteria named and the first pass had only reasoned about.

**Verified and discarded, with the reason:**

- *"chai raises an AssertionError for `expect(5).to.include({})`"* - it does not. chai's refusal (`val !== Object(val)`) tests the **argument**, not the target, so the empty-keys case reaches the key walk and passes. Kept as a divergence in our favour, per point 1 above, rather than as a chai-compatibility fix.
- *"`hasOwnProperty("nope")`, `in` and `getOwnPropertyDescriptor` now throw on an expectation"* - true, and not separable: QuickJS routes a *has* and a *get* through the same `get_own_property` hook, so an armed chain cannot answer one without the other. `pm.response.to` has behaved this way since #487; making it consistent, not novel. Not fixed.
- *"usage errors should be `AssertionError`, not `TypeError`"* - chai does raise `AssertionError` for a wrong-typed value. Kept as `TypeError` for consistency with this file's stated rule and with #998's `body()` refusal two commits ago, and now **recorded in both docs** rather than left implicit, which is what the reviewer actually asked for.
- `.throw(errorInstance)`, `expect_length`'s float message, `Symbol.toStringTag` spoofing a `Date`, `.throw(arrowFn)` - pre-existing or out of #999's scope, and each still throws rather than passing.

**One pre-existing false pass the review surfaced is filed as #1056**: `pm.expect(fn).to.throw()` reads the script's wall-clock deadline as a throw, so a target that never returns reports PASS. It predates this branch and the fix belongs where the interrupt is raised, not in this diff.

Verification after the round: 2598 engine tests green, `clang-format` clean, docs-compile test green.

---

Issue #1064, opened 2026-08-28
app(variables): the variable popover has no &quot;bound data row&quot; origin, so &quot;why is this the value&quot; is wrong while a row is bound


athrvk:
Follow-up to #1007.

## Problem

`app/src/components/ui/variable-popover.tsx` explains a token by listing the definitions it beat (`ShadowedBy`, driven by `VariableOrigin[]`: environment > chain > global). #1007 added a tier *above* all three - a bound row's column - and the popover has no origin for it, so while a row is bound it explains a value the send will not use. Its `data.*` branch also says "Bound per iteration by the run's data file. Not a variable - defining one with this name would change nothing", which stays true for the prefixed spelling and is false for a bare column name: defining a variable of that name does something, it just loses to the row.

`VariableInput` has the matching gap by design: #1007 paints a bare token as a bound column only when *no* scope defines the name, deliberately leaving the shadowing paint state as a design question (recorded in `docs/app/variable-resolution.md`, D18).

## Fix / solution pathway

Give `VariableOrigin` a row-shaped case and render it at the top of `ShadowedBy` when a row is bound, then decide the paint state for the shadowing case - probably the variable's existing accent treatment plus a note, rather than a fourth colour. It needs the same "a row is bound" signal as #1062, so it is cheapest taken after that one, which is where the signal gets plumbed.

## Tests

`variable-popover.test.tsx` / `EditableVariable.test.tsx`: with a row bound, a column that also names an environment variable shows the row as the winning origin and the variable as shadowed beneath it; with no row bound, neither appears and the popover reads exactly as it does today. Mutation-check by removing the origin.

## Docs (same commit)

`docs/app/variable-resolution.md` (D18's preview paragraph) and `docs/app/COMPONENTS.md`'s token-state list.

## Sequencing

`app/` only, after #1007, and best after #1062. Parallel-safe with #1063.

## Acceptance criteria

- [ ] While a row is bound, the popover names the row as the origin and shows the shadowed variable beneath it.
- [ ] The `data.*` terminal message stays as it is; the bare spelling gets its own.

---
Generated by [Claude Code](https://claude.ai/code)

athrvk:
Claiming this issue - PR will follow from this run.

[...]

Issue #1063, opened 2026-08-28
app(scripting): pm.variables.get completion and the script &quot;Referenced:&quot; chips do not know a bound row answers bare column names


athrvk:
Follow-up to #1007.

## Problem

#1007 made `pm.variables.get` / `.has` / `.toObject` read a bound row's bare column names first, above the environment (`engine/src/runtime/script_engine.cpp`, `iteration_cell`). Two renderer surfaces still describe the pre-#1007 world:

1. `app/src/hooks/useScriptVariableCompletionProvider.ts` builds the completion list inside `pm.variables.get("` from the three scopes alone (`getAllVariables()`), so a declared data column - a name that accessor now answers - is never offered. `pm.environment.get` / `pm.globals.get` / `pm.collectionVariables.get` must keep their scope-only lists: those accessors really do read one scope, which is the distinction `app/src/lib/script-variable-completion.ts` was built around.
2. `app/src/lib/referenced-variables.ts` tags `pm.<scope>.get()` calls and `{{name}}` mentions for the script panel's "Referenced:" chips (`ScriptPanel.tsx`). A `pm.variables.get("email")` that reads a column is invisible to it - the same gap `pm.iterationData.get` had before #600.

Neither is wrong output exactly; both are *missing* output, which is the quiet half of this repo's written-but-never-read defect.

## Fix / solution pathway

Blend the collection's declared columns into the `variables` accessor's completion list only, labelled so the reader can tell a column from a variable, reusing `useDataContract` - the same source `pm.iterationData`'s completion already uses, rather than a second column list. For the chips, treat a `pm.variables.get("x")` whose `x` is a declared column the way `pm.iterationData.get("x")` is treated today.

## Tests

`variable-completion-scope.test.tsx`: the columns appear for `pm.variables.get("` and do **not** appear for `pm.environment.get("`. `script-data-chip.test.tsx`: a column read through `pm.variables.get` produces the same chip as one read through `pm.iterationData.get`. Mutation-check both by reverting the blend.

## Docs (same commit)

`docs/app/pm-api-compatibility.md` if the completion behaviour is described there; `docs/app/COMPONENTS.md` for the chips.

## Sequencing

`app/` only, after #1007. Parallel-safe with the other #1007 follow-ups - different files.

## Acceptance criteria

- [ ] `pm.variables.get("` offers declared columns; the single-scope accessors do not.
- [ ] A column read through `pm.variables.get` is chipped like one read through `pm.iterationData.get`.

---
[...]

Issue #1062, opened 2026-08-27
app(request-builder): the Send-with-row preview resolves a bare column from the environment while the send defers it to the row


athrvk:
Follow-up to #1007, which landed the bare-column tier (a bound row's column names answer `{{name}}` above the environment) but wired only the *payload* side of the renderer.

## Problem

`composeForSend` now sends `dataColumns` on `POST /compose` for a Send-with-row (`app/src/modules/request-builder/index.tsx`), so the engine defers `{{username}}` and binds it from the row. The preview does not: `useVariableResolver`'s `resolveString` calls `resolveTemplate(input, lookup)` with no `boundColumns`, so the UrlBar preview, the tab title and the unresolved-token painting all show the *environment's* value for a name the send will answer from the row. The same gap paints the token in `VariableInput`: a bare name a scope defines keeps painting as that variable, since nothing there knows a row is bound.

That is the shape `docs/app/COMPONENTS.md` describes for the `data.*` painter - "a resolved token carrying a value the engine will never send" - reached through the bare spelling, and it is only visible while a row is picked. #1007 recorded it in `docs/app/variable-resolution.md` (D18, "What the builder's preview does not show yet") rather than leaving it silent, and routed the collision warning to the Data tab's column audit in the meantime.

## Fix / solution pathway

Thread the picked row - or the parsed file's column names - from `SendWithRowDialog` / `useSendWithRow` to `useVariableResolver`, and pass it as `resolveTemplate`'s third argument. The resolver already takes the parameter (#1007 added it and pinned it against the engine with the conformance fixture), so this is wiring, not semantics.

Decide one thing explicitly: whether the preview shows the *row's value* - it has the row, unlike composition - or the deferred token. Showing the value is closer to what the send does and is the recommendation. Whichever is chosen, `{{data.column}}` and the bare spelling must agree, because they are one bind.

## Tests

`app/src/modules/request-builder/`: a Send-with-row whose row carries a column an environment variable also defines - the preview shows the row's answer, not the environment's. Mutation-check: drop the argument and the assertion fails on the environment's value. A second case with no row picked, pinning that the preview is unchanged there.

## Docs (same commit)

`docs/app/variable-resolution.md` - D18's "What the builder's preview does not show yet" paragraph is the one this closes, so it is rewritten rather than appended to. `docs/app/COMPONENTS.md` if the token states change.

## Sequencing

[...]

Issue #1051, opened 2026-08-27
engine(compose): two header names that resolve to the same name silently drop one


athrvk:
Found while reviewing #1050 (issue #1008). Pre-existing, not introduced there - the residual pass inherits the behaviour because it deliberately mirrors composition's.

## Problem

Composition resolves header **names** as well as values (`resolve_compose_head`, `engine/src/http/request_composer.cpp:791`: `resolved[resolve_header_template (key, vars, refusal)] = …`). Two names that resolve to the same string land on the same key, and the second assignment overwrites the first with no word to anybody:

```
{{tenant_header}}: acme      with tenant_header = "X-Tenant"
X-Tenant: legacy
```

goes out as one header, and which one survives depends on the iteration order of the pre-resolution map. `Headers` compares case-insensitively, so `{{h}}` resolving to `authorization` collides with an `Authorization` the user typed too.

The same shape is in `resolve_header_names` (`engine/src/http/request_exchange.cpp`, #1008's residual pass), which rebuilds the map the same way and for the same reason - it is the same rule, not a second one.

This is the quiet-wrong-request class the header rules exist to refuse elsewhere: `#738` refuses a value that would *forge* a header, and `unsendable_header_text` refuses one that would end the line, but a value that makes a header *disappear* passes both. Neither path is tested for it (`request_composer_test.cpp` has no case).

## Fix / solution pathway

Decide the rule once, in composition, and let the residual pass follow it:

1. **Refuse**, in the shape #738 established - a `400` from `POST /compose` naming both spellings and the variable that collided them - if a collision is judged a mistake the user must see. This is the more house-consistent answer: refusal rather than repair.
2. Or **define** last-writer-wins deliberately, in which case it needs a comment saying so and a test pinning which one wins, because "whichever the map iterated last" is not a rule.

Whichever is chosen, the residual pass must answer the same way; its error channel is the one open question, since it has no `400` to return (see #1050's second deviation note - the pre-send gate is what refuses there today).

## Tests

`request_composer_test.cpp`: two header names resolving to one, in both spellings of the collision (two templates, and a template colliding with a literal header); the chosen rule asserted, mutation-checked. A sibling in `residual_token_test.cpp` for the execute-time pass.

## Docs (same commit)

`docs/engine/api-reference.md` (`POST /compose`'s refusal list, if it becomes one) and `docs/app/variable-resolution.md` beside the #738 header rule it sits next to.

## Sequencing

[...]

PR #1065, opened 2026-08-28
feat(engine): a bound data row answers bare column names - #1007

### LINKED ISSUE #1007: engine(variables): dataset columns never bind to bare {{column}} names or pm.variables.get - every imported data-driven collection silently sends empty values
Sub-issue of #996. Verified at `e5277b6`. **This revisits a recorded design decision** (the `{{data.*}}` reserved namespace, `docs/app/variable-resolution.md:48-63`) under the owner's standing Postman-compatibility direction - the decision section below records the recommended resolution; the owner can override on this issue before pickup.

## Problem

Postman binds CSV/JSON dataset columns to **bare** names: `{{username}}` in a request and `pm.variables.get("username")` both read the current row (verified via learning.postman.com; precedence: local > **data** > environment > collection > global). Vayu binds rows only under the disjoint `{{data.column}}` namespace (`is_data_variable_name`, `engine/src/http/request_composer.cpp:193-199`, `:309-311`) and `pm.variables.get` never consults the row (`engine/src/runtime/script_engine.cpp:3975-3982` walks environment/chain/globals only).

Failure mode for every imported data-driven collection, fully silent: `{{username}}` resolves to `""` in the URL/body (unknown ordinary name → empty, `request_composer.cpp:318`) and `pm.variables.get("username")` returns `undefined`. No error fires anywhere on the single-send path - the `data.*`-without-file 400 guard never triggers because imported collections do not use the `data.` prefix.

## Throughput guard (owner constraint, 2026-08-26)

**This change must not reduce load-testing throughput.** Bare-column binding under load rides the SAME compose-time detection the `{{data.*}}` path uses: at plan composition, record which parts reference bound column names; runs with no dataset - and datasets whose columns no part references bare - add zero per-iteration work. The per-iteration bind for flagged parts happens at the existing `apply_data_template` point, same bounded cost. `pm.variables.get`'s row consult is deferred-script only (post-run, off the hot path). PR body carries before/after numbers on the `docs/engine/benchmarks.md` harness shape: dataset-off throughput unchanged; dataset-on cost reported.

## Decision (recommended)

Keep `{{data.column}}` as the canonical, collision-proof Vayu spelling (its escaping/error rules are strictly better - `scenario_data.hpp`). **Additionally** bind bare column names at Postman's precedence position - above environment - in both readers:

- `pm.variables.get(name)`: consult the bound row before the environment walk.
- `{{name}}` composition: when a data row is bound, resolve names against the row first (`build_variable_values` layering, `request_composer.cpp:222-239` - the row becomes the top layer).
- Collision rule stated in docs and tested: a column shadows an environment variable of the same name **only while a row is bound** (exactly Postman's semantics); the reserved `{{data.*}}` namespace keeps its cannot-be-shadowed guarantee either way.
- `pm.iterationData` unchanged (already compatible).

## App changes

`docs/app/data-driven-runs.md` consumers: the column-audit UI (`app/src/services/data-files/column-audit.ts`) may warn on column/environment collisions - verify what it checks and extend the warning to say the column wins (state the finding). Autocomplete for bare columns where `{{data.col}}` completes (`app/src/lib/script-variable-completion.ts`).

## Tests

Engine gtest: bare `{{column}}` resolves from the bound row in url/header/body; `pm.variables.get(column)` returns the typed row value; column shadows same-named environment variable while bound and stops shadowing unbound; `{{data.column}}` behavior byte-identical to today; **a datasetless run skips the machinery (flag empty, path unchanged)**; conformance fixture (`engine/tests/fixtures/variable-resolution-conformance.json`) extended so the renderer mirror stays pinned. `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`; `cd app && pnpm test`. Clean under #928 families.

## Docs (same commit)

`docs/app/variable-resolution.md` (the precedence ladder gains the data layer; the D-decision record gets its amendment note) and `docs/app/pm-api-compatibility.md`, `docs/app/data-driven-runs.md`.

## Sequencing

`request_composer.cpp` + `script_engine.cpp` variable readers - rebase-level with #993-#995 (which move where rows bind) and the other #996 subs. Land after #993 so single-request runs benefit immediately, and inherit its throughput-guard plumbing.

## Acceptance criteria

- [ ] Imported Postman data-driven collections resolve bare columns in requests and scripts at Postman's precedence.
- [ ] `{{data.*}}` semantics unchanged; the collision rule is tested both ways and documented.
- [ ] **Benchmark evidence in the PR: dataset-off throughput unchanged; dataset-on cost reported.**

---

athrvk:
## Description

Postman binds a dataset's columns to **bare** names - `{{username}}` in a request reads the current row - and Vayu bound them only through the reserved `{{data.column}}` namespace (#402). So every imported Postman data-driven collection sent a request its author never wrote: `{{username}}` reached the wire as the literal token (since #1009 an unresolved ordinary name keeps its braces), and `pm.variables.get("username")` answered `undefined`.

**The plan.** Root cause is a design decision, not a bug: #402 deliberately made `data.*` a namespace *disjoint* from the variable tiers, which dissolves the precedence question instead of answering it - and Postman answers it, putting a data column above the environment. So `{{data.*}}` is left exactly as it is and the bare spelling is added beside it as a fourth, conditional tier: while a row is bound, that row's column names answer a bare `{{name}}` above the active environment; with no dataset, nothing changes. The implementation is **deferral, not a second substitution**: composition cannot substitute a bound column (a plan is composed once, a row is bound per iteration), so it leaves the token written as it stands - the same thing it already does for `data.*` - and the existing per-row bind joins *both* spellings through the same walk, which is what makes a bare column inherit the identical JSON/XML escaping, missing-column and null-cell refusals, and header CRLF/collision rules. Which bare names a bind owns travels as `http::BoundColumnNames`, filled by the engine wherever it knows the dataset (a scenario plan, a single-request load run, a send carrying one row) and stated by a client through a new optional `dataColumns` field on `POST /compose`.

**The alternative I rejected** is the one the issue sketches - "the row becomes the top layer" in `build_variable_values`. Composition holds no row, so a row layer there would write iteration 0's values into the payload every iteration then reuses; and binding bare names only *after* composition (no `dataColumns` field) would fix the common case but leave a same-named environment variable shadowing the column, which is exactly the precedence the issue asks for. Deferral is the only shape that puts the row above the environment without composing a row into a plan.

### Engine

- `lookup_variable` gains one arm: a name in `BoundColumnNames` resolves to nothing, checked after the two reserved namespaces (`data.*`, and #994's `$vu` / `$iteration`) and before the scopes and the generator table - so a column named `$guid` is deferred rather than generated, and one named `$vu` cannot take the identity's meaning away from it. `resolve_template` / `resolve_json_strings` take the set with an empty default, so every caller with no dataset is byte-identical.
- `POST /compose` accepts `dataColumns` (names, never values). Absent or `null` = composition as it was; a non-array, a non-string entry or an empty name is a `400 invalid_compose_request`.
- The split learns the third token kind: `tokenize_bindable_fields` / `tokenize_auth_fields` take the column set beside the reserved-namespace predicate they already had, so one split still serves a field carrying one of each. #994's `DataFieldTemplate` already keeps every token's name **as written**, which is exactly what a bare column needs - an error quotes `{{username}}` or `{{data.username}}` because the name it holds is the one the request spelled.
- `column_of` strips the `data.` prefix only where there is one, which is the whole of what the joiner needed: every refusal, encoding and header rule around it is unchanged.
- `core::bound_columns_of` is the one place a column set is read off a row or a dataset; the plan (`resolve_step`), the single-request load run (`read_load_data_set` → `context->load_template`) and the single send (`bind_data_row`, `plan_send_row_auth`) all use it.
- `first_oauth2_data_token` scans the bare spelling too - **not asked for by the issue**, but without it a bound column in an OAuth 2.0 config is deferred by composition and then reaches the token endpoint as literal text, which is the silent-wrong-request class this whole issue is about.
- Scripts: `pm.variables.get` / `.has` / `.toObject` / `replaceIn` read a bound row's bare column names first, above the environment - typed exactly as `pm.iterationData` types them for the readers, rendered as text for `replaceIn`. `data.`-prefixed names are still not variable names (`get("data.x")` stays `undefined`); a bare name the row does not carry still falls through to the scopes, so one script runs with and without a data file.

### App

- `resolveTemplate` mirrors the engine's new arm, in the engine's order, pinned by four new cases in the cross-language conformance fixture (which grew an optional per-case `boundColumns`).
- `composeForSend` sends the row's own keys for a Send-with-row; `LoadTestConfigDialog` carries the picked file's columns for a load run. A collection run sends none - the engine composes those steps itself.
- The column audit counts a bare `{{email}}` as a reference to a declared `email` column (without this it would report a working column as "declared but not referenced", the verdict that gets a working column deleted) and states the column/variable collision on the audit surface.
- `{{` autocomplete offers both spellings, labelled; a bare token naming a declared column that no scope defines paints as a bound-column runtime token instead of an undefined variable offering to create one.

### Merged with master (2026-08-28)

master landed #994 (the iteration identity) and #1051 (two header names that resolve to one) in this same machinery, so the merge was a real reconciliation rather than a textual one - `023d8d0` carries the reasoning per hunk. The short version: #994's refactor absorbed this branch's token-spelling work outright (`DataColumnRef` is gone), the column set became a second axis on the split it already had, and `resolve_header_block` takes it so a bound column in a header name is deferred and #1051's collision refusal still sees the resolved name.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Re-run on the merge commit, not just before it:

- Engine: `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` - **2681 passed, 0 failed** (15 new here: composer deferral and the `dataColumns` 400s, bare-column bind/escaping/refusal cases, the header-collision refusal asserted for the bare spelling, the plan-level end-to-end pair, the wire test, the OAuth refusal, the `pm.variables` row tier, and 4 conformance-fixture cases).
- App: `pnpm test` **5833 passed across 538 files**, plus `pnpm type-check`, `pnpm lint` and `pnpm format:check` - all clean.
- `mkdocs build --strict` clean; `clang-format-19` clean over the whole of `engine/{src,include,tests}` and `clang-tidy-19` clean on every touched engine source.
- **Throughput guard** (`docs/engine/benchmarks.md` harness shape: the repo's Go mock server on loopback, release build, 10s at concurrency 64, three runs per arm, and the no-dataset arm repeated in reversed order to cancel drift). Measured before the merge, against a release build of the then-current master:

  | arm | baseline (req/s) | this branch (req/s) |
  |---|---|---|
  | no dataset | 85727 / 87154 / 87651, then 87505 / 86424 / 85906 | 88600 / 86484 / 85221, then 87472 / 86112 / 85718 |
  | dataset, `{{data.id}}` | 83809 / 86079 / 82196 | 83907 / 85505 / 82256 |
  | dataset, bare `{{id}}` | (does not bind) | 84221 / 83322 / 81177 |

  Dataset-off medians: 86765 baseline against 86298 here, 0.5% apart with the sign flipping between the two rounds - inside the run-to-run spread this harness shows, which is what "unchanged" looks like on it. That is what the code says too: a run with no dataset carries no column set, so the split and the join are the ones it always ran, and the only new work anywhere is one `empty()` test per token at compose time. Dataset-on: binding a bare column costs what binding `{{data.id}}` costs - the same template, the same join.

## Checklist

- [x] I have added tests

## Related Issues

Closes #1007

### Deviations from the issue, and deferred work

1. The issue's Problem section says an unbound `{{username}}` "resolves to `""`". That was true at `e5277b6`; #1009 has since made an unknown ordinary name keep its braces, so the failure today is a literal token on the wire rather than a silent blank. Same fix, corrected framing - and the reason the deferral works at all.
2. The issue points the autocomplete work at `app/src/lib/script-variable-completion.ts`; that file completes `pm.*` accessor arguments. The `{{` completion the issue means lives in `useVariableCompletionProvider.ts` and `variable-autocomplete.tsx`, which is where the bare spelling was added.
3. `first_oauth2_data_token` was widened as described above - slightly more than the issue asked for, and the one place where deferring a name without widening the refusal would have created a new silent failure.
4. Deferred, each filed rather than left as a "later":
   - #1062 - the Send-with-row preview still resolves a bare column from the environment while the send defers it to the row. The resolver takes the parameter and the fixture pins it; nothing passes it yet, because no preview surface knows which row a run will bind. Disclosed in D18 and stated by the Data tab's audit in the meantime.
   - #1063 - `pm.variables.get("` completion and the script panel's "Referenced:" chips do not know that accessor now reads a bound row.
   - #1064 - the variable popover has no "bound data row" origin, so its shadowing explanation is incomplete while a row is bound.


athrvk:
Resolve merge conflicts

athrvk:
Conflicts resolved in `023d8d0` - master merged in, nothing rebased or force-pushed.

They were real rather than textual: master landed #994 (the `{{$vu}}` / `{{$iteration}}` identity) and #1051 (two header names that resolve to one) in exactly the machinery this branch rewires. How each side survived:

- **#994's refactor absorbed half of this branch.** `DataFieldTemplate` now keeps every token's name *as written*, which is precisely what `DataColumnRef` existed for here - so that struct is gone, and an error quotes `{{username}}` or `{{data.username}}` because the name it holds is the one the request spelled.
- **The column set became a second axis on the split master already had**, rather than a second split: `tokenize_bindable_fields` / `tokenize_auth_fields` take it beside their reserved-namespace predicate, so a field carrying `{{data.id}}` next to `{{$vu}}` next to a bare `{{username}}` is still split once and joined once - which is the property #994's own comment says two templates would break.
- **The joiner needed one line**: `column_of` strips the `data.` prefix only where there is one. Every refusal, encoding and header rule around it is master's, untouched.
- **Precedence is ordered explicitly**: `lookup_variable` answers the two reserved namespaces before the bound columns, so a column named `$vu` or `data.x` cannot take a reserved token's meaning away from it. The renderer's mirror is ordered to match, and the conformance fixture pins the pair.
- **#1051's `resolve_header_block` takes the column set too**, so a bound column in a header name is deferred and the collision refusal still sees the resolved name.
- **The single-request load run's template moved** off `LoadDataSet` to `context->load_template` in #994, because the identity binds with no row at all; the run's columns go to that one split instead.
- Docs: master's identity section and this branch's D18 record now sit side by side, and the pm-api table keeps `pm.info.vu` beside the row-first `pm.variables` note.

Verified on the merge commit rather than before it: **2681 engine tests** and **5833 app tests** green, `pnpm type-check` / `lint` / `format:check` clean, `mkdocs build --strict` clean, and clang-format 19 clean over the whole engine tree. CI is running now.

One thing worth flagging: I briefly ran clang-format over the conflicted Markdown and TypeScript as well as the C++, which mangled links in four docs. I restored those from the conflicted state and re-resolved them by hand - `mkdocs build --strict` passing is what says they are clean, and the formatter now touches only `.cpp`/`.hpp` here.

---

PR #1069, opened 2026-08-28
feat(engine): guarded VACUUM at startup returns freed pages to disk - #990

### LINKED ISSUE #990: engine(db): freed pages never return to disk - guarded VACUUM after the startup prune

athrvk:
## Description

Deleting rows moves their pages to SQLite's freelist, where the engine reuses them and the filesystem never sees them again - so a workspace that once held weeks of heavy runs kept that high-water mark forever, whatever the run prune, the trash purge and the `metrics` drop freed. `Database::init()` now ends with a `VACUUM` that gives those pages back, guarded so a database with nothing worth reclaiming pays three PRAGMA reads and nothing else.

**Root cause, approach, and the alternative rejected.** Nothing in the engine ever ran a bare `VACUUM`, `PRAGMA auto_vacuum` or `incremental_vacuum` - the only `VACUUM` in `engine/src` is the backup's `VACUUM INTO`, which compacts a *copy* and leaves the live file alone. The one place that had considered it declined in a comment: a rewrite under a write lock, and "startup is the worst possible moment to pay that". That objection is what the fix had to answer rather than override, and it turns out not to hold *at this point in startup*: `daemon.cpp` runs `init()` before it starts the HTTP listener, and the lock file has already refused a second engine, so the write lock has nothing waiting behind it. So the pass sits at the end of `init()`, gated on both thresholds the issue recommends - at least 25% of the file's pages free **and** at least 10 MiB in them. Each threshold is what makes the other safe: the fraction alone rewrites a nearly-empty workspace on every start (a database holding one collection is mostly freelist the moment anything is deleted from it), the byte floor alone rewrites four gigabytes to win back eleven megabytes. The rejected alternative is `auto_vacuum`: switching an existing database's mode requires a full `VACUUM` anyway, and it then changes page bookkeeping for every write thereafter - a permanent per-write cost to avoid an occasional startup one.

**Three deliberate deviations from the issue text, all argued:**

1. **It runs last of the startup passes, not immediately after the prune.** The issue says "right there", after `prune_runs_configured()`. But `purge_expired_trash_configured()` (#988) runs after it and frees pages too, and a reclamation placed between them would give back the prune's share and leave the purge's for next time. Last is the only position that sees what every sweep freed.
2. **The `VACUUM` is followed by a `PRAGMA wal_checkpoint(TRUNCATE)`, which the issue does not mention.** Under WAL the rewritten image lands in the `-wal` file and the database is not resized until a checkpoint copies it back - so without it this pass frees every page it set out to and leaves the file *exactly as large as it found it*. That is not a theory: the first version of the test measured it. Best-effort, and stated as such at the site: a checkpoint that cannot truncate has still committed the rewrite, and the next one returns the space.
3. **The below-threshold tests assert the file does not shrink, not that it is byte-identical.** The issue asks for "no size change"; empirically a start writes a page of its own (196,608 -> 200,704 bytes, measured), so byte-equality is an assertion about `init()`'s other writes rather than about this pass. What each negative test pins instead is the threshold it is about, and the mutation checks below show it is the guard doing the declining.

**One repo-level tidy inside the diff, not beside it:** the backup's `vacuum_into` and this pass both need a second connection for a statement sqlite_orm cannot run on the connection it holds. That open / `busy_timeout` / close sequence is now `open_workspace_connection`, called by both, rather than written twice a thousand lines apart in one file.

**Two consumer-trace findings, one of which does not hold.** A trace of every surface that could see this suggested `WorkspaceBackupCard.tsx` displays a database-derived file size and would change. It does not: `VACUUM INTO` already writes a compacted copy, so the snapshot's `sizeBytes` was never a function of the live file's freelist. The second finding is real and is now written down: `db.init()` returns before `/health` answers and the app gives the engine **45 seconds** to come up (`ENGINE_HEALTH_MAX_ATTEMPTS` x `ENGINE_HEALTH_POLL_INTERVAL_MS`), so the rewrite is on that budget. The thresholds bound how *often* it runs, not how long one takes, which is exactly why the outcome is logged at info level - a slow start is explainable rather than mysterious. No app change; nothing reads the live file's size, and nothing parses engine startup logs.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`: **2606 engine tests, all green** (2603 before, three new). `mkdocs build --strict -f .github/mkdocs.yml` clean. `clang-format-19 --dry-run -Werror` clean over the whole of `engine/{src,include,tests}`; `clang-tidy-19` clean on both touched translation units. No pre-existing failures reproduced on the base commit. The app is untouched, so its suite was not run - no `app/` file is in this diff and no renderer, MCP or CLI surface reads the live database's size.

Three new cases in `db_test.cpp`, one per quadrant of the guard, each verified by reverting the thing it is supposed to pin:

| Test | Pins | Mutation check, run |
|------|------|---------------------|
| `StartupReclaimsAHeavilyPrunedDatabase` | 24 MiB pruned to one: both thresholds crossed, so the file shrinks past half | Reclamation call removed from `init()` -> **fails** (file stays at 25 MB) |
| `StartupLeavesADatabaseBelowTheByteFloorAlone` | 4 MiB pruned to one: ~95% free, under 10 MiB, so nothing runs | `VACUUM_MIN_RECLAIMABLE_BYTES` -> 0 -> **fails** |
| `StartupLeavesADatabaseBelowTheFreelistFractionAlone` | 60 MiB pruned to 48: 12 MiB free but a fifth of the file, so nothing runs | `VACUUM_MIN_FREELIST_PERCENT` -> 0 -> **fails** |

Each mutation was applied, rebuilt and run - the table records measured results, not expectations. The two negative cases are what makes the third mutation meaningful: with both thresholds zeroed they redden while the positive case stays green, which is the guard being the thing that declines rather than the pass being absent.

One flakiness guard worth naming: the seeded runs are stamped from the wall clock rather than from a counter. Stamped in 1970 they are all older than `runRetentionDays`, so the *next* start's prune deletes every one of them and each case ends up measuring an empty database - which is how the fraction test first passed for the wrong reason.

## Checklist

Docs move in the same commit: `docs/engine/db-schema.md` gains a **Reclaiming freed pages** section under the startup heading (thresholds and why both, the placement, the second connection, the checkpoint, the 45-second budget, the rejected `auto_vacuum`), the run-retention paragraph links to it, and the `metrics`-drop paragraph - which asserted as settled fact that the file "does not shrink, because a `VACUUM` would rewrite the whole database under a write lock at startup" - now says what actually happens. The code comment that made the same claim points at the pass instead.

## Related Issues
Closes #990

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nn5BdHaf3yonsvTq1HM7fr)_

### DIFF
--- a/docs/engine/db-schema.md
+++ b/docs/engine/db-schema.md
@@ -107,6 +107,40 @@ endpoint reports it as a `recovery` node, absent on a clean start - see
 dismissible banner naming the database path and, when there is one, the
 quarantined file and the `.recover` command that reads it.
 
+### Reclaiming freed pages
+
+Deleting rows - a prune, a trash purge, the `metrics` drop - moves their pages to SQLite's
+freelist, where the engine reuses them and the filesystem never sees them again, so a workspace
+that once held weeks of heavy runs keeps that high-water mark forever. `Database::init()` ends
+with a guarded `VACUUM` that gives them back (issue #990).
+
+Guarded, because a `VACUUM` rewrites the whole database: it runs only when **both** thresholds are
+crossed - at least **25%** of the file's pages are free **and** they hold at least **10 MiB**
+(`VACUUM_MIN_FREELIST_PERCENT` / `VACUUM_MIN_RECLAIMABLE_BYTES`, compile-time, deliberately not
+configurable). The fraction alone would rewrite a nearly-empty workspace on every start; the byte
+floor alone would rewrite four gigabytes to win back eleven megabytes. Neither threshold crossed
+means the pass reads three PRAGMAs and returns.
+
+Three things worth knowing about where it sits:
+
+- **Last in `init()`**, after the run prune and the trash purge, so it sees the pages every sweep
+  freed rather than only the first one's.
+- **On a connection of its own**, like `POST /workspace/backup`'s `VACUUM INTO` - sqlite_orm
+  exposes no way to run these statements on the connection it holds. Nothing contends for it:
+  `init` runs before the HTTP listener starts, and the lock file has already refused a second
+  engine.
+- **The `VACUUM` is followed by a `wal_checkpoint(TRUNCATE)`.** Under WAL the rewritten image
+  lands in the `-wal` file and the database is not resized until a checkpoint copies it back, so
+  without it the pass would free every page it set out to and leave the file exactly as large as
+  it found it.
+
+The outcome is logged at info level (`Reclaimed N KB of freed database pages`), because the cost
+is paid on the startup path - `db.init()` returns before `/health` answers, and the app gives the
+engine 45 seconds to come up. A rewrite slow enough to matter is one this line explains.
+Best-effort like the sweeps before it: a failure is a warning, never a daemon that will not start.
+`auto_vacuum` was rejected as the alternative - switching an existing database's mode requires a
+full `VACUUM` anyway, and it changes page bookkeeping for every write thereafter.
+
 ---
 
 ## Tables
@@ -842,7 +876,8 @@ backlog cannot stall `/health`, SSE, or the runs poll. The cascade itself lives 
 is wired into both at once. `prune_runs_configured()` reads the two
 knobs (config, `data_retention`, defaults 200 / 30) and runs at **startup** (`Database::init`) and
 after a run reaches a **terminal** status (design mode's `store_result`, and the load-run
-completion/failure paths in `run_manager.cpp`).
+completion/failure paths in `run_manager.cpp`). What a prune frees is disk the file keeps until
+[Reclaiming freed pages](#reclaiming-freed-pages) gives it back.
 
 ---
 
@@ -976,8 +1011,8 @@ along with the `MetricName` enum and `db::Metric`. `sync_schema()` only syncs th
 storage still declares - it never drops one that was removed from it - so `Database::init()`
 issues an explicit `DROP TABLE IF EXISTS metrics`, and an upgraded database sheds the table and
 its rows on first start. The freed pages return to SQLite's freelist for reuse; the file itself
-does not shrink, because a `VACUUM` would rewrite the whole database under a write lock at
-startup.
+shrinks only when the guarded reclamation at the end of `Database::init()` decides they are worth
+a rewrite - see [Reclaiming freed pages](#reclaiming-freed-pages).
 
 ---
 
--- a/engine/include/vayu/core/constants.hpp
+++ b/engine/include/vayu/core/constants.hpp
@@ -914,6 +914,26 @@ constexpr int MAX_BACKUPS_RETAINED = 5;
  * spares is reclaimed by the next one rather than kept.
  */
 constexpr int64_t SPEC_DOCUMENT_SWEEP_GRACE_MS = 600'000;
+/**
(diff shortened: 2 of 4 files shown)


athrvk:
A second commit fixes two `clang-tidy-19` findings in the new tests - `bugprone-implicit-widening-of-multiplication-result` on the `MIB` constant and `readability-math-missing-parentheses` on the run stamp. Both are in code this PR added, and both would have failed the engine tidy gate, which lints whole changed files.

The PR body's "clang-tidy clean on both touched translation units" was written from the run over `database.cpp`; the run over `db_test.cpp` finished afterwards and found these two. It is true as of `31cd039` - re-run over both files, no findings.

---

Issue #173, opened 2026-07-28
engine,app: flat vs nested error-body mismatch - CRUD validation errors surface as a bare "HTTP 400"


athrvk:
Base: master `61a3b3e` (2026-07-28). All anchors verified at this commit; the `/import/apply` addendum verified at `4136c66` (2026-07-29).

## Problem

The engine emits error bodies in **two shapes**, and the app&#39;s shared HTTP client only understands one of them - so every validation and not-found message built by the flat-shape routes is discarded and the user sees a generic `HTTP 400: Bad Request`.

**The flat shape** - `send_error` is a single inline helper, `engine/include/vayu/http/routes.hpp:43-46`:

```cpp
inline void send_error (httplib::Response&amp; res, int status, const std::string&amp; message) {
    res.status = status;
    res.set_content (nlohmann::json{ { &#34;error&#34;, message } }.dump (), &#34;application/json&#34;);
}
```

It has **42 call sites across 9 route files** (verified by `rg -c send_error engine/src/http/routes/*.cpp`): `runs.cpp` (10), `requests.cpp` (8), `execution.cpp` (7), `collections.cpp` (5), `metrics.cpp` (5), `environments.cpp` (3), `config.cpp` (2), `globals.cpp` (1), `scripting.cpp` (1). This includes every 400/404 from the resource CRUD verbs (`POST`/`PUT` `/requests`, `/collections`, `/environments`) - missing required field, `null` on a no-default field, unknown id, malformed entries.

**The nested shape** - `{&#34;error&#34;: {&#34;code&#34;: ..., &#34;message&#34;: ...}}`:

- `config.cpp:50-54` `config_error()`, whose own comment states the defect outright (`config.cpp:47-49`): *&#34;Build the nested error body the app&#39;s http-client reads (`errorData.error.message`); the flat `{&#34;error&#34;: &#34;...&#34;}` shape is dropped by the client and surfaces only as a bare &#34;HTTP 400&#34;.&#34;*
- `oauth.cpp` throughout (`:68`, `:156`, `:168`, `:190`), with the header comment (`oauth.cpp:16-17`): *&#34;Errors use the nested shape ... which the app&#39;s http-client parses into ApiError.errorCode.&#34;*
- One path in `execution.cpp:565` (OAuth preflight failure detail).

**The single client reads only the nested shape** - `app/src/services/http-client.ts:87-92`:

```ts
const errorData = await response.json().catch(() =&gt; ({}));
[...]

PR #1072, opened 2026-08-28
feat(app): a Trash view, an undo toast, and MCP trash tools over the engine's soft delete - #989

### LINKED ISSUE #989: app(collections): a Trash view and an undo toast on top of the engine's soft delete
App half of the trash feature. **Blocked by #988** (engine soft delete), which defines the endpoints this consumes. From the 2026-08-25 UX sweep, verified at master `96385b2`.

## Problem

Deletes in the collections tree are guarded only by a confirm dialog (`app/src/modules/collections/useTreeCrud.ts:415-460` - `DeleteConfirmTarget`, hard mutations, no undo affordance). Once #988 makes the engine side soft, the app should give users the two affordances that make soft delete real: an immediate undo, and a place deleted things live until purge.

## Fix / solution pathway

1. **Undo toast**: after a successful delete mutation, show a toast with an Undo action for ~8s calling `POST /trash/{id}/restore` and invalidating the tree query. Copy the repo's existing toast pattern (grep `toast` in `app/src` for the primitive - use the primitive, not a hand-rolled copy, per the repo rule).
2. **Trash view**: a modest surface listing `GET /trash` roots (name, type, deleted-at, item count), with per-row Restore and Delete forever (confirm), plus retention copy ("kept 30 days"). Placement: follow where History lives as a sidebar module (`app/src/modules/history/sidebar/HistoryList.tsx` is the closest structural sibling to copy); wire route/navigation the way that module does.
3. **Tree interactions**: restored items reappear via query invalidation; no optimistic tree surgery needed in v1.

## App changes

This issue is the app change. Engine changes: none (consumes #988's endpoints as shipped; if a gap surfaces, it goes back on #988 before this lands).

## Tests

Vitest: undo toast appears on delete and its action calls restore + invalidates; Trash view renders rows, restore and purge paths fire the right mutations; empty state. Follow the store/component test patterns beside `useTreeCrud`. `cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check`. Mind the repo rules: no platform assertions, stub both branches where platform-dependent.

## Docs (same commit)

If `docs/app/` documents the collections tree interactions (verify with `rg -l "delete" docs/app/`), add the trash/undo behaviour there; otherwise none.

## Sequencing

**After #988 merges.** Touches `app/src/modules/collections/` (shared with nothing currently open on the app side - #467 stage 2 is tooling-only) and adds a module. Parallel-safe against all engine work.

## Acceptance criteria

- [ ] Deleting shows an undo toast that actually restores.
- [ ] The Trash view lists, restores, and purges; copy states the retention window.
- [ ] App suite, type-check, lint, format:check all green.

---

### LINKED ISSUE #1070: app(collections): an undone delete does not reopen the tab the delete closed
Found while building #989 (the Trash view and the undo toast), and deliberately left out of that PR rather than widened into it.

## Problem

Deleting a request closes any tab open on it, and deleting a collection closes the tabs of everything its cascade took: `app/src/modules/collections/useTreeCrud.ts` calls `closeTabsForEntities(affected)` on success. Undo restores the rows - the tree repopulates through query invalidation - but the tabs stay closed, so a user who deletes the request they were editing and immediately takes it back is left looking at whatever tab the close fell through to, and has to find the request in the tree again.

Two smaller consequences of the same seam, both in `app/src/queries/collections.ts`:

- `useDeleteRequestMutation` calls `useResponseStore.getState().clearResponse(deletedId)`, and `closeTabsForEntities` clears it again. A restored request comes back with an empty response pane even if a response was on screen a second earlier.
- `useDeleteCollectionMutation` calls `useDataFileStore.getState().clearDataFile(deletedId)`, so a restored collection has lost its remembered data-file path (#599).

None of this is wrong for a *delete*, which is what those calls were written for. It is only visible now that a delete can be taken back.

## Fix / solution pathway

The honest shape is probably not "reopen everything": restoring a collection whose cascade closed nine tabs should not reopen nine tabs. Two options worth weighing in the PR that takes this:

1. **Reopen only what the undo can justify** - the tab that was *active* when the delete happened, if it was one of the closed ones. Requires the delete path to remember that one id alongside the trash entry.
2. **Let the restore say so instead** - the undo toast already exists; a restored request could simply be reopened by clicking the row, and the toast could say nothing about tabs. Cheapest, and arguably right.

The response body is a session artifact and probably should not be preserved; the data-file path is stored state and is the stronger candidate for restoring alongside the row.

## App changes

This issue is the app change. Engine changes: none - `POST /trash/{id}/restore` already returns the restored entry.

## Tests

Vitest, beside the existing `CollectionTree.undo-toast.test.tsx`: delete a request with its tab open, undo, assert whatever behaviour is chosen (tab reopened, or explicitly not reopened and the row clickable). Mutation-check it.

## Docs (same commit)

`docs/app/state-management.md` describes what the delete mutations clear; if the restore path starts preserving any of it, that paragraph moves with the code.

## Sequencing

**After #989 merges** - it touches the undo path that PR introduces. Parallel-safe against all engine work; touches `app/src/modules/collections/` and `app/src/queries/collections.ts`.

## Acceptance criteria

- [ ] The chosen behaviour is implemented and stated in the PR body, with the rejected alternative and why.
- [ ] A restored collection does not silently lose its data-file path, or the PR says why that is correct.
- [ ] App suite, type-check, lint, format:check green.

---

### LINKED ISSUE #1071: app(mcp): the delete tools promise a Trash that the MCP surface cannot list, restore from, or empty
Found by the consumer trace for #989 (the app-side Trash view), and left out of that PR because it is a different surface.

## Problem

`app/electron/mcp/tools.ts` already tells an agent about the Trash. `delete_collection`'s description says the deleted item

> "goes to Vayu's Trash, where the user can restore it until the retention window (`trashRetentionDays`, 30 days by default) runs out - not something to undo from here."

and `delete_request` says the same. That sentence is accurate about the engine (soft delete shipped in #988 / PR #1045) and is currently the only mention of the Trash on the MCP surface: there is no `list_trash`, no `restore_*`, no purge tool, and `app/electron/mcp/engine-client.ts` has no method for any of the three routes. So an agent working through MCP can put things in the Trash and can neither see what it put there nor take it back - "not something to undo from here" is true only because nobody wired it.

The engine side needs nothing: `GET /trash`, `POST /trash/{id}/restore` and `DELETE /trash/{id}` have been serving since #1045 and are documented in `docs/engine/api-reference.md` under Trash.

## Fix / solution pathway

1. Three methods on `app/electron/mcp/engine-client.ts`, mirroring the renderer's `apiService.listTrash` / `restoreTrashEntry` / `purgeTrashEntry` (`app/src/services/api.ts`).
2. `list_trash` and `restore_trash_entry` as tools. **Purge is the one to think about** rather than add by reflex: `DELETE /trash/{id}` is the only irreversible call in the set, and an agent that can both delete and purge can destroy a collection permanently in two calls with no human in between. Either leave it out and say so in the delete tools' descriptions, or gate it the way any other destructive tool here is gated - decide in the PR, do not leave it implicit.
3. Update the `delete_*` descriptions once restore exists, since "not something to undo from here" stops being true.
4. Check whether `app/electron/mcp/resources.ts` should expose the trash as a resource alongside collections, or whether a tool is the whole answer.

## App changes

This issue is the app change. Engine changes: none.

## Tests

Follow the existing MCP tool tests in `app/electron/mcp/`: each new tool's happy path, the 404 and 409 a restore can answer with (a request whose collection is itself deleted), and the purge decision from item 2 asserted rather than assumed.

## Docs (same commit)

`docs/engine/mcp.md` lists the MCP surface and must gain whatever ships.

## Sequencing

Parallel-safe. Independent of #989 (a different process - Electron main, not the renderer) and of #1070. Land whenever.

## Acceptance criteria

- [ ] An agent can list the Trash and restore from it through MCP.
- [ ] The purge decision is made explicitly and stated in the PR body.
- [ ] The `delete_*` tool descriptions no longer claim an undo the surface now has.
- [ ] `docs/engine/mcp.md` updated; app suite, type-check, lint, format:check green.

---

athrvk:
## Description

The engine's delete went soft in #988 (PR #1045): a deleted collection or request is stamped and kept, filtered out of every live read, and swept by a startup purge after `trashRetentionDays`. The app never grew a reader for any of it - and had already been rewritten to promise one. The confirm dialog has been telling users their collection *"will be moved to the Trash, where they can be restored"* while no Trash existed and no restore was reachable. This is the repo's most-repeated defect in its most user-visible form: one layer writes, no layer displays, and here the UI actively promises the missing half.

**Root cause and approach.** Nothing was broken in the engine; the app was missing the surfaces over an API that already worked. So this adds the reader, in the shape the existing code already had room for: wire types and an `apiService` call per route, a `queryKeys.trash` family with hooks beside the others, a sixth drawer view structured after History (`modules/trash/`), and an undo toast raised by the delete itself using the toast primitive's existing `action` slot. The alternative I rejected was client-side archiving - snapshotting the deleted subtree in the renderer so a restore could rebuild the tree optimistically. It loses to the same argument `state-management.md` already makes about cascade deletes: which rows a delete or a restore touched is engine-side knowledge (the cohort is defined by a `deleted_at` timestamp the client never sees), so a client that re-derives it drifts from the engine's definition of "descendant". Both directions invalidate coarsely instead.

**Scope: three issues, at your request on this PR** ("Impl 1070 and 1071 in this PR itself"):

- **#989** - the Trash drawer view (⇧⌘T) and the undo toast.
- **#1070** - the undo now gives back the tab the delete closed and the collection's remembered data-file path. Both were discarded on the way out and cannot be read back afterwards, so the undo closure captures them before the mutation runs. **Only the tab that was focused**: restoring a collection whose cascade closed nine tabs must not reopen nine, and the complaint this answers is the narrower one. The response body is deliberately not carried - it is a session artifact, and holding a body plus its raw copy alive against a maybe-undo is the leak `closeTabsForEntities` exists to prevent.
- **#1071** - `list_trash`, `restore_trash_entry` and `purge_trash_entry` on the MCP surface, whose `delete_*` tools have described the Trash since #1045 while saying restoring was "not something to undo from here" - true only because nobody had wired it. **The gating is the part worth reviewing**: `restore_trash_entry` puts data back, so it carries the write toggle alone with `destructiveHint: false` - a confirmation step for an undo prompts for the wrong thing. `purge_trash_entry` is the Trash's own hard delete and is double-gated exactly like `delete_collection`: writes enabled, then a confirmation whose prompt names the entry and what it holds, read off `GET /trash` first so an unreadable id refuses rather than prompting with counts nobody verified.

Three fields the engine sends now have a reader, each chosen because nothing else on screen could say it: **the cascade counts** (a folder in the Trash is indistinguishable from an empty one otherwise), **`reparentedToRoot`** (a collection whose parent is gone comes back at the top level - the engine moving the user's folder), and **`trashRetentionDays`** (read from config rather than written, so it stays true when the setting changes; `0` renders as "kept until you delete them", because the engine's purge returns before computing a cutoff - the plausible reading is exactly backwards).

**Deliberate deviations from the issue text**, both argued rather than silent:

1. **The undo toast takes the `info` default (4s, scaled by the user's notification setting) rather than #989's "~8s".** A hardcoded duration overrides a preference the user set - including "never", which means "wait for me" - to buy something the feature already has: the toast is the fast path, and the Trash view is the same undo with the retention window behind it. It also matches the drag-and-drop undo already in the tree. Say the word and I will pin a longer window.
2. **The Trash row's actions stay visible instead of hover-revealed.** The design system hides row actions because a drawer row has a primary action of its own that they must not compete with; a Trash row has none. Recorded in `docs/design-system.md` beside the rule it qualifies, using that rule's own test.

`TrashEntry` deliberately omits the engine's `parentId`, the way `RequestExample` omits the `order` and timestamps it does not use. A review pass caught it as dead state first.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Every gate run locally on the final commit, after merging current `master` in:

- `cd app && pnpm test` - **540 files, 5844 tests, green** (baseline on `7105525` was 532 / 5780)
- `pnpm type-check` - renderer, electron and electron-tests all clean
- `pnpm lint` - zero errors, zero warnings
- `pnpm format:check` - clean
- `mkdocs build --strict` - clean

No pre-existing failures were encountered. CI on `888dbf2` was fully green; the one red "CI gate" on `ba0c086` was that run being superseded mid-flight by the merge push, and a CodeQL failure on `056812f` was `ECONNRESET` while downloading the CodeQL bundle, before any analysis ran - re-run once, passed.

New cases are mutation-checked rather than assumed, and three of them exist because a check found the code green where it should not have been:

- Removing `offerUndo` reddens 3 of 4 undo cases; removing the `reparentedToRoot` branch reddens exactly one; removing the client-state restore reddens the tab case.
- Emptying either `onSuccess` in `queries/trash.ts` used to leave the whole suite green - every view test replaces `@/queries` wholesale, so nothing exercised the invalidation that makes a restored row reappear. `app/src/queries/trash.test.ts` drives the real hooks against a real `QueryClient` with only `apiService` stubbed.
- `retention.test.ts` covers the `0` and negative cases against the engine's own `retention_days <= 0` behaviour.

**Three defects were caught by review and fixed rather than argued away.** The undo path was discarding the response the Trash view reads, so a re-parented folder was explained or not depending on which button restored it - the "two funnels" defect `app/CLAUDE.md` names, reproduced; both callers now go through one `restoreNotice()`. A Trash row expressed collection-vs-request only through an `aria-hidden` icon, so a screen reader got a name and a date and no way to tell them apart. And in the MCP work, the refusal unwrapper read the flat `{"error": "..."}` body while these routes send the nested `{"error": {"code", "message"}}` that #173 settled on - a 409 would have reached the agent as raw JSON instead of the sentence naming which collection to restore first, and its test used the same wrong shape and passed.

## Checklist

## Related Issues

Closes #989
Closes #1070
Closes #1071


athrvk:
Impl 1070 and 1071 in this PR itself

athrvk:
Both folded in, as asked - #1070 and #1071 are now part of this PR rather than follow-ups, and the body and title are updated to match (it closes all three).

**#1070** - the undo gives back the tab the delete closed and the collection's remembered data-file path. Both were discarded on the way out and cannot be read back afterwards, so the undo closure captures them before the mutation runs. Two calls worth your eye: only the tab that was **focused** comes back (restoring a collection whose cascade closed nine tabs must not reopen nine - the rest are one click away in the tree the restore just repopulated), and the response body is deliberately not carried, since holding a body plus its raw copy alive against a maybe-undo is the leak `closeTabsForEntities` exists to prevent. Nothing is put back until the engine confirms the restore, so a refused undo reopens no tab - its own case.

**#1071** - `list_trash`, `restore_trash_entry`, `purge_trash_entry`, and both `delete_*` descriptions now point at them instead of saying restoring is "not something to undo from here". The gating decision the issue asked to be made explicitly: `restore_trash_entry` carries the write toggle alone with `destructiveHint: false`, because a confirmation step for an undo prompts for the wrong thing; `purge_trash_entry` is double-gated exactly like `delete_collection` - writes enabled, then a confirmation whose prompt names the entry and what it holds, read off `GET /trash` first so an unreadable id refuses rather than prompting with counts nobody verified. I did not add a trash resource to `resources.ts`: that file is ambient-context data, and the trash is an occasional action-oriented query `list_trash` already answers.

One defect worth naming, because it is the kind that passes review by looking right: the MCP refusal unwrapper read the flat `{"error": "..."}` body, while these routes send the nested `{"error": {"code", "message"}}` that #173 settled on - so a 409 would have reached the agent as raw JSON instead of the sentence naming which collection to restore first, and its test used the same wrong shape and passed. Fixed against `routes.hpp`, with the test reddened when the nested branch is removed.

Also merged current `master` in, so this is no longer behind. All gates green on the final commit: 540 test files / 5844 tests, type-check, lint, format:check, `mkdocs build --strict`. Earlier CI noise was not this PR's - one "CI gate" red was that run being superseded by the merge push, and one CodeQL red was `ECONNRESET` fetching the CodeQL bundle before any analysis ran, re-run once and green.

---

Issue #1055, opened 2026-08-27
engine(loadtest): {{$vu}} in a credential goes out written as it stands - the identity stops at the request because deferral needs rows


athrvk:
Deferred from #994, which added the `{{$vu}}` / `{{$iteration}}` identity and deliberately stopped it at the request.

## Problem

The identity binds every bindable field of a request - the URL, header names and values, the body, both halves of every form field - and **not** a credential. A basic-auth username of `user-{{$vu}}`, or a bearer token carrying `{{$iteration}}`, is base64-encoded or written into the header line with the token text still in it, and goes out that way. Loud by #186's rule (the literal reaches the server and comes back as a 401 rather than resolving to something wrong), but it is not what the author asked for.

The reason is structural rather than an oversight, and it is why #994 did not just add the namespace to `tokenize_auth_fields`. A credential is encoded when the request is *built*, so a value can only reach one if the build was **deferred** (`AuthResolution::Defer`, issue #591) - and a build is deferred only for a run that carries `data` rows, because that is the case the deferral was added for. Splitting identity tokens out of credentials as things stand would therefore bind `{{$vu}}` in a data-driven run and silently not bind it in every other run, which is a rule nobody can hold in their head. One rule that is wrong everywhere beats two rules that differ by whether a data file happens to be attached.

Anchors at the commit that filed this: `engine/include/vayu/core/scenario_data.hpp` (`tokenize_auth_fields`, which states the exclusion and points here), `LoadDataSet::auth_resolution()` in `engine/include/vayu/core/load_strategy.hpp`, and the deferral decision in `plan_send_row_auth` / `build_load_request`.

## Fix / solution pathway

Make the deferral decision depend on **what the credentials carry** rather than on whether the run has rows:

1. Split the credentials for both reserved namespaces (`tokenize_auth_fields` over the same predicate `tokenize_bindable_fields` uses), so a credential carrying either kind of token yields a non-empty template.
2. Derive `AuthResolution::Defer` from that template alone, at every site that decides it today - the single-request load path (`build_load_request`), the scenario plan (`resolve_scenario`, where a step's `auth_template` already means "this step's auth was deferred"), and `POST /execute` (`plan_send_row_auth`).
3. `bind_iteration` then joins the credentials against the whole `IterationBinding` rather than against the row alone, which is the one-line change once (1) and (2) are in place - the joiner already resolves both namespaces.
[...]

PR #1085, opened 2026-08-28
feat(engine): generate the {{$guid}} family per iteration under load - #995

### LINKED ISSUE #995: engine(loadtest): dynamic variables freeze at composition - regenerate the {{$guid}} family per iteration on the load path
Sub-issue of #992 (gap 3 - the parent carries the full context and the out-of-scope list). Verified at master `96385b2`.

## Problem

Dynamic variables generate once per load run, not per iteration: plan steps are composed - including all `{{$guid}}`-family substitution - once, before the first send (`engine/include/vayu/core/scenario_plan.hpp:73`; the generator table is `engine/src/http/request_composer.cpp:142-176`). So a run configured with `{{$randomUUID}}` in the body sends the SAME uuid on every request from every VU for the whole run. `docs/app/variable-resolution.md:332-337` records this as a deliberate scope cut ("per-iteration interpolation on a 60k+ RPS hot path ... deliberately kept out of #226's scope") - this issue is the follow-through, with the hot-path cost controlled rather than assumed.

The practical sting: uniqueness is exactly what these variables exist for. Idempotency keys, unique emails, correlation ids - all silently degrade to one shared value precisely when concurrency makes collisions matter.

## Fix / solution pathway

The cost-control insight: **pay only when a generator is present, and only for the parts that contain one.**

1. At plan composition, instead of substituting generators, detect them: scan each composed part (url, each header value, body) for `{{$...}}` occurrences and record, per step, WHICH parts carry generators (a small bitmask/list on the plan step). Runs with none - the overwhelming majority - take today's path untouched, zero per-iteration cost. This mirrors how `apply_data_template` already re-binds only data-carrying parts per iteration (`engine/src/core/scenario_load.cpp:415-423`).
2. For flagged parts, keep the pre-generator template text on the plan step and re-run generator substitution per iteration at the same bind point, reusing the existing generation functions (same table, same escaping rules as `{{data.*}}` - `engine/include/vayu/core/scenario_data.hpp`).
3. Thread-safety: generation under load runs on worker threads - the generator implementations must not share unsynchronized state (check how `request_composer.cpp`'s generators get randomness; `thread_local` engines per the `metrics_collector.cpp:39` pattern if needed).
4. **Benchmark gate, stated in the PR with numbers**: a generator-free run must show no measurable regression (the detection is composition-time only); a generator-heavy run's cost is reported honestly. `docs/engine/benchmarks.md` has the harness conventions.
5. Design mode is untouched (already generates per send).

## App changes

None functional, verified - the renderer mirror (`app/src/lib/dynamic-variables.ts`) documents behaviour but does not simulate load timing. `docs/app/variable-resolution.md` is the consumer to fix (below).

## Tests

Engine gtest: a two-iteration run with `{{$randomUUID}}` in the body observes DISTINCT values at the mock listener (this is the mutation check for the whole feature - today it fails); a generator-free plan takes the untouched path (assert the flag is empty); escaping in JSON body per `scenario_data.hpp` rules; concurrent generation produces no torn/duplicate values across VUs in a same-instant burst (statistical, keep honest about what it can prove). `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`. Clean under the #928 tidy families.

## Docs (same commit)

`docs/app/variable-resolution.md:332-337` - replace the once-per-run rule with the new per-iteration rule and its zero-cost-when-absent design; note the closed loop on the old #226 scope cut.

## Sequencing

Last of #992's subs - builds on the unified binding path from the rows sub-issue and pairs naturally with the identity sub. Touches `scenario_plan.*`, `scenario_load.cpp`, `request_composer.cpp`. The tidy paydown program is complete and `readability-function-cognitive-complexity` is enabled since #1021 - new code must be clean under the full enabled set, including the complexity threshold.

## Acceptance criteria

- [ ] `{{$guid}}`-family values differ per iteration under load; generator-free runs demonstrably skip the machinery.
- [ ] No regression on the generator-free benchmark; generator cost reported with numbers in the PR.
- [ ] `variable-resolution.md` states the new rule.

---

athrvk:
## Description

A run's request half is composed once and sent many times, so every generator the table resolved at composition reached the wire as one value repeated for the whole run - the uniqueness `{{$randomUUID}}` is written for, lost exactly where concurrency makes collisions matter.

**Root cause.** `lookup_variable` (`engine/src/http/request_composer.cpp`) resolves a generator name in the table's own position, and every load run's request half is composed exactly once - engine-side for a scenario plan step, through `POST /compose` for a single-request run started by the app or MCP. The reserved namespaces `{{data.column}}` and `{{$vu}}` / `{{$iteration}}` already survive composition and are joined per iteration by one shared template (`tokenize_bindable_fields` -> `apply_iteration_template`), and every executor - scenario load, single-request load, design-mode collection, `POST /execute` - already drives that join.

**Approach.** Make a generator token a third kind the same template carries, and defer generation at composition only where the composed payload is *repeated*. Composition takes an explicit `DynamicResolution`, and `POST /compose` exposes it as `deferDynamicVariables`; plan resolution sets it for every step, and the app and MCP set it when they compose a payload for a run. `keeps_reserved_namespace` gains a generator arm so those tokens split into the same `StepDataTemplate`, and `TemplateJoiner::value_of` gains the one arm that *computes* rather than reads - fresh per occurrence, into a scratch member. The cost guard is structural rather than measured: a request spelling no generator has nothing left for the split to find, so its template stays empty and every submission still pays the one `empty()` test it always did.

**The alternative I rejected:** reserving the generator names globally, the way `{{$vu}}` is - one rule, no flag. It changes what `POST /compose` returns for every caller, makes `pm.variables.replaceIn("{{$guid}}")` and a pre-request script's `pm.request` read the token instead of a value, and moves the shared conformance fixture. That is a far wider blast radius than the gap justifies, and it would make the preview and the Send worse to fix the run.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Re-run on the merged head (`7c0090f`, after merging `master` in): `python build.py -e -t && cd engine && ctest --preset linux-dev` - **2705 passed, 0 failed**; `cd app && pnpm test` - **544 files, 5960 tests passed**; `pnpm type-check`, `pnpm lint`, `pnpm format:check` and `mkdocs build --strict` all clean. (Before the merge, on this branch's own work alone: 2701 engine, 5872 app.) clang-format 19 and clang-tidy 19 clean over every changed file, whole-file scope. No pre-existing failures on the base commit.

Engine tests added, all mutation-checked - reverting the deferral and the split predicate turns five of them red, verified by actually reverting both and re-running:

- `LoadDataTest.ADeferredGeneratorProducesADistinctValuePerIteration` - six iterations, six distinct ids read off the listener (the headline case).
- `ScenarioLoadTest.EveryIterationOfEveryUserGeneratesItsOwnValue` - eight sends across four virtual users, all distinct.
- `ScenarioRunnerTest.AGeneratorBindsPerPassOfADesignRunRatherThanFreezing` - the design-mode collection run, where deferring and then not binding would put the literal token on the wire.
- `LoadDataTest.TwoGeneratorsInOneFieldAreTwoValues`, `ScenarioLoadTest.AGeneratedValueBindsIntoAJsonBodyAsAStringsContent`, and two `ScenarioDataXmlBodyTest` cases (escaped in character data, refused inside a comment - a generator binds under a cell's rules, both directions).
- The cost guard, asserted structurally: `LoadDataTest.ARunWithoutAGeneratorCarriesAnEmptyTemplate` and `ScenarioPlanTest.AStepWithoutAGeneratorCarriesAnEmptyTemplate`.
- Composition itself: `DeferLeavesEveryGeneratorNameWrittenAsItStands` (both values of the option, in one test), `DeferIsTheTableOnly`, `ADefinedVariableStillOutranksTheGeneratorUnderDefer`, and the `400` on a non-boolean field.

`IdentitySplit.ARequestCarryingNoReservedTokenSplitsNothing` used a `{{$guid}}` as its example of a token that never survives composition. That is the rule this issue changes, so it moved to an ordinary variable and gained a sibling asserting the new one - noted rather than quietly edited.

### Benchmark gate

Full method and numbers in `docs/engine/benchmarks.md`. On this 4-core container, running the Go mock server beside the engine:

- **Feature on**, four arms interleaved and rotated per round (c=32, 5 s, 8 rounds): off 20 007, literal 21 367, identity (`{{$iteration}}`) 21 990, on (`{{$randomUUID}}`) 20 854 req/s medians. Nothing is resolvable - `on` lands above `off`, and the arms' own min-max spans are wider than any distance between them. An unrotated first pass had `on` 26% below `off`; it did not survive rotation, which is exactly what the method rotates for.
- **Feature off, against master**, paired and alternating (10 rounds, 5 s, c=32): **-0.01%** on medians. The null control - the same harness with this branch's binary in *both* arms - reports **-1.51%**, so the harness moves the number further than the change does.

The guard the throughput constraint actually rests on is structural, not this measurement: a request spelling no generator has an empty template, and the submission path tests that before copying or binding anything.

## Checklist

## Deliberate deviations from the issue

1. **The issue says "App changes: None functional, verified". That is wrong, and the app change is required.** The renderer and MCP compose a single-request load run's payload themselves (`POST /compose`) and hand the composed result to `POST /runs`, so the engine never sees the template a run repeats and cannot defer on its own. Both now pass `deferDynamicVariables: true`; a Send is deliberately untouched. With the issue's premise trusted, the headline case - a load test with `{{$randomUUID}}` in the URL - would still have frozen.
2. **The load dialog's "Dynamic variables are generated once" warning is deleted, not reworded**, along with the prop, the scan behind it and the scan's tests - the warning was their only caller - and `containsDynamicVariable`, which the scan was the last reader of.
3. **A generator inside an auth credential is still generated once, at composition.** `apply_auth` encodes a credential when the request is built, so a token left for the bind would go out as base64 of its own text; deferring the build is what #1055 carries, and this is the same reason a credential does not carry `{{$vu}}`. So acceptance criterion 1 holds for every bindable field of the request and not for a credential - stated in `scenario_data.hpp`, `variable-resolution.md` and `api-reference.md` rather than left to be discovered.
4. **Plan resolution defers unconditionally, not only for load runs.** Both executors join a step's template before every send, so one rule covers both modes; a design-mode collection run over several iterations now gets a fresh value per pass too, which is the same fix one shape over.

Closes #995


athrvk:
The red `CI gate` on the timeline is against `1e761e1`, and it is not a test failure: its log reads `engine: cancelled  <-- failed`, with `quality: success`, `app: success` and `engine-format: success` beside it. The engine jobs were cancelled mid-run by my own push of the `master` merge (`7c0090f`) about twenty seconds earlier, and the gate counts `cancelled` as a failure. Nothing was re-run to make that go away - the merge push started a fresh run on the current head, which is the one that matters.

For the record, that merge was revalidated locally before it went up: engine `ctest --preset linux-dev` 2705/2705, app 544 files / 5960 tests, plus type-check, lint, `format:check` and `mkdocs build --strict`.

---

Issue #994, opened 2026-08-25
engine(loadtest): give load runs an iteration identity - {{$vu}}/{{$iteration}} templates and pm.info under load


athrvk:
Sub-issue of #992 (gap 2 - the parent carries the full context and the out-of-scope list). Verified at master `96385b2`.

## Problem

A load run has no addressable identity for "which VU am I, which iteration is this": no `{{$vu}}` or `{{$iteration}}` exists in the generator table (`engine/src/http/request_composer.cpp:142-176` is the complete 15-entry list), and `pm.info.iteration` is populated only on scenario paths while `pm.info.iterationCount` stays absent under load (`engine/src/runtime/script_engine.cpp:3441-3476` `setup_pm_info`). Without a dataset, there is no way to make VU 7's request differ from VU 12's - no unique usernames, no per-VU idempotency keys, no "user-{{$vu}}@example.com".

The VU already knows both numbers: `VirtualUser` holds `step` and `iteration` (`engine/include/vayu/core/scenario_load.hpp:124-152`), and every result is annotated per record (`engine/src/core/load_strategy.cpp:46-47` region). The identity exists; it is just not exposed.

## Throughput guard (owner constraint, 2026-08-26)

**This change must not reduce load-testing throughput.** Use #995's detection pattern: at compose time, record per step WHICH parts (url, header values, body) contain `{{$vu}}`/`{{$iteration}}` tokens - plans containing none (the overwhelming majority) take today's path with zero per-iteration work. Flagged parts pay one bounded substitution at the same bind point `apply_data_template` already runs. No per-iteration allocation beyond the substituted part itself. PR body carries before/after numbers on the `docs/engine/benchmarks.md` harness shape: tokens-off throughput unchanged, tokens-on cost reported. The `pm.info` half is free by construction - it runs in the deferred post-run script, never on the hot path.

## Fix / solution pathway

1. **Templates**: reserve `{{$vu}}` (1-based VU index) and `{{$iteration}}` (0-based per-VU iteration counter) as load-path substitutions. They cannot live in the composition-time generator table (plans compose once, `scenario_plan.hpp:73`) - bind them at the same per-iteration point `apply_data_template` runs (`scenario_load.cpp:415-423`), gated by the compose-time part flags above, with the same reserved-namespace/no-shadowing rules `{{data.*}}` has (`scenario_data.hpp:1-80`). In design mode and single sends they resolve to `1`/`0` (a run of one) rather than erroring - state this rule in the docs.
[...]

Issue #992, opened 2026-08-25
engine(loadtest): data-driven load for the common case - rows without a scenario, an iteration identity, per-iteration generators


athrvk:
Tracking issue for the 2026-08-25 scripting/load sweep's central finding, verified then at master `96385b2`: **the canonical load-test shape - one request, N virtual users, each carrying distinct data - was not expressible in Vayu.** Three verified gaps compounded; each became a sub-issue.

## Program status (refreshed 2026-08-28, verified at master `98d2ab0`): all three sub-issues landed. What keeps this open is the exit criteria below.

The three gaps as they now stand - the original claims no longer describe HEAD, so they are recorded here as *what changed*:

1. **Rows without a scenario - closed by #993.** `ResultAnnotations::data_row_index` is now "Carried by either shape since issue #993: a single-request run binds rows too" (`engine/include/vayu/core/load_strategy.hpp:127-131`).
2. **Iteration identity - closed by #994** (and extended into scripts by #1057/PR #1088). `{{$vu}}`/`{{$iteration}}` resolve ahead of every scope (`engine/src/http/request_composer.cpp:633-638`), `pm.info.iteration`/`vu` are set for a load run's deferred script (`engine/src/core/run_manager.cpp:130-131`).
3. **Per-iteration dynamic variables - closed by #995 (PR #1085).** Composition defers the `{{$guid}}` family (`request_composer.cpp:610-611`), the per-iteration join computes it (`engine/src/core/scenario_data.cpp:778-797`), and `docs/app/variable-resolution.md:655-680` states the new rule where the old "generated once, deliberate" paragraph stood. Benchmark evidence is checked in at `docs/engine/benchmarks.md:249-300`: feature-off paired against master at -0.01% with a -1.51% null control - the throughput guard held.

**Known, deliberate residue** (each with its own tracking): a generator or identity token inside an **auth credential** is still composed once - the deferral decision is keyed on "does this run carry rows", not on what the credentials hold. That is **#1055** (open, claimed 2026-08-28), stated in `request_composer.cpp:1342-1349` and `docs/app/variable-resolution.md:165-171`.

## Remaining work (the reason this stays open) - verified unmet at `98d2ab0`

[...]

PR #1088, opened 2026-08-28
feat(engine,app): resolve {{$vu}} / {{$iteration}} from the script resolver - #1057

### LINKED ISSUE #1057: engine(scripting): pm.variables.replaceIn does not resolve {{$vu}} / {{$iteration}}, so a script has two ways to ask and only one answers
Deferred from #994, which added the `{{$vu}}` / `{{$iteration}}` identity to the request bind and deliberately stopped short of the script resolver.

## Problem

`pm.variables.replaceIn("user-{{$vu}}")` returns `user-{{$vu}}`. The identity names are reserved ahead of the variable scopes (`lookup_variable`, `engine/src/http/request_composer.cpp`), which is what keeps a scope from freezing them at composition - and `replaceIn` goes through that same lookup (`js_pm_variables_replace_in`, `engine/src/runtime/script_engine.cpp`), so the token keeps its braces there too.

That is defensible and documented (`docs/engine/scripting.md`, the `replaceIn` section): the identity belongs to the send, and a script reads it as `pm.info.vu` / `pm.info.iteration`. It is also a seam. `replaceIn` already resolves the *other* reserved namespace - `{{data.column}}` binds there against the iteration's row (#890) - so a user who learned that rule reasonably expects the sibling namespace to behave the same way, and gets a literal instead. The `{{`-autocomplete inside a script's `replaceIn` string offers `{{$guid}}` and the ordinary variables, and #994 deliberately did **not** add the identity names there for exactly this reason: offering a token that cannot resolve is worse than not offering it.

## Fix / solution pathway

`ContextData` already carries both numbers - #994 added `vu` beside the existing `iteration`, and `setup_pm_info` reads them (`script_engine.cpp`). So the resolver has the values; what is missing is the arm that answers with them.

1. In `js_pm_variables_replace_in`, resolve `$vu` / `$iteration` from the context data before falling through to the scopes, mirroring how the data row is handled there today.
2. **Decide what a script with no identity gets, and state it.** An ordinary Send has no run: either the token keeps its braces (consistent with `pm.info.vu` being `undefined` there) or it resolves to the run-of-one `1` / `0` the request bind uses (consistent with what the request beside it was bound with). The second is the better answer and the harder one to justify - take it only with the reasoning written down, because the two APIs must not disagree about the same send.
3. Once (1) and (2) are settled, add the two names to the script editor's template-mode completion (`app/src/hooks/useScriptVariableCompletionProvider.ts`, which reads `DYNAMIC_VARIABLES` today) beside the app's `ITERATION_VARIABLES` table from #994 - the app half is a few lines once the engine half answers.

## App changes

Required, same PR as the engine change and not before it: the completion entry above. Offering a token whose resolution has not landed is the false affordance this issue exists to remove.

## Tests

Engine gtest: a pre-request script's `replaceIn("{{$vu}}/{{$iteration}}")` reads the same numbers the request beside it was bound with, on a scenario load run (two virtual users, so a wrong source shows), on a single-request load run's deferred script, and on a design-mode collection run; the no-run case asserts whatever (2) decided, in both directions. App vitest for the completion entry. Commands as usual; clean under the enabled tidy families.

## Docs (same commit)

`docs/engine/scripting.md` - the `replaceIn` section's "not readable from `replaceIn`, and that is deliberate" paragraph is **removed** rather than reworded, and the `pm.info` section stops being the only way to read the identity. `docs/app/variable-resolution.md` if the no-run rule changes what a preview should show.

## Sequencing

After #994. Shares `script_engine.cpp` with the #996 program's subs - serial with them, in any order.

## Acceptance criteria

- [ ] `pm.variables.replaceIn` resolves both identity names to the numbers the request beside it was bound with, on every run shape.
- [ ] The no-run answer is decided, tested in both directions, and written down.
- [ ] The script editor offers the two names in template mode, and only after the resolution lands.
- [ ] The deliberate-exclusion paragraph in `scripting.md` is gone.

athrvk:
## Description

`pm.variables.replaceIn("user-{{$vu}}")` returned `user-{{$vu}}` while the request beside it went out as `user-1`. Root cause: the identity names are reserved *ahead of the scopes* in `lookup_variable` (`engine/src/http/request_composer.cpp`), which is what makes them bindable at all - a scope that could answer would freeze one value into the composed request for the whole run - and the script resolver bottoms out at that same function. So a script had two ways to ask which iteration it was, `replaceIn` and `pm.info`, and only one answered.

The approach is a **caller-supplied identity** rather than a change to that rule. `resolve_template` and `resolve_template_with_data` take an optional `IterationIdentity`, defaulted absent, answered ahead of every scope and ahead of a bound row; exactly one caller passes one, `resolve_script_template`. Composition, which has no iteration, is unchanged, and the three conformance-fixture cases pinning `{{$vu}}` keeping its braces at compose time still pass untouched.

**The alternative I rejected** was injecting `$vu` / `$iteration` into the variable map the script resolver already builds. It reads as the smaller change and does not work: `lookup_variable` refuses the identity names *before* it consults the scopes, so a map entry is never reached - and made to work, it would be exactly the shadowing the reservation exists to prevent.

**The no-run answer is `1` and `0`, decided rather than defaulted** (the issue's item 2). `POST /execute` binds those into every send that carries no row (`execution.cpp`, issue #994: a single send is a run of one), and it binds them *before* the pre-request script runs - so they are what the request the script is handed already carries. `pm.info.vu` stays `undefined` there, and that is not the two APIs disagreeing: `pm.info` answers which iteration of which run this is, and there is no run; a token answers what it resolves to here. Both directions are tested and all three docs say so.

`IterationIdentity` moved from `core/scenario_data.hpp` down to `http/request_composer.hpp`, beside the two names it answers for, with `core::IterationIdentity` a using-declaration of it - the script sandbox resolves from a layer below core, the same move `render_data_value` already made for the data namespace's spelling rule. No bind site changes spelling, and the run-of-one numbers the script layer falls back to are not a second default: they are that struct's own member initialisers, the ones `POST /execute` binds with.

Two consequences worth calling out, neither of them scope creep:

- **`pm.sendRequest` answers the identity too**, because it resolves through the same function - URL, header values, raw body and auth credentials. Tested off the listener rather than in a string, since a resolution that ran after the URL was taken would still read right inside a script.
- **The script editor now offers the two names in `replaceIn`'s `{{`-completion**, from the same `ITERATION_VARIABLES` table the URL and body editors read. The comment that had withheld them named #1057 as the thing that would make the list right to extend; it is deleted rather than softened. Unlike a generator, the identity is never withheld for a same-named scope variable, since the engine resolves these two names ahead of every scope.

**Merged with `master` (`8d952ca`) in `70d5a67`.** #995 landed while this was open and added its own optional trailing parameter to `resolve_template`, in the slot this one added `identity` to. Both survive, with #995's `DynamicResolution` first so every positional caller of it is untouched; the two are opposite questions and they compose, and the script resolver passes `Generate` explicitly beside its identity - a script renders its own `{{$guid}}` here and now while resolving the identity it was sent as. Details in the comment below.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

All green, re-run in full on the merge commit `70d5a67` rather than inferred from either side:

- `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` - **2717 tests, 2717 passed, 0 failed** (2714 on `master`; the seven new engine tests land in four files).
- `cd app && pnpm test` - **545 test files, 5970 tests, all passing** (four new cases in one new file).
- `cd app && pnpm type-check`, `pnpm lint` (zero warnings), `pnpm format:check` - all clean.
- `mkdocs build --strict -f .github/mkdocs.yml` - clean, which is what validates the new cross-page anchors.
- `clang-format-19 --dry-run -Werror` on every touched engine file, and `scripts/pre-commit`'s clang-tidy over all staged C++ - exit 0.

New tests, and the run shape each pins - the issue's Tests section names three and all three are covered end to end:

| Test | Shape |
|---|---|
| `ScriptEngineTest.ReplaceInResolvesTheIdentityTheRequestBesideItWasBoundWith` | the numbers answer, a variable named `$vu` does not, `{{$vus}}` keeps its braces, and the answer agrees with `pm.info` |
| `ScriptEngineTest.ReplaceInAnswersTheRunOfOneWhereNothingSetsTheIdentity` | the no-run rule, both directions: `1` / `0` rendered **and** `pm.info.vu` still `undefined` |
| `ScriptEngineTest.ReplaceInReadsTheIdentityAheadOfABoundRowsColumns` | both reserved namespaces in one resolution; a column a file names `$vu` loses to the reserved name |
| `ScenarioIdentityTest.AStepScriptResolvesTheIdentityItRanAs` | a real two-VU, four-iteration scenario load run |
| `LoadIdentityTest.ADeferredScriptResolvesTheIdentityItWasSentAs` | a single-request load run's deferred script |
| `ScenarioRunnerTest.AStepScriptRendersTheIdentityItsOwnRequestWasBoundWith` | a collection design run, read off the wire: the script renders the identity into a header while the URL beside it carries what the bind wrote, on every pass |
| `SendRequestTest.TheIdentityReachesTheWireAsTheRequestBesideItCarriesIt` | `pm.sendRequest`'s URL and body, read off the listener |

Every one of them fails if the change is reverted - the tokens come back with their braces and each assertion compares against a number.

Two read-only reviewers went over the full diff, one against the acceptance criteria and one against `CLAUDE.md`. Three findings survived my own verification and are fixed in `2094a2c`:

1. The replacement paragraph in `scripting.md` said the numbers are `1` and `0` "outside a load run" and named a collection design run as one of those cases. Wrong: a design run's `{{$iteration}}` advances with the pass, which `scenario_runner_test.cpp` has asserted on the wire since #994. The claim now points at the per-shape binding table, and the two copies of the same sentence (the completion comment and its documentation string) say "a plain Send" instead.
2. The issue names three run shapes to test and the first pass covered one. The other two now have theirs - the last two rows above.
3. `docs/app/variable-resolution.md`'s autocomplete section lists, name by name, which script list offers what, on the rule that a list offers exactly what the call resolves. `replaceIn` gaining two names belongs in it; it now has a bullet, and the reserved-names section says a script can resolve them.

One reviewer finding I discarded, with the reason: that the run-of-one `{1, 0}` in the script layer and the `IterationIdentity{}` `POST /execute` binds with are "two independently chosen defaults landing on the same numbers". They are not - since the move they are the same struct's own member initialisers, one definition, which is part of what the move was for. A second finding, that `TemplateJoiner::value_of` in `scenario_data.cpp` is a duplicate of `identity_value`, I also left: the reserved-name test is already the one shared primitive both call, and the two differ in what they return because one feeds a generic token walk and the other the load path's typed join. Consolidating those two lines would put a `std::string` allocation on the hot path #992 guards.

## Checklist

## Related Issues

Closes #1057


athrvk:
resolve merge conflicts

athrvk:
Done - `70d5a67` merges `master` (`8d952ca`) in and resolves it.

The conflict was a real one rather than a textual near-miss: #995 landed on `master` while this branch was open and added its own optional trailing parameter to `resolve_template`, in the same slot this one added `identity` to. Both survive, with #995's `DynamicResolution` first so every one of its positional callers is untouched:

```cpp
std::string resolve_template (const std::string& input,
const VariableValues& vars,
const BoundColumnNames& bound_columns            = {},
DynamicResolution dynamic                        = DynamicResolution::Generate,
const std::optional<IterationIdentity>& identity = std::nullopt);
```

They are opposite questions and they compose, which is what the merged doc block now says: `dynamic` asks whether the generator family is this composition's to resolve, `identity` supplies the one answer for the reserved identity names. The script resolver passes `Generate` explicitly beside its identity rather than letting it default - a script renders its own `{{$guid}}` here and now, while resolving the identity it was sent as, and #995's own doc names `replaceIn` as a `Generate` caller. `resolve_template_with_data` merged clean; nothing else in the two diffs overlapped.

Re-verified on the merge commit, not inferred from the two sides:

- engine `ctest --preset linux-dev` - **2717 passed, 0 failed** (was 2695 here, 2714 on `master`)
- `cd app && pnpm test` - **545 files, 5970 tests, all passing**
- `pnpm type-check`, `pnpm lint`, `pnpm format:check`, `mkdocs build --strict` - all clean
- `clang-format-19 --dry-run -Werror` on every engine file the merge touched - clean

---

Issue #1011, opened 2026-08-25
docs(scripting): ten verified discrepancies in pm-api-compatibility.md - stale pm.info claims, an undocumented shipped assertion, the pre-#959 cycle text


athrvk:
Sub-issue of #996. Docs only. The sweep verified every claim in `docs/app/pm-api-compatibility.md` against the code at `e5277b6`, both directions; 16 of 20 table rows and all 9 "Not (yet) supported" entries verified clean. These are the discrepancies:

## Corrections

1. **`pm.info` iteration claims contradicted by the code AND the same page**: `pm-api-compatibility.md:151-153` says iteration/iterationCount are "set by the collection runner and by nothing else" - but send-with-row sets 0 of 1 (`engine/src/http/routes/execution.cpp:1250-1258`, streaming `:1109-1112`), a load run's deferred script gets `iteration` (`engine/src/core/run_manager.cpp:129`), and the page itself says so at `:524`. The table at `:141-142` carries the same stale cells. Note the code comments cite #300 for the CURRENT behavior - the doc cites #300 for the opposite reading; reconcile to the code.
2. **`pm.response.to.have.body` exists but appears in NO doc** (`script_engine.cpp:2621-2622`, `:2345`): missing from `pm-api-compatibility.md:30` and `scripting.md:396-399`. Add it - coordinated with #998, which changes its semantics; write the doc against #998's post-fix contract if that lands first, else document today's substring form with #998 linked.
3. **Response-row omissions** at `pm-api-compatibility.md:25`: `responseTimeWire`, `responseTimeQueueWait`, `errorCode`, `errorMessage` are bound (`script_engine.cpp:3263-3273`) and even USED by the page itself at `:797` without being declared. Same for `dataTruncated` on streamed-event entries (`:28` vs `script_engine.cpp:3224-3225`).
4. **Stale cycle-detection claim in BOTH docs**: "a cyclic value raises a RangeError after 64 levels" (`pm-api-compatibility.md:345-346`, `scripting.md:178-179`) predates #959's fix - cycles now report at the depth they close (`script_engine.cpp:557-580`); the 64-level cap is a backstop for merely-deep structures with a different message. Cite #959.
5. **Data-file formats understated**: `:517` says "CSV or JSON"; the product accepts CSV/TSV/JSON/JSONL (`scripting.md:1020`, `data-driven-runs.md:3`).
[...]

Issue #601, opened 2026-08-14
engine/app(data): phase 3 - send with row: POST /execute binds one data row, the builder picks it, and pre-request scripts finally see pm.iterationData without a run


athrvk:
Phase 3 of #598; requires phase 1 (#599). Independent of phase 2 (#600). This is the phase that closes the epic's motivating gap: a pre-request script using `pm.iterationData` is untestable today without launching a full collection run.

## Engine: `data` on POST /execute

- New optional field `"data": { &lt;one row object&gt; }` on the execute payload. Parser: a `read_data_row` free function returning `{ok, error, value}`, exactly the `read_transient_flag` / `read_stream_flag` shape (`execution.cpp:239-252, :260-343`) - unit-drivable directly, since the suite has no in-process route harness. Validation: must be a JSON object of name/value pairs (the `parse_scenario_request` per-row check at `scenario_plan.cpp:192-200` - extract or restate it; it is file-static today), bounded by `maxScenarioDataBytes` for the single row.
- Bind: **the primitive already exists** - `bind_data_row(request, row, 0)` (`scenario_data.cpp:208-210`), whose header says verbatim it exists "for a caller that holds no template". Call it after `build_request` lands the composed request (`execution.cpp:829`), before the stream branch (`:860`) and `execute_exchange` (`:946`).
- Bind failure: **400 before any run row exists**, body carrying the binder's own message (`{{data.x}} names a column data row 0 does not have (columns: ...)`) - matching the scenario philosophy (nothing sent, named error) and the constraint that validation precedes run-row creation. The partially-bound request is discarded, never sent (`scenario_data.hpp:122-125`).
- A payload with `data.*` tokens and **no** `data` field keeps today's behaviour (tokens go out literal - #595 item 4 owns the warning discussion; do not couple it here). `transient` + `data` and `stream` + `data` both compose fine - bind first, then the normal path; say so in tests.
[...]

PR #1104, opened 2026-08-28
fix(engine): the pm.info completion states #994's iteration rule - #1096

### LINKED ISSUE #1096: engine(scripting): the pm.info completion string still states the pre-#994 iteration rule, and the editor shows it
Found while fixing #1011 (PR #1094), by sweeping for surviving copies of each claim that issue corrected. It is an engine source change, not a docs one, so it was reported rather than folded into a docs-only diff.

**Anchor refresh + scope addition, 2026-08-28, verified at master `98d2ab0`:** the completion string moved to `scripting.cpp:856-863` (the first refresh comment's `:860-861` still holds inside that block), the fixture copy to `script-typedefs.d.ts:766`, and the 2026-08-28 closure verification found two more surviving copies of the same claim - one engine comment and one docs table cell #1094 missed. Both folded in below.

## Problem

`engine/src/http/routes/scripting.cpp:856-863` is the `documentation` string for the `pm.info` completion entry (the entry block is `:854-864`; the offending sentences begin at `:860`), and it ends:

&gt; iteration / iterationCount are set by the collection runner and by nothing else. A load test's Tests script runs once per sampled response rather than once per iteration, so it has no iteration number to report and reads undefined for both.

Both sentences are now wrong, in the direction that matters - they describe a *narrower* surface than the one that ships, so a script author is told a field is unreadable where it reads:

1. **`iteration` is set for a load run's deferred script** (#994). `engine/src/core/run_manager.cpp:130-131` sets `script_ctx.iteration` and `script_ctx.vu` from the sample; the comment there argues the case explicitly. So "reads undefined for both" is false about the exact mode the sentence names.
2. **`iterationCount` is not the collection runner's alone.** A send that binds a row sets it to `1` - `engine/src/http/routes/execution.cpp:1490` on the buffered path, `:1323` and `:1409` on the streaming one - because that send is row 0 of 1.

`run_manager.cpp:126-129` is the one part still true: `iterationCount` does stay absent under load.

**Two more copies of the runner-alone claim, same fix, same commit** (found 2026-08-28):

3. The per-field entry beside it: `scripting.cpp:915-916` - "undefined outside a collection run, and set by the runner alone".
4. An engine source comment: `engine/src/runtime/script_engine.cpp:5453-5455` - "present only for a scenario run's steps - the runner is the one caller that sets them". False against `execution.cpp:1481-1490` and `run_manager.cpp:123-129`.

**This is more user-facing than the doc pages that carried the same claim.** The completion table is served at `GET /scripting/completions`, the `.d.ts` at `GET /scripting/types` is generated from it, and `docs/app/pm-api-compatibility.md` names that table as the single source of truth for this surface - "do not add a fourth place that lists `pm.*` names". So this string is what Monaco shows on hover over `pm.info.iterationCount`, and what every MCP agent reads through `vayu://scripting/completions`. The prose in both doc pages was *mostly* corrected in #1011 - see the Docs section for the one cell it missed.

The checked-in fixture carries it too, at `engine/tests/fixtures/script-typedefs.d.ts:766`, which is the generated artifact rather than a second source - it moves when the table does.

## Fix / solution pathway

Rewrite the two sentences in `scripting.cpp`'s `pm.info` documentation string to the rule the code holds, in the shape the corrected docs use: `iteration` and `vu` are reported by every run, load runs included; `iterationCount` comes from the collection runner and from a send that bound a row (`1`, row 0 of 1), and is absent under load because a duration-bounded run has no total to report. Keep it to the two sentences - the surrounding "every field is optional, test with `typeof`" half is correct and is the part a completion popup most needs.

Apply the same correction to the per-field entry at `:915-916` (the sibling entries `pm.info.iteration` at `:891` and `pm.info.vu` at `:901` should be swept for the same wording while there) and to the comment at `script_engine.cpp:5453-5455`.

Then regenerate the fixture, with the repo's own command rather than by hand:

```
VAYU_UPDATE_SCRIPT_TYPEDEFS=1 ctest --preset linux-dev -R ScriptTypes
```

## App changes

None. The app reads both endpoints and displays what they say; nothing there restates the rule.

## Tests

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`. The existing `ScriptTypes` fixture test is the mutation check by construction: the fixture is pinned to the generator, so a changed string that was not regenerated fails, and a regenerated fixture that does not match the source fails the other way.

## Docs (same commit)

One cell, previously believed done: **`docs/app/pm-api-compatibility.md:189`**, the `iteration` row's "Reads `undefined` when" column, still says a single Send never sets it. #1094 corrected the `iterationCount` cell beside it (`:191` now says "An **ordinary** Send") but left this one, and the page now contradicts its own line `:771` ("`pm.info.iteration` is then `0` of `1` - the send is row 0 of 1"). One word plus the send-with-row case in the "Set to" column closes it; use `scripting.md:1006`'s careful spelling ("An ordinary Send still reads `undefined`"). Nothing else - the rest of both pages was corrected in #1011 and this change makes the completion table agree with them. Verify rather than assume: if `docs/engine/api-reference.md`'s completions section quotes the string, it moves too.

## Sequencing

`engine/src/http/routes/scripting.cpp` - not `script_engine.cpp` (item 4's comment edit there is a comment, conflict-free at rebase level), so it is disjoint from #1000 and #1003, which serialise on `script_engine.cpp`. **#1098 edits the same completion table and regenerates the same fixture: take both in one PR, or land this first and #1098 strictly after** - two in-flight PRs on this file will conflict. Otherwise land whenever; it gates nothing and nothing gates it.

## Acceptance criteria

- [ ] The `pm.info` completion documentation states #994's rule: `iteration` and `vu` are reported by every run; `iterationCount` comes from the collection runner and from a send that bound a row, and is absent under load.
- [ ] No copy of "set by the collection runner and by nothing else" / "set by the runner alone" / "the runner is the one caller that sets them" survives in `engine/`, the regenerated fixture included.
- [ ] `docs/app/pm-api-compatibility.md:189`'s `iteration` cell no longer claims a single Send never sets it, and the page no longer contradicts its own `:771`.
- [ ] The completion string does not contradict `docs/app/pm-api-compatibility.md` or `docs/engine/scripting.md` at the merge commit.

---

athrvk:
## Description

The `pm.info` completion `documentation` string ended with two sentences that describe a **narrower** surface than the one that ships, so a script author is told a field is unreadable where it reads:

> iteration / iterationCount are set by the collection runner and by nothing else. A load test's Tests script runs once per sampled response rather than once per iteration, so it has no iteration number to report and reads undefined for both.

**Root cause, and why the fix is a sweep rather than one string.** #994 widened who reports the identity, and #601 had already widened who reports the count, but the completion table was never swept for the claim - so the same sentence survived in five engine places while both doc pages were corrected by #1011. `run_manager.cpp:130-131` sets `iteration` and `vu` from the sample the response was actually sent in; `execution.cpp:1323` / `:1411` / `:1492` set `iterationCount` to `1` for a send that bound a row, that send being row 0 of 1. The only part still true is the narrow half `run_manager.cpp:126-129` argues: `iterationCount` stays absent under load, because a duration-bounded run has no total to report.

**The alternative I rejected:** correcting only the `pm.info` umbrella string the issue's title names. The claim exists in five engine copies with four different spellings, and acceptance criterion 2 asks that none survive - a fix that leaves four is the same defect one grep away, which is how this claim reached its third issue. The sweep is comment-and-string only; no behaviour changes.

This string is the most user-facing carrier of the claim: served at `GET /scripting/completions`, generated into the `.d.ts` at `GET /scripting/types`, shown by Monaco on hover, and read by every MCP agent through `vayu://scripting/completions`.

### What changed

`engine/src/http/routes/scripting.cpp` - four entries, rewritten in the spelling `docs/engine/scripting.md:991-1021` already uses:

- `pm.info` - the two sentences.
- `pm.info.iterationCount` - "set by the runner alone".
- `pm.info.iteration` and `pm.info.vu` - both said `undefined on a single Send`, where a send that bound a row reports `0` and `1`. Swept per the issue.

Three further engine copies of the same claim, per acceptance criterion 2:

- `engine/src/runtime/script_engine.cpp:5452-5455` - `setup_pm_info`'s comment.
- `engine/include/vayu/runtime/script_engine.hpp:139-141` - `ScriptContext`'s own field doc, "The collection runner sets it and nothing else does". Not named in the issue; found by sweeping for the *claim* rather than the wording, and it is the doc every other copy defers to ("see ScriptContext").
- `engine/tests/scenario_runner_test.cpp:13` and `engine/tests/script_engine_test.cpp:5107` - two test header comments stating the pair as the scenario runner's alone. The tests themselves are correct and unchanged: what they pin is that a context declaring neither reports neither, which is still true.

`engine/tests/fixtures/script-typedefs.d.ts` - regenerated with `VAYU_UPDATE_SCRIPT_TYPEDEFS=1`, never edited by hand. Four entries move and nothing else does.

### Docs (same commit)

`docs/app/pm-api-compatibility.md` - the `iteration` row said "A single Send" while the `iterationCount` row directly below it already said "An ordinary Send", and the page contradicted its own `:771` ("`pm.info.iteration` is then `0` of `1` - the send is row 0 of 1"). Corrected, and with it the `vu` row and the paragraph at `:208-214` restating both, which carried the identical defect one line down - leaving them would have made the page contradict the completion string this PR corrects (acceptance criterion 4).

**Deliberate deviation from the issue spec, stated:** the issue scopes the docs to the one `iteration` cell and "nothing else". The `vu` cell and the paragraph are the same sentence about the same pair, so this is three one-word-class changes rather than one. Nothing else on either page moved; `docs/engine/scripting.md` needed nothing, its prose already being the target spelling.

Verified rather than assumed, both of the issue's "check this" items:

- `docs/engine/api-reference.md` does not quote the completion strings - `:4848-4851` defers to `scripting.md#script-identity-pminfo` for which run shapes report which. No change.
- The app restates nothing: `script-variants.tsx:144-153` / `:192-201` is the only hand-authored `pm.info` copy in `app/src`, and it covers `eventName` / `requestId` / `requestName` only. **App changes: none**, as the issue states.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

All run locally on this branch, all green:

- `python build.py -e -t` then `ctest --preset linux-dev --output-on-failure` - **2744 tests, 2744 passed, 0 failed** (41.4s).
- `cd app && pnpm test` - **5970 tests across 545 files, all passed**. Included because `app/src/hooks/script-typedefs.docs-compile.test.ts` compiles the doc pages' JavaScript blocks against the regenerated fixture.
- `cd app && pnpm type-check` - clean on all three projects (renderer, electron, electron-tests).
- `clang-format-19 --dry-run -Werror` - clean on every touched C++ file. (Note for anyone reproducing: the environment's *unpinned* `clang-format` is 18 and reports a false violation in `script_engine.cpp` at a line this PR does not touch. 19 is the pin, and 19 is clean.)

**The mutation check is the fixture pin, and it was observed rather than reasoned about.** `ScriptTypesTest.TheCheckedInDeclarationsMatchTheGenerator` compares the generator's output to the checked-in `.d.ts` byte for byte. It **failed** on the first commit (strings changed, fixture not yet regenerated) and **passes** on the second - so a future edit to any of these four strings that skips the regeneration reddens, and a regenerated fixture that does not match the source reddens the other way. That is why this PR adds no new test: the pin already covers the changed surface, and a second assertion on the same string would be the fourth place listing `pm.*` names that `pm-api-compatibility.md:1131` warns against.

No pre-existing failures to report - the suite was green on the base commit apart from this same fixture test, which failed only because the strings had already changed in the working tree.

## Checklist
- [x] I have added tests (if applicable) - see the mutation-check note above

## Related Issues
Closes #1096

### DIFF
--- a/docs/app/pm-api-compatibility.md
+++ b/docs/app/pm-api-compatibility.md
@@ -186,8 +186,8 @@ truthful value for it, so a script tests with `typeof` rather than assuming:
 | `requestId` | The saved request the send is filed under | An ad-hoc request (MCP's `run_request` with no `requestId`, a load run started from a URL) |
 | `requestName` | The request's name **as the client sent it** - the name in the editor, which for an unsaved edit differs from the stored row | A request with no name, and an ad-hoc one |
 | `eventName` | `"prerequest"` in the Pre-request tab, `"test"` in the Tests tab | Never, for a script Vayu runs - both hooks set it |
-| `iteration` | The 0-based pass this response was sent in - a collection run's pass, or the iteration a load run's sampled response carried | A single Send, which is one request rather than a pass of anything |
-| `vu` | The 1-based virtual user that sent it. Spans the concurrency in a collection load run; `1` everywhere else, because one request repeated is one user's iterations | A single Send |
+| `iteration` | The 0-based pass this response was sent in - a collection run's pass, the iteration a load run's sampled response carried, or `0` for a send that bound a row, which is row 0 of 1 | An ordinary Send, which is one request rather than a pass of anything |
+| `vu` | The 1-based virtual user that sent it. Spans the concurrency in a collection load run; `1` everywhere else, because one request repeated is one user's iterations, and `1` for a send that bound a row, beside its iteration `0` | An ordinary Send |
 | `iterationCount` | How many passes that run will make - a collection run's total, and `1` for a send that bound a row, which is row 0 of 1 | An ordinary Send, and a load run |
 
 ```javascript
@@ -210,8 +210,9 @@ That is not a reversal of issue #300's ruling but the case it excluded: what
 #300 refused was reporting a *reservoir position* as an iteration number, and
 neither of these is one - a load run's submission claims its iteration and its
 virtual user before it is sent, and both travel with the response into the
-sample, exactly as the bound data row does. A single Send reads `undefined` for
-both.
+sample, exactly as the bound data row does. An **ordinary** Send reads
+`undefined` for both; a send that bound a row reports iteration `0` of `1`, the
+same exception `iterationCount` makes above.
 
 ### `pm.sendRequest` is synchronous, callback-only, and not available to agents
 
--- a/engine/include/vayu/runtime/script_engine.hpp
+++ b/engine/include/vayu/runtime/script_engine.hpp
@@ -137,7 +137,8 @@ struct ScriptContext {
      * `iterationCount` stays narrower than either: a duration-bounded run has
      * no total to report, and a field readable from one mode and not another is
      * worse than one that is never readable at all. The collection runner sets
-     * it and nothing else does.
+     * it, and so does the `data` row send above - `1`, the total that send's
+     * row 0 of 1 belongs to.
      *
      * `iteration` is 0-based (Postman's convention); `vu` is 1-based, because
      * it is a name for a user rather than an index into anything;
--- a/engine/src/http/routes/scripting.cpp
+++ b/engine/src/http/routes/scripting.cpp
@@ -861,8 +861,10 @@ nlohmann::json get_script_completions () {
     "collection run's pass, the iteration a load run's sampled response was "
     "sent in, and a send that bound a data row, which is row 0 of 1. "
     "iterationCount is narrower - the collection runner's total, or that "
-    "send's 1 - and stays undefined under load, where a duration-bounded run "
-    "has no total to report. An ordinary Send reports none of the three." },
+    "send's 1 - and no load run reports it: a duration-bounded one has no "
+    "total, and a field readable from one load shape and not another is worse "
+    "than one that is never readable there. An ordinary Send reports none of "
+    "the three." },
     { "sortText", "0_pm_info" } });
 
     completions.push_back ({ { "label", "pm.info.requestId" }, { "kind", KIND_FIELD },
@@ -918,8 +920,10 @@ nlohmann::json get_script_completions () {
     { "documentation",
     "How many passes the run will make in total: the collection runner's "
     "total, or 1 for a send that bound a data row, which is row 0 of 1. "
-    "undefined on an ordinary Send and under load, where a duration-bounded "
-    "run has no total to report.\n\nExample:\nconsole.log("
+    "undefined on an ordinary Send, and on every load run - a duration-bounded "
+    "one has no total to report, and a field readable from one load shape and "
+    "not another is worse than one that is never readable there."
+    "\n\nExample:\nconsole.log("
     "'pass ' + (pm.info.iteration + 1) + ' of ' + pm.info.iterationCount);" },
     { "sortText", "1_pm_info_iterationCount" } });
 
--- a/engine/src/http/routes/scripting.cpp
+++ b/engine/src/http/routes/scripting.cpp
@@ -857,10 +857,12 @@ nlohmann::json get_script_completions () {
     "What this script is attached to and which hook it is running in. Every "
     "field is optional - an ad-hoc request has no id, and an unsaved one has a "
     "name no stored request carries - so test with typeof rather than assuming "
-    "a value.\n\niteration / iterationCount are set by the collection runner "
-    "and by nothing else. A load test's Tests script runs once per sampled "
-    "response rather than once per iteration, so it has no iteration number to "
-    "report and reads undefined for both." },
+    "a value.\n\niteration and vu are reported wherever a real one exists: a "
+    "collection run's pass, the iteration a load run's sampled response was "
+    "sent in, and a send that bound a data row, which is row 0 of 1. "
+    "iterationCount is narrower - the collection runner's total, or that "
+    "send's 1 - and stays undefined under load, where a duration-bounded run "
+    "has no total to report. An ordinary Send reports none of the three." },
     { "sortText", "0_pm_info" } });
 
     completions.push_back ({ { "label", "pm.info.requestId" }, { "kind", KIND_FIELD },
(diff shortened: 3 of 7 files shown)


athrvk:
Review-pass note, for the record.

Two read-only reviewers ran over the diff - one against the acceptance criteria line by line, one against the repo's conventions. The conventions pass came back clean on every category (no em-dashes, no AI co-authorship, docs in the same commit as the code, fixture generated rather than hand-edited, completion strings in the file's plain-prose house style), with one observation: the `vu` row and the paragraph in `pm-api-compatibility.md` go beyond the issue's "Nothing else". That is the deviation already argued in the PR body, so it stands.

**One finding survived verification and is fixed in `d794cb0`.** Condensing the docs' rule into the completion string had dropped the clause that carries the general case, leaving a reason that does not cover it:

> ...stays undefined under load, where a duration-bounded run has no total to report.

That reads as though the field would be reportable for a load run that *does* know its total - and there is one. `LoadStrategy::create` returns `IterationsLoadStrategy` for an `iterations` mode (`load_strategy.cpp:958-971`), a scenario load run has the same shape (`scenario_load.cpp:461-478`), and both compute a real `requests_expected`. `iterationCount` is undefined there too, for a different reason: `ResponseSample` (`metrics_collector.hpp:64-89`) carries `iteration`, `vu` and `data_row_index` and no total at all, so the deferred script has none to read whatever the mode.

`docs/engine/scripting.md:1018-1021` and `run_manager.cpp:126-129` both give **two** clauses - the duration-bounded case, and that a field readable from one load shape and not another is worse than one never readable there - and the second is what makes the rule general. Restored in the three places this PR newly carried the shortened form; `ScriptContext`'s field doc already stated both and is untouched. Verified against the strategy factory and the sample struct rather than taken on the reviewer's report.

Worth naming because it is the exact defect this issue exists to close - a doc string stating a narrower rule than the code holds - reintroduced while fixing it, one clause at a time.

Nothing else was discarded: the acceptance-criteria pass found all four criteria met, with the three unrelated "and by nothing else" hits in `engine/` (`run_manager.hpp:484` live ticks, `load_strategy.hpp:134` step index, `script_engine.hpp:208` the `in_scenario` flag) confirmed as true statements about other fields rather than surviving copies of the claim.

Re-verified on `d794cb0`: 2744 engine tests pass, `ScriptTypes` fixture regenerated with the repo's own command, `clang-format-19` clean.

The earlier red checks on this PR were all on `80c002c`, the first commit - `ScriptTypesTest.TheCheckedInDeclarationsMatchTheGenerator`, failing because the strings had changed and the fixture had not yet been regenerated. That is the mutation check working, and `4d6415f` regenerated it.

---

Issue #1003, opened 2026-08-25
engine(scripting): pm.request.body is a bare string - make it Postman's RequestBody shape, #991-style


athrvk:
Sub-issue of #996. Verified at `e5277b6`. Sibling of #991 (same shape problem, same file, same decision posture: Postman compatibility preferred).

## Problem

`pm.request.body` is a plain string, absent when there is no body (`setup_pm_request`, `engine/src/runtime/script_engine.cpp:3415-3422`). Postman's `pm.request.body` is a RequestBody object with `.mode`, `.raw`, `.urlencoded`, `.formdata`, `.toString()`. Imported-script failure modes: `pm.request.body.mode === 'raw'` guards are silently skipped (`.mode` is `undefined` - silent wrong); `JSON.parse(pm.request.body.raw)` throws (loud). `.toString()` accidentally works via String.prototype.

Also anchor the existing divergences this issue must preserve: body assignment is writable in pre-request via write-back (`:3596-3699`, further than Postman's read-mostly contract - keep it), and assigning on a form-data request is refused by name (`:3667-3677` - keep that refusal).

## Fix / solution pathway

Mirror #991's approach exactly (it is the precedent for "string becomes Postman object without breaking string-ish uses"):

- `pm.request.body` becomes an object: `.mode` (`"raw"`, `"urlencoded"`, `"formdata"` - derived from the composed request), `.raw` (the string, for raw mode), `.urlencoded`/`.formdata` as read-only arrays of `{key, value}` where the engine's composed request has them (verify what the composed Request carries for those modes - `rg -n "form" engine/include/vayu/types.hpp` - and expose what exists; state what cannot be exposed).
- String-context mitigation per #991: own `toString()`/`Symbol.toPrimitive`/`toJSON` returning the raw text; prototype `String.prototype` so generic string methods keep working; known breaks (`===`, `typeof`) documented in the same migration note style.
- Writes: assignment of a string keeps today's semantics (raw-mode write-back); assignment on form-data keeps its named refusal; `.raw = "..."` also accepted for symmetry.
- No body → `body` is `undefined` (matches both products today - verify Postman treats absent body as undefined, MED; if Postman gives an empty RequestBody, state the divergence in the doc rather than chase it).

## App changes

Typedef surface regenerated (`engine/src/http/routes/scripting.cpp` typedefs, `engine/tests/fixtures/script-typedefs.d.ts`); autocomplete additions if `pm.request` members are enumerated (verify, state - same check as #991).

[...]

PR #1117, opened 2026-08-28
feat(engine): pm.response.status is Postman's reason phrase - #1000

### LINKED ISSUE #1000: engine(scripting): pm.response.status and sendRequest res.status are numbers - adopt Postman's reason-phrase contract, code stays numeric
Sub-issue of #996. Verified at `e5277b6`. **Owner standing direction (2026-08-26): prefer Postman compatibility - this issue records that decision for the status/reason family, a breaking change with a migration note, same posture as #991.**

## Problem

Postman splits status into `code` (number) and `status` (reason phrase, `"OK"`); Vayu spells both as the number:

- `pm.response.status` is numeric (`engine/src/runtime/script_engine.cpp:3254-3256`, pinned by `script_engine_test.cpp:902-904`; documented as deliberate at `docs/app/pm-api-compatibility.md` line ~197).
- `pm.sendRequest`'s callback response sets `status` to `status_code` too (`:4356-4357`, re-verified - the comment beside it already acknowledges "Postman spells the reason phrase there instead").

Imported-script failure modes, both silent: `if (res.status === 'OK')` is always false (guard silently skipped); `pm.expect(pm.response.status).to.eql("OK")` fails where Postman passes; `res.status !== 200` guards invert.

## Throughput guard (owner constraint, 2026-08-26)

**This change must not add per-response work to the load path.** The reason phrase is needed only when a script reads it - and scripts run deferred, post-run, on sampled responses. Therefore: derive the phrase **lazily at script-bind time** from a static canonical table keyed by `status_code` (a `constexpr` lookup, zero allocation, zero hot-path presence). Do NOT thread a per-response `std::string reason` through the transfer/completion path just for this - if the engine does not already capture the wire's reason line, use the canonical table exclusively and document that rule (HTTP/2 has no reason phrase anyway, so canonical-by-code is what Postman effectively shows there too). If the engine DOES already carry the wire phrase somewhere reachable at script-bind time, using it is free - verify and state which case applies.

## Fix / solution pathway

1. `pm.response.status` and the sendRequest response `status` become the **reason phrase string**; `code` stays the number (it already exists on both). Phrase source per the throughput guard above: existing capture if reachable, else the canonical IANA table.
2. `pm.response.to.have.status(...)`: number compares `code`, string compares the phrase - coordinate with #998, which touches the same function; whichever lands second rebases.
3. **Migration note** (docs + PR body for release-notes curation): Vayu scripts comparing `pm.response.status` to a number switch to `pm.response.code`. Loud in the common case (`status === 200` becomes always-false → test fails visibly), but note the silent case honestly: `status >= 400`-style arithmetic guards degrade silently - the note must name that pattern.
4. Regenerate the typedef surface (`engine/src/http/routes/scripting.cpp` typedefs, `engine/tests/fixtures/script-typedefs.d.ts`) - `status` changes type.

## App changes

Verify whether any renderer surface reads `status` from script-visible payloads expecting a number (`rg -n "\.status\b" app/src/hooks/script-typedefs*` and the Console/test-result renderers) - state the finding; expected none beyond typedef text.

## Tests

Engine gtest: `pm.response.status === "OK"` for a 200; an HTTP/2 (or phrase-less) response gets the canonical phrase; `code` unchanged; sendRequest response mirror; `to.have.status("OK")` and `(200)` both pass, `("200")` fails (Postman verdict). Update the pinning test at `script_engine_test.cpp:902-904` to the new contract. `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`. Clean under #928 families.

## Docs (same commit)

`docs/engine/scripting.md` and `docs/app/pm-api-compatibility.md`: the row that documents the numeric spelling as deliberate flips to the Postman contract + migration note.

## Sequencing

`script_engine.cpp` - serial with the other #996 subs; coordinate with #998 on `js_response_have_status`.

## Acceptance criteria

- [ ] `status` is the reason phrase on both objects; `code` the number; canonical fallback for phrase-less wires.
- [ ] **No per-response allocation or field added to the transfer/completion path - the phrase is derived at script-bind time** (stated and anchored in the PR).
- [ ] `to.have.status` accepts number and reason string with Postman verdicts.
- [ ] Migration note names both the loud and the silent legacy patterns; typedefs regenerated.

---

athrvk:
## Description

Postman splits the status line in two - `code` is the number, `status` is the reason phrase - and vayu spelled both as the number. A script lifted from Postman read the split wrong in silence: `if (res.status === 'OK')` was always false so its guard was skipped, and `pm.expect(pm.response.status).to.eql('OK')` failed where Postman passes.

**Root cause and approach.** The engine already had every part of the answer and only the binding was wrong. `response_reason_phrase` (`script_engine.cpp:3217`) has answered `reason()` and `to.have.status("OK")` since #1049 - wire phrase where the status line carried one, the registered `vayu::http::status_text` for the code where it did not - so `status` becomes a third reading of that one helper rather than a new derivation. Both spellings flip together: `pm.response.status` and the `pm.sendRequest` callback's `res.status`, with `code` untouched on both.

Flipping both is what keeps `build_send_response`'s standing rule true. That comment argued for the numeric spelling on the grounds that *"two objects called a response inside one sandbox disagreeing about what `status` means is the worse divergence"* - which is an argument about agreement, not about which of the two spellings they agree on. **The alternative I rejected** was flipping `pm.response.status` alone and leaving `sendRequest`'s `res.status` numeric: that is the smaller diff and the smaller break, but it manufactures exactly the intra-sandbox disagreement the existing design went out of its way to avoid, and it is not what #1000 specifies.

**The throughput guard the issue sets is met by construction.** Nothing is added to the transfer or completion path: the phrase is derived at script-bind time from `Response::status_text`, a field the response already carries, which `client.cpp` fills from the wire (or from the canonical table) as the transfer completes. The issue's open question - "does the engine already carry the wire phrase?" - is answered *yes*, so this is "existing capture, canonical fallback", not "canonical exclusively"; no new field, no per-response allocation, no hot-path presence.

`pm.response.to.have.status(...)` needed no change: it has taken a number or a phrase since #998/#1049 and decides by the argument's type. Acceptance criterion 3 was already satisfied by landed code.

**App changes:** none needed, and verified rather than assumed. Neither the renderer nor MCP holds a copy of the scripting surface - `useScriptTypeDefinitions.ts` and `useScriptCompletionProvider.ts` fetch the `.d.ts` and the completions from the engine, and `app/electron/mcp/resources.ts` re-serves the same two. The only app-shipped script snippets that touch the status use `pm.response.code` and `pm.response.to.have.status(200)`, neither of which changes. The many `response.status` reads under `app/src/modules/` are the REST layer's own numeric `status` field on `POST /execute` - a different field, deliberately untouched.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Breaking change
- [x] Documentation update

## Testing

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` - **2782 tests, 2782 passing, 0 failing.** No pre-existing failures on this base (`01d70d3`).

`cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check` - green; the doc-compile guard type-checks every `pm.*` fenced block in both edited pages against the regenerated declarations.

Tests added or changed:

- `ResponseStatusCode` - the pin flips from `pm.response.status === 200` to `=== "OK"`, beside the unchanged assertion on `code`.
- `ResponseStatusIsThePhraseAndCodeIsTheNumber` (new) - the split itself: phrase, `typeof` string, `code` still `404`, and agreement with `reason()`. Mutation-checked - reverting the binding to `JS_NewInt32` fails three of its four cases, and it is what would notice a `status` that quietly went back to the code while every assertion on `code` still passed.
- `ResponseStatusPrefersTheWirePhrase` (new) - a server's own wording (`"Totally Fine"` on a 200) wins over the registered text.
- `ResponseStatusFallsBackToTheRegisteredPhrase` (new) - an empty `status_text` (HTTP/2 carries no phrase) reads `"Service Unavailable"` for a 503, agreeing with `to.have.status`; and vayu's synthetic `0` reads `"Error"`.
- `SendRequestTest.DeliversTheResponseToTheCallback` - the callback response's `status` pin flips to `'OK'` over a real wire.

## Checklist

## Related Issues

Closes #1000

---

**Migration note** (for release-notes curation - this is a breaking change to a script-visible type):

`docs/engine/scripting.md` gains a `Was | Reads now | Use instead` subsection, on #1003's precedent, with `docs/app/pm-api-compatibility.md` cross-referencing it rather than duplicating the table. It names both legacy shapes honestly:

- **Loud.** `pm.response.status === 200` and `pm.expect(pm.response.status).to.equal(200)` become always-false, so the test that made them goes red and names itself. Use `pm.response.code`.
- **Silent.** `if (pm.response.status >= 400)` is not an assertion - JavaScript answers `false` rather than throwing, so the branch simply stops being entered and nothing reports that it stopped. The note says to search a ported script for arithmetic on `status` before trusting a green run.

The declarations fixture (`engine/tests/fixtures/script-typedefs.d.ts`) is regenerated with `VAYU_UPDATE_SCRIPT_TYPEDEFS=1`, not hand-edited; the diff is the three lines the completion table changed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FpjHMCMBzCbT13mocM9m7v)_

### DIFF
--- a/docs/app/pm-api-compatibility.md
+++ b/docs/app/pm-api-compatibility.md
@@ -252,8 +252,12 @@ capability being off. `res` carries `code`, `status`, `responseTime`,
 `headers` with `get()`/`has()`/`each()`/`all()`/`count()`/`toObject()`/`one()`/
 `indexOf()` - the same read methods as `pm.response.headers` - `json()` and
 `text()`, a subset of `pm.response` with no assertion chain. `status` is the
-numeric code there too, so the two objects called a response do not disagree
-inside one sandbox.
+reason phrase there too, so the two objects called a response do not disagree
+inside one sandbox - `code` is the number on both. Both spellings changed
+together in #1000; a script reading `res.status` as a number wants `res.code`,
+and [the migration
+note](../engine/scripting.md#status-is-the-reason-phrase-code-is-the-number)
+names the one pattern that degrades silently.
 
 **It is bounded, and both bounds throw.** The request's timeout is clamped to
 whatever is left of the script's own time budget (`scriptTimeout`, 5s by
@@ -496,6 +500,22 @@ on the `Date` and `RegExp` sides, and the values `include`, `oneOf`, `members`,
 test as the script's own error rather than as a difference. Chai reports a
 difference for some of these, which under `.not` is a pass; a test that cannot
 read its subject has not passed.
+
+### `code` is the number, `status` is the reason phrase
+
+Postman splits the status line in two and Vayu answers the same split:
+`pm.response.code` is `200`, `pm.response.status` is `"OK"` - the phrase the
+status line carried, or the registered text for the code where the wire had
+none (HTTP/2 carries no phrase at all). `pm.response.reason()` answers the
+same string, and `pm.response.to.have.status(...)` compares a number against
+the code and a string against the phrase.
+
+Vayu spelled `status` as the number until #1000, so this is a break for a
+script written against the old spelling rather than against Postman.
+`docs/engine/scripting.md` carries [the migration
+note](../engine/scripting.md#status-is-the-reason-phrase-code-is-the-number),
+including the one legacy pattern - arithmetic on `status` - that goes on
+compiling, stops matching, and reports nothing.
 
 ### Response timings, and the two error fields
 
--- a/docs/engine/scripting.md
+++ b/docs/engine/scripting.md
@@ -333,7 +333,7 @@ Access HTTP response data:
 
 ```javascript
 pm.response.code              // Status code (number, e.g., 200)
-pm.response.status            // Alias for code
+pm.response.status            // Reason phrase (string, e.g., 'OK') - see below
 pm.response.responseTime      // Perceived latency in ms (submit → completion).
                               // In load tests this includes generator-side
                               // queue wait. For pure server wire time, use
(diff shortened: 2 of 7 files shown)


athrvk:
One thing worth knowing when reading this PR's checks: **`App on ${{ matrix.os }}` and `App quality checks` report `skipped` here, and that is not because the app suite is irrelevant to this change.**

This PR edits `docs/engine/scripting.md` and `docs/app/pm-api-compatibility.md`, which are exactly the two files `app/src/hooks/script-typedefs.docs-compile.test.ts` reads and type-checks against the declarations fixture - so a bad doc example in this diff is something that suite, and only that suite, would catch. But `pr-tests.yml`'s `app` filter routes on `app/**` / `shared/**` / `VERSION` minus all `*.md`, and `docs/**` is in no area filter at all, so neither page can trigger the job that compiles it.

So the suite was run by hand instead, on this exact commit:

- `cd app && pnpm test` - **546 files, 5982 tests, all passing** (the docs-compile guard included).
- `mkdocs build --strict -f .github/mkdocs.yml` - clean, which is what says the new cross-page anchor resolves.

The routing gap itself is filed as #1118 rather than fixed here - it is CI plumbing, not this issue's scope, and widening the diff to reach it would put an unrelated workflow change under a scripting-surface review.

---

Issue #1152, opened 2026-08-30
app(dashboard): per-flush chart efficiency - three transforms computed twice, the ramp-up scatter escapes bucketing, and the live-dot animation names keyframes that do not exist


athrvk:
Sub-issue of #1143. Verified at master `e1c578a`, all three adversarially re-verified. Three low findings in the same chart surface, one work item.

## Problem

1. **Double compute.** `MetricsView.tsx:82, 84-85, 87-88` build `buildLatencyChartData`, `buildPercentileChartData` and `buildStatusOverTime` in useMemos consumed only as render gates (`:289`, `:363-364`, `:406`); the built arrays are never passed down, and each chart re-runs the identical O(n) transform from the same history in its own useMemo (`TimeSeriesCharts.tsx:113, :155`; `StatusCodesOverTimeChart.tsx:33`). The store replaces the array reference every flush (`dashboard-store.ts:180`), so both layers recompute each commit - the three heaviest series transforms paid twice at ~2Hz.
2. **The scatter escapes the bucketing design.** `MetricsView.tsx:356-360` hands raw `historicalMetrics` to the ramp-up scatter; `ScatterAndDistribution.tsx:41-56` maps per-tick objects, sorts O(n log n), builds three more per-point arrays, and uPlot draws one point per tick per redraw (`UPlotChart.tsx:205-212, 389-390`) - while every other live chart buckets first, and `live-window.ts:100-104` documents bucketing as the retention design's assumption. Bounded (50k cap, ramp_up mode only, opt-in full-run window) - real waste, low impact.
3. **Dead keyframe reference** ("written but never read" class): `UPlotChart.tsx:416` sets `animation: "vayuPulse 1.6s ..."` but only `vayu-pulse` exists (`index.css:785-793`, `tailwind.config.js:124,139`; CSS keyframe names are case-sensitive; repo-wide grep: no `vayuPulse` definition). The live-dot renders static on every live chart (`TimeSeriesCharts.tsx:136,185,272,307,341,410`; `StatusCodesOverTimeChart.tsx:101`) - the pulsing intent is stated in the comment at `:393-394`.

## Fix / solution pathway

1. Compute each series once in MetricsView and pass it down as a prop, or drop the MetricsView copies and let the charts' own empty-gates decide (`TimeSeriesCharts.tsx:128,177`, `StatusCodesOverTimeChart.tsx:93` already return null on short data).
[...]

Issue #1143, opened 2026-08-30
App performance program - startup latency, idle cost, live-run CPU/GC, RAM growth, per-platform costs (2026-08-30 sweep)


athrvk:
Tracking issue for the 2026-08-30 app-performance sweep, run at master `e1c578a`: startup time, idle CPU/RAM, GC pressure, live-run resource use - the app as a product, on all three platforms. Method: seven readers, two measurement agents, one adversarial verifier per finding; 28 candidates, 5 refuted.

## Program status (refreshed 2026-09-02, verified at master `fa5439e`)

**19 of 19 original subs closed, every closure verified at acceptance-criteria level. Exit criterion 2 is MET. Two late subs and two follow-ups remain.**

~~#1144~~ (window alongside engine, ramped poll, non-fatal ceiling) - ~~#1145~~ (MCP lazy) - ~~#1146~~ (React.lazy + lazy monaco; +~~#1172~~) - ~~#1147~~ (monaco slim import, -2.2MB dist) - ~~#1148~~ (ESRCH-first probe) - ~~#1149~~ (six fonts self-hosted) - ~~#1150~~ (prefetch-once, engine summary cache) - ~~#1151~~ (incremental anomaly detector) - ~~#1152~~ (predicate gates, bucketed scatter, live dot) - ~~#1153~~ (scenario steps batched, O(1) summary fold, memo cards) - ~~#1154~~ (execution resources released at retain) - ~~#1155~~ (reservoir byte budget) - ~~#1156~~ (response cache bounded at 24) - ~~#1157~~ (2MB display gate + engine 32MB /execute cap) - ~~#1158~~ (design SSE commits per flush through the shared batcher, windowed and memoised list; 20k/100k microbenchmark in PR #1247) - ~~#1159~~ (first update check held back 60 s after launch, unref'd, tested) - ~~#1160~~ (exe blockmap uploaded; release-level proof lands with v0.26.0) - ~~#1162~~ (measurement harness; dispatch and schedule paths both proven - schedule run #8 succeeded 2026-08-31) - ~~#1164~~ (engine `starting` is a state of its own with a 45 s grace window).

[...]

PR #1178, opened 2026-08-31
perf(app): the anomaly detector derives what arrived, not the whole buffer - #1151

### LINKED ISSUE #1151: app(dashboard): detectAnomalies rescans the entire retained tick history with slice+sort medians on every store commit - O(n*k) twice a second on the renderer main thread
Sub-issue of #1143. Verified at master `e1c578a`, adversarially re-verified.

## Problem

Every store commit (default 500ms, fastest 250ms - `load-test-service.ts:126,153`, `config/metrics.ts:22`, `client-settings.ts:62-67`) recomputes anomaly detection from scratch over the whole retained tick buffer: `MetricsView.tsx:103` memoizes `detectAnomalies(chartWindow)` on `chartWindow`, which IS `historicalMetrics` (`MetricsView.tsx:80`), and `dashboard-store.ts:180` builds a new array reference on every `addMetricsBatch` - so the memo never holds across commits. Per invocation, `trailingMedians` (`detectAnomalies.ts:96-100`) allocates a slice plus a spread copy and comparator-sorts 15 elements *per index*, over three series (p99 `:133-134`, rps and concurrency `:203-204`) - ~6n allocations and 3n sorts per commit - and `detectFirst5xx` (`:232-239`) runs `Object.entries` per tick over the whole history every commit for the life of a healthy run.

n is ~3,000 ticks at the default 5-min window and the engine's 10Hz tick (`constants.hpp:216`, `run_manager.cpp:1954-1955`), user-raisable to 15/30-min or full-run, hard-capped at 50,000 (`live-window.ts:115`). At defaults: ~9,000 slice+sort medians and ~18,000 short-lived arrays twice per second on the renderer main thread; at a full-run window during a soak, up to ~150,000 medians (~300,000 allocations) per commit - plausibly a large fraction of a core, plus steady GC churn.

Not deliberate: the "deliberately dumb" header defends detector *semantics*, not cost; `dashboard-store.ts:85-90` documents this codebase's own stance of folding to O(1) aggregates precisely to avoid full-history scans, which this path violates uncommented. Nuance: the history-view call site (`LoadTestDetail.tsx:158`) runs once per loaded run and is fine - this is the live path only.

## Fix / solution pathway

Make detection incremental: history is append-only between commits (trim only drops the front), so keep a rolling 15-tick window per series (small ring + incremental median, or recompute medians only for newly-arrived indices) and cache anomalies for the unchanged prefix; give `detectFirst5xx` a resume cursor. Alternatively (smaller): run detection at 1Hz on a stable snapshot instead of per commit. Follow the O(1)-aggregate pattern `dashboard-store.ts:85-90` already documents.

## App changes

This IS the app change.

## Tests

`cd app && pnpm test` - detectAnomalies has unit tests; add cases proving incremental output equals from-scratch output over a randomized append/trim sequence (property-style), plus a mutation check that the incremental path is actually taken (spy on the from-scratch helper).

## Docs (same commit)

None expected - internal transform. Verify `docs/app/COMPONENTS.md` does not describe the detector's mechanics; if it does, it moves.

## Sequencing

`MetricsView.tsx` + dashboard utils: lands FIRST in the dashboard pair, before the per-flush chart-efficiency issue (filed next in #1143). Parallel-safe against everything else.

## Acceptance criteria

- [ ] Per-commit detection cost no longer scales with total retained history (only with new ticks), proven by the equivalence + mutation tests.
- [ ] Anomaly output unchanged for identical inputs.

---

athrvk:
## Description

`MetricsView.tsx:103` memoized `detectAnomalies(chartWindow)` on the store's `historicalMetrics`, and `dashboard-store.ts:180` builds a new array reference on every `addMetricsBatch` - so the memo never held across commits and the detector re-derived the entire retained buffer twice a second (250ms at the fastest setting). Per pass that was a 15-element slice-plus-spread comparator sort *per index, per series* over three series - ~9,000 sorts and ~18,000 short-lived arrays at the default 5-minute window, up to ~150,000 sorts during a full-run soak - plus `detectFirst5xx` running `Object.entries` per tick over the whole history for the life of a healthy run. All of it on the renderer's main thread. Every anchor in the issue was re-verified against master `db346d3`; all seven were confirmed, with only cosmetic line drift.

**The plan.** `detectAnomalies` stays exactly what it was: the pure one-shot over a whole series, which is what `LoadTestDetail` wants and what every rule is defined against. Alongside it, `createAnomalyDetector()` holds its derivation across calls, and the live view keeps one. The split that makes the two agree is the whole design: what a tick **and its predecessors** decide - a trailing median, a failure-rate delta, whether a rule is satisfied - is computed once when the tick arrives and cached under an *absolute* tick index that no trim renumbers, so it can never need revisiting; what a tick's **position in the buffer** decides - no baseline for the first `BASELINE_TICKS`, no delta for the first tick - is re-applied when the buffer is trimmed, by clipping the leading window's start to the oldest tick the rule can still speak for. A position only ever falls, so hotness is monotone: a trim clips a window and can never split one, and a window whose lead-in has rolled out goes quiet exactly as a from-scratch pass over the shortened buffer does. The next buffer is recognised as the last one with ticks appended and the oldest dropped, the offset found by searching elapsed time and then confirmed by object identity at *both* ends; anything that does not confirm is refused and re-derived from scratch rather than guessed at, since a wrong alignment would silently report another run's windows.

**The alternative I rejected.** The obvious one is the store's own latched-aggregate pattern - fold anomalies into `addMetricsBatch` beside `peakConcurrency` and `breakpoint`, whose rationale comment (`dashboard-store.ts:84-90`) is precisely this "O(1) per tick rather than a full scan" stance. It does not fit: those two are *latched* on purpose ("peak only grows, breakpoint sticks on first crossing, so they survive old ticks rolling out of the retention window"), and anomalies are the opposite - a window must stop being reported when the history it rested on is trimmed away. A latched store field would have changed what the dashboard reports, failing the issue's second acceptance criterion. It also could not serve `LoadTestDetail`, which detects over a fetched series that never touches the store. So the derivation stays in the view; only its *state* is now carried between commits.

A second, smaller alternative the issue offers - run detection at 1Hz on a stable snapshot - was rejected because it leaves per-commit cost scaling with total retained history, which is what acceptance criterion 1 rules out.

**Note on the residual cost.** Per-commit work is now O(dropped + appended + number of open anomaly windows). The last term is why this is not flatly O(new ticks): assembling the report walks the run list, and a run's list is proportional to the anomalies in the buffer, not to its length. On a healthy run that list is empty and the commit costs only the arriving ticks. The recovery tail - the one piece of state that grows unboundedly while a degradation stalls - is resumed rather than rescanned, and there is a test for that specifically.

**Adjacent, not touched:** `LATENCY_RECOVERY_FACTOR` is exported but read only inside this module, and `Anomaly.magnitude` is consumed only by the over-cap ranking - no renderer reads it (the label carries the number as text). Both are small instances of the repo's "written but never read" pattern and both are out of this issue's scope; flagged here rather than folded into the diff.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

Closest to a performance change; ticked "new feature" for the added entry point. No behaviour change: the reported anomalies are identical.

## Testing

Local, all green:

- `pnpm test` - **556 files, 6073 tests passed**. The detector's own file goes from 14 to 25 tests.
- `pnpm type-check` - clean across all three projects (renderer, electron, electron-tests).
- `pnpm lint` - clean, zero warnings.
- `pnpm format:check` - clean.

No engine build or `ctest` run: the diff touches no C++ and no engine input, so nothing there could go from green to red (root `CLAUDE.md`, "scale the verification to what the change could actually break"). No pre-existing failures were observed on the base commit.

The 11 new cases are:

- **Equivalence** - two randomized append/trim replays (seeded, so a failure names one reproducible sequence) driving the store's real buffer discipline: append a batch, front-trim past the time window, then past the hard cap, and compare against a from-scratch pass after *every* commit. One replay uses a cap smaller than the run so every commit trims; the other a window wide enough to hold whole anomaly windows. Both assert the sequence actually produced all four anomaly kinds - an agreement between two empty lists is not an agreement, and the first draft of this test silently was one until that assertion was added.
- **Trim semantics** - a window whose lead-in has rolled out falls silent, matching a rescan of the shortened buffer.
- **Buffer recognition** - starting over on a buffer that is not a continuation; refusing one whose head is a different tick at the same elapsed time; refusing one that shares only its first tick; and returning the identical result (and spending zero sorts) when handed the same array twice, which is what React does under StrictMode.
- **Cost** - counting the comparator sorts, `Object.entries` scans and `latency_p99_ms` reads a single commit spends, against a from-scratch pass over the same buffer for contrast.

Each was mutation-checked by breaking the implementation and confirming a failure: disabling the front clip (3 tests fail), dropping the clip's cache invalidation (2), restarting rather than resuming the recovery tail (1), removing the head alignment check (1), removing the tail alignment check (1).

Two mutations *survived* the first draft, and rather than adding tests for them I removed what they proved was redundant: the position rules were being applied in three places at once - a placeholder at push, a check on read, and the front clip - and only the clip was observable. The read-time checks are gone and the front clip is now the single mechanism, documented as load-bearing.

## Checklist

Docs, same commit: the detector's mechanics moved out of `docs/app/COMPONENTS.md` (as the issue asked - it did describe them) into `app/src/modules/dashboard/README.md`, which now carries the rules, the two entry points and the invariant a future rule has to honour. `COMPONENTS.md` keeps a one-line pointer.

## Related Issues

Closes #1151

Unblocks #1152 (the per-flush chart efficiency issue), which the wave map in #1143 sequences after this one in the dashboard pair.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XrWfwFDVtRBBcM2wLYSTEQ)_

### DIFF
--- a/app/src/modules/dashboard/README.md
+++ b/app/src/modules/dashboard/README.md
@@ -62,8 +62,8 @@ arrives:
 - **`detectAnomalies(history)`** is the one-shot, pure over the whole series.
   `LoadTestDetail` uses it: a stored run's series arrives once, complete.
 - **`createAnomalyDetector()`** holds its derivation across calls, for a caller
-  handing over the same growing buffer again and again. `MetricsView` uses it,
-  in a ref. The live buffer's array identity changes on every store commit
+  handing over the same growing buffer again and again. `MetricsView` holds one
+  for the run's life. The live buffer's array identity changes on every commit
   (twice a second by default), so a from-scratch pass re-derived a trailing
   median per tick per series over the whole retained window - ~9,000
   fifteen-element sorts twice a second at the default 5-minute window, up to
--- a/app/src/modules/dashboard/README.md
+++ b/app/src/modules/dashboard/README.md
@@ -47,3 +47,33 @@ consumes it.
   `components/stats/ModeStatsRow.tsx` decides which four cards each mode shows,
   and its file comment tabulates them. Add a mode there, next to the branch that
   renders it.
+
+## Anomaly detection (`utils/detectAnomalies.ts`)
+
+The run's degradation windows - latency spikes, error bursts, throughput drops
+and the first 5xx - scanned out of the same per-tick series the charts plot,
+with fixed factors over a trailing median and nothing tunable. Two readings of
+one detection: the charts shade the windows, and the history Overview's
+`RunEvents` card states them in words. Both are absent for a clean run.
+
+It has two entry points, and which one a view wants depends on how its series
+arrives:
+
+- **`detectAnomalies(history)`** is the one-shot, pure over the whole series.
+  `LoadTestDetail` uses it: a stored run's series arrives once, complete.
+- **`createAnomalyDetector()`** holds its derivation across calls, for a caller
+  handing over the same growing buffer again and again. `MetricsView` uses it,
+  in a ref. The live buffer's array identity changes on every store commit
+  (twice a second by default), so a from-scratch pass re-derived a trailing
+  median per tick per series over the whole retained window - ~9,000
+  fifteen-element sorts twice a second at the default 5-minute window, up to
+  ~150,000 during a full-run soak, on the renderer's main thread (#1151).
+
+The two answer identically for the same buffer, and the rule that keeps them
+that way is in the file's header comment: what a tick and its predecessors
+decide is cached once under an index no trim renumbers, and what a tick's
+_position in the buffer_ decides is re-applied when the buffer is trimmed. A new
+rule that reads a tick's position without honouring the second half will pass
+its own unit test and disagree with the one-shot the first time the window
+fills; `detectAnomalies.test.ts` replays randomized append/trim sequences
+against a from-scratch pass to catch exactly that.
--- a/app/src/modules/dashboard/components/MetricsView.tsx
+++ b/app/src/modules/dashboard/components/MetricsView.tsx
@@ -107,7 +107,9 @@ function MetricsView({
 	 * window (#1151). Handed the same buffer again it takes only what arrived.
 	 * It answers exactly what the one-shot `detectAnomalies` would, including for
 	 * a buffer that is not a continuation (a second run, a remount), which it
-	 * starts over on - so nothing here depends on the ref surviving.
+	 * starts over on - so nothing here depends on the instance surviving. Held in
+	 * `useState` rather than a ref because the state initialiser is the one hook
+	 * that runs exactly once without reading anything during render.
 	 */
 	const [detector] = useState(createAnomalyDetector);
 	const anomalies = useMemo(() => detector.detect(chartWindow), [detector, chartWindow]);
--- a/app/src/modules/dashboard/components/MetricsView.tsx
+++ b/app/src/modules/dashboard/components/MetricsView.tsx
@@ -21,7 +21,7 @@
  * derives - it does not render metric chrome inline.
  */
 
-import { memo, useMemo } from "react";
+import { memo, useMemo, useState } from "react";
 import { Activity } from "lucide-react";
 import { formatNumber } from "@/utils";
 import { useDashboardStore } from "@/stores";
@@ -39,7 +39,7 @@ import {
 	buildStatusOverTime,
 	latestThroughputMbps,
 } from "../utils/metricsTransforms";
-import { detectAnomalies } from "../utils/detectAnomalies";
+import { createAnomalyDetector } from "../utils/detectAnomalies";
 import { HeroRow } from "./hero/HeroRow";
 import { ModeStatsRow } from "./stats/ModeStatsRow";
 import {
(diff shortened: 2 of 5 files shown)


athrvk:
Correcting the "Adjacent, not touched" paragraph in the description above - I checked both claims properly afterwards and neither is what I called it. Recording it here so nobody opens an issue against a non-defect.

**`LATENCY_RECOVERY_FACTOR`** - I said it was exported but read only inside the module, and filed it under "written but never read". The first half is true, the label is not. It is read by the rule it names, and its *value* is pinned behaviourally: setting it to 2.5 or to 1.1 fails three tests (`keeps a window open through a slow recovery instead of closing it at 3x`, plus both open-tail cases). The file's stated contract - "every constant below is exported so the tests pin the rule rather than restate it" - holds for it. It is pinned by the series the tests construct rather than by an `expect(LATENCY_RECOVERY_FACTOR).toBe(...)`, which is the stronger form of the two. (It now has an external reader as well: the oracle added in #1179.)

**`Anomaly.magnitude`** - I said no renderer reads it. That is true and stays true: `RunEvents` reads `kind`/`label`/`startSeconds`/`endSeconds`, and `anomalyAnnotations` reads `startSeconds`/`endSeconds`/`label`/`kind`. But it is not an unread field - it is what the over-`MAX_ANOMALIES` triage ranks by, which is real behaviour, and the number reaches the UI inside the `label` string by design. A structured field with an internal consumer and a rendered text form is not the defect pattern.

So there is no follow-up hiding here, and no issue filed for one. The genuine follow-up from this PR was the recovery-tail cost, which is #1179.

---

Issue #1218, opened 2026-09-01
app(layout, collections): focus falls to `<body>` after closing a tab from the keyboard and after confirming a tree delete


athrvk:
Part of #1211 (keyboard accessibility program). Verified at master `5677ed5`.

## Problem

Two destructive keyboard actions strand focus.

1. **Tab close.** `TabStrip.tsx:108-111`: Delete/Backspace on a focused tab calls `closeTab(tab.id)`; `Shell.tsx:191-195` does the same for the close-tab chord. The store picks the next *active* tab (`stores/tabs-store.ts:343-358`) but nothing moves DOM focus; the closed `role="tab"` element (roving `tabIndex`, `TabStrip.tsx:88`) unmounts and focus drops to `<body>`. `TabStrip` has no `useEffect` at all and its only `.focus()` is the arrow-navigation path (`:244`). `TabStrip.keyboard.test.tsx:170-195` asserts store state, not `document.activeElement`. (verified)
2. **Tree delete.** `DeleteConfirmDialog` is rendered controlled with no trigger (`CollectionTree.tsx:358-377`); it handles open-focus (`delete-confirm-dialog.tsx:93-98`, focus Cancel) but not close. Radix FocusScope then restores focus to the previously focused element, which was the row's menu item or the row itself (the roving Delete key clicks `[data-tree-delete]`, `useRovingTreeFocus.ts:220-224`) - both gone after the mutation, so focus drops to `<body>`. `useRevealActiveSelection.ts:166-183` refocuses only *moved* rows. `rg onCloseAutoFocus app/src` (non-test) returns nothing. (plausible - the mechanism is verified in code, the resulting activeElement was not run.)

Severity: medium. A keyboard user must Tab from the document start after every close or delete; screen readers lose their place.

## Fix / solution pathway

1. In the tab's Delete/Backspace handler, focus the neighbouring tab element before calling `closeTab` - the tablist already has the `tabs[next].focus()` logic at `TabStrip.tsx:215-244` to reuse. For the chord from inside content, focus the newly active tab (`[data-tab-id]`) in `Shell` after `closeTab`, or move that responsibility into the strip via a store subscription so both paths share it.
2. Pass `onCloseAutoFocus` on `DeleteConfirmDialog` (Radix prop; the component already uses `onOpenAutoFocus` at `:95`) to focus the next sibling row through the roving hook's `focusItem`. The Trash view's restore flow should get the same treatment if it shares the dialog.

## App changes

This is the app change.

[...]

Issue #1214, opened 2026-09-01
app(request-builder): the XML body mode mounts `xml` but the `{{` completion provider covers only json, plaintext and graphql - and the guard test enumerates the same three


athrvk:
Part of #1211 (found in the editors pass). Verified at master `5677ed5`.

## Problem

- `BodyPanel.tsx:427-428` (`app/src/modules/request-builder/components/RequestTabs/panels/`) mounts the editor with `language="xml"` when `request.bodyMode === "xml"`.
- `BODY_LANGUAGES = ["json", "plaintext", "graphql"]` (`app/src/hooks/useVariableCompletionProvider.ts:70`) is the list the `{{` completion provider is registered for (`:239-241`). XML bodies get no variable completion.
- The guard written for exactly this drift ("Add a body mode that mounts, say, `xml`, and completion is simply absent", `body-editor-completion.test.tsx:20-22`) covers only `json`, `text`, `graphql` in its `it.each` (`:101-105`) and iterates the same three in the dead-entry mirror (`:119`), so the XML mode landed under it.

Severity: low. A feature is missing in one mode; nothing misbehaves.

## Fix / solution pathway

Add `"xml"` to `BODY_LANGUAGES` and `["xml", 1]` to both loops in the test. Then make the test derive the mode list from the body-mode table (`BodyPanel.tsx:92`, the `{ value, label, contentType }` entries) rather than a literal, so the next mode cannot slip under it either.

## App changes

This is the app change.

## Tests

`cd app && pnpm test -- body-editor-completion`. Mutation check: remove `"xml"` from `BODY_LANGUAGES` and the test must fail.

## Docs (same commit)

`docs/app/COMPONENTS.md:271` says xml "buys highlighting and the header rather than an editor" - reword once completion is in.

## Sequencing

One constant and one test; parallel-safe with everything under #1211, including the Monaco-variable-tokens sub-issue (which registers a different provider).

## Acceptance criteria

- [ ] Typing `{{` in an XML body offers variables.
- [ ] The guard test enumerates every body mode from the mode table, not a literal list.

---
Generated by [Claude Code](https://claude.ai/code)

athrvk:
Claiming this issue - PR will follow from this run.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #1211, opened 2026-09-01
Keyboard accessibility program - dead menu path in the tree, Monaco Tab trap, focus loss after close/delete, region navigation, variable tokens in editors, tooling (2026-09-01 sweep)


athrvk:
Tracking issue for the keyboard-accessibility revisit. Sweep at master `5677ed5`, four read-only passes (global shortcuts and focus, composite widgets, editors, tooling and primitives), every finding below re-verified in code by the reviewer before filing. Sub-issues are implementer-ready and grouped by conflict surface; this issue only tracks.

## What is already right (do not re-do)

- One shortcut registry (`app/src/constants/shortcuts.ts`) with platform derived once (`lib/platform.ts:29`), strict shift/alt matching, and tests pinning single definition, listing and platform. Menu, Dock, palette and UrlBar all format chords from it.
- The collection tree is a real WAI-ARIA tree with a roving tabindex: arrows, Home/End, Right/Left, Enter/Space, F2, Delete/Backspace, `*`, typeahead, Alt+Arrow reorder (`useRovingTreeFocus.ts`). Hover-only row actions reveal on `group-focus-within` and a source guard enforces it (`row-action-reveal.test.ts`).
- The tab strip is a correct hand-rolled tablist (`TabStrip.tsx`) with arrow/Home/End, Delete/Backspace close, an overflow menu, and keyboard tests.
- Key-value editors have a labelled enable checkbox, a trailing blank row so focus is never lost while typing, and a focus-revealed Remove button.
- Every `<Button size="icon">` carries an accessible name (100% of 16 raw sites plus 31 `TooltipIconButton` uses), enforced by `icon-button-labels.test.tsx`.
- A global `:focus-visible` outline covers every interactive element at specificity 0 (`index.css:842-855`); every Radix primitive either paints its own ring or uses a background-fill indicator.
- Radix dialogs and popovers handle initial focus, Escape and focus return; three destructive dialogs deliberately focus Cancel.

## What is wrong, by severity

| Sev | Finding | Sub-issue |
|---|---|---|
| High | Collection tree row menu (Duplicate, Move to, Run, Add, Export) cannot be opened from the keyboard: Shift+F10 / ContextMenu / Shift+Enter fire `.click()` on a Radix trigger that listens only to pointerdown and keydown | #1212 |
[...]

PR #1235, opened 2026-09-02
fix(VariableInput): own the suggestion list, give the field combobox semantics, and stop hiding the tokens

### LINKED ISSUE #1215: app(VariableInput): the suggestion list steers whichever cmdk root the document finds first, the input has no combobox semantics, tokens are tab stops inside an aria-hidden overlay, and the secret-variable popover cannot be entered by keyboard
Part of #1211 (keyboard accessibility program). Verified at master `5677ed5`. Four defects on one surface, `app/src/components/shared/VariableInput/index.tsx` and `app/src/components/ui/variable-popover.tsx`.

## Problem

1. **Arrow and Enter are forwarded to the first cmdk root in the document.** `VariableInput/index.tsx:268-297`: ArrowUp/Down build a synthetic `KeyboardEvent` and dispatch it at `document.querySelector("[cmdk-root]")`; Enter/Tab click `document.querySelector('[cmdk-item][data-selected="true"]')`. Four `Command` roots exist in the app (`CommandPalette.tsx`, `suggestion-list.tsx`, `variable-autocomplete.tsx`, `command.tsx`). If another root is mounted earlier in DOM order, the keys steer the wrong list and Enter picks the wrong item. Not asserted anywhere.
2. **No combobox semantics.** `rg 'role="combobox"|aria-activedescendant|aria-expanded' app/src --type tsx` (non-test) returns nothing. Assistive technology has no relationship between the input and the suggestion list.
3. **Focusable tokens inside `aria-hidden`.** Each `{{name}}` token is `role="button" tabIndex={0}` (`variable-popover.tsx:336-337`, deliberately, per `keyboard-reachability.test.tsx:16-21`) but the overlay that contains them is `aria-hidden="true"` (`VariableInput/index.tsx:600-608`). Screen readers are told to ignore an element that receives focus (axe rule `aria-hidden-focus`), and a URL with five variables adds five Tab stops between the URL and Send.
4. **Secret-variable popover branch has no entry point.** The popover sets `onOpenAutoFocus={(e) => e.preventDefault()}` (`variable-popover.tsx:373`), compensated by `autoFocus` on the value input (`:488`) and the new-variable input (`:577`) - but the "secret, not yet revealed" branch (`:462-476`) renders a `readOnly` input plus `RevealButton` with neither. The popover is non-modal and portaled (`popover.tsx` passes no `modal`), so Tab from the trigger does not enter it and focus-outside dismisses it. `keyboard-reachability.test.tsx:56-62` asserts the dialog appears, not that focus can reach it.

Severity: medium. Item 1 is a correctness bug waiting on DOM order; 4 makes secret variables unreachable by keyboard; 2 and 3 are semantics.

## Fix / solution pathway

1. Own the list: hold a ref to the `Command` root that `VariableAutocomplete` renders and dispatch to it (or, better, drive cmdk through its own `value`/`onValueChange` controlled API and call the highlighted item's `onSelect` directly). Do not query the document.
2. Give the input `role="combobox"`, `aria-expanded`, `aria-controls` (the list id) and `aria-activedescendant` (the highlighted item's id); mark the list `role="listbox"` if cmdk does not already.
3. Either drop `aria-hidden` from the overlay and make the token strip one roving Tab stop (arrow keys between tokens - `useRovingTreeFocus` is the existing roving primitive; see `docs/app/COMPONENTS.md:273`), or keep `aria-hidden`, make tokens `tabIndex={-1}` and give the input an `aria-describedby` that names the variables. Choose one and record it; the current state is a deliberate trade for reachability (`keyboard-reachability.test.tsx`), so preserve reachability.
4. In the hidden-secret branch give `RevealButton` `autoFocus` (or stop preventing open-autofocus in that branch), and extend the reachability test to assert `document.activeElement` is inside the popover after opening.

## App changes

This is the app change.

## Tests

`cd app && pnpm test -- VariableInput variable-popover keyboard-reachability` and `pnpm type-check`, `pnpm lint`.

- Mount two `Command` roots with the VariableInput second; ArrowDown must move the highlight in the attached list (fails today).
- Combobox attributes present and `aria-activedescendant` tracks the highlight.
- Secret branch: after opening, focus is on the reveal button; Tab/Escape behave.
- Token strip: one Tab stop (or `tabIndex=-1` with `aria-describedby`), per the choice made; Enter/Space on a token still opens the popover.

## Docs (same commit)

`docs/app/COMPONENTS.md:1472` token-state table and the VariableInput section; `docs/app/variable-resolution.md:261-264` (tooltip/popover parity paragraph).

## Sequencing

Files: `VariableInput/index.tsx`, `variable-popover.tsx`, `variable-autocomplete.tsx`. Nothing else under #1211 touches these. Parallel-safe.

## Acceptance criteria

- [ ] Suggestion navigation is bound to the input's own list regardless of DOM order; test proves it.
- [ ] Input carries combobox semantics with a live `aria-activedescendant`.
- [ ] No focusable element sits inside an `aria-hidden` subtree in VariableInput.
- [ ] A secret variable's popover receives focus on open from the keyboard.

---

athrvk:
## Description

Four defects on one surface, all of them assembled out of two earlier fixes that never met.

**Root cause.** The suggestion list was never *owned* by the field that steers it: `VariableInput` synthesised a keydown at `document.querySelector("[cmdk-root]")` and selected by clicking `document.querySelector('[cmdk-item][data-selected="true"]')` - the first cmdk list in the whole document, of which the app has four. And the token popover had been made keyboard-operable (`role="button"`, `tabIndex={0}`, Enter/Space) while the overlay it renders into was marked `aria-hidden="true"` as a visual duplicate; each commit was right alone, and together they hid a focusable control.

**Approach.** The ordering is extracted to `lib/variable-suggestions.ts` and read by both sides, so "the next row" is answerable without touching the DOM: `VariableAutocomplete` draws that list, `VariableInput` steps through the same one via cmdk's controlled `value` / `onValueChange` and calls the selected entry's handler directly. On that ownership the input can carry real combobox semantics. For the overlay we took the first of the issue's two options - drop `aria-hidden`, one roving Tab stop for the token strip - because the second (keep `aria-hidden`, tokens `tabIndex={-1}`) withdraws the reachability `keyboard-reachability.test.tsx` exists to protect; the text segments and run-time tokens carry `aria-hidden` individually instead, since those genuinely do restate the input's own value and neither is focusable.

**Rejected alternative:** replacing cmdk inside `VariableAutocomplete` with a hand-rolled listbox, which would have handed us the option ids directly instead of having the list report them. Rejected because `suggestion-list.tsx` documents the opposite move as a deliberate correction ("a hand-rolled copy of a primitive does not receive the primitive's fixes"), and cmdk still owns the highlight paint and scroll-into-view. `CommandListboxProbe` reports the ids from inside the list instead, subscribed to cmdk's own store - `state.selectedItemId` could not be used, because cmdk only maintains it for a list it highlights itself and writes a controlled `value` straight into the store.

- **Plain header suggestions are now navigable too.** They never were - cmdk reads arrows off its own `Command.Input`, which neither list renders - so that list was mouse-only and the comment in `VariableInput` claiming cmdk handled it was wrong. Special-casing the fix to the variable list would have been more code, and advertising a navigable list to a screen reader while the arrows do nothing is worse than the original bug.
- **`role="combobox"` changes the input's accessible role**, app-wide: every `VariableInput` with a `variables` scope (URL bar, headers, params, auth) now reports it. That is what criterion 2 asks for, but it is worth knowing the reach. Two existing tests caught it and were corrected rather than bumped: 4 `getByRole("textbox", …)` queries in `DesignRunView.test.tsx`, and `auth-editor-parity.test.tsx`, which counted every `role="combobox"` as a proxy for "the location control is a `Select`" and now counts `Select` triggers - changing its expected count 2 → 4 would have kept it passing for the wrong reason.
- **Revealing a secret now moves focus to the field** (second commit). The masked and revealed branches are the same `<Input>` in the same slot, so React reconciled rather than remounted them and `autoFocus` never fired; keying them apart completes the flow the issue's fix pathway describes, and makes the comment beside the reveal button true.
- **The overlay's test hook changed** from `[aria-hidden="true"]` to `data-variable-overlay` in six token tests. The old selector returns the first match in the tree, so it was one decorative icon away from silently measuring the wrong node.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check` - all green, re-run after merging `master` in.

- **578 files, 6355 tests passing** on the merged tree. **+20 are this PR's**; the rest of the increase over the original base is the two PRs merged into `master` while this was open (#1232, #1233). No pre-existing failures on `master`.
- New `app/src/components/shared/VariableInput/keyboard-navigation.test.tsx` - 17 cases: own-list navigation with a decoy `Command` root mounted first, clamping at the ends, Enter selection, ⌘Enter left alone for the Send chord, plain-list navigation, combobox attributes with a live `aria-activedescendant` tracking both an arrow-moved and a pointer-moved highlight, the plain-textbox case, no focusable element under `aria-hidden`, the parts that stay hidden, and the roving token strip (one Tab stop, arrow/Home/End movement, no wrap, disabled leaves the strip entirely, popover still opens).
- `keyboard-reachability.test.tsx` - 4 new cases: focus lands inside the popover, a hidden secret lands on the reveal eye, revealing hands focus to the field, and Escape closes without disclosing. Tab itself is not simulated - jsdom implements no sequential focus navigation and `userEvent` is not a dependency here (the call `runtime-token-interaction.test.tsx` already records) - so what Tab was needed *for* is asserted instead.
- **Mutation-checked.** Reverting each of the eight fixes turns the covering test red, verified by patching them back one at a time: the `document.querySelector` arrow dispatch, the Enter click, `aria-hidden` on the overlay, the roving `tabIndex`, `role="combobox"`, `aria-activedescendant`, the secret branch's `autoFocus`, and the branch keys.
- `mkdocs build --strict` passes for the three docs pages.
- The engine is untouched, so no engine build or `ctest` run was made - CI's engine jobs skipped for the same reason. Per CLAUDE.md, a test that cannot change colour is not worth running.

### Noted, not changed

`renderOverlayContent` is a long, branch-heavy function (five ordered token classifications plus a text fallback). It was already that shape and its ordering is deliberately pinned to the resolver's precedence; extracting a `classifyToken` discriminated union would be a good next touch of this file, but it is not this fix's call and would bury the diff. Filed as **#1239**, sequenced after this and after **#1238** (run-time token tooltips are still mouse-only - this PR narrowed the overlay's `aria-hidden` rather than removing it, and that reasoning holds only until those tokens become focusable, so the two changes have to land together).

## Checklist

## Related Issues
Closes #1215


athrvk:
Two things this change deliberately left open now have issues behind them rather than a note in the description:

- **#1238** - a run-time token's tooltip (`{{$guid}}`, `{{data.email}}`, `{{$vu}}`) is its entire content and still cannot be opened from the keyboard. This PR narrowed the scope of the `aria-hidden` on the overlay rather than removing it everywhere, and put it on the run-time wrappers specifically because they are not focusable. That reasoning holds exactly until they become focusable, so the two changes have to happen together. Filed under #1211.
- **#1239** - `renderOverlayContent` holds five ordered token classifications and five near-identical JSX shapes in one 180-line function. Both reviewers landed on it independently. It is why the roving strip heals its Tab-stop index off the rendered DOM instead of counting tokens: the classification is not available without rendering, and restating it would have been the duplication `app/CLAUDE.md` warns about. Sequenced after this PR and after #1238, since all three touch the same function.

Neither is in scope here - the first is a different set of tokens, the second is a refactor that would bury this diff.

---

athrvk:
Merged `master` in - the branch had gone `behind` while this was open, and the two PRs that landed (#1232 for #1218, #1233 for #1214) both touch `docs/app/COMPONENTS.md` and `docs/design-system.md`. No conflicts, and no source overlap.

One thing did need saying rather than just merging. #1232 extracted "move the tree's one tab stop and focus the row" out of `useRovingTreeFocus` into a shared `focusTreeRow` (`tree-focus.ts`). That is a far closer neighbour to this PR's roving token strip than the hook was, so both doc paragraphs explaining why the strip is hand-wired now have a second question to answer - and the answer is a different one: the hook cannot *find* a token (`[role="treeitem"]`, `data-tree-*`), whereas `focusTreeRow` would find it and then move the stop by writing `tabIndex` onto DOM nodes. That is correct for tree rows, which render `tabIndex={-1}` and do not re-render for it - `tree-focus.ts` says so itself - and wrong here, where the stop *is* a React prop and the write would sit on a vdom value React still believes.

Re-verified on the merged tree: **578 files, 6355 tests passing**, lint / type-check / format clean, `mkdocs build --strict` clean. The +20 this PR adds are unchanged; the rest of the increase is the two merged PRs' own tests.

---

PR #1263, opened 2026-09-02
fix(app): derive the status-colour families from index.css, and correct what the doc says about amber - #1253

### LINKED ISSUE #1253: app(design): `--status-warning-text` exists but the design system says it does not, with a ratio table that does not reproduce - and the bare-fill guard's family list is hardcoded past it
Found while implementing #1225 (PR #1252), which measured the status family's `-text` pairs and did not touch the amber one. Filed rather than folded in: it is a different rule (never use a bare fill as a foreground) from the one that PR applies (which family a status surface belongs to), and fixing it means re-measuring a table.

## Problem

Three claims about `--status-warning` disagree with `index.css`.

**1. The doc says the token has no `-text` variant. It has one.** `docs/design-system.md:298-299`:

> `status-warning` is the exception: it has no `-text` variant, because the bare token already measures 17.72 / 15.78. Leave it as-is.

`app/src/index.css` declares `--status-warning-text` in both blocks (`38 90% 30%` light, `40 92% 60%` dark) and `@theme inline` maps `--color-status-warning-text` (`:69`), so `text-status-warning-text` is a real utility - and ten components already use it (`toast.tsx:136`, `ResponseStatusBar.tsx:223`, `ScriptSection.tsx:51`, `ContractCoverage.tsx:88`, `SettingsPanel.tsx:202`, `LoadTestDetail.tsx:239,304`, `SpecSync.tsx:551`, `ScenarioStepsTab.tsx:84`). A reader following the doc's "leave it as-is" would paint the bare token as text.

**2. The ratios in that row do not reproduce.** `docs/design-system.md:288` reports `status-warning` at 17.72 light / 15.78 dark, in a table whose stated method is "Measured on `bg-card` in the running app". Recomputing with the WCAG 1.4.3 formula (the same script reproduces the neighbouring rows: `status-success` `-text` 5.68 against the doc's 5.71):

| | light (`--card` `0 0% 100%`) | dark (`--card` `240 6% 11%`) |
|---|---:|---:|
| bare `--status-warning` (`38 92% 36%`) | **3.98** | 4.35 |
| `--status-warning-text` (`38 90% 30%` / `40 92% 60%`) | 5.46 | 9.79 |

So the bare token fails the 4.5 floor for normal text in light, which is exactly what the doc's row says it does not do, and the `-text` variant it says does not exist is what clears it. 17.72 is not a ratio either value produces against either card.

**3. The `css` block in that section lists four status tokens; there are seven.** `docs/design-system.md:505-509` shows `--status-success`, `-error`, `-running`, `-stopped`. `index.css` also declares `--status-redirect`, `--status-no-response` and `--status-warning`, all with `-text` pairs, and the HTTP-status vocabulary further down the page already refers to them.

**4. The guard's family list is hardcoded past the same three.** `app/src/components/ui/status-color-tokens.test.ts:68-76` enumerates `["destructive","success","warning","status-success","status-error","status-stopped","status-running"]`, with a comment (`:40-42`) repeating the doc's claim that `status-warning` has no `-text` variant. So `text-status-warning`, `text-status-redirect` and `text-status-no-response` used as a bare foreground are currently caught by nothing. No live instance exists today - the gap is future drift, as with the icon-button guard in #1216.

## Fix / solution pathway

Derive the list rather than restating it. `design-system-doc.test.ts` already parses every `--token: value;` declaration out of `index.css` (`declaredValues()`), so the same parse answers "which families have a `-text` pair": any `--X` whose `--X-text` is also declared. Feeding that into `status-color-tokens.test.ts`'s `FAMILIES` closes items 4 and keeps closing it as tokens are added, and it is the derivation `app/CLAUDE.md` asks for over a hardcoded list. Keep the existing non-empty floor.

Then correct the prose: drop the "no `-text` variant" sentence and its comment twin in the test, re-measure the amber row against all four surfaces the section names (`--card`, `--background`, `--muted`, the token's own `/10` tint) rather than only the card, and extend the `css` block to the three missing tokens. If the recomputed table disagrees with any other row, say so in the PR rather than quietly restating it - the numbers there were measured by hand and only the bare HSL triples are machine-checked (`design-system-doc.test.ts`, and `app/CLAUDE.md:184`: "Prose is not - if you change a value, read the sentence around it").

Worth deciding while there, and not before: whether `--status-warning`'s indicator value should follow the `-500` convention its four siblings use. The index.css comment (`:315-320`) already explains that it cannot - amber-500 measures 2.14 on a light card, under the 3.0 icon bar - so this is a documentation question, not a retune.

## App changes (required)

`app/src/components/ui/status-color-tokens.test.ts` only. No component changes: no bare status token is used as a foreground today.

## Tests

`cd app && pnpm test -- status-color-tokens design-system-doc`. Mutation-check the derived list: add a `text-status-warning` foreground to a fixture string and the guard must fail; remove the `-text` declaration for a family from the parse and that family must drop out rather than throwing.

## Docs (same commit)

`docs/design-system.md`'s "The bare token is the fill" table and the "Run / Status Indicator Colors" `css` block. `mkdocs build --strict`.

## Sequencing

After #1252, which edits the same section of `docs/design-system.md` (the added paragraph on which family a surface belongs to) and would conflict textually, not logically. Parallel-safe against everything else; touches no store, no query, no engine surface.

## Acceptance criteria

- [ ] `status-color-tokens.test.ts` derives its families from `index.css` and so covers every token that has a `-text` pair, with the non-empty floor kept.
- [ ] `docs/design-system.md` no longer claims `--status-warning` has no `-text` variant, and its ratios for that row are recomputed and reproducible from the declared values.
- [ ] The `css` block in the status section lists every `--status-*` token `index.css` declares, or says why it lists a subset.

athrvk:
## Description

**Root cause.** `status-color-tokens.test.ts` enforced its rule over a list of families written by hand, and the stylesheet outgrew it. `index.css` declares ten families with a `-text` pair; the list held seven, so `status-redirect`, `status-no-response` and `status-warning` had a real foreground token and no guard over their bare one. The comment explaining the last one's absence - "it has no `-text` variant, because bare already measures 17.72 / 15.78" - and its twin at `docs/design-system.md:298` were describing a stylesheet that no longer exists: `--status-warning-text` is declared in both themes (`index.css:342,652`) and ten source files already use it.

**Approach.** Derive the list instead of restating it. A family is a token that declares *both* halves as colours in the base `:root` palette. That rule answers the `primary` question on its own: `--primary-text: var(--primary)` is an alias, so `primary` declares no second, more readable colour and there is nothing to ban - and on the day it is given one, it joins the guard rather than waiting to be noticed. The parse itself moves to `lib/css-tokens.testkit.ts`, because `design-system-doc.test.ts` was already doing it and a second copy would not receive the first one's fixes (its value regex has already had to learn that the doc aligns values in columns).

**Rejected alternative:** deriving from every `--token:` declaration in the file, which is the obvious reading of "any `--X` whose `--X-text` is also declared". It pulls in `primary`, because the graphite accent scheme overrides `--primary-text` with a literal (`index.css:546`) - and `text-primary` is a correct, deliberate utility in 47 places, so the guard would have gone red on landing. The base palette is where a token's own definition lives; scheme blocks re-tune it.

The amber ratios were then recomputed rather than restated, and `--status-warning` turns out to be the family's worst case, not its safe one.

### The measurements

WCAG 1.4.3 from the declared HSL triples. The method reproduces every existing row of the doc's table within **0.04**, which is what makes the eleventh row's disagreement a finding rather than a different ruler:

| family | doc light bare / `-text` | recomputed | doc dark bare / `-text` | recomputed |
|---|---|---|---|---|
| `destructive` | 4.87 / 5.48 | 4.87 / 5.49 | 1.73 / 5.40 | 1.73 / 5.36 |
| `success` | 3.33 / 5.71 | 3.35 / 5.68 | 7.46 / 8.81 | 7.45 / 8.80 |
| `warning` | 2.13 / 5.46 | 2.14 / 5.46 | 8.13 / 9.81 | 8.09 / 9.79 |
| `status-success` | 2.30 / 5.71 | 2.30 / 5.68 | 7.53 / 8.81 | 7.53 / 8.80 |
| `status-error` | 3.78 / 5.88 | 3.78 / 5.87 | 4.59 / 5.85 | 4.57 / 5.83 |
| `status-stopped` | 2.79 / 5.73 | 2.78 / 5.74 | 6.23 / 7.40 | 6.21 / 7.37 |
| `status-running` | 3.64 / 5.99 | 3.63 / 5.98 | 4.77 / 6.75 | 4.77 / 6.76 |
| **`status-warning`** | **17.72 / 17.72** | **3.98 / 5.46** | **15.78 / 15.78** | **4.35 / 9.81** |

**No other row is restated**, per the issue's instruction: the deltas above are all <= 0.04 and are the browser-versus-formula difference, not drift, so the hand-measured figures stand as they are.

Two independent corroborations that 3.98 is right and 17.72 is not: `docs/design-system.md:227` already records **3.98** on the card for `--series-warning`, which is the identical triple `38 92% 36%`; and `index.css:319-324` records amber-500 at 2.14, the neighbouring value. 17.72 is not a ratio either amber produces against any surface `index.css` declares.

The new row's `-text` columns repeat the `warning` row's figures deliberately: `--warning-text` and `--status-warning-text` hold the *same* HSL in both themes, and a second number for one colour is the drift #1252 spent a diff removing.

`status-redirect` and `status-no-response` gain rows too, since the guard now covers them and the table is what justifies the ban per family. All three fail 4.5 in **both** themes - a value tuned to read as a dot on either card is by construction near the middle - which contradicted the paragraph under the table claiming `destructive` is the only family failing in dark. That paragraph is rewritten rather than left standing next to a table that now disproves it.

### Deliberate deviations from the issue

- **The comment's copy of the ratio table is deleted, not extended.** The issue asks only that the "no `-text` variant" sentence go. But the header comment carried a second copy of the doc's eight-row table, and keeping it meant either adding a duplicate `status-warning` row or leaving the copy knowingly short. It now cites `docs/design-system.md` as the one record - the same "one number for one fact" move #1252 made.
- **The caption above the table said "the two worst also fail the 3.0 floor for icons".** Four of its values are under 3.0 (1.73, 2.13, 2.30, 2.79), and were before this change. Corrected in passing because it is the sentence introducing the table whose numbers this diff changes, and `app/CLAUDE.md:184` asks for exactly that: "if you change a value, read the sentence around it."
- **The issue names `app/src/components/ui/design-system-doc.test.ts`.** The file is at `app/src/design-system-doc.test.ts`.
- **The issue lists nine `text-status-warning-text` call sites**; there are ten - `app/src/constants/http-status.ts:108` is the tenth. No code change was needed at any of them.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`cd app && pnpm test` - **591 files, 6511 tests, all passing.** The base commit (`a332853`) was run first and is green at 591 / 6504, so there are no pre-existing failures to report and the delta is exactly the 7 cases added: 4 named cases plus 3 more per-family scans, the families having gone from 7 to 10. `pnpm type-check`, `pnpm lint` and `pnpm format:check` are clean. `mkdocs build --strict` builds.

Three mutation checks, each run by breaking the thing and confirming the red:

1. **The newly covered family is really covered.** Changing `http-status.ts:108` from `text-status-warning-text` to the bare `text-status-warning` reds "uses no bare `text-status-warning`, which is the fill token", naming the file and line. Under the old hardcoded list that edit was silent.
2. **The scan is really scanning.** Making `bareForeground`'s pattern unmatchable reds "reports the bare token where a source uses one" and nothing else - so that fixture case is load-bearing rather than decorative.
3. **The new doc values are really checked.** Changing one digit of `--status-redirect` in the doc's `css` block reds `design-system-doc.test.ts` with `doc says "258 60% 61%", index.css has "258 60% 62%"`.

The derivation itself is pinned by two cases over a fixture stylesheet rather than the real one, so they state the rule instead of the current answer: a family with both halves derives, the same family with the `-text` declaration removed drops out (rather than throwing), and a family whose `-text` only aliases the bare token never appears.

Two floors are kept, both of which fail loudly rather than silently: the `-text` utility count (> 50, now built from the derived families rather than a second hardcoded alternation baked into a regex literal), and a new assertion that the derived list is non-empty - an empty one would make every per-family case *vanish* rather than fail, which is worse than the drift being fixed.

Not run: the engine build or `ctest`. Nothing under `engine/` is touched, per the root `CLAUDE.md` rule on scaling verification to what a change could break.

## Checklist

Docs in the same commit: `docs/design-system.md` gains the corrected amber row and two more, a rewritten paragraph under the table, a replacement for the "it has no `-text` variant" claim that records the four-surface measurement the issue asked for (3.98 card / 3.63 background / 3.38 muted / 3.54 on its own `/10` tint for the bare token; 5.46 / 4.98 / 4.64 / 4.86 for `-text`), the three missing tokens in the `css` block, and a line saying the guard derives its families. `docs/app/COMPONENTS.md` and `app/CLAUDE.md:125-128` describe the three-token split and the guard's rule, both unchanged by this diff. No heading renamed, no page, link or nav entry added or moved.

The new module reads only `app/src/index.css`, which is inside `app/`, so it needs no entry in `routed-inputs.testkit.ts`; `routed-inputs.test.ts` passes unedited.

## Related Issues

Closes #1253

https://claude.ai/code/session_01P8wDS1rZKDGtQ9vKewHZRR

---
_Generated by [Claude Code](https://claude.ai/code/session_01P8wDS1rZKDGtQ9vKewHZRR)_

### DIFF
--- a/app/src/components/ui/status-color-tokens.test.ts
+++ b/app/src/components/ui/status-color-tokens.test.ts
@@ -17,7 +17,9 @@
  * Using the fill token as a foreground fails AA. The measured ratios live in
  * `docs/design-system.md` ("The bare token is the fill"), which is the one
  * record of them - every family fails at one end or the other, `destructive`
- * in dark, `status-warning` in both, the rest in light. That spread is the tell
+ * in dark, the three mode-consistent HTTP-status indicators (`status-warning`,
+ * `status-redirect`, `status-no-response`) in both, the rest in light. That
+ * spread is the tell
  * that this is not a dark-mode bug: it is the fill token standing in for the
  * foreground one, and which mode it breaks in just depends on where that fill
  * sits relative to the surface behind it.
--- a/app/src/components/ui/status-color-tokens.test.ts
+++ b/app/src/components/ui/status-color-tokens.test.ts
@@ -14,22 +14,13 @@
  * the readable *foreground*, darkened in light mode and lightened in dark so it
  * clears AA as small text or a small glyph.
  *
- * Using the fill token as a foreground fails AA. Measured live on `bg-card`
- * (contrast ratio; 4.5 is the floor for normal text, 3.0 for icons):
- *
- *   family           light bare   light -text   dark bare   dark -text
- *   destructive         4.87         5.48         1.73         5.40
- *   success             3.33         5.71         7.46         8.81
- *   warning             2.13         5.46         8.13         9.81
- *   status-success      2.30         5.71         7.53         8.81
- *   status-error        3.78         5.88         4.59         5.85
- *   status-stopped      2.79         5.73         6.23         7.40
- *   status-running      3.64         5.99         4.77         6.75
- *
- * `destructive` is the only family that fails in dark; the rest fail in light.
- * That inversion is the tell that this is not a dark-mode bug - it is the fill
- * token standing in for the foreground one, which shows up in whichever mode the
- * fill happens to sit closest to the surface behind it.
+ * Using the fill token as a foreground fails AA. The measured ratios live in
+ * `docs/design-system.md` ("The bare token is the fill"), which is the one
+ * record of them - every family fails at one end or the other, `destructive`
+ * in dark, `status-warning` in both, the rest in light. That spread is the tell
+ * that this is not a dark-mode bug: it is the fill token standing in for the
+ * foreground one, and which mode it breaks in just depends on where that fill
+ * sits relative to the surface behind it.
  *
  * So `text-<family>` is banned, including `hover:`/`focus:` prefixes and the
  * `/NN` opacity forms - a faded `-text` is no safer than a solid bare token, and
@@ -37,9 +28,13 @@
  * `bg-*`, `border-*` and `*-foreground` (the paired foreground for a solid fill)
  * are all correct uses of the bare token and are left alone.
  *
- * `status-warning` is deliberately absent from the list below: it has no `-text`
- * variant, because bare already measures 17.72 / 15.78 - banning it would only
- * force churn towards a token that does not exist.
+ * **The families are read out of `index.css`, not listed here.** A hand-written
+ * list held seven while the stylesheet had grown to ten, so `status-redirect`,
+ * `status-no-response` and `status-warning` were guarded by nothing - and the
+ * comment that had explained `status-warning`'s absence ("it has no `-text`
+ * variant") was describing a stylesheet that no longer existed (#1253). A
+ * family is now whatever declares both halves in the base palette, so the next
+ * token pair arrives guarded.
  *
  * **Known blind spot: inline styles.** This scans for the `text-<family>` class,
  * so `style={{ color: "hsl(var(--warning))" }}` walks straight past it.
@@ -57,6 +52,7 @@
  */
 
 import { describe, it, expect } from "vitest";
+import { familiesWithTextPair, indexCss } from "@/lib/css-tokens.testkit";
 
 const sources: Record<string, string> = import.meta.glob("/src/**/*.{ts,tsx}", {
 	query: "?raw",
@@ -64,16 +60,8 @@ const sources: Record<string, string> = import.meta.glob("/src/**/*.{ts,tsx}", {
 	eager: true,
 });
 
-/** Families with a `-text` foreground variant. One line each to add another. */
-const FAMILIES = [
-	"destructive",
-	"success",
-	"warning",
-	"status-success",
-	"status-error",
-	"status-stopped",
-	"status-running",
-];
+/** Families with a `-text` foreground variant, as `index.css` declares them. */
+const FAMILIES = familiesWithTextPair();
 
 /**
  * `text-<family>` not followed by `-` or a word character, so `-text` and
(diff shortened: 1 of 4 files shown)


athrvk:
Two adversarial review passes over the diff, one against #1253's acceptance criteria and one against the repo's conventions. Both criteria sets came back met; three findings survived my own verification and are fixed in `c12403a` and `2a82ba7`:

- **The guard's own comment disagreed with the doc it defers to.** It said `status-warning` is the one family failing in both themes. Three do: `status-redirect` measures 4.34 light / 3.99 dark and `status-no-response` 3.98 / 4.34, which is what makes the three of them a set and what the rewritten paragraph in the doc already said. Recomputed and corrected.
- **Two figures for one measurement, undisclosed.** `design-system.md:268` records the light-amber floor on `--muted` as 4.63; the new paragraph cross-references it and computes 4.64. Both are right - the first was measured in a browser, which quantises to 8-bit channels, the second computed from the declared triple - so the page now says so once, rather than leaving the next reader to "correct" one into the other. This is the one place the recomputation disagrees with something already on the page, so it is stated rather than restated.
- **The derivation's own doc comment opened by listing the families it derives.** Accurate today, and exactly the sentence that would not fail CI on the day an eleventh family lands. Removed.

Discarded, with the reason: a finding that the amber row's four-surface measurement covers light mode only. True, and deliberate - the dark values are 9.79 / 11.25 / 8.53 / 8.78, all far clear of both floors, so the four-surface breakdown exists to show where the *light* value lands against the 4.5 line. The section it sits in states the same asymmetry (`-text` tokens are solved against `--muted`/`--accent` in light and `--card` in dark), so a dark breakdown would be four numbers making one point that the table's dark column already makes.

Filed rather than folded in: **#1264** - `status-token-contrast.test.ts` and `color-schemes.test.ts` still hand-roll the `readFileSync` of `index.css` that `css-tokens.testkit.ts` now exports as `indexCss`. Not two copies of a parser (both need their own per-theme parsing), but two more copies of the workaround for vitest stubbing a CSS import to `""`, which is the part that carries the trap.

---

Issue #1139, opened 2026-08-29
Mock server: route through a pre-routing handler, not a `.*` regex catch-all


athrvk:
Deferred from the vcpkg baseline bump in the 0.24.0 release PR (athrvk/vayu#1138), which moved cpp-httplib 0.53.0 -> 0.53.1.

## What changed upstream

0.53.1 guards regex route matching against stack exhaustion. `RegexMatcher::match` now refuses before ever calling `std::regex_match`:

```cpp
inline bool RegexMatcher::match(Request &request) const {
  request.path_params.clear();
  // See CPPHTTPLIB_REGEX_ROUTE_PATH_MAX_LENGTH: an overlong path is treated as
  // a non-match rather than risking a stack overflow in std::regex_match.
  if (request.path.length() > CPPHTTPLIB_REGEX_ROUTE_PATH_MAX_LENGTH) {
    return false;
  }
  return std::regex_match(request.path, request.matches, regex_);
}
```

`CPPHTTPLIB_REGEX_ROUTE_PATH_MAX_LENGTH` defaults to 256. The upstream comment is explicit that libstdc++'s backtracking recurses roughly once per matched character for quantified patterns "such as `(.*)`".

## Why it lands on us

`start_mock_server` in `src/http/routes/mock_server.cpp` registers a regex catch-all on every method:

```cpp
// Every method cpp-httplib routes, on every path: the route table decides
// what answers, not httplib's own matcher.
svr.Get (".*", responder); // also serves HEAD
svr.Post (".*", responder);
...
```

`.*` carries regex metacharacters, so it stays a `RegexMatcher` and is subject to the limit. Two consequences:

- **Net win, unasked for.** Before 0.53.1 that catch-all was exactly the vulnerable shape: a long enough request path to a running mock could exhaust the thread stack. The bump closes that.
- **Behaviour change to record.** A mocked request path longer than 256 characters now fails to match and falls through to a 404, without reaching `serve_mock_request` - so Vayu's own route table never gets to decide. The comment quoted above ("on every path") stops being true.

Raising `CPPHTTPLIB_REGEX_ROUTE_PATH_MAX_LENGTH` is the wrong lever: our pattern is the quantified one the limit exists to protect, so a higher ceiling buys back the stack-exhaustion risk.

## What the engine's own routes do

[...]

PR #1276, opened 2026-09-02
fix(engine): an inbox records a path past the regex route limit

### LINKED ISSUE #1140: Inbox drops a webhook whose path is over 256 characters
Found while fixing #1139. The inbox listener carries the same `.*` regex catch-all the mock server did, so it inherits the same defect from cpp-httplib 0.53.1 - but it **cannot take the same fix**.

## The defect

`src/http/routes/inbox.cpp`:

```cpp
// Every method cpp-httplib will route, on every path. CONNECT, TRACE and
// PRI are the remainder and are not routable - a webhook is never one.
httplib::Server& svr = inbox->listener.server ();
svr.Get (".*", capture); // also serves HEAD
svr.Post (".*", capture);
...
```

Since 0.53.1, `RegexMatcher::match` returns false for any path longer than `CPPHTTPLIB_REGEX_ROUTE_PATH_MAX_LENGTH` (256) before calling `std::regex_match`. So a webhook delivered to a path over 256 characters gets httplib's bare 404 and **is never recorded** - no row, nothing in the inbox list, no indication anything arrived.

This is more reachable than the mock-server instance was. An inbox exists to receive whatever an external sender posts, and long paths are ordinary in that traffic: signed callback URLs, per-delivery tokens in the path, deeply nested tenant routes. The comment above the registrations - "on every path" - is exactly what stopped being true.

## Why #1139's fix does not port

The mock server moved to `set_pre_routing_handler`, which cpp-httplib runs at the top of `Server::routing`, **before** the request body is read. That was free for the mock because `serve_mock_request` reads only `req.method` and `req.path`.

The inbox's `capture` is the opposite - recording the body is its entire purpose:

```cpp
capture_row.body_bytes     = static_cast<int64_t> (req.body.size ());
capture_row.body_truncated = capture_row.body_bytes > raw->limits.max_body_bytes;
capture_row.body           = capture_row.body_truncated ?
       req.body.substr (0, static_cast<size_t> (raw->limits.max_body_bytes)) :
       req.body;
```

Under a pre-routing handler `req.body` is empty, so every capture would silently record a zero-byte body. That is a worse bug than the one being fixed. The inbox also calls `set_payload_max_length (constants::inbox::MAX_PAYLOAD_BYTES)`, whose enforcement happens during the body read the pre-routing handler precedes.

## Options, none obviously right

- **Raise `CPPHTTPLIB_REGEX_ROUTE_PATH_MAX_LENGTH`** via a compile definition. It is a global compile-time macro, so this also raises it for the management server's `([^/]+)` routes, and `[^/]+` backtracks per character the same way. It buys back exactly the stack exhaustion the limit exists to prevent.
- **A non-regex catch-all.** `make_matcher` sends any metacharacter-free pattern to `PathParamsMatcher`, but a `:param` captures only as far as the next `/`, so there is no multi-segment catch-all spelling that avoids regex.
- **Pre-routing plus a manual body read.** The handler has no stream access, so this means reaching past the public API.
- **A content-reader handler** registered per method, which sees the body as it streams - but registration still goes through the same matcher, so the path limit still applies first.

Worth deciding deliberately rather than in a release PR, which is why this is filed rather than folded into athrvk/vayu#1138 alongside the mock fix.

A test belongs with whichever option wins: POST to an inbox on a 300-character path, assert a row is recorded with the body intact.

athrvk:
## Description

The inbox registered its catch-all as a `.*` route per method, and since cpp-httplib 0.53.1 a regex route refuses any path longer than `CPPHTTPLIB_REGEX_ROUTE_PATH_MAX_LENGTH` (256) before it reaches `std::regex_match`. A webhook delivered to a longer path drew httplib's own 404 and was **never recorded** - no row, nothing in the inbox list, no sign anything arrived. Long paths are ordinary in the traffic an inbox exists to receive.

**Root cause and approach.** The inbox was asking cpp-httplib's *route matcher* to decide whether a request is captured, and it has nothing to route: one handler answers every method and path. So rather than ask the matcher a question whose answer is always yes, the listener routes every request **as** one literal path and the handlers register on that literal - which `make_matcher` answers with a `PathParamsMatcher`, a string compare with no length limit. No regex runs against an inbox's traffic at all now, which removes the bound rather than raising it. The one hook cpp-httplib gives a mutable `Request` ahead of routing is `process_request`'s `setup_request` callback, and overriding `process_and_close_socket` reaches it - a *private* virtual, which is no obstacle (access control governs who may call a virtual, not who may override it) and is how cpp-httplib's own `SSLServer` extends the same function. A `static_assert` on the pinned version makes a cpp-httplib bump re-read that override rather than drift past it silently. `ManagedListener` already owned its server through a `unique_ptr`, so it gained a constructor taking one its owner built. The capture then reads the path off `req.target` through httplib's public `decode_path_component`, which is exactly what its own request-line parser puts in `Request::path` - the sibling of the `query_of` the row already used.

**The alternatives, and why each was rejected** - the issue asked for this decision to be made deliberately:

- **Raise `CPPHTTPLIB_REGEX_ROUTE_PATH_MAX_LENGTH`.** Wrong twice over. It is a compile-time macro read by an `inline` function, so a value for one translation unit alone is an ODR violation; and raising it globally hands `std::regex_match` up to httplib's 8192-byte request-target cap of attacker-controlled path - the per-character recursion the limit exists to bound - on the one listener that may bind past loopback. `mock_server.cpp` already recorded the same objection for #1139.
- **A non-regex catch-all.** `PathParamsMatcher`'s marker is `/:` and a param captures only as far as the next `/`, so there is no multi-segment spelling. Registering one route per segment depth bounds capture by depth instead of by length and still drops the deep case silently - a stopgap, not a fix.
- **Port #1139's pre-routing handler.** It runs before the body is read, and recording the body is what a capture is for; it also takes a `const Request&`, so it cannot rewrite the path either.
- **Content-reader handlers.** Registration goes through the same matcher, so the length limit applies first.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- New in `engine/tests/inbox_test.cpp`, beside the existing catch-all test:
  - `RecordsAPathPastTheRegexRouteLimitWithItsBody` - a 307-character path, asserting the row exists with its path, query and **body** intact, plus a body-less GET on the same path.
  - `RecordsThePathDecodedAsHttplibItselfWould` - a percent-encoded path, asserting the stored path is still the decoded one and the query is still the raw string, since the path now comes off the target.
- New in `engine/tests/managed_listener_test.cpp`, covering the new constructor: a null server is refused at construction, and a server handed over is the one routes register on and requests are served by.
- **Mutation-checked**: putting the `.*` registrations and the plain `ManagedListener` back reddens `RecordsAPathPastTheRegexRouteLimitWithItsBody` (nothing is stored) while the existing `RecordsWhatArrivedOnAnyMethodAndPath` stays green - the defect reproduces exactly where the issue says it does.
- `python build.py -e -t` then `ctest --preset linux-dev` - **2865 tests, all passing** (2861 on the base commit, so +4 and nothing else moved).
- `clang-format-19 --dry-run -Werror` clean on all five changed C++ files; `clang-tidy-19 -p engine/build` clean on `inbox.cpp`, `managed_listener.cpp`, `inbox_test.cpp` and `managed_listener_test.cpp`.
- `mkdocs build --strict` was **not** run - mkdocs is not installed in this environment. The docs edits are prose inside existing sections, with no new link, heading or nav entry.

## Checklist

Docs in the same commits: `docs/engine/api-reference.md`'s Webhook Inbox "Bounds" paragraph - the request-target cap is named as the third non-setting bound, and "any path" is stated as literal below it; and `docs/engine/architecture.md`'s shared-listener paragraph, whose enumeration of what stays per-manager gains the `httplib::Server` a manager may now hand it. One doc was checked and needs no change: `docs/engine/db-schema.md`'s `path` column still reads "Decoded path, no query" - the derivation moved, the value did not.

## Related Issues
Closes #1140

### DIFF
--- a/docs/engine/api-reference.md
+++ b/docs/engine/api-reference.md
@@ -3249,12 +3249,19 @@ started, and a running listener keeps what it was started with:
 | `inboxMaxCaptures` | 500 | 1 - 10000 | Captures retained per inbox, oldest evicted first. Also the ceiling on one `limit` of the capture list |
 | `inboxLivePollIntervalMs` | 250 | 25 - 5000 | How often a watched inbox checks for new captures - the delay between a webhook landing and its event |
 
-Two are **not** settings, deliberately. A request over **8 MiB** is refused at
+Three are **not** settings, deliberately. A request over **8 MiB** is refused at
 the transport with a `413` and recorded nowhere: that bounds what an
 unauthenticated remote caller can make the engine buffer, which is not the local
 user's preference to spend. The canned response's **`delayMs` is capped at
 30000**: it holds a listener thread for its whole duration and a stop waits on
-that join, so it bounds how long a stop can be made to take.
+that join, so it bounds how long a stop can be made to take. The **request
+target** - path and query together - is capped at **8192 bytes** by the
+transport, past which it answers `414`.
+
+Below that cap the path's length does not matter, and "any path" is meant
+literally: an inbox routes every request to its capture without matching the
+path against anything, so a signed callback URL, a per-delivery token in the
+path or a deeply nested tenant route is recorded like any other (issue #1140).
 
 ### POST /inbox/start
 
--- a/docs/engine/architecture.md
+++ b/docs/engine/architecture.md
@@ -154,7 +154,11 @@ alive at that point. All four managers run that lifecycle - bind, wait for the a
 join, release - through one shared `ManagedListener`
 (`engine/include/vayu/http/managed_listener.hpp`), so a fix to it reaches every listener rather than
 one of four copies; what stays per-manager is the route registration, the error each bind failure
-answers with, and the OAuth attempt TTL sweep. The mock server is the one built on the helper rather
+answers with, and the OAuth attempt TTL sweep. Since #1140 a manager may hand it the
+`httplib::Server` those routes are registered on, too: the inbox hands it a subclass that routes
+every request as one path, so that a path too long for cpp-httplib's regex matcher still reaches its
+capture, and the other three take the plain server the listener builds for itself. The mock server
+is the one built on the helper rather
 than ported to it: #505 extracted it before this listener existed, which was the whole point of
 extracting it when the third one landed.
 
--- a/engine/include/vayu/http/managed_listener.hpp
+++ b/engine/include/vayu/http/managed_listener.hpp
@@ -55,6 +55,16 @@ namespace vayu::http {
 class ManagedListener {
     public:
     ManagedListener ();
+    /**
+     * Take a server the owner built, for a listener that needs one a plain
+     * `httplib::Server` cannot be - the webhook inbox routes every request as
+     * one path through a `Server` subclass, so that a path cpp-httplib's regex
+     * matcher would refuse still reaches its handler (see routes/inbox.cpp).
+     *
+     * Throws `std::invalid_argument` on a null server rather than letting it
+     * read back as a listener `stop()` has already released.
+     */
+    explicit ManagedListener (std::unique_ptr<httplib::Server> server);
     /// Stops and joins, so a manager destroying its records never leaks a thread.
     ~ManagedListener ();
 
--- a/engine/src/http/managed_listener.cpp
+++ b/engine/src/http/managed_listener.cpp
@@ -55,7 +55,16 @@ ClaimRegistry& registry () {
 } // namespace
 
 ManagedListener::ManagedListener ()
-: server_ (std::make_unique<httplib::Server> ()) {
+: ManagedListener (std::make_unique<httplib::Server> ()) {
+}
+
+ManagedListener::ManagedListener (std::unique_ptr<httplib::Server> server)
+: server_ (std::move (server)) {
+    if (!server_) {
+        throw std::invalid_argument (
+        "ManagedListener was handed no server - every accessor would then read "
+        "as a listener stop() had already released");
+    }
 }
 
 ManagedListener::~ManagedListener () {
--- a/engine/src/http/routes/inbox.cpp
+++ b/engine/src/http/routes/inbox.cpp
@@ -123,9 +123,11 @@ std::string path_of (const std::string& target) {
  * recording the body is what a capture is for; it also takes a `const
  * Request&`, so it cannot rewrite the path either. `process_request`'s
  * `setup_request` callback is the one hook that sees a mutable `Request` ahead
- * of routing, and overriding `process_and_close_socket` - the protected virtual
- * cpp-httplib's own `SSLServer` extends the same way - is how a caller reaches
- * it.
+ * of routing, and overriding `process_and_close_socket` reaches it. That one is
+ * *private* in `httplib::Server`, which is no obstacle: access control governs
+ * who may call a virtual, not who may override it, and it is how cpp-httplib's
+ * own `SSLServer` extends the same function. `process_request` and the members
+ * the override reads are protected.
  */
 class AnyPathServer final : public httplib::Server {
     public:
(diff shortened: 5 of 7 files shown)


athrvk (commented):
Reviewed at head `46d8ba4` against cpp-httplib 0.53.1's source and with a local build.

**Verdict: correct, and the best available within the library's public API.** One addition asked for below; nothing blocking.

What I checked:

- **The override is a faithful mirror.** `AnyPathServer::process_and_close_socket` matches `Server::process_and_close_socket` in `httplib.h` at 0.53.1 line for line, the only difference being the `setup_request` lambda where the base passes `nullptr`. The `static_assert` pins that to the release it copies.
- **The path derivation reproduces httplib's own.** `parse_request_line` erases any `#fragment` from `req.target`, splits at `?`, and sets `req.path = decode_path_component(lhs)`; the 414 check on the 8192-byte target runs before `setup_request`, which runs before routing. `path_of(req.target)` yields the same string httplib would have put in `req.path`, and the decoded-path test pins it.
- **No length bound remains.** A pattern with no `:param` is `request.path == pattern()` in `PathParamsMatcher::match`, so `"/"` matches every rewritten request and no regex runs on inbox traffic.
- **The rejected alternatives are rightly rejected, plus one the PR did not list.** `set_error_handler` looked like a public catch-all that sees every unmatched request, but 0.53.1 reads a regular handler's body only inside `dispatch_request` after a route has matched, so an unmatched request reaches the error handler and the post-routing handler with an empty body. That closes the last public-API route; `setup_request` via the protected virtual is the only door, and it is the one `SSLServer` uses.
- **Local run:** `python build.py -e -t` on this head, then `ctest --preset linux-dev -R Inbox`: 34/34 passed, the two new cases included. I did not re-run the mutation check; the test's comment names it and the PR body records it. Engine CI: macOS green, Ubuntu and Windows in progress at the time of writing. Base is behind master (`87d0d99`) with no conflict.

**One ask before merge.** The mirror plus the version assert is a stopgap by construction: every cpp-httplib bump now stops the build until someone re-reads `process_and_close_socket`, and the baseline-bump cadence is deliberately routine. That is acceptable only with an exit on record. I filed #1283 for it (upstream a public request-setup hook or a matcher-free default handler, retire the override on the first release that carries one). Please reference #1283 from the `AnyPathServer` comment or the assert message, so the next person who trips the assert finds the plan rather than re-deriving it.

Documented trap worth keeping in mind, no change needed: after the rewrite every hook on this server sees `req.path == "/"`, so a logger or error handler added to the inbox listener later must read the target, as `capture` now does. The comment beside `capture_row.path` says so.

---

athrvk:
Done in `26aede0`. Both places now point at #1283: the `AnyPathServer` comment gains a closing paragraph saying the mirror has an exit on record rather than by default, and the assert message itself ends with "or retire the mirror entirely if the release carries a public request-setup hook (issue #1283)" - so the plan is in front of whoever the build stops, not only in the class comment above it. Comment and assert-message text only; no behaviour change.

Thanks for the `set_error_handler` check - that was the one public-API door I traced but did not write down. Confirming your reading independently: in 0.53.1 `routing()` reads a regular handler's body inside `dispatch_request`, after the matcher has said yes, so an unmatched request reaches both the error handler and the post-routing handler with `req.body` empty. Worth having on the record in #1283, since it is the same reason the pre-routing option fails and the next person will likely reach for it too.

One correction to my own PR body, already fixed in `fe95bba`: it described `process_and_close_socket` as a *protected* virtual. It is **private** in `httplib::Server`, and `SSLServer`'s override is private too - which changes nothing, since access control governs who may call a virtual rather than who may override it, but the assert asks for a re-read on every bump and "protected" would have sent that reader looking for the wrong thing. The PR body and the code comment both say private now, and the override is declared private to mirror the two it follows.

---

Issue #604, opened 2026-08-14
app(variables): a runtime token's tooltip can never be shown - the overlay is pointer-events:none - and the script panel still paints {{data.*}} red


athrvk:
Two findings from #592 / #603, both adjacent to that fix and deliberately left out of its diff.

## 1. `RuntimeToken`'s tooltip is unreachable

`VariableInput` paints its overlay with `pointerEvents: "none"` so clicks fall through to the transparent input underneath and place the caret (`components/shared/VariableInput/index.tsx`, the overlay `div`). `EditableVariable`'s wrapper re-enables them explicitly - `style={{ pointerEvents: "auto" }}`, commented "Make variable tokens clickable" - because it has a popover to open.

`RuntimeToken`'s wrapper does not. So neither `{{$guid}}` nor `{{data.email}}` receives a pointer event, and the Radix tooltip on it - the token's entire content, since there is nothing to click - can never open. This predates #603: the generator token has been in this state since it was added, and the data token inherits it.

**What makes it awkward** rather than a one-line fix: `pointerEvents: "auto"` alone makes the token swallow the click that used to place the caret, and `handleContainerClick` early-returns on anything inside `[data-variable-token]`, so clicking the token would do nothing at all. The fix is the pair - re-enable pointer events *and* have the token focus the input (or drop out of the early-return) so caret placement survives.

Test-wise this is the repo's "written but never read" class one more time: a tooltip in the markup that no test hovers and no user can reach. A behavioural test needs a real hover, so pin it with `userEvent.hover` rather than a source scan.

## 2. `ScriptPanel`'s "Referenced" chips paint a data column as broken

`referencedVariables` (`lib/referenced-variables.ts`) picks `{{name}}` out of a script, and both `ScriptPanel` and the collection `ScriptTab` chip each name with `variant={allVariables[varName] ? "secondary" : "destructive"}`. A `{{data.email}}` in a script string therefore still gets the red treatment #592 removed from the builder.

Narrower than #592 was: these chips offer no Create, so there is no trap - just the wrong colour, and only for authors who write a `{{data.*}}` template inside a script rather than using `pm.iterationData`. Left out of #603 because it needs its own decision (what does an informational chip look like in that row?) and #592's fix pathway enumerated the surfaces without it.

Worth deciding alongside it: the engine does not interpolate `{{}}` inside a script at all, so arguably *every* template chip in that row overstates what will happen.

## Sequencing

[...]

PR #1301, opened 2026-09-03
fix(app): warn when a scoped script read's own scope answers emptily and another scope holds the value

### LINKED ISSUE #1196: app(scripting): a scoped read of an empty, shadowed variable looks healthy in the editor - pm.collectionVariables.get returns '' while {{name}} resolves from the environment, and nothing warns
Owner-reported, hit in real use, verified at master `72bef0d`. A test script asserted `pm.collectionVariables.get("shop_domain")` against the value that `{{shop_domain}}` visibly resolves to in the URL bar - and got `''`, failing the test, with no cue anywhere in the app about why.

## What actually happened (engine verified CORRECT - do not "fix" it there)

`pm.collectionVariables.get` reads the collection scope's own chain, exactly as Postman does: `js_pm_scope_get` (`engine/src/runtime/script_engine.cpp:6269-6283`) returns `JS_UNDEFINED` when no enabled definition exists in the chain and the stored value - including an empty string - when one does (`lookup_variable`, `:6240-6252`, looks past disabled rows, stops at enabled ones). The failing get returned `''`, which proves an **enabled `shop_domain` row with an empty value exists at collection scope** (a common leftover of a Postman import's initial-value placeholders, or a created-then-never-filled row), while the environment holds the real value. `{{name}}` resolution uses full scope precedence, so the URL gets the environment's value; the scoped accessor honestly answers with the collection's empty one. Postman behaves identically; the unset-vs-empty distinction and scope precedence were verified Postman-compatible in the #996 program. **Engine is out of scope here.**

## The product gap

The trap is invisible at every surface the author looks at:

- The URL token's hover tooltip shows the winning value and its environment source - nothing hints a shadowed empty collection row exists underneath.
- The script editor's chips and completion know exactly which scope each accessor reads - `SCOPE_BY_ACCESSOR` maps `collectionVariables` to `"collection"` (`app/src/lib/script-variable-completion.ts:97-103`) and the referenced-variables machinery discriminates reads - yet a chip for `pm.collectionVariables.get("shop_domain")` paints exactly like a healthy read when that scope's own answer is empty/absent and a different scope holds the non-empty value.
- The variables popover does list origins and renders an empty one as the word "empty" (`app/src/components/ui/variable-popover.tsx:729`) - but only if the user thinks to open it on the token, which is precisely what someone debugging a script assertion does not know to do.

Result: a silently wrong test verdict and a debugging session that ends in the engine being suspected. The cost is real - this exact shape burned the owner today.

## Fix / solution pathway

Teach the script surfaces the cross-check they already have the data for. When a name is read through a single-scope accessor (environment/globals/collectionVariables per `SCOPE_BY_ACCESSOR`) and that scope's own enabled answer is empty or absent while a different scope defines the name non-empty:

1. Paint the chip in the existing warning tone (the tone ladder from the data-column chips work) with a tooltip stating the fact plainly, e.g. "empty at collection scope - the non-empty value comes from the environment; this read returns ''" (or "not defined at collection scope - returns undefined"). The referenced-variables `reads` discriminator plus the resolver's per-scope origins are the inputs; no new state needed.
2. Same signal in the completion detail for the accessor's name argument, so the trap is visible at typing time, not only after a failed run.
3. Optional third surface, judged by the implementer: the popover's origin list could badge the shadowed-empty row more loudly than the plain "empty" word when the winning value is non-empty - the two-line case that says "this is probably not what your script wants".

Keep the semantics untouched everywhere - this is paint and prose only. Do not add heuristics engine-side; do not warn on `pm.variables.get` (the merged read is the correct escape and returns the winning value).

## App changes

This IS the app change.

## Tests

`cd app && pnpm test` - the chip machinery has per-tone tests (`script-data-chip`-style): add cases for scoped-read-empty-with-non-empty-shadow (warning tone + tooltip text), scoped-read-defined (unchanged), `pm.variables.get` (never warns). Mutation-check by dropping the cross-check.

## Docs (same commit)

`docs/app/pm-api-compatibility.md` / `docs/engine/scripting.md` only if either documents the chip tones; otherwise none - verify and state. The scoped-accessor semantics are already documented correctly.

## Sequencing

Script-surface files (`referenced-variables`, chip rendering, completion provider) - disjoint from every open issue (#1194/#1195 touch MethodBadge and EditableVariable; #1164 the Dock; #1188/#1190/#1192 elsewhere). Parallel-safe.

## Acceptance criteria

- [ ] The reported scenario (enabled empty collection row + non-empty environment value + `pm.collectionVariables.get`) shows a warning-toned chip whose tooltip names both facts, test-pinned.
- [ ] Healthy scoped reads and merged reads are visually unchanged.
- [ ] Engine behavior byte-identical (no engine diff in the PR).

---

athrvk:
## Description

**Root cause.** `referencedVariables` recorded `reads: "scope"` but threw away *which* scope - `PM_GET`'s accessor capture was consumed inline to pick a `PmRead` and never stored. Both script chip surfaces were left asking `allVariables[name]`, which answers about the **winner of the full ladder**. So an enabled, empty `shop_domain` row at collection scope painted the healthy accent chip while `pm.collectionVariables.get("shop_domain")` honestly returned `''` and `{{shop_domain}}` in the URL bar resolved the environment's value: a failing assertion with nothing on screen connecting it to its cause. A Postman import's initial-value placeholder is the usual way in.

**Approach.** Each reference now carries its accessor's single `scope`, and `describeScopedRead` joins it against the resolver's existing `getVariableOrigins` - already provided on both surfaces and read by neither, the quiet half of this repo's written-but-never-read defect. It warns when the accessor's own scope answers emptily or not at all **while the ladder's winner is a different scope holding a value**. Amber, not destructive: the read works and the name resolves, it just does not resolve *there* - the line `describeColumnToken` already draws, and the paint #604 took off this row.

The winner is the test rather than "some other scope has a value", and the two part company exactly where the read's own scope wins: an empty collection row beside a non-empty global is the collection's answer *and* the merged answer, and an enabled environment row wins whatever it holds. Warning on those would be amber on healthy scripts. (My first cut got this wrong in both directions; `f6a3bd2` fixes it and pins the three cases where the rules diverge.)

**Alternative rejected.** Deriving the warning in each panel from `getAllVariables` plus a scope guess. That is the copy-per-surface shape `describeColumnReference` exists to prevent, and the winner map cannot express "empty here, non-empty there" at all. One decision in `lib/`, two consumers.

Tooltips print sources (`environment - Staging`), never values, because a shadowing definition may be a secret.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Locally on this branch: `pnpm test` **6655 passed / 603 files, 0 failed**; `pnpm type-check` clean; `pnpm lint` clean; `pnpm format:check` clean.

17 new cases. `referenced-variables.test.ts` (44 total) covers the reported scenario, the undefined-vs-empty split, disabled rows looked past, nearest-enabled-in-chain, secret masking, the three precedence cases above, and every read that must stay silent. `script-data-chip.test.tsx` and `ScriptTab.chips.test.tsx` each render the trap, a healthy scoped read that another scope still wins, and the merged read. Mutation-checked twice: neutering `describeScopedRead` reddens 5, dropping the precedence guards reddens 5.

Both chip suites mocked a resolver without `getVariableOrigins` and went red as soon as the panels read it; the harness now supplies origins. `script-panels.test.tsx` gets an empty list, since no name there can be the shadowed-empty read.

### One commit here is not this issue's work

`763cb7b` makes `app/CLAUDE.md` prettier-clean. `format:check` covers `md` and had been failing on that file since #1299, which predates this branch - `pr-tests.yml` triggers on `pull_request` only, so nothing re-checks `master` and the break sat there blocking every PR's `App quality checks`, and `CI gate` with it.

I had left it out at first, since `CLAUDE.md` says to format only files you touched, and flagged it for the repo owner in [this comment](https://github.com/athrvk/vayu/pull/1301#issuecomment-5520805510) instead. Folded in at the owner's direction.

It was not only a red tick. The quality job runs `Type-check frontend` *after* the format step, so **a failing format check had been skipping the type-check on every pull request** - the repo has had no CI type-check since #1299 merged. This commit restores it. The change itself is emphasis markers only (`_x_` where the page wrote `*x*`), 7 lines, no prose touched.

## Checklist

- **Fix pathway item 2 (the same signal in the completion `detail`) is not implemented.** It has no reachable case today. `useScriptVariableCompletionProvider` filters candidates by `info.scope === context.scope` where `info.scope` is the *winner's* scope, so under `pm.collectionVariables.get(` the trapped name - won by the environment - is never offered at all; and the configuration that *is* offered (the empty definition winning) is not the trap, because then `{{name}}` resolves to `""` as well. Making it reachable means changing which names a list offers, which is a semantics change the issue explicitly scopes out ("paint and prose only"). Filed as #1302, which reuses `describeScopedRead` rather than adding a second cross-check.
- **Fix pathway item 3 (badging the shadowed row in the popover) is not implemented.** The issue marks it optional and implementer's judgment; the popover already renders `empty` and a shadowed-by list, so the reader who opens it is not the one being trapped.
- `PM_ACCESSORS` merges this module's own two accessor tables into one. `SCOPE_BY_ACCESSOR` in `script-variable-completion.ts` still keys the same five names; it predates this and has not diverged, and folding them together belongs to the next change that touches both rather than to a paint fix.
- `mkdocs build --strict` could not run locally (mkdocs is not installed in this environment). The docs edit is prose inside an existing section - no new heading, link or nav entry - so there is nothing for `--strict` to reject.

## Related Issues
Closes #1196

### DIFF
--- a/app/CLAUDE.md
+++ b/app/CLAUDE.md
@@ -82,7 +82,7 @@ Forgetting it fails loudly (`document is not defined`), never silently.
 sets no `testTimeout`, so every case gets the default 5s. A case that builds
 thousands of rows and renders them spends most of that budget idle and crosses
 it under machine or CI-shard load, failing over the environment rather than
-the code (#846, #1280). Ask what the fixture size is *for* before reaching for
+the code (#846, #1280). Ask what the fixture size is _for_ before reaching for
 a timeout: a property such as "past the grid's first window" needs a fixture
 one step past that window, not hundreds of rows. Only where a case genuinely
 needs the large fixture, give it an explicit `it(name, { timeout: N }, fn)`
@@ -101,7 +101,7 @@ missing from the other.** `responseFromExecuteResult` (live `/execute` body)
 and `responseFromRunResult` (a stored trace, from History) both build
 `ResponseState` and neither knows about the other. When you add a field to
 one, add it to both and assert they agree (`validation-funnels.test.ts` is the
-shape: drive both with the same input, compare). The engine stores the *same*
+shape: drive both with the same input, compare). The engine stores the _same_
 object it returned rather than letting each side derive its own.
 
 **Before measuring or changing a class, `rg` for it in the components.** A
@@ -110,7 +110,7 @@ conclusion about a combination the app never renders is not a finding.
 ## UI rules (enforced by tests - breaking one fails CI)
 
 - **Status colours have three tokens:** `--status-*` (dot/icon/tint),
-  `--status-*-text` (when the colour *is* the text), `--status-*-fill` (solid
+  `--status-*-text` (when the colour _is_ the text), `--status-*-fill` (solid
   chip under a white label). Using the bare fill as a foreground is the most
   common colour bug here. → `status-color-tokens.test.ts`
 - **`--primary` vs `--primary-fill`:** `--primary` is text/ring/chart and
@@ -174,9 +174,9 @@ conclusion about a combination the app never renders is not a finding.
   → `focus-indicator.test.ts`, `icon-button-labels.test.tsx`,
   `a11y-suppressions.test.ts`, plus the lint itself
 
-## Borders: what a rule sits *on*, never what it is
+## Borders: what a rule sits _on_, never what it is
 
-**A border is invisible or not depending on what it sits *on*, never on what it
+**A border is invisible or not depending on what it sits _on_, never on what it
 is.** `--border` is tuned for the canvas (1.14) and is the same colour as
 `--card` in dark (1.00), so a rule inside a card is simply absent. A card's own
 outline on `border-border` is correct, because that edge faces the canvas; both
@@ -195,7 +195,7 @@ vs 1.16, and the pair inverts in light), which is why sunken uses an alpha of
 
 The mistake is **enumerable, not impossible**: a `border-rule` under no
 declared surface silently falls back to the invisible default. So guard the
-*declarations* (`surface-rule.test.tsx`, `ImportModal.surface-rule.test.tsx`);
+_declarations_ (`surface-rule.test.tsx`, `ImportModal.surface-rule.test.tsx`);
 asserting `border-rule` is present proves nothing. The surface classes are in
 use across the tree; a new surface declares one rather than picking a token.
 On an element whose primitive already sets a background utility
@@ -232,6 +232,6 @@ later on an unrelated change. Those lists and the `app_doc_fixtures` /
 by `routed-inputs.test.ts`, which fails if either side changes alone. Take the
 path from the registry rather than spelling it again in the guard.
 
-Module READMEs carry the *why* for their feature and are easy to miss:
+Module READMEs carry the _why_ for their feature and are easy to miss:
 `app/src/modules/README.md`, plus one each for `welcome/`, `request-builder/`
 and `dashboard/`.
--- a/app/src/lib/referenced-variables.test.ts
+++ b/app/src/lib/referenced-variables.test.ts
@@ -17,16 +17,23 @@
  */
 
 import { describe, it, expect } from "vitest";
-import { describeColumnReference, referencedVariables } from "./referenced-variables";
-import type { DataContractScope } from "@/types/domain";
+import {
+	describeColumnReference,
+	describeScopedRead,
+	referencedVariables,
+} from "./referenced-variables";
+import type { DataContractScope, VariableOrigin } from "@/types/domain";
 
 /** The names alone, for the cases that are about extraction and not syntax. */
 const names = (script: string) => referencedVariables(script).map((r) => r.name);
 
 describe("the pm API", () => {
 	it.each(["environment", "globals", "collectionVariables"])("reads pm.%s.get", (bucket) => {
+		const scope = { environment: "environment", globals: "global", collectionVariables: "collection" }[
+			bucket
+		];
 		expect(referencedVariables(`pm.${bucket}.get("token")`)).toEqual([
-			{ name: "token", via: "pm", reads: "scope" },
+			{ name: "token", via: "pm", reads: "scope", scope },
 		]);
 	});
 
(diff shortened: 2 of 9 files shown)


athrvk:
**`App quality checks` carries a failure that is not this PR's: `app/CLAUDE.md` is not prettier-clean on `master`.**

`pnpm format:check` (run from `app/`, glob `**/*.{...,md,...}`) fails on `CLAUDE.md` alone. I verified it reproduces on the base commit `8b9f7f2` with this branch's changes stashed - the file is byte-identical to `master` here and this PR never touches it. Prettier wants `_emphasis_` where the page writes `*emphasis*`, in at least 5 places:

```
85:  the code (#846, #1280). Ask what the fixture size is *for* before reaching for
104: shape: drive both with the same input, compare). The engine stores the *same*
113:   `--status-*-text` (when the colour *is* the text), `--status-*-fill` (solid
177: ## Borders: what a rule sits *on*, never what it is
179: **A border is invisible or not depending on what it sits *on*, never on what it
```

It landed with #1299 and was not caught because `pr-tests.yml` triggers on `pull_request` only - there is no run of it on `master`, so nothing re-checks the merge commit. On this PR the job stopped at `pnpm lint` before reaching `Check frontend formatting`, which is why the earlier red pointed at my own file instead; that one is fixed and pushed.

I am not fixing it here. It is one file outside this diff, and `CLAUDE.md`'s rule is explicit: *"Never run prettier or `eslint --fix` repo-wide; format only files you touched that were clean before."* The whole fix is one command, which I have deliberately left for you:

```bash
cd app && pnpm exec prettier --write CLAUDE.md
```

Worth knowing it blocks every PR's `App quality checks`, not just this one, until it lands somewhere.

Everything this PR is responsible for is green locally: `pnpm test` 6652 passed / 603 files, `pnpm lint` clean, `pnpm type-check` clean, and `format:check` clean on every file except the one above.

---

Issue #1237, opened 2026-09-02
app(tree): ArrowLeft from any row but the first lands on its first sibling, not on the parent - `parentItem` stops at the group's first treeitem


athrvk:
Found while wiring the variables sidebar onto the same hook (#1217, PR #1236). Pre-existing, in the shared primitive, so it affects the collection tree as well - not introduced by that PR, and deliberately left out of its diff.

## Problem

`parentItem` (`app/src/modules/collections/useRovingTreeFocus.ts:106-115`) walks up from the focused row and, at each ancestor, returns the first `[role="treeitem"]` that ancestor contains, skipping only the case where that treeitem is the row itself:

```ts
let node = current.parentElement;
while (node && node !== tree && tree?.contains(node)) {
    const first = node.querySelector<HTMLElement>(ITEM);
    if (first && first !== current) return first;
    node = node.parentElement;
}
```

For the **first** row of a group that is correct: its own wrapper contains only itself, so the walk continues to the section wrapper and finds the parent. For every **later** row the first ancestor it reaches is the `role="group"` wrapper, whose first treeitem is a *sibling* - so the walk stops there and returns a sibling as the parent.

Verified at master `9951d0e` with two throwaway tests (neither kept):

- The collection tree's DOM shape (folder row + `role="group"` of per-row wrappers, the hook's own fixture shape at `useRovingTreeFocus.test.tsx:56-78`): ArrowLeft on the second request focuses **"Req one"**, the first request, rather than the folder.
- The variables sidebar at PR #1236's head: ArrowLeft on "Production", the second environment row, focuses **"Staging"** rather than the "Environments" header.

Consequences, in order of how visible they are:

1. **ArrowLeft does not leave a list.** From the third row of a folder it moves to the first row of that same folder; a second press then does reach the parent, so the key looks like it "hesitates" rather than like it is broken. `docs/design-system.md` Tree Navigation states "Left collapses then moves to the parent", which is true only for a group's first row.
2. **`*` groups by the same key.** The expand-siblings case (`useRovingTreeFocus.ts:255-260`) calls two rows siblings when `parentItem` returns the same element, so for rows past the first it partitions by "first sibling" instead of by parent. Two rows in the same group still agree, so this is likely benign today - worth an assertion rather than an assumption.
3. **ArrowRight's first-child check is unaffected**: it compares `parentItem(next) === current` for `next` being the row *after* the parent, which is that group's first row - the case `parentItem` gets right.

[...]

Issue #1224, opened 2026-09-01
app(graphql): the schema controls change home with the explorer's state and sit on the wrong side - Refresh moves between the explorer header and an invisible hover button, the open/close toggle is on the right edge of a pane that opens on the left, and a third Refresh stands in the context bar


athrvk:
Owner-reported: the refresh-schema button moves, and the toggle for a pane that opens on the left sits on the right. Verified at master `5677ed5`. (Corrected after the owner pointed out the first version kept the toggle on the wrong side.)

## Problem

The same three facts about one subject (schema status, refresh it, show or hide the browser) are laid out differently depending on whether the explorer pane is open, and the toggle is on the far side from the pane it controls.

| State | Query header (right side) | Explorer header (pane is the **left** panel) | Context bar (if open) |
|---|---|---|---|
| Explorer **open** | operation picker only (`GraphQLBody.tsx:728-741`: the chip renders only when `!showExplorer`) | status badge, Show-descriptions, **Refresh**, Hide-schema (`graphql-explorer/SchemaExplorer.tsx:223-249`) | **Refresh** (`context-bar/GraphQLSection.tsx:95-101`, standing) |
| Explorer **closed** | `SchemaChip` on the far right: status + "Browse schema", plus a **Refresh that is `opacity-0` until hover or focus** (`GraphQLBody.tsx:284-300`) | not rendered | **Refresh** (standing) |

Four consequences, each verified in code:

1. **The toggle is on the wrong side.** The explorer is the first `ResizablePanel` in a horizontal group (`GraphQLBody.tsx:835-866`), so it opens to the **left** of the query pane. The control that opens it is the chip at the **right** edge of the Query header, and both icons say the opposite side: `PanelRightOpen` on the chip (`GraphQLBody.tsx:25,280`) and `PanelRightClose` on the explorer's Hide button (`SchemaExplorer.tsx:41,246`). A user reaches for the left edge, where the pane will appear, and finds nothing; the icon then points the wrong way.
2. **Refresh relocates.** Open: second of three icons in the left pane's header. Closed: an invisible button after a chip on the far right of the Query header. Same action, two positions, one undiscoverable except by hovering the right spot.
[...]

PR #1315, opened 2026-09-03
fix(app): the GraphQL schema explorer's search and insertions lead somewhere

### LINKED ISSUE #1226: app(graphql): the schema explorer's search flattens the tree into unlabelled duplicates with a dead chevron, refuses insertions silently, writes a bare fragment for a type click, and dumps whole descriptions - it does not behave the way a user expects a schema browser to
Owner-reported against a large production Admin API schema: searching for a field name showed three identical `accessScopes` rows with no categories, clicking the highlighted row did nothing, and clicking the type row wrote a fragment instead of a query. Verified at master `5677ed5`; every mechanism below is in the code, the runtime behaviour was reproduced by reading the paths the screenshots exercise. (The 400 the owner saw alongside this was the request going out as a GET with a body - #1228 - not the fragment; corrected from the first version.)

## Problem

Files: `app/src/modules/request-builder/components/RequestTabs/panels/body/graphql-explorer/SchemaExplorer.tsx`, `app/src/lib/graphql/schema-tree.ts`, `app/src/lib/graphql/insert-skeleton.ts`, `GraphQLBody.tsx`.

1. **Search results lose both the categories and the owner.** In search mode `rows` are `searchSchema(...)` matches mapped to `depth: 0` (`SchemaExplorer.tsx:146-152`); the Query / Mutation / Types headings are gone. `ExplorerRow` renders name, signature and description (`:401-470`) and never the node's `ownerTypeName` or `branch`, although the node carries both (`schema-tree.ts:83,86`). Three fields named `accessScopes` on three different types are three indistinguishable rows; the user cannot tell `App.accessScopes` from `AppInstallation.accessScopes` or which is reachable from Query.
2. **The chevron in search results is dead.** Expandable nodes render the toggle (`:405-421`) and it writes `toggleExpanded` (`:316`), but search-mode rows are derived from the index and ignore `expanded` (`:146-159`). The icon flips to "expanded" and nothing appears beneath it. That is the "clicking the highlighted row does nothing" in the report: the row shows `aria-expanded=true` with no children.
3. **A refused insertion is invisible.** A field found by search under a non-root type has `rootPath: null` (`buildSearchIndex`, `schema-tree.ts:392-399`). With an empty document, `insertField` finds no enclosing selection set that owns it and returns a refusal (`insert-skeleton.ts:524-560`). `handleExplorerInsert` then calls `announce(reason)` (`GraphQLBody.tsx:646-649`), which writes to a `sr-only` live region (`:380,873`). The docs call this "refused out loud" (`COMPONENTS.md:294`); for a sighted user it is silence. Clicking the row does nothing visible.
4. **A type click writes a fragment, not a query.** `insertionForNode` maps `kind === "type"` to `insertFragment` (`insert-skeleton.ts:742,634-665`), which appends `fragment XFields on X { …scalars }` to the document. On an empty document the result parses (the module's contract, `COMPONENTS.md:294`) but contains no operation, so it cannot execute. Nothing in the pane says "this inserts a fragment"; a user's model is "click the thing, get a query that returns it", which is how GraphiQL's explorer plugin and Postman's GraphQL client behave. The explorer also offers no way to find which root fields return a type.
5. **A name match still dumps the whole description.** `full = showDescription || descriptionStart >= 0` (`SchemaExplorer.tsx:362-371`), and `searchSchema` reports `descriptionStart` for every match regardless of which tier matched (`schema-tree.ts:468-470`). A type whose name matched *and* whose 1,100-character description happens to mention the term is drawn in full, filling the pane above the results the user wanted. The rule was written for tier-3 (description-only) matches and applies to all three.

Severity: medium-high for a feature that exists to be clicked. Items 2 and 3 make the primary interaction look broken; item 4 produces a document that cannot run; items 1 and 5 make results unreadable on a real schema.

## Fix / solution pathway

Follow the browsers users already know (GraphiQL's docs explorer and its explorer plugin, Postman's GraphQL schema panel). Solved problem, do not invent grammar.

1. **Grouped, owner-labelled results.** Render search matches under the same three headings as the tree (`branch` is on every node), each field row prefixed with its owner in muted text (`App.accessScopes`), root-reachable rows first within a group. The index already carries everything; this is a render change plus a small group-by in the row model.
2. **No dead affordance.** In search mode either omit the toggle, or make it *reveal in tree* (clear the search, expand the path, focus the row - GraphiQL's jump-to). Recommend reveal-in-tree for rows with a `rootPath`, omit otherwise. Never render a toggle whose click changes nothing on screen.
3. **Refusals and outcomes visible.** Keep the live region and add a visible notice in the pane (an `InfoChip` or the existing `Toaster` primitive - not a hand-rolled banner) naming what happened and what to do. Better, remove the most common refusal: for a field with `rootPath: null`, compute the root paths that reach its owner type (`Query.app → App`, `Query.currentAppInstallation → AppInstallation`), insert the shortest as a query skeleton when there is one, and offer a chooser when there are several. Record the decision; the minimum acceptable is a visible refusal with a next step.
4. **Type rows lead somewhere.** Under a type row, list *Returned by* (root fields whose result type resolves to it) as children, and make activating the type row insert via the shortest root path, the same as 3. Keep the fragment as an explicit secondary action (a row-menu item, or the primary action only when the cursor is inside an existing operation), and tighten the insertion contract from "the document still parses" to "the document still parses **and is executable**" - a click must not produce a fragment-only document. Add the corresponding refusal for the cases where that cannot hold.
5. **Full description only when the description is why the row matched.** Pass the match tier (or set `descriptionStart` to -1 when the name or signature matched) so the full-text rule applies to tier-3 matches alone.

## App changes

This is the app change; no engine or MCP involvement.

## Tests

`cd app && pnpm test -- SchemaExplorer schema-tree insert-skeleton explorer-store GraphQLBody.explorer` and `pnpm type-check`, `pnpm lint`. Existing coverage to extend: `SchemaExplorer.test.tsx:149-200` (search), `insert-skeleton.test.ts` (refusals), `schema-tree.test.ts`.

- Search renders group headings and owner prefixes; two same-named fields are distinguishable by text (mutation check: drop the owner prefix and the test fails).
- No `[data-tree-toggle]` in search results for path-less rows; for rooted rows, clicking it clears the search and focuses the revealed tree row.
- A refused insertion produces visible text in the pane (assert on the visible node, not the `sr-only` region).
- Activating a search-found non-root field on an empty document inserts an executable operation reaching it (or the chooser appears when several paths exist); `insert-skeleton.test.ts` gains "never produces a document with no operation" over the fixture's type rows.
- Type row shows *Returned by* children; activation inserts a query, fragment insertion is reachable only through its explicit action.
- Name-matched row with a term also in its description renders one line; description-only match renders in full (both directions pinned).

## Docs (same commit)

`docs/app/COMPONENTS.md:273` (search behaviour, "flat result list", full-description rule), `:293` (`schema-tree.ts`: search index and grouping), `:294` (`insert-skeleton.ts`: the contract becomes parses-and-executable; type activation semantics; visible refusals). `mkdocs build --strict`.

## Sequencing

Same files as #1224 but different regions (this: rows, search, insertion; #1224: the headers). Either order; small merge. Independent of #1220 (Monaco variable tokens register providers, not explorer code) and of #1228 (method and transport). Parallel-safe otherwise.

## Acceptance criteria

- [ ] Search results are grouped by Query / Mutation / Types and every field row names its owner type; no result row carries a toggle that does nothing.
- [ ] Every activation produces a visible outcome: an insertion, an "already selected" reveal, a chooser, or a refusal with a next step shown in the pane.
- [ ] No activation can leave the document without an executable operation; a type row exposes the root fields that return it and inserts a query by default.
- [ ] A row is drawn with its full description only when the description is the reason it matched, or when Show-descriptions is on.
- [ ] `docs/app/COMPONENTS.md` describes the grouped search, the visible refusals and the executable-document contract.

---

### LINKED ISSUE #1318: app(graphql): a fragment can only be added to a selection set of exactly its type, so the interface and union case - the one fragments exist for - is refused
Folded into PR #1315 rather than deferred - kept here as the record of what it covered.

Falls out of #1226, which made `insertFragment` write a fragment definition **and its spread** as one edit, because a definition nothing spreads is rejected with the whole request (`Fragment "X" is never used`, a rule every spec-conformant server enforces).

## Problem

`app/src/lib/graphql/insert-skeleton.ts`, `insertFragment`. The spread's host was found by **exact** type-name match against the enclosing selection sets. That is conservative and always legal - a `fragment F on X` spread inside an `X` selection set is valid without qualification - but it refuses the case named fragments are most often written for: **a concrete type inside an interface or union selection.**

```graphql
query {
  node(id: "1") {   # selection set's type is Node
    id
    # a fragment on Post belongs exactly here, and was refused
  }
}
```

## Fix, as landed

The host test is fragment-spread **compatibility**, computed by `doTypesOverlap` - the same predicate `PossibleFragmentSpreadsRule` validates with, so it cannot write a spread the server then rejects. An exact match still wins over an overlapping one: that is the set the user is in, not merely one the spread is legal in. With no host anywhere on the cursor's chain it still refuses, visibly.

Rejected: writing an **inline** fragment (`... on Post { … }`) instead. Smaller and needs no name, but a type row's fragment is the reusable one; inline fragments are what a *field* click would want, and mixing the two behind one activation is the ambiguity #1226 set out to remove.

## Acceptance criteria

- [x] A fragment on a concrete type can be added from inside a selection set of an interface it implements, or a union it belongs to, and the result validates.
- [x] Compatibility is decided by `doTypesOverlap`, not a hand-rolled subtype walk.
- [x] A type with no overlapping set on the cursor's chain is still refused visibly, unchanged.
- [x] `docs/app/COMPONENTS.md` and `docs/app/graphql.md` state the widened rule.

Tests in `insert-skeleton.test.ts`: the interface case, the union case, exact-match preference over overlap, and the unchanged refusal. Mutation-checked - restoring the equality test reddens the first two.

---

### LINKED ISSUE #1319: app(graphql): "Returned by" reads only a field's declared result type, so on a Relay-style schema the types reached through Node have no route at all
Folded into PR #1315 rather than deferred - kept here as the record of what it covered.

Falls out of #1226, which added `rootFieldsReturning` - the index behind a type row's **Returned by** children and behind the query a type row inserts.

## Problem

`app/src/lib/graphql/schema-tree.ts`, `returnedByIndex` keyed a root field under the type it *declares* and nothing else. So `Query.node(id: ID!): Node` was keyed under `Node` alone, and on a schema built the Relay way - one `node` field, every concrete type reached through it - `rootFieldsReturning(schema, "Post")` was empty even though `Query.node` reaches a `Post` on every real request. The consequences were all the ones the reverse index was added to remove: no **Returned by** children on the `Post` row, a type row falling through to the fragment path, and a searched `Post.title` refused exactly as it was before #1226.

The one-hop stop is deliberate and documented; this was not that. `Query.node → Post` is still one hop - it was a gap in what one hop was taken to mean.

## Fix, as landed

Each root field is keyed under its declared type **and** under the concrete types that type resolves to (`schema.getPossibleTypes`), one hop still. An entry reached that way carries a `narrowTo`, which the route's `FieldStep` passes on, because the query it inserts is a different shape:

- `renderSteps` writes the inline fragment the narrowing needs - `node(id: $id) { ... on Post { title } }` - and selects the *narrowed* type's fields, not the interface's. This is not the "probably compatible" guessing the module's header refuses; it is the narrowing without which the selection would not be legal at all.
- A **Returned by** row says where it leads (`(id: ID!): Node → Post`) and expands into that type, so it does not read as returning a `Node` and open `Node`'s one field.
- A route that returns the type outright outranks one that narrows to it, ahead of the deprecated-last rule, so a schema of direct returns is byte-identical to before.

Rejected: keying only the interface and letting the type row show its interfaces instead. Smaller, and it hands the user a second hop to make by hand - which is the browsing the explorer exists to do for them.

## Acceptance criteria

- [x] A concrete type reachable only through an interface or union root field lists that route under **Returned by**, labelled so routes are distinguishable.
- [x] Activating such a row, or a field found by search beneath it, inserts an operation that narrows with an inline fragment and passes `validate`.
- [x] A direct route always outranks a narrowed one, and a schema with only direct routes is unchanged.
- [x] `docs/app/COMPONENTS.md` and `docs/app/graphql.md` state what one hop now covers.

Tests in `schema-tree.test.ts` and `insert-skeleton.test.ts`; the narrowing is mutation-checked - dropping the inline fragment reddens with "Cannot query field title on type Node".

---

athrvk:
## Description

The explorer's five defects share one root cause: **the pane throws away what it already knows.** The search index carries every row's `branch` and `ownerTypeName` and the renderer drew neither; `searchSchema` knew which of three haystacks matched and reported only offsets; the schema knows which root fields return a `Post` and nothing asked it. So three indistinguishable `accessScopes` rows, a chevron that flipped with nothing under it, a click whose only answer went to an `sr-only` region, and a type click that left a fragment no Send can run. The fix surfaces what is already there rather than adding machinery: report the match tier beside the offsets, group results under the tree's own headings with the owner named on each row, reconstruct a search row's tree address from the same id builders the tree uses, and materialise one reverse index - which root fields answer with a given type - so a type row inserts the query that reaches it and a path-less field borrows that route instead of refusing.

**Rejected: keeping the fragment as a type row's default with a "this inserts a fragment" hint.** A hint on an unrunnable document is still an unrunnable document, and the user's model ("click the thing, get a query that returns it") is the one GraphiQL and Postman already teach.

Per issue item 5, the description-tier rule now asks the **tier** rather than the offset: both offsets are reported whatever the tier, so `Query.search` - named `search`, described "Search across users and posts." - answered the old question yes and drew its whole paragraph over the results being read.

### What review changed

Three findings were verified and fixed rather than filed:

- **An appended fragment invalidated the whole request.** A definition nothing spreads fails `Fragment "X" is never used`, which rejects the entire document - including the operation the user already had and did not touch. `insertFragment` now writes the definition and its `...XFields` spread as one edit, and refuses where the spread cannot go. This also retired a parses-only `hasOperation` check that could not tell a mid-edit document from one with no operation.
- **Reveal could silently render nothing.** The growing window resets to one step whenever the row count changes, and leaving the results for the tree changes it - so on a schema with more types than a step the row was expanded in the store and never rendered, which is the dead click this pane exists to end.
- **The refusal notice outlived its request.** Nothing above `GraphQLBody` is keyed on the request, so switching tabs left a notice about one request's schema describing another's. It now carries the request it was written for, compared during render.

Two findings were discarded with reason: the `Callout` severity being `info` rather than `warning` - these refusals explain, they do not warn, and a warning-toned banner on an ordinary click on a subscription row over-signals; and the group heading using `EYEBROW_CLASS` rather than the `Eyebrow` component - the class *is* the one shared definition, and the component takes no `role`/`data-*`, which this needs for `role="presentation"` inside the tree.

### Two follow-ups folded in rather than deferred

Both were filed, then implemented here at the owner's direction; the issues are closed with the record of what they covered.

- **#1318 - a fragment now lands where it is legal, not only where the name matches.** The spread's host was found by type-name equality, which refused the one case named fragments exist for: a fragment on a concrete type inside a selection of the interface it implements. The host is decided by `doTypesOverlap`, the predicate `PossibleFragmentSpreadsRule` validates with, so this cannot write a spread the server then rejects; an exact match still wins, since that is the set the user is in rather than merely one the spread is legal in.
- **#1319 - a route through an interface or union is still a route.** The reverse index keyed a root field only under the type it declares, so on a Relay-shaped schema - one `node` field, every concrete type behind it - `Post` had no route at all. Each root field is now keyed under the members of what it declares too, still one hop, carrying the narrowing the route needs; `renderSteps` writes that narrowing as the inline fragment without which the selection would not be legal (`node(id: $id) { ... on Post { title } }`), and the **Returned by** row says where it leads (`: Node → Post`) and expands into that type. A route that returns the type outright outranks one that narrows, so a schema of direct returns is byte-identical to before.

### Remaining deliberate deviations from the issue spec

- **The "chooser" is the tree, not a menu.** The issue allows "a chooser when there are several" routes. Rather than a new popup, a type row's **Returned by** children list every root field that answers with it, so the alternatives are browsable and activatable in place; a click takes the best route deterministically (direct before narrowed, then deprecated last, then declaration order).
- **There is no row-menu action for fragments.** The issue offered "a row-menu item, or the primary action only when the cursor is inside an existing operation". A third target inside a 24px row is the composite-row hit-area trap this pane's own docs name, so the primary click reaches the fragment path exactly when no root field returns the type.
- **No engine build or ctest run.** The diff is `app/` and `docs/` only - the issue states "no engine or MCP involvement" - so nothing in the engine could go from green to red.

### Merged with master

`master` moved under this branch (#1316 landed, reworking the same explorer header). Merged rather than rebased, and resolved by hand in two files:

- `SchemaExplorer.tsx` - master retired the pane's own two-row header, its `onRefresh`/`onClose` props and its Refresh, moving them to the Query header (#1224). Master's single-row header is kept as-is; this branch's `notice` prop and the reveal-window reset are carried onto it.
- `docs/app/COMPONENTS.md` - the explorer paragraph was rewritten on both sides. Master's new opening (and its `aria-level` parentage note, #1237) is kept, with this branch's four additions re-applied into it: the grouped and owner-labelled search, the description-tier rule, the no-toggle / show-in-tree control, and the visible refusal.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check` on the merge commit - **6909 tests in 611 files, all passing**; type-check, ESLint and Prettier clean. `mkdocs build --strict -f .github/mkdocs.yml` builds clean. No pre-existing failures on the base commit.

New coverage, mutation-checked (each was confirmed to redden with the fix reverted, then restored):

| Area | Pins |
|---|---|
| `schema-tree.test.ts` | routes to a type (direct before narrowed, deprecated last, subscriptions excluded, selectable types only); routes through an interface and a union, and a direct-only schema left exactly as it was; the **Returned by** container, its children, and a narrowed row's label, target type and path; branch grouping without reshuffling the ranking; the tier a match landed in, including `tier === "name"` while `descriptionStart === 0`; `treeLocationOf` driven against `childNodes` itself, so a reconstructed address no row answers to fails |
| `insert-skeleton.test.ts` | a type row inserts a query not a fragment, uses a mutation when that is the route, skips the deprecated route; a search-found non-root field reaches through its owner's route and matches the Types-branch result exactly; the inline fragment a narrowed route needs, and that it selects the narrowed type's fields not the interface's; the fragment spread into exact, interface and union hosts with exact preferred, and the whole document validating each time; refusals for no route, no host, and a mid-edit document; "never leaves the document without an operation" over every fixture type row, with a non-empty floor on what it scanned |
| `SchemaExplorer.test.tsx` | group headings in the tree's order; three `id` rows distinguished by `data-tree-owner`; the owner kept out of `data-tree-name` so highlighting still marks the name; no toggle and no `aria-expanded` on a result; reveal clears the search and focuses the revealed row; reveal on a 250-type schema still renders the row (the growing-window floor); a container row opening on activation; a name-matched row with the term in its description staying clipped |
| `GraphQLBody.explorer.test.tsx` | the refusal asserted on the visible `Callout`, not the live region; the notice cleared by the next landing activation |
| `explorer-store.test.ts` | `revealPath` unions rather than toggles, and spends the search term |

## Checklist

## Related Issues
Closes #1226
Closes #1318
Closes #1319


athrvk:
**CI status, so the red check on this PR is not read as the current state.**

`PR CI - run tests` ran **once**, on the first commit (`60c846e`), and has not run since. The three pushes after it produced no `synchronize` run - only `Apply path labels` fires, so the head commit has no test result at all rather than a passing one. That looks like the usual "a PR opened by a token does not re-trigger `pull_request` workflows on that token's own pushes"; `pr-tests.yml` has only a `pull_request:` trigger, so I have no `workflow_dispatch` to start one, and re-running the existing run would re-run it against `60c846e`, not against the head.

**What failed on `60c846e`, and why it is not this PR's current state.** One test, on every platform:

```
src/modules/request-builder/components/RequestTabs/panels/body/GraphQLBody.explorer.test.tsx:489
  what the pane says out loud > inserts a type as a fragment
  expected 'mutation CreatePost($input: CreatePos…' to contain 'fragment PostFields on Post'
```

`60c846e` is the model-layer half of the change: it made a type row insert the query that returns the type, while that test still pinned the old fragment behaviour it was replacing. The test was rewritten in the very next commit (`c11651a`) as *"inserts a type as the operation that returns it, not a fragment"*, with the mutation check stated in the body. Nothing else failed in that run.

**Local verification on the head commit** (`917b277`): `pnpm test` is **6071 passing in 550 files**, and `pnpm type-check`, `pnpm lint` and `pnpm format:check` are clean; `mkdocs build --strict` builds the two edited pages. The diff is `app/` and `docs/` only, so no engine job is affected.

Nothing here needs a fix pushed - the failure is a superseded commit's. Re-running CI on the head needs a push from an account whose events trigger the workflow, or one `Re-run all jobs` from the Actions tab against the current head.

---

athrvk:
Closing the loop on the comment above - it is now out of date, and nothing needs doing by hand.

`master` had moved under this branch (#1316, which reworked the same explorer header) and GitHub had marked the PR `dirty`. I merged `master` in and resolved the two conflicts by hand:

- **`SchemaExplorer.tsx`** - master retired the pane's own two-row header, its `onRefresh`/`onClose` props and its Refresh, moving all three to the Query header (#1224). Master's single-row header is kept as it stands; this branch's `notice` prop and the reveal-window reset are carried onto it.
- **`docs/app/COMPONENTS.md`** - the explorer paragraph was rewritten on both sides. Master's new opening and its `aria-level` parentage note (#1237) are kept, with this branch's four additions re-applied into it: the grouped and owner-labelled search, the description-tier rule, the show-in-tree control replacing the dead toggle, and the visible refusal.

Re-verified on the merge commit: **6909 tests in 611 files passing** (master brought 61 more test files with it), type-check, ESLint, Prettier and `mkdocs build --strict` all clean. All the cases this PR added still run.

The merge commit also did what the plain pushes did not - `PR CI - run tests` is queued against the head now, so the check will report on its own and no manual re-run is needed.

---

Issue #1323, opened 2026-09-03
app(request-builder): the Body and script editors are a fixed 320px/350px box inside a pane that has the whole window's height - the editor should fill the pane


athrvk:
## Problem

On the Body tab the editor stack (Query plus Variables for GraphQL, the single editor for JSON, XML, text and JSON-RPC) stops about a third of the way down the pane, with a drag handle under it and empty panel below. The request pane is side by side with the response pane, so it has the window's full height; the editor uses a fixed slice of it and leaves the rest dark. Verified on master `7f377096`.

The height is a constant in component state, not a fraction of the pane:

```ts
// BodyPanel.tsx:202-208
useResizable({ defaultSize: 320, min: 160, max: 800, direction: "vertical" });
```

`useResizable` holds the size in `useState` (`app/src/hooks/useResizable.ts:53`), so the drag is forgotten on every tab switch (Radix unmounts the inactive tab) and every request. The editor box is `style={{ height: editorHeight }}` (`BodyPanel.tsx:425` and `:465`) inside a `TabsContent` that is `flex-1 overflow-y-auto p-4` (`RequestTabs/index.tsx:122`), a scroll container that never needs to scroll for the editor because the editor never grows. The script panels have the same shape with a different constant: `height="350px"` in `ScriptPanel.tsx:276`, `height="320px"` in the collection's `ScriptTab.tsx:286`. `BodyPanel` is the only caller of `useResizable`.

The layout that results is a form field where the pane wanted a workspace: on a 1440px-tall window the body editor is 320px and the pane below it is empty.

## Fix pathway

1. **The editor fills the pane.** For the modes that host an editor (`json`, `text`, `xml`, `jsonrpc`, `graphql`), `BodyPanel`'s root becomes a flex column that fills `TabsContent`, and the editor box becomes `flex-1 min-h-0` with a floor of 160px. `TabsContent` for the Body and script tabs gets `flex flex-col` in addition to its scroll (the scroll still serves the form-data and urlencoded tables, which can be taller than the pane). The mode picker, the Content-Type notice and the Source/Resolved toggle keep their intrinsic height above the editor; the editor takes the remainder. The GraphQL stack's vertical `ResizablePanelGroup` already fills whatever height it is given (`GraphQLBody.tsx:748`, `className="h-full"`), so it needs nothing.
[...]

PR #1324, opened 2026-09-03
feat(scripts): snippets you can insert, instead of prose you retype

### LINKED ISSUE #1223: app(scripts): the Quick Reference under the script editor is an always-on wall of 14 lines and nine paragraphs that cannot be inserted, collapsed or hidden, while the editor itself is pinned to 350px - redesign as an insertable, collapsible snippets surface with one source
Owner-reported: the block under the Pre-request and Tests editors needs product thinking, not a polish pass. Verified at master `5677ed5`.

## Problem

`app/src/modules/request-builder/components/RequestTabs/panels/script/ScriptPanel.tsx:248-283` renders, under every script editor and on every visit:

- a `<pre>` of `config.quickReference` - **14 lines** for the pre-request variant and 7 for tests (`script/script-variants.tsx:77-92,178-188`; the code comment at `ScriptPanel.tsx:258` still says "six lines"), and
- a bulleted list of `config.notes` - **nine paragraphs, ~4,100 characters** for pre-request (`script-variants.tsx:93-170`), four for tests.

What that block is not:

- **Not insertable.** Nothing in `app/src` calls `executeEdits` or a snippet controller (verified: zero hits outside tests). A user reads a line and retypes it. Postman's Snippets panel, the product convention here, inserts at the cursor on click.
- **Not collapsible, not remembered.** No toggle, no store key. The collection-level script tab has its *own* second implementation - a collapsible four-card `QUICK_REF` grid, default hidden, component state only (`modules/collections/CollectionDetail/ScriptTab.tsx:60-65,71,265-296`) - with different content and a different design. Two hand-rolled copies of one concept, neither with the other's strengths.
- **Not the primary surface.** The editor is `height="350px"` (`ScriptPanel.tsx:249`) inside a panel that is `flex-1 overflow-y-auto` (`RequestTabs/index.tsx:122`), so on a tall window the editor stays 350px and the reference gets the rest. `BodyPanel.tsx:423` and both GraphQL panes use `height="100%"`. The one place a user writes code is the one panel where the editor does not grow.
- **Mostly a duplicate.** The rule-level notes already live where an author meets them: the docs (`docs/engine/scripting.md:966` "add refuses a name that is already there"; the auth-order, case-sensitivity, `sendRequest` bounds and `pm.info` facts are all in `docs/engine/scripting.md` and `docs/app/pm-api-compatibility.md`, which the "Scripting docs" link opens) and the editor's own IntelliSense: the engine's completion table carries `detail` and `documentation` for every `pm.*` member and 12 `KIND_SNIPPET` templates (`engine/src/http/routes/scripting.cpp:2014-2020`, e.g. "Test: Status code"), which Monaco surfaces through `useScriptCompletionProvider.ts:42-100` and the generated typedefs (#1209) surface as hover docs. The panel restates a subset of that in prose the user cannot act on.
- **Content-pinned by tests as prose.** `script-panels.test.tsx:233-256` asserts substrings of the notes (`lower-cases every key`, `case-insensitive`, `add.*throws`, `load tests do not run pre-request scripts`, `writing to it does nothing`) and `:296-312` pins "three sunken slabs". Those facts are worth pinning; pinning them as panel prose is what makes the wall load-bearing.

Severity: medium, UX. Nothing is broken; the panel spends its space on the wrong thing and offers no interaction.

## Fix / solution pathway

Product model, following the convention users already know (Postman: editor first, a snippets list beside it that inserts on click; VS Code: the reference is hover and completion, not a sidebar):

1. **Editor first.** `height="100%"` in a flex column like `BodyPanel`; the editor takes the panel. The "Names referenced" chips and inherited-scripts notice stay above it as today.
2. **One `ScriptSnippets` surface, collapsible and remembered.** A rail or a collapsible strip under the editor, default collapsed, with its state in `layout-store` beside `graphqlVariablesCollapsed` (`stores/layout-store.ts:65,121,154`) - the persisted-collapse precedent `docs/app/COMPONENTS.md:271` documents. The whole header is the control (the composite-row hit-area rule). Used by both `ScriptPanel` and the collection `ScriptTab`; delete `QUICK_REF` and the `quickReference` arrays.
3. **Snippets come from the engine's completion table, and insert.** The `KIND_SNIPPET` entries are the single source (no third list in the renderer): add a `context` field (`pre`, `test`, `both`) to the table so each variant shows the right set, group them (Variables, Request, Response, Tests, Signing, Logging), and on click or Enter insert at the cursor through Monaco's snippet controller so placeholders (`${1:200}`) work exactly as they do from completion. Add the pre-request templates the current 14 lines imply and the table lacks (query upsert, URL assign, body raw, `sendRequest` callback shape, `pm.info`). `script_types_test.cpp` and `pm-api-compatibility.md` already gate the table; the `context` field is one more column they check.
4. **Rules go where they are read.** Each `notes` fact moves to the member's `documentation` in the completion table (so hover and completion show "throws on a name already there" on `add`, the case rule on header indexing, the bounds on `sendRequest`) or is confirmed present in the two docs pages. Keep at most one contextual line in the panel per variant - the one fact a user cannot discover from the editor: for pre-request, "load tests do not run pre-request scripts"; for tests, the `pm.response.to` vs `pm.expect` split - as an `InfoChip` or a single muted sentence, with the docs link beside it.
5. **Re-home the pins.** The content assertions in `script-panels.test.tsx` move to a test over the completion table's `documentation` (renderer side, via the fixture the docs-compile suite already reads, `engine/tests/fixtures/script-typedefs.d.ts`) so a runtime member losing its rule still fails a test. The "three slabs" test becomes "two" or is rewritten for the new surface.

Rejected: keeping the wall and adding a collapse toggle. A toggle on unread prose is still unread prose; the value is in insertion and in the editor's height.

## App changes

This is the app change; the engine change is the `context` field and a few added templates in `scripting.cpp` (plus regenerated `script-typedefs.d.ts` if templates carry docs - snippets are excluded from the typedefs, so likely none).

## Tests

`cd app && pnpm test -- script-panels ScriptTab ScriptSnippets layout-store` and `pnpm type-check`, `pnpm lint`; `cd engine && ctest --preset linux-dev -R "ScriptTypes|Scripting"`.

- `ScriptSnippets`: renders only the variant's `context` set; click and Enter insert at the cursor with placeholders (mock the editor and assert the snippet controller call); collapse state round-trips through `layout-store`; both hosts render the same component (a source-scan that `QUICK_REF` and `quickReference` no longer exist, with a non-empty floor on the files scanned).
- Table: every `KIND_SNIPPET` has a `context`; every pre-request template runs in the real engine under `eventName: prerequest` (the runtime `typeof` pattern in `script_types_test.cpp`).
- Re-homed facts: the five substrings above found in the table's `documentation` or the docs pages (mutation check: drop one from the table and the test fails).
- Editor fills the panel: `height="100%"` asserted the way `BodyPanel` tests do.

## Docs (same commit)

`docs/app/COMPONENTS.md:262` (says the quick reference is data in `script-variants.tsx`) and the collection detail section; `docs/design-system.md:421` (names "the script panels' Quick Reference blocks" as the fill-not-border example - substitute another example); `docs/engine/scripting.md` gains the snippet `context` column in the completion-table section; `mkdocs build --strict`.

## Sequencing

Files: `ScriptPanel.tsx`, `script-variants.tsx`, `CollectionDetail/ScriptTab.tsx`, new `ScriptSnippets`, `layout-store.ts`, `scripting.cpp`, tests, docs. Touches `ScriptPanel.tsx` after #1213 (which adds `ariaLabel` at every `CodeEditor` mount) - land after it, or rebase over it. Independent of #1209 and #1220 (different files), though #1209's single-interface typedefs make the hover docs in step 4 land on every chain member.

## Acceptance criteria

- [ ] The script editor fills its panel; nothing below it is visible by default except the one contextual line and the docs link.
- [ ] A collapsible, remembered snippets surface (one component, used by request and collection script hosts) inserts grouped, variant-appropriate templates at the cursor with working placeholders; its content comes from the engine completion table.
- [ ] Every fact the old notes carried is present in the completion table's `documentation` or the docs pages, pinned by a test that fails when one is removed.
- [ ] `quickReference`, `notes` and `QUICK_REF` no longer exist; docs updated as listed.

---

### LINKED ISSUE #1330: app(collections): the collection Script tab's editor is still a fixed 320px while the request panels' fills the panel
Filed from PR #1324 (issue #1223) rather than deferred without a record. That PR gave the *request* script panels' editor the panel and left this one as it was, deliberately; this is the half it did not do.

## Problem

`app/src/modules/collections/CollectionDetail/ScriptTab.tsx` mounts its `CodeEditor` at `height="320px"`, hardcoded, the way `ScriptPanel.tsx` mounted one at `height="350px"` before #1223. On a tall window the collection's pre- and post-request scripts - which run in front of *every* request in the subtree, so they are often the longer of the two kinds - get 320px and the rest of the screen goes to whitespace, the SaveFailed slot and a Clear button.

#1223's height criterion named the request panel, where the defect was sharper (a wall of reference prose took the space the editor did not). The reasoning applies here too; the shape does not transfer unchanged, which is why this is its own issue:

- The tab is a form. Below the editor sit the snippets list (shared with the request panels since #1324), `SaveFailed`, and a **Clear** button; those have to stay reachable, so the editor cannot simply take `flex-1` in a column that scrolls.
- The editor is wrapped in a bordered box with its own filename header row (`pre-request.js` / `post-request.js`), which the request panel has no equivalent of.
- The tab autosaves on editor blur (`handleEditorFocusOut`), so a layout change that alters where focus goes on a click below the editor changes when a save fires.

## Fix / solution pathway

Two options, in preference order:

1. **Fill the tab.** `CollectionDetail`'s tab body becomes a flex column with a definite height, the editor box takes `flex-1 min-h-0` with a floor (`min-h-48`, as `ScriptPanel` uses), and the row of controls stays below it as a `shrink-0` footer. The snippets list is collapsed by default, so the default view is editor plus a two-row footer.
2. **Resizable, like `BodyPanel`.** `useResizable` with a persisted size, if a fixed fill turns out to fight the tab's other content. This is the heavier answer and buys a preference nobody has asked for yet.

Rejected: raising 320 to a bigger number. It is the same defect with a different constant, and a 27-inch monitor and a laptop want different constants.

## App changes

This is the app change. Engine: none.

## Tests

`cd app && pnpm test ScriptTab CollectionDetail` and `pnpm type-check`.

- The editor's height is asserted the way `script-panels.test.tsx` asserts the request panel's (`data-height` off the mocked `CodeEditor`, plus the wrapper carrying `flex-1`), mutation-checked: put `320px` back and the test reddens.
- The autosave cases in `ScriptTab.autosave.test.tsx` still pass unchanged - that is the guard that the focus-out contract survived the layout change.

## Docs (same commit)

`docs/app/COMPONENTS.md`'s CollectionDetail tab table, if the row's description names the editor's sizing.

## Sequencing

After PR #1324 merges (it is what leaves the tab as the last fixed-height script editor). Parallel-safe against everything else; nothing is blocked on it.

## Acceptance criteria

- [ ] The collection Script tab's editor grows with the window, with a floor, and the Clear button and save-failure notice stay visible without scrolling on an ordinary window.
- [ ] The blur-to-save contract is unchanged, proven by the existing autosave suite passing untouched.
- [ ] No hardcoded pixel height remains on that `CodeEditor` mount.

athrvk:
## Description

The block under every script editor was a wall nobody could act on: a 14-line `<pre>` of `pm.*` calls plus nine paragraphs of rules in the request panels, and a second, differently shaped four-card grid on the collection Script tab - two hand-rolled copies of one idea, drifted from each other and from the runtime, neither insertable.

**Root cause and approach.** Those lines already exist, better, in the engine's completion table: every one is a template with placeholders and documentation that Monaco offers as you type. What the table could not say was *which script kind* a template belongs in, so it could not be listed - which is why the renderer grew its own copies. The fix says it (`context`, plus a `group` heading, on snippet entries only), and adds the five pre-request templates the retired 14 lines implied and the table lacked. One `ScriptSnippets` component - used by both hosts - filters the table to the editor it sits under, groups it, and inserts at the cursor through Monaco's snippet controller, so `${1:200}` becomes a tab stop exactly as it does when the same template is accepted from completion. The rules the prose carried move to where the author already meets them: a member's rule into the completion table's `documentation` (hover and completion), a hook's rule into the two published pages the *Scripting docs* link opens.

**The alternative rejected:** keeping the wall and adding a collapse toggle. A toggle on unread prose is still unread prose - the value is in insertion. A third snippet list in the renderer was rejected for the same reason the two old ones are being deleted: the engine table is the single source, and MCP re-serves it.

### Merged with master, and #1330 folded in

`master` landed #1323 ("the Body and script editors fill their pane") while this branch was in flight. That is the same ground, so its layout wins on both hosts: `ScriptPanel` takes master's `flex min-h-0 flex-1 flex-col` column and `min-h-40 flex-1` editor box (and drops the `height="100%"` this branch passed - #1323 made it the `CodeEditor` default), and `ScriptTab` takes master's `min-h-0 flex-1` editor frame. The height assertion this branch had added went with it: master's own case pins the same claim against the classes that now carry it.

That also settles **#1330**, which this run filed for the collection tab's then-fixed 320px editor: on this branch the tab's editor grows with a floor, the snippets list sits between it and the tab's own controls, and no pixel height remains on that mount. Verified here, pinned by master's case in `ScriptTab.chips.test.tsx`.

### Other deliberate deviations from the issue spec

- **The collapsed snippets header is visible below the editor**, so strictly the panel shows two things below it rather than the issue's "one contextual line and the docs link". A collapsible surface that the issue's own second criterion asks for has to have a header to be openable; the wall is what went, not the affordance. It is pinned as a count, so a third block cannot creep back in unnoticed.
- **`ContextBarSectionFrame` was not reused** for the collapsible header. It lives under `components/layout/context-bar/`, is written for that bar's section registry and has no slot for the trailing count; generalising it means changing the context bar in a PR about script panels. `Collapsible`, `Command` and the layout store are reused as-is.
- `docs/app/state-management.md` documents the new store field. Its layout-store block was already missing `graphqlVariablesCollapsed` / `graphqlVariablesSize` before this PR; not fixed here to keep the diff to this issue.

## Type of Change
- [x] New feature
- [x] Documentation update

## Testing

Run on the merged tree (`79d8d69`):

- `cd engine && ctest --preset linux-dev --output-on-failure` - **2877/2877 passed**.
- `cd app && pnpm test` - **7007 tests in 619 files, all passed**.
- `cd app && pnpm type-check`, `pnpm lint`, `pnpm format:check` - clean.

New tests:

- Engine (`script_completions_test.cpp`): every snippet declares a `context` and a `group` from the allowed sets; no non-snippet entry carries either; every `pre`/`both` template runs under `execute_prerequest` with its placeholders stripped; every `test`/`both` template runs under `execute_test` against a JSON response and its assertions pass; each re-homed rule is found by name in the documentation string that now carries it.
- App: `script-snippets.test.ts` (the pure filter/group read), `ScriptSnippets.test.tsx` (variant-appropriate templates, insertion with placeholders intact, collapse round-tripping through `layout-store`, the engine-unreachable state, and a source scan with a non-empty floor that `quickReference` / `notes` / `QUICK_REF` are gone), `editor-snippet.test.ts` (the snippet controller is what gets driven, after focus), `script-rules.test.ts` (every rule the retired prose carried, read out of the pinned generated declarations and the two docs pages - delete one from the engine table and it reddens), plus a count of what may sit under the editor.

Two review passes ran over the diff (acceptance criteria, repo conventions). What they found and what happened:

- The snippets box was `surface-sunken` around a `Command` that declares `bg-card surface-card` itself - `bg-transparent` cannot strip a surface class, so the filter field's `border-rule` resolved to the card's value inside a sunken box. **Fixed**: one surface for the box and its contents.
- Only one test template was ever executed by CI. **Fixed**: the `execute_test` guard above.
- The "Response (post only)" fact from the retired grid was in no documentation string. **Fixed**: `pm.response` now says a response exists only in a test script, pinned both sides.
- `docs/engine/scripting.md` had no snippet section. **Fixed**: it has one.
- A `className` prop no caller passed. **Removed.**

Pre-existing failures: none; both suites were green on the base commit too.

## Checklist

## Related Issues
Closes #1223
Closes #1330


athrvk:
**CI has not scheduled a run for the last two commits on this branch, and that is not something this PR can fix from here.**

What happened, in order:

| Head | `pull_request` workflows | Outcome |
|---|---|---|
| `61bcc73` | ran | cancelled by the next push |
| `0118c00` | ran | app suite red - `script-rules.test.ts` asserted an engine fact that landed one commit later |
| `16b3e81` | ran | app suite red - six suites render the script panels, which now read the completions query, with no `QueryClientProvider`; fixed in `722b697` |
| `722b697` | **none** | only `Labeler` (a `pull_request_target` workflow) fired |
| `dff8afd` | **none** | same |

`PR CI - run tests`, `PR title`, `Docs site` and `CodeQL` all trigger on `pull_request`, and none of them created a run for either commit - while another open PR (#1326) received its full set normally at 18:47. So the events are being delivered for the repository but not for this pull request. There is no run to re-run, and I am not going to push an empty commit or close and reopen the PR to provoke one.

Both failures CI did report are fixed, and the whole tree is verified locally at `dff8afd`:

- `cd engine && ctest --preset linux-dev --output-on-failure` - 2877/2877 passed
- `cd app && pnpm test` - 6953 tests in 615 files, all passed
- `mkdocs build --strict -f .github/mkdocs.yml` - clean

A push to the branch, or re-running the workflow from the Actions tab against this head, should be enough to get the checks reporting again.

---

PR #1326, opened 2026-09-03
feat(app): paint Monaco's widgets from the app's colour tokens

### LINKED ISSUE #1321: app(editors): Monaco runs on the stock `vs`/`vs-dark` themes, so its suggest, find and hover widgets paint VS Code's palette inside Vayu's - define one Monaco theme per mode from the app's tokens
## Problem

`CodeEditor` hands Monaco the built-in theme names and nothing else (`app/src/components/ui/code-editor.tsx:295`, `theme={isDark ? "vs-dark" : "vs"}`); no `defineTheme` call exists in the app. The editor surface itself is tolerable because the token colours are Monaco's syntax palette, but every **widget** Monaco draws is in VS Code's chrome: the suggest list the three completion providers open (`useVariableCompletionProvider`, `useScriptCompletionProvider`, `useScriptVariableCompletionProvider`), the find and replace widget, the hover card, the context menu. In dark mode that is `#252526` panels with `#454545` borders and VS Code's blue selection, next to Vayu's `--popover` (`240 6% 11%`), `--border` and `--primary`. In light mode `vs` gives white-on-white against `--background` `#f4f4f5`. The variable hover was the first place this was reported (#1320 moves that one hover onto the app's tooltip), but the palette mismatch is on every widget, and #920 already had to fight Monaco once for the scrollbars for the same reason.

The design system's colour rules (`docs/design-system.md`) hold for every surface the app draws; Monaco's widgets are the remaining surface they do not reach.

## Fix pathway

- Add `lib/monaco-theme.ts` that defines two themes, `vayu-light` and `vayu-dark`, via `monaco.editor.defineTheme`, each inheriting from `vs` or `vs-dark` for syntax and overriding the widget colours from the app's tokens: `editor.background` and `editor.foreground` from `--background` and `--foreground`; `editorWidget.background`, `editorWidget.border`, `editorSuggestWidget.*`, `editorHoverWidget.*` from `--popover`, `--popover-foreground`, `--border`; `editorSuggestWidget.selectedBackground` and `list.*` from `--accent`; `editor.selectionBackground` and `focusBorder` from `--primary` at the app's own alpha; `editorLineNumber.foreground` from `--muted-foreground`; scrollbar slider colours from whatever #920 settled on so the two sources agree.
- Tokens are HSL triplets in CSS custom properties, and Monaco wants hex. Read them once per definition from `getComputedStyle(document.documentElement)` and convert; re-define on theme and scheme change (the colour-scheme picker changes `--primary`, so the theme must follow it, not be computed once at module load). `useMonacoTheme` or the existing `isDark` source in `CodeEditor` is the trigger.
- Register once, at the same point `useLoadedMonaco` reports Monaco ready, so every editor instance and every `fixedOverflowWidgets` overlay picks it up.
- Font: widgets should use the app's UI font, not Monaco's default, via the `.monaco-editor .suggest-widget` and `.monaco-editor .find-widget` rules in `index.css` beside the existing `.vayu-variable-token-*` rules, radius from `--radius`.

## Tests

- A unit test over the theme builder: given a stub token map, the returned `IStandaloneThemeData` carries the expected hex for `editorWidget.background`, `editorSuggestWidget.selectedBackground` and `editor.selectionBackground`, in both modes; a token missing from the map falls back to the base theme rather than to `#000000`. Mutation check: drop one override and confirm the assertion fails.
- `code-editor` test: the `theme` prop passed to `@monaco-editor/react` is `vayu-dark` or `vayu-light` by mode, never `vs` or `vs-dark`. Stub the scheme both ways and assert both branches (the repo rule that a test never asserts the host platform applies to the mode too).
- A source-scan guard that no file passes `"vs-dark"` or `"vs"` as a theme name outside `monaco-theme.ts`; it must assert it scanned a non-empty file set.
- `pnpm test monaco`, `pnpm test code-editor`, `pnpm type-check`, `pnpm lint`.

## Docs, same commit

- `docs/design-system.md`: a short section stating that Monaco widgets take their colours from the same tokens through `monaco-theme.ts`, and which token each widget colour derives from, so a future token change knows it reaches the editor.
- `docs/app/COMPONENTS.md` entry for `CodeEditor`.

## Sequencing

Independent of #1320, which removes one hover from Monaco; the theme still covers the suggest and find widgets that remain. Touches `code-editor.tsx` and `index.css`; #1213 (editor tab focus) touches `code-editor.tsx` too, so land whichever is ready first and rebase the other.

## Acceptance criteria

- Triggering variable completion in a body editor opens a suggest list whose background, border and selected-row colour match a Radix popover and the palette's active-row colour on the same screen, in both modes.
- Cmd/Ctrl+F inside an editor opens a find widget in the app's palette and radius.
- Switching the colour scheme in Settings changes the editor's selection colour without a reload.
- No `vs`/`vs-dark` string is passed as a theme anywhere in `app/src` outside the theme module.

---

athrvk:
## Description

`CodeEditor` handed Monaco the built-in theme names (`theme={isDark ? "vs-dark" : "vs"}`) and the app defined none of its own, so every widget Monaco draws - the suggest list the three completion providers open, the find and replace widget, the hover card, the context menu - painted VS Code's chrome inside Vayu's.

**The plan.** Root cause: Monaco's widget colours come only from theme data, and the app never supplied any. `lib/monaco-theme.ts` builds `vayu-light` / `vayu-dark` from the same tokens every other surface reads, inheriting `vs` / `vs-dark` so the syntax palette is untouched and only the chrome is ours. Two constraints decide where the registration sits: `getComputedStyle` reports only the values live on the element, so just the mode the app is currently wearing can be read (one theme per mode, redefined on every change); and Monaco answers a theme name it does not know by falling back to `vs` in silence, after which a later `defineTheme` re-applies only the theme already showing. So `monaco-setup` defines the theme while it composes the instance, before any editor can be created, and `useMonacoTheme` redefines it inside the MutationObserver callback rather than in an effect - a child's effects run before its parent's, and `<Editor>` is the child. The rejected alternative was pure CSS over `.monaco-hover` / `.suggest-widget`: Monaco writes theme colours onto those nodes as inline styles and paints selection and find matches on its own canvas, where no rule reaches them, so that answer covers part of one widget and needs `!important` to do it.

Font and radius have no key in `IStandaloneThemeData`, so two rules in `index.css` carry them - **outside `@layer` and prefixed with `html`**, which is the part the review caught: Tailwind v4 compiles `@layer utilities` to a real cascade layer, an unlayered declaration beats a layered one at any specificity, and Monaco's unlayered `suggest.css` / `findWidget.css` declare the same properties on the same two-class selectors. Layered, the radius half of this change would have been dead on arrival; unlayered it ties, and Monaco's stylesheet loads after this one with the editor chunk, so `html` is the step that wins it. That rule is now written down in `design-system.md` for the next person who has to out-rank a Monaco widget.

Deliberate deviation from the issue spec, one: the issue asks for the app's UI font on the widgets. The rules do that, then restore the mono stack on the parts of them that are code (a completion's label, a hover's sample) - a proportional font on `pm.response.json()` is harder to scan than the problem it fixes.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Full app suite: **618 files, 6981 tests, all passing**. `pnpm type-check`, `pnpm lint`, `pnpm format:check` clean; `mkdocs build --strict` clean.

New and changed coverage:
- `lib/monaco-theme.test.ts` (13 cases): HSL-to-hex including the alpha byte and the not-a-triple cases; both modes' base theme; widget colours from a stub token map; a missing token dropping its key rather than defaulting to black; the two `index.css` widget rules read back out of the stylesheet (unlayered, `html`-prefixed) because no jsdom test can see a cascade jsdom does not implement; and the source scan keeping `"vs"` / `"vs-dark"` out of every file but the theme module, with a floor under the scanned set.
- `components/ui/code-editor.theme.test.tsx` (4 cases): the theme name in both modes, a mode switch defining before it renames, and an accent-scheme change redefining without a remount.
- `lib/monaco-setup.contributions.test.ts` gains a sixth concern: the theme is defined while Monaco is composed, observed through a spy on the `editor.api` namespace the setup module spreads.
- `code-editor.chords.test.tsx` and `code-editor.loading.test.tsx`: their Monaco stubs gain the one `editor.defineTheme` member the wrapper now reaches for.

Mutation-checked, each reverted and confirmed red: `theme={themeName}` back to `"vs-dark"` (5 cases), deleting the `registerMonacoTheme` call in `monaco-setup` (the sixth concern), a `#000000` fallback in the builder, and the widget rules moved back inside `@layer utilities` (2 cases).

Reviewed adversarially against the acceptance criteria and against CLAUDE.md; one finding was verified and discarded rather than fixed: the guard's `/["']vs(-dark)?["']/` regex could in principle fail on a future unrelated `"vs"` literal. Narrowing it to a `theme=`-shaped context would trade a loud, cheap false positive for a quiet false negative, which is the worse failure for a guard whose whole job is to notice.

## Checklist

## Related Issues
Closes #1321


athrvk:
Note for the reader, not tracked work (no user-visible change, so per CLAUDE.md it belongs here rather than in an issue): `lib/monaco-theme.ts`'s `hslTokenToHex` and `charts/uplot/uplotTheme.ts`'s `wrap`/`raw` now both resolve the app's HSL triples from the DOM, for different outputs - Monaco's theme data is parsed with `Color.fromHex` and takes hex only, uPlot takes any CSS colour and gets `hsl(...)`. They could share a reader one day; neither could have used the other as-is today.

CI is green on `7ccd426` (app shards on all three platforms, quality checks, docs build).

---

Issue #1280, opened 2026-09-02
app(tests): `send-with-row.test.tsx` "opens the list on the row that step bound" renders a 600-row grid and runs at 3.6-5.3s against vitest's 5s default - it fails under shard load


athrvk:
Found while verifying the last wave's closures at master `fa5439e`: one test failed in a batch run and passed alone, on both the new and the old master, so it is a pre-existing flake rather than a regression. Filed because a test that is red at random is the one kind CI cannot be trusted with.

## Problem

`app/src/modules/request-builder/components/UrlBar/send-with-row.test.tsx:732-734`, "opens the list on the row that step bound, without a click", stubs a 600-row CSV (`bigCsv(600)`) and renders the send-with-row grid over it in jsdom. Measured: 5.3s in a 17-file batch (timed out at vitest's 5000ms default), 3.6s alone at `fa5439e`, 4.6s alone at `5677ed5`. `app/vitest.config.ts` sets no `testTimeout`, and the file sets none either (zero `timeout` occurrences). CI runs the app suite sharded on three OSes under contention, which is exactly the condition that pushes it over.

## Fix / solution pathway

Decide what the test proves and size the fixture to that. If the point is "the list opens on the bound row even when that row is past the first window", the smallest CSV that puts the bound row beyond `useGrowingWindow`'s first window is enough (read `ROW_WINDOW` in the grid and use window plus a handful); 600 rows is not the property under test. If the test genuinely needs a large fixture, give it an explicit per-test timeout with a comment saying why, rather than the global default. Prefer the smaller fixture.

## App changes

The test file only.

## Tests

`cd app && pnpm test -- send-with-row` and, as the proof, the same invocation with `--reporter=verbose` showing the case under one second. Mutation check for the property: bind a row inside the first window and confirm the assertion still discriminates (the list must open on the bound row, not merely open).

## Docs (same commit)

None.

## Sequencing

Parallel-safe.

## Acceptance criteria

- [ ] The case completes in well under a second in a full-suite run, or carries an explicit, commented timeout.
- [ ] The assertion still fails when the bound-row behaviour is broken.

---
Generated by [Claude Code](https://claude.ai/code)

athrvk:
Claiming this issue - PR will follow from this run.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #846, opened 2026-08-19
app/tests: the 5,000-step ScenarioRunView cases sit on vitest's default 5s timeout, so a loaded machine fails one of them - PR #845 hit the unnamed version of this


athrvk:
Filed from the #803 implementation run (2026-08-19), which reproduced it twice and then measured what makes it happen. PR #845's body carries the undiagnosed version of the same event - *"one run reported `1 failed | 5805 passed` and the tail I captured did not name the file"* - so this issue exists to name the file and the mechanism.

## Observed

`src/modules/history/main/run-view-navigability.test.tsx > a run at the storage cap > counts the filtered list, not the whole one, in what is still to come` fails in a full-suite run with:

```
Error: Test timed out in 5000ms.
 ❯ src/modules/history/main/run-view-navigability.test.tsx:332:2
```

It passes when the file is run alone (18/18, ~11s for the file). Both cases in that `describe` build a 5,000-result report and render `ScenarioRunView`; the failing one then clicks a filter chip, so it pays the render twice.

## What actually decides it: how loaded the box is

Four full-suite runs on the same 4-core box, same tree except where noted:

| run | tree | machine | wall | result |
|---|---|---|---|---|
| 1 | #803 branch | engine build + ctest running alongside | 565s | **1 failed** / 5780 passed |
| 2 | #803 branch | engine build + ctest running alongside | 495s | **1 failed** / 5780 passed |
| 3 | `598227f` (base) | idle | 330s | 5778 passed |
| 4 | #803 branch | idle | 334s | 5781 passed |

The suite runs `pool: "threads"` with vitest's default `testTimeout` of 5000ms (`app/vitest.config.ts` sets neither). When the four cores are shared with something else the whole run takes ~1.7x as long, and this test - the most render-heavy in the suite - is the first thing to cross 5s. Nothing about the branch matters; run 4 is the same tree as runs 1-2, green when nothing competes.

That makes it a **wall-clock assertion nobody wrote on purpose**: the test asserts a count, but the 5s default asserts that a 5,000-row render finishes in five seconds on whatever machine is running it. CI runners are shared machines, and the app job is unsharded on ubuntu and macOS.

## Fix pathway

Not prescribing one - the honest options differ in what they give up:

1. **Give the two storage-cap cases an explicit timeout** (`it(..., { timeout: 20_000 })`). Smallest change, keeps the coverage, and states the intent: this is a windowing assertion, not a performance budget.
[...]

PR #1332, opened 2026-09-03
fix(scripts): a clicked snippet reaches the editor, and lands where the explorer's insertions do

athrvk:
## Description

Owner-reported against #1324: clicking a snippet did nothing. Fixing that turned up two more defects in the same three lines - where the template lands, and whether anyone is told it landed. Three commits, each with its own root cause.

### 1. The insertion never reached Monaco

`editor.action.insertSnippet` is a VS Code **workbench** command. The standalone editor this app embeds registers no action of that name - it registers `SnippetController2` as the `snippetController2` *contribution* - and `trigger` resolves its argument against the editor's action map and returns silently when nothing matches. The call was well-formed, threw nothing, logged nothing, and inserted nothing.

Measured in headless Chromium against the shipped `monaco-editor@0.55.1`, driving one editor both ways:

```json
{ "viaAction": "", "hasController": true,
  "viaController": "pm.test(\"name\", function () {\n    \n});" }
```

**The test was the other half of that defect.** It mocked the editor as a bare object with a `trigger` spy and asserted the call the app makes - never that Monaco answers it - so it stayed green over a feature that did nothing. The fake is now shaped like the editor: `trigger` resolves nothing, as Monaco's does, and the only working door is the contribution. Reverting the fix reddens three of its cases. `editor-snippet.contract.test.ts` covers what a fake cannot, reading the installed package for the id we ask for, so a Monaco upgrade that renames it fails in a case that says why.

### 2. Where it landed was wrong twice

- **The caret's column is not a landing site.** A caret parked inside `const t = Date.now();` - where it is whenever the author typed there and then reached for the list - cut the statement: `const t = D` above the template, `ate.now();` below. Padding with newlines is what *produced* that split. The template now goes to the end of the caret's line, taking a leading newline only when that line has code on it; a blank line is where it belongs already, indentation included.
- **A selection was being replaced.** `SnippetSession` builds its edit from `getSelections()`, which is right when the completion popup accepts a snippet over the prefix being completed and wrong from a list: with `keep` selected, a click produced `const pm.test("name", …); = 1;`. The selection is collapsed to its end first. The click adds code; it never removes any.

### 3. It now behaves like the schema explorer, which solved this first

The GraphQL explorer inserts into an editor from a list beside it - the same problem - and `lib/graphql/insert-skeleton.ts` treats the cursor as **context rather than a splice point**: the selection set the cursor is in, failing that an enclosing one, failing both a new operation **appended to the end of the document**. A caret at offset 0 of a non-empty document gets that last case, because it is Monaco's default rather than the author's choice.

A script has no selection sets, but it has that case, and now gives the same answer: 1:1 in a script with code in it appends after that code; an empty script still takes it at 1:1. Two more conventions came from the same place - the explorer **announces** every insertion into a keyed live region (keyed because a live region only speaks when its text changes, so the same template twice would otherwise be silent), and shows a **refusal on screen**, since an `sr-only` refusal is a click that to a sighted user did nothing - which is this PR's own bug in miniature. `ScriptSnippets` does both, in the same words, which is why `insertSnippetAtCursor` returns *where it landed* rather than a boolean, as `insertionForNode` returns a `placement`.

**One difference kept deliberately**: the explorer's already-present handling (select the existing field rather than insert a duplicate) needs a parsed document to know a field is already in a selection set. Arbitrary JS has no equivalent, and a repeated statement is often what the author wants.

## Type of Change
- [x] Bug fix

## Testing

- `cd app && pnpm test` - full suite green on the branch.
- `pnpm type-check`, `pnpm lint`, `pnpm format:check` - clean. Engine untouched.
- Real-editor probe (headless Chromium, the `monaco-editor` in `node_modules`), covering: the action id inserting nothing, the contribution inserting with tab stops, a caret mid-statement leaving it whole, a selection surviving, an indented blank line keeping its indentation, and an untouched editor appending after existing code. A scratch script rather than a committed test - Monaco does not run under jsdom, and standing a browser up for one insertion is a suite this repo does not have; the contract test is the committed guard that stands in for it.

`App on windows-latest (shard 1/2)` went red once on two 5-second timeouts in unrelated suites (`LoadTestConfigDialog`, `UrlBar/send-with-row`) - diagnosed in a comment below, then **green on the one permitted re-run of the same commit**, which settles it as the shard-load case `app/CLAUDE.md` documents (#846, #1280).

## Checklist
- [x] I have updated documentation - not needed: no documented behaviour changes, the feature now does what #1223's docs already describe

## Related Issues

Follow-up to #1223 / #1324.

### DIFF
--- a/app/src/components/shared/ScriptSnippets.test.tsx
+++ b/app/src/components/shared/ScriptSnippets.test.tsx
@@ -98,25 +98,33 @@ function open() {
 
 describe("ScriptSnippets", () => {
 	it("starts collapsed, because the editor is what the panel is for", () => {
-		render(<ScriptSnippets context="pre" onInsert={() => {}} />);
+		render(
+			<ScriptSnippets context="pre" onInsert={() => ({ placement: "cursor" as const })} />
+		);
 
 		expect(screen.queryByPlaceholderText(/filter snippets/i)).not.toBeInTheDocument();
 	});
 
 	it("remembers being opened, in the store that survives the tab unmount", () => {
-		const { unmount } = render(<ScriptSnippets context="pre" onInsert={() => {}} />);
+		const { unmount } = render(
+			<ScriptSnippets context="pre" onInsert={() => ({ placement: "cursor" as const })} />
+		);
 		open();
 
 		expect(useLayoutStore.getState().scriptSnippetsCollapsed).toBe(false);
 
 		// The Radix tab switch: the panel goes away and comes back.
 		unmount();
-		render(<ScriptSnippets context="pre" onInsert={() => {}} />);
+		render(
+			<ScriptSnippets context="pre" onInsert={() => ({ placement: "cursor" as const })} />
+		);
 		expect(screen.getByPlaceholderText(/filter snippets/i)).toBeInTheDocument();
 	});
 
 	it("offers a pre-request editor its own templates and the shared one", () => {
-		render(<ScriptSnippets context="pre" onInsert={() => {}} />);
+		render(
+			<ScriptSnippets context="pre" onInsert={() => ({ placement: "cursor" as const })} />
+		);
 		open();
 
 		expect(screen.getByText("Set a header")).toBeInTheDocument();
@@ -127,7 +135,9 @@ describe("ScriptSnippets", () => {
 	});
 
 	it("offers a test editor the assertions instead", () => {
-		render(<ScriptSnippets context="test" onInsert={() => {}} />);
+		render(
+			<ScriptSnippets context="test" onInsert={() => ({ placement: "cursor" as const })} />
+		);
 		open();
 
 		expect(screen.getByText("Test: Status code")).toBeInTheDocument();
@@ -135,7 +145,7 @@ describe("ScriptSnippets", () => {
 	});
 
 	it("hands the caller the template, placeholders and all", () => {
-		const onInsert = vi.fn();
+		const onInsert = vi.fn(() => ({ placement: "cursor" as const }));
 		render(<ScriptSnippets context="pre" onInsert={onInsert} />);
 		open();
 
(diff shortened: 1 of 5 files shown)


athrvk:
**`App on windows-latest (shard 1/2)` went red on two 5-second timeouts, neither of them in this diff.** Re-running that job once; if it fails again I will treat it as real rather than call it a flake.

```
FAIL  LoadTestConfigDialog.test.tsx > load profile → fields
      > shows rate and duration for Constant RPS, and no connection count
      Error: Test timed out in 5000ms.
FAIL  UrlBar/send-with-row.test.tsx > picking a row
      > reads the file only when the picker is opened
      Error: Test timed out in 5000ms.
```

Why this is not the change under review:

- The diff is three files - `src/lib/editor-snippet.ts` and its two test files. Neither failing suite imports any of them (`grep` for `editor-snippet` / `ScriptSnippets` / `script-snippets` in both: no hits), and nothing here touches the load-test dialog or the URL bar.
- Both pass locally, in a full run of the suite on this commit: **620 files, 7012 tests, 0 failures**. The same two cases pass in about a second when run on their own.
- The failure mode is a timeout rather than an assertion, which is the case `app/CLAUDE.md` writes up: a render-heavy case gets the default 5s and "crosses it under machine or CI-shard load, failing over the environment rather than the code" (#846, #1280). The other three app jobs - ubuntu, macOS, Windows shard 2 - were green on the same commit.

One thing I cannot rule out and would rather name than hide: this PR adds two test *files*, and vitest's sharding splits by file, so the shard's composition changed even though its content did not. If the re-run is green, that is the reading. If it reddens on the same two cases, the honest conclusion is that these two are now sharing a shard and are over budget where they were previously under it - and the fix is theirs to carry (an explicit `it(name, { timeout }, fn)` with the measured cost, per that same guide), not a global timeout bump and not anything hidden in this diff.

---

Issue #939, opened 2026-08-22
app(inputs): Enter-commit fields ignore IME composition - committing a CJK buffer fires the action with a half-composed value


athrvk:
From the 2026-08-23 keyboard review at `cc8cf61`. One defect class, seven sites, one shared guard.

## Problem (verified)

No Enter-to-act input checks `e.nativeEvent.isComposing` (proven: `rg 'isComposing' app/src` matches nothing outside node_modules). For users typing with an IME (Chinese/Japanese/Korean), the Enter that **commits the composition buffer** is a keydown the handler cannot distinguish from a submit - so the action fires with a half-composed value. Sites:

- `SaveAsExampleDialog.tsx:96-98` - Enter saves the example (named by an uncommitted romaji buffer)
- `ImportModal.tsx:739-741` - Enter starts the URL fetch
- `McpSettingsPanel.tsx:798-803` - Enter adds the allowlist host
- `SettingControls.tsx:483-485`
- `VariablesCategoryTree.tsx:263-268` and `:355-363` - create/rename environment
- `variable-popover.tsx:215-232`

## Fix / solution pathway

One shared helper (e.g. `isCommitEnter(e)` in `lib/` returning `e.key === "Enter" && !e.nativeEvent.isComposing` - the 229-keyCode fallback only if support for older engines is demonstrated necessary; Electron's Chromium does not need it) applied at all seven sites. New Enter handlers use the helper by convention - one sentence in `app/CLAUDE.md`'s conventions if a natural home exists (verify).

## Tests

One representative behavioural test (fire keydown Enter with `isComposing: true`, assert no action; with false, action fires - jsdom supports constructing the event) on the shared helper plus one wired site; mutation-check by removing the guard. `cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check`.

## App changes / Docs

The seven files + the helper. Engine/MCP: none. Docs: none beyond the possible CLAUDE.md convention line.

## Sequencing

Parallel-safe against everything open; overlaps #935 on `SaveAsExampleDialog.tsx`/`ImportModal.tsx` Enter handlers (both issues touch the same `if (e.key === "Enter")` lines - land either first, the other rebases; or one PR takes both, stating so).

## Acceptance criteria

- [ ] A composition-commit Enter triggers no action at any of the seven sites; a plain Enter still does.
- [ ] The guard is one shared definition, not seven copies.

---
Generated by [Claude Code](https://claude.ai/code)

Issue #1347, opened 2026-09-03
app(window): under a bare X server the frameless window never paints - the app comes up whole and shows nothing, with no error


athrvk:
Filed from PR #1345 (issue #1165) rather than left as a workflow workaround with no record.

## Problem

Measured on this repository's own `ubuntu-latest` runner, twice, at PR #1345's `681d54d`: launched under `xvfb-run -a` with no window manager, the packaged app starts completely and never becomes visible.

The evidence is the app's own output, captured by `scripts/perf/measure-app.mjs`:

```
[Sidecar]   Data Dir: /home/runner/.config/vayu-client
[Engine] Database initialized at /home/runner/.config/vayu-client/db/vayu.db
[Engine] Listening on http://127.0.0.1:9876
[Sidecar] Engine is ready
[Main] Engine started successfully at http://127.0.0.1:9876
[Main] MCP server listening at http://127.0.0.1:9877/mcp
[Engine] GET /config - Fetching configuration entries
[Engine] GET /environments - Fetching all environments
[Engine] GET /collections - Fetching all collections
[Engine] GET /scripting/types - Generating script type declarations
```

So the main process ran its whole `whenReady` chain, the engine came up, and the renderer executed far enough to mount and issue its prefetches. What never happened is a frame: `ready-to-show` did not fire in 90 seconds, and `createWindow`'s handler is where the app calls `mainWindow.show()` (`app/electron/main.ts:406-412` on master `75c7277b`, the `show()` at `:407`), so nothing was ever displayed. No error, no dialog, no log line - the process simply sits there looking healthy.

It is specific to the app's own window, not to Electron in that environment: `scripts/perf/startup-harness.cjs` opens a plain `BrowserWindow` in the same session and reaches `ready-to-show` in 222 ms blank and 285 ms holding the built renderer. The difference between them is the window: the app's is `frame: false` with `titleBarStyle: "hidden"` (`main.ts:350-351`), `show: false` (`:394`), a `backgroundColor` read from `nativeTheme` (`:393`), `titleBarOverlay` (`:369`) and, since Electron 44 landed, an explicit `roundedCorners: true` (`:357`); the harness's is default.

[...]

PR #1383, opened 2026-09-04
fix(app): keep the create-variable popover's focus in the value field

### LINKED ISSUE #1380: app(variables): picking a scope in the create-variable popover moves focus to the chip - typing then goes nowhere and Enter re-presses the chip instead of creating
## Problem

Owner-reported from the URL bar: click an undefined `{{base_url}}`, pick **Collection** in the "create in" row, and the value field is no longer where keystrokes go. Reproduced on master `711f359f` by mounting `VariablePopover` in a Chromium harness and driving it with Playwright:

| Step | Focus | Typing `abc` | Enter |
|---|---|---|---|
| Click the token | value field (`Value for new variable base_url`) | lands in the field | creates `base_url="abc"` in environment, popover closes |
| Click the token, then click the **Collection** chip | the chip `<button>` | goes nowhere, field stays empty | re-presses the chip; nothing created, popover stays open |

The chips are plain `<button>`s (`app/src/components/ui/variable-popover.tsx:660-680`) with `onClick={() => setCreateScope(scope)}` and nothing else, so a click takes focus the way any button does, and the field's `autoFocus` (`:643`) only ran once on mount. The screenshot shows the resulting state exactly: Collection highlighted, the value field without a focus ring. A user who chose the scope first, then typed, gets an empty field and a Create button that writes an empty value if they reach for it with the mouse.

The default selection itself is correct: Environment is preselected when an environment is active (`SCOPE_PREFERENCE`, `:303`), and Collection in the screenshot is the result of the click.

## Fix pathway

- After `setCreateScope(scope)`, return focus to the value field (`ref` on the `Input`, `.focus()` in the chip's click handler). Picking a scope is a modifier on the value, not a destination.
- Keep the chips reachable by keyboard: `Tab` from the field lands on them, arrow keys move the selection between them (a `radiogroup` with `aria-checked`, which is what three mutually exclusive choices are, rather than three `aria-pressed` buttons), and Enter on a chip creates the variable with that scope, since a chip that has focus and a value already typed has nothing else Enter could mean.
- Disable Create while the value is empty, or make an intentional empty value explicit; today the same click creates `base_url=""` in the chosen scope.

## Tests

`variable-popover.test.tsx`, create branch (the file already covers "create in Environment"):

- Clicking a chip leaves `document.activeElement` on the value field. Mutation check: drop the refocus and the assertion fails.
- After clicking Collection, `userEvent.type("abc")` puts `abc` in the field and Enter calls `onValueChange("base_url", "abc", "collection")` once.
- With focus on a chip (via Tab), Enter creates with that chip's scope.
- Arrow keys move the selection between chips; the group reads as a radio group.
- Create is disabled with an empty value (or the chosen alternative is asserted).
- `pnpm test variable-popover`, `pnpm type-check`, `pnpm lint`.

## Docs, same commit

`docs/app/COMPONENTS.md` entry for `VariablePopover` (grep it), one sentence on the create row's keyboard model, if the entry describes the create branch.

## Sequencing

Touches `variable-popover.tsx` and its test only. #1320 (merged) rewrote the edit branch of the same file; the create branch is untouched by anything open.

## Acceptance criteria

- Click token, click Collection, type, Enter: the variable is created in the collection with the typed value; the field keeps focus throughout.
- Tab to a chip and press Enter with a value typed: the variable is created with that chip's scope.
- Create with an empty field is refused or explicitly confirmed; it never silently writes an empty variable.

---

athrvk:
## Description

Picking a scope in the create-variable popover moved focus to the chip, so typing afterwards went nowhere and Enter re-pressed the chip instead of creating (#1380).

**Root cause.** The "create in" chips are plain `<button>`s carrying only `onClick={() => setCreateScope(scope)}` (`variable-popover.tsx:672-688`), so a click takes focus the way any button does, and the value field's `autoFocus` had run once at mount with nothing left to give. Nothing in the component put focus back - the file has no DOM refs and no `useEffect` at all. Two defects travelled with it: Create on an empty field called `onValueChange(name, "", scope)`, and `mergeVariable` stores what it is handed, so the name got defined as `""` - a definition that *resolves*, so the token stops reading as undefined and answers with nothing; and the create field's Enter was a bare `e.key === "Enter"`, which #939 exists to prevent (an IME's composition commit arrives as an ordinary Enter keydown).

**Approach.** The chips become the WAI-ARIA radio group they always were - three mutually exclusive choices were announcing as three independent toggles and standing as three Tab stops in a two-control form. Roving tabindex makes the group one stop, `role="radio"` + `aria-checked` replaces `aria-pressed`, arrows move selection and focus together wrapping at the ends, and Enter on a focused chip creates with that chip's scope. A mouse click hands focus back to the value field: picking a scope is a modifier on the value, not a destination.

**The alternative rejected.** `ToggleGroup` (`components/ui/toggle-group.tsx`, Radix-backed) is the repo's segmented-control primitive and would have brought roving focus for free, but its single-select arm lets the active item be *deselected* to no value, and its arrows move focus without selecting - neither is what "exactly one scope, chosen as you arrow" needs - and adopting it would have restyled the 10px scope-tinted chips into a segmented control the issue never asked for. The hand-rolled group follows `ProfilePicker.tsx`, which the design system names as the canonical radiogroup.

**One deliberate deviation from that pattern:** the arrow handler sits on each `role="radio"` button rather than on the container. Keydown from the focused chip bubbles either way, but a handler on the container makes it an unfocusable interactive element that `jsx-a11y` can only be silenced about at the line - and the repo's suppression ceiling is at 15/15. Options that are already focusable do not need that trade, which is the sentence added to the design system.

**One deliberate deviation from the issue spec:** the empty-value refusal is `length > 0`, not a trim. A value of spaces is a value the user typed, and the variables editor stores the same string without trimming it, so trimming here would make the two disagree about what the user wrote. The behaviour is asserted rather than left to be rediscovered.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`pnpm test` - **647 files, 7332 tests, all passing** on the first commit; the popover's own file then goes 39 → 51 with the second. `pnpm type-check`, `pnpm lint` and `pnpm format:check` clean. Engine untouched, so no C++ build or ctest run: no engine test could go from green to red.

Twelve new cases in `variable-popover.test.tsx`, mutation-checked by reverting the fix and confirming the failure:

| Mutation reverted | Test that goes red |
|---|---|
| drop `createValueRef.current?.focus()` from the chip click | hands focus back to the value field when a chip is clicked |
| drop the `canSubmitCreate` guard and `disabled` | refuses the write and says so by disabling Create; refuses it on Enter too |
| `isCommitEnter` → bare `e.key === "Enter"` | ignores an Enter that only commits an IME buffer |

jsdom does not move focus on a click the way a browser does, so the focus cases focus the chip first - that is the browser behaviour under test, and where focus lands afterwards is the only thing that distinguishes the fix.

The two existing create tests move from `getByRole("button", …)` to `getByRole("radio", …)` for the chips; no other test in the repo asserts on this row.

### Driven in the running app, not only in jsdom

Both states were exercised in the real Electron app (engine connected, a request in a collection with `https://{{base_url}}/v1/orders` in the URL bar, the red token clicked), driving the same interaction the issue reports: click the **Global** chip, then type `api.staging.internal`. The pre-fix component was checked out into the same running app to capture the left column.

| What was read off the live popover | Before (master `3a83608`) | After (this PR) |
|---|---|---|
| Chip group role | no radiogroup; two `aria-pressed` toggles | one `role="radiogroup"`, chips `role="radio"` |
| Tab stops in the group | 2 (`tabindex=0`, `tabindex=0`) | 1 (`tabindex=0` on the selected chip, `-1` on the rest) |
| `document.activeElement` after clicking a chip | `BUTTON[Global]` - focus stuck on the chip | `INPUT[Value for new variable base_url]` |
| What the field holds after typing 20 characters | `""` - the keystrokes went nowhere | `api.staging.internal` |
| Create with an empty field | enabled, ready to write `base_url=""` | disabled |
| ArrowRight on the last chip | no handler | wraps to the first, selection and focus together |

The screenshots show it plainly: before, the focus ring sits on the **Global** chip with the value field still showing its `value…` placeholder after typing and Create solid blue; after, the ring is back on the field holding `api.staging.internal`, with Global selected.

## Checklist

`docs/app/COMPONENTS.md`, beside the prose that already describes when the create offer appears and when it is withheld: the create row's keyboard model and the empty-value refusal. `docs/design-system.md`, Accessibility section: the radio-group-of-buttons case beside the existing composite-widget rule, since this group deliberately places its handler where that rule does not.

## Related Issues

Closes #1380

### DIFF
--- a/app/src/components/ui/variable-popover.test.tsx
+++ b/app/src/components/ui/variable-popover.test.tsx
@@ -257,7 +257,7 @@ describe("an undefined variable can define itself", () => {
 			writableScopes: ["global", "environment"],
 		});
 		const panel = open();
-		fireEvent.click(within(panel).getByRole("button", { name: "Global" }));
+		fireEvent.click(within(panel).getByRole("radio", { name: "Global" }));
 		fireEvent.change(within(panel).getByLabelText(/value for new variable/i), {
 			target: { value: "abc" },
 		});
@@ -271,11 +271,9 @@ describe("an undefined variable can define itself", () => {
 		// that appears to work and does not.
 		renderPopover({ varInfo: null, resolved: false, writableScopes: ["global"] });
 		const panel = open();
-		expect(within(panel).getByRole("button", { name: "Global" })).toBeInTheDocument();
-		expect(
-			within(panel).queryByRole("button", { name: "Environment" })
-		).not.toBeInTheDocument();
-		expect(within(panel).queryByRole("button", { name: "Collection" })).not.toBeInTheDocument();
+		expect(within(panel).getByRole("radio", { name: "Global" })).toBeInTheDocument();
+		expect(within(panel).queryByRole("radio", { name: "Environment" })).not.toBeInTheDocument();
+		expect(within(panel).queryByRole("radio", { name: "Collection" })).not.toBeInTheDocument();
 	});
 
 	it("falls back to explaining, when nothing at all is writable", () => {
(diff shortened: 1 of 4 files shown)


athrvk:
### The same interaction, both builds

Driven in the real Electron app (engine connected, a request in a collection, `https://{{base_url}}/v1/orders` in the URL bar, the red token clicked), doing what the issue reports: **click the Global chip, then type `api.staging.internal`.** The left column is master's component checked out into the same running app.

| Before (master `3a83608`) | After (this PR) |
|---|---|
|  |  |
| Focus ring on the **chip**. Twenty characters typed and the field still shows its `value…` placeholder - the keystrokes went nowhere. Create is solid blue, one click from writing `base_url=""`. | Ring back on the **field**, holding `api.staging.internal`, with Global selected. |

The popover as it opens, before anything is typed - Create is refused while the field is empty:



Arrows move selection and focus together, wrapping at the ends (ArrowRight from the last chip):



And in the window it lives in:



The readings behind these shots are in the PR body's *Driven in the running app* table. Screenshots come off the renderer through CDP rather than an X11 grab, so a frameless window that never paints under a bare X server (#1347) could not have handed back a blank frame unnoticed.

The images are hosted on `claude/screenshots-1380`, a branch that carries nothing but these six PNGs and is not part of this PR. Delete it whenever you like - the pictures above stop rendering, nothing else changes.

---

PR #1384, opened 2026-09-04
fix(app): reach the application menu on Windows and Linux (#1361)

### LINKED ISSUE #1361: app(shell): on Windows and Linux the application menu is set but never shown - Help, About, Check for Updates and the View items have no mouse route
## Problem

`createApplicationMenu` in `app/electron/main.ts:392-579` builds the full template on every platform and installs it with `Menu.setApplicationMenu`. The window is frameless (`frame: false`, `titleBarStyle: "hidden"`, `main.ts:247-248`), and on Windows and Linux a frameless window draws no menu bar, so the template's only effect there is its accelerators. Everything without an accelerator is unreachable by mouse:

| Menu item | macOS | Windows and Linux |
|---|---|---|
| Help > Documentation, Scripting Guide, Report an Issue | menu bar | nowhere (docs link exists only on the Welcome screen footer) |
| Help > About Vayu | app menu | nowhere |
| Help > Check for Updates | app menu | Settings > Updates card only |
| View > Back, Forward, Zoom, Full Screen | menu bar | accelerators only; the title-bar arrows cover Back and Forward |
| File > Settings | menu bar | Ctrl+, and the Dock |
| Edit roles | menu bar | shortcuts only, which Chromium handles anyway |

Two consequences. A Windows user cannot find out what version they run or where the docs are without knowing a shortcut. And the About panel options set at `main.ts:1023-1029` are tuned for macOS (`copyright`, `iconPath`, `website` are read on Linux and Windows too, per the Electron docs, but `authors` is not set, and nothing calls `showAboutPanel` on those platforms outside the invisible menu).

## Fix pathway

1. **One template, two surfaces.** Keep `createApplicationMenu`'s template as the single definition. On Windows and Linux, the renderer's title bar gets a menu button (the app icon is the natural trigger: on Windows it already opens the system menu on right-click, `TitleBar.tsx:147-179`; make left-click open the application menu and keep right-click for the system menu, so the icon does what the same icon does in Windows apps). The button sends `window:appMenu { x, y }` and main pops `Menu.getApplicationMenu().popup({ window, x, y })`. No second template in React; the renderer never knows the items.
2. **Keyboard.** Alt, or F10, opens the same popup on Windows and Linux, which is what a native menu bar answers to. Register through the existing chord table in `constants/shortcuts.ts` so the Keyboard Shortcuts panel lists it.
3. **About on Windows and Linux.** Add `authors: ["Atharva Kusumbia"]` to `setAboutPanelOptions` and confirm `showAboutPanel` renders on both (the docs say it does; verify once on each). If the Windows panel is judged poor, a small in-app About dialog with version, engine version and the two links is the fallback, driven from the same menu item.
4. Leave macOS exactly as it is; the menu bar there is the right surface and the button does not render.

## Tests

- `TitleBar`: on `win32` and `linux` (platform stubbed both ways) the app icon has an accessible name "Application menu", left-click sends `window:appMenu` with a position, right-click on Windows still sends `window:systemMenu`; on `darwin` no menu button renders. Mutation check: remove the click handler and the IPC assertion fails.
- Main: `window:appMenu` pops `Menu.getApplicationMenu()` at the given point; a unit test with a stubbed `Menu` asserts `popup` is called with the window and coordinates, and that no second `buildFromTemplate` happens for the app menu.
- `shortcuts.listed` test: the new chord appears in the panel's groups.
- `pnpm test TitleBar shortcuts startup-order`, `pnpm type-check`, `pnpm lint`.

## Docs, same commit

- `docs/app/COMPONENTS.md` title bar entry: the application menu button on Windows and Linux, and that it shows the same template as the macOS menu bar.
- `docs/architecture.md`: one line beside the menu description.

## Sequencing

Touches `TitleBar.tsx`, `main.ts` (one IPC handler and the About options), `preload.ts`, `electron.d.ts`, `constants/shortcuts.ts`. #1347 touches `createWindow`; #1355 touches `webPreferences`; both trivial rebases. Independent of the context-menu issues filed alongside.

## Acceptance criteria

- On Windows and Linux, clicking the app icon in the title bar opens a menu with File, Edit, View, Window and Help, matching the macOS menu bar item for item; Alt or F10 opens the same menu.
- Help > About Vayu shows the version and links on all three platforms.
- Help > Documentation opens the docs site in the system browser from Windows and Linux without any shortcut.
- On macOS nothing changes: no button, the native menu bar as before.

---

athrvk:
## Description

The window is frameless (`frame: false`, `titleBarStyle: "hidden"`), so Windows and Linux draw no menu bar: the template `createMenu` installs contributed accelerators there and nothing else, and every item carrying none - Help > Documentation, Scripting Guide, Report an Issue, About Vayu, Check for Updates - had no mouse route at all. A Windows user could not find out which version they ran without knowing a shortcut.

**Plan.** Root cause is one template with only one surface. The fix keeps `createMenu`'s template as the single definition and gives it a second surface: the title-bar app icon renders on Windows *and* Linux, and left click / Enter / Space / F10 / a tap of Alt send `window:appMenu`, which pops `Menu.getApplicationMenu()` at the icon. On Windows right click keeps the system menu - what the same icon does in every Win32 app - and macOS renders no button and registers no listener, its menu bar drawing the same template already. **Rejected:** building a menu in the renderer (an HTML menu bar or a dropdown fed by a second template). It would be a second definition of the menu, free to drift from the one macOS draws, and would have to re-implement Electron's roles (undo/redo, zoom, fullscreen) and accelerators by hand.

`electron/app-menu.ts` holds the decision and the popup, extracted for the reason `context-menu.ts` gives: main.ts creates windows and starts the engine at import time, so no unit test can import it. `showApplicationMenu` names no Electron type - the menu and the window arrive as the slices of them it reads, which a real `Menu` satisfies and a test's fake satisfies too - so the popped shape is asserted directly rather than inferred. Alt is not a `Chord`: it is both modifier and key, and only a press that ends without another key is a menu request, so `lib/alt-tap.ts` is a small state machine that Alt+Tab, Alt+← and a pointer press all disarm. Both keyboard routes ask `isModalOpen()` first, per the window-chord rule.

The About panel is fixed in both halves rather than one: its fields are platform-scoped, and `website`/`authors` are read on Linux alone, so macOS and Windows showed a version and nothing to follow. `authors` is added (the issue's ask) and `credits`, which those two platforms read, now names the docs site and the repository.

The issue's anchors had drifted and the PR follows the live code: the function is `createMenu` (`main.ts:454-640`), not `createApplicationMenu`; the icon rendered on Windows only and opened the *system* menu on left click as well as right, so left click is what changes there.

- **The chord is F10 only; Alt is handled outside the registry.** `matchesChord` matches a key arriving *with* modifiers, which the Alt gesture has no way to be. Its label reads "Application menu (Windows and Linux)" so the Keyboard Shortcuts panel does not advertise a key macOS has no binding for; the alternative - filtering the row out on macOS - would make `shortcuts.listed.test.ts` pass or fail depending on the host it runs on.
- **`docs/architecture.md` is not touched.** It contains no menu description to sit beside (zero occurrences of "menu"); the menu bullet the issue means lives in `docs/app/architecture.md`, which is updated.
- **Right click on Linux is left to the app's context menu.** There is no system menu there to offer, and swallowing the event would remove the only menu that surface has.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Local, on this branch:

- `pnpm test` - full app suite green: **649 files, 7361 tests** (re-run after the review pass; CI has since run the same suite on ubuntu, macOS and both Windows shards, all green).
- `pnpm type-check` - green (renderer, electron, electron-tests projects). `pnpm lint` - zero errors, zero warnings. `pnpm format:check` - clean.
- New `electron/app-menu.test.ts` - **15 cases**: the three refusals and their order, the point's rounding, every malformed point the channel can deliver, the popup driven against a fake menu (asserting it is popped over the window at the rounded point, and at the pointer when the point is unusable), plus a characterization read of main.ts's wiring - it pops the installed menu, builds no second template, and the About options carry a link field for every platform.
- New `src/lib/alt-tap.test.ts` - **7 cases**: the tap, and the six ways a press is not one.
- Rewritten `src/components/layout/app-icon-menus.test.tsx` (formerly `app-icon-system-menu.test.tsx`, renamed because the icon now serves both menus) - **32 cases**, each platform stubbed both ways rather than read from the host.

Mutation-checked, three guards, each reverted and confirmed red: dropping `isModalOpen()` reddens "stays shut while a dialog owns the window" on both platforms; dropping the `armed &&` in the Alt watcher reddens four watcher cases and "does not open it for Alt used as a modifier"; dropping the `isWindows` guard on right click reddens "is not offered on Linux, which has none".

No pre-existing failures were observed on the base commit. The environment has no GUI, so the popped menu and the About panel were verified by unit test and by Electron's documented platform scoping, not by opening a window on each OS.

## Checklist

## Related Issues

Closes #1361


athrvk:
### Before / after, from the running app

Captured from the real app on Linux (engine connected, v0.25.0) on a virtual X display in the dev container: the base commit `3a83608` for **before**, this branch for **after**. The menus below are Electron's own popped menus, grabbed from the X screen - not renderer markup.

**The title bar**

Title bar before and after

On the base commit there is no app icon on Linux at all. Clicking where it now sits changes nothing (`compare` reports 0 pixels different against the idle frame), and F10 does nothing either - the menu template exists but has no surface.

**Left click pops the installed menu**

Application menu open

File, Edit, View, Window, Help - the same template the macOS menu bar draws, popped at the icon's bottom-left.

**The items that had no mouse route**

Help submenu

Documentation, Scripting Guide, Report an Issue, Check for Updates and About Vayu - the five the issue names.

**Help > About Vayu**

About panel

Version, the Website link and a Credits button, from the `authors` / `credits` fields this PR adds.

Not shown: Windows and macOS. The container is Linux-only, so the Windows right-click system menu and "macOS renders no button" remain covered by the unit tests that stub the platform both ways, not by a screenshot.

The images are served from commit 993cab9 on this branch, which added them and whose successor removed them again - the diff under review stays code-only.

---

PR #1392, opened 2026-09-04
fix(app): compose the create-variable row from the app's primitives - #1391

### LINKED ISSUE #1391: app(variables): the create-variable popover's "create in" row overflows the card by 21px, and its scope chips are a hand-rolled segmented control - use the primitives the rest of the app uses
## Problem

Owner-reported from the URL bar and visible in the same screenshot as #1380: the Create button's right edge sits past the popover's right border. Measured on master `f5ad0ed7` by mounting `VariablePopover` in Chromium with all three scopes writable:

| | px |
|---|---|
| Card width (`w-72`) | 288 |
| Row content width (`p-2.5` each side) | 266 |
| Row `scrollWidth` | 298 |
| Create button right edge minus card right edge | **21** |

Same numbers in light and dark. The row at `app/src/components/ui/variable-popover.tsx:733-802` is one `flex items-center gap-1.5` line holding the "create in" label, three chips (74, 63 and 44 px with the app's UI face) and `ml-auto` Create. Nothing wraps or shrinks, and `PopoverContent` (`popover.tsx:34`) does not clip, so the button paints outside the card. It overflows whenever all three scopes are writable, which is the common case: a request in a collection with an active environment.

The overflow is the symptom. The cause is that this row is built from nothing the rest of the app uses:

- **The scope chips are a hand-rolled segmented control.** Three `<button role="radio">` elements with per-chip active classes and a hand-written arrow-key handler (`:750-778`, from #1380). `app/src/components/ui/toggle-group.tsx` exists for exactly this, and its header comment describes the shape being repeated here as the defect it replaced: "three ghost buttons inside a div, each repeating the same active/inactive string", no shared radius, loose gaps, the active state written out per segment. Radix ToggleGroup already provides roving focus and arrow keys; #1380 rebuilt them by hand one file over. Every other one-of-few choice in the app is a `ToggleGroup`: the export format (`ExportSpecDialog.tsx:118`), the code section's mode (`CodeSection.tsx:145`), the response body view (`ResponseBody.tsx:269`), the load-test metrics format, the certificate format.
- **The action does not sit where actions sit.** The same file's edit branch puts Cancel and Save in a `flex justify-end gap-2` footer row (`:686`), and every dialog ends in a `DialogFooter` with the primary action right-aligned. The create branch is the one place in the app where the primary action shares a row with a control.

## Fix pathway

Compose from the primitives; write no new layout rule.

1. **Scope choice: `ToggleGroup` size `xs`**, `type="single"`, one `ToggleGroupItem` per writable scope, labelled with `VARIABLE_SCOPE_CONFIG[scope].full`. Delete the hand-rolled radio buttons and the arrow-key handler; Radix supplies both. The segments are neutral, as every segmented control in the app is; the chosen scope's colour appears on the token and its badge once the variable exists, which is where the scope colour lives everywhere else. Keep #1380's two behaviours by wiring them to the primitive: `onValueChange` refocuses the value field, and Enter on a focused segment creates (a `keydown` on the group, not per item).
2. **Action row: the footer the edit branch already has.** `<div className="flex justify-end gap-2">` under the choice row with the Create button (`size="sm"`, disabled while the value is empty, from #1380). Same row shape as manual-mode Cancel/Save at `:686`, so the two branches read as one component.
3. **"create in" label** stays as the row's leading text, or becomes the group's `aria-label` if the visual label is dropped; either way the `radiogroup` labelling from #1380 is replaced by the primitive's own.
4. `w-72` stays. With the button on its own row the widest line is the three-segment control, 181px of content, inside 266px.

## Tests

`variable-popover.test.tsx`, create branch:

- The scopes render as a `ToggleGroup` (assert the Radix `role="group"` root with `data-orientation`, and one `role="radio"` item per writable scope rendered by the primitive, not by local buttons). Mutation check: restore the local buttons and the root assertion fails.
- The Create button is not inside the group's row (assert its closest `flex` ancestor differs from the group's), and a source-shape guard that the file contains no `role="radio"` literal, since the primitive owns that attribute. Mutation check: move the button back and the first assertion fails.
- The #1380 behaviours stay green: focus returns to the field after a segment is chosen, Enter on a focused segment creates with that scope, Create is disabled while the value is empty.
- `pnpm test variable-popover toggle-group`, `pnpm type-check`, `pnpm lint`, `pnpm format:check`.

## Docs, same commit

- `docs/app/COMPONENTS.md:1850-1861` (the `VariablePopover` create-row paragraph #1383 wrote): the choice is a `ToggleGroup`, the button is a footer row.
- `docs/design-system.md`: if the segmented-control section lists its call sites, add this one.

## Sequencing

Touches `variable-popover.tsx` and its test. #1380 is merged; this replaces its chip implementation with the primitive while keeping its behaviour. Nothing open touches the file.

## Acceptance criteria

- With Environment, Collection and Global all offered, every pixel of the Create button is inside the popover's border, at the default UI font and at the Inter and system fonts the appearance settings offer, in light and dark.
- The scope choice looks and behaves like the export dialog's JSON/YAML control: one track, radius that follows the Roundedness setting, arrow keys move the selection, no per-segment class copies.
- Focus returns to the value field after choosing a scope; Enter on a focused segment creates; Create is disabled while the value is empty.
- `rg 'role="radio"' app/src/components/ui/variable-popover.tsx` returns nothing.

---

athrvk:
## Description

The create offer's row held the "create in" label, three scope chips and an `ml-auto` Create button on one unwrapping flex line - 298px of content in a 266px row - and `PopoverContent` does not clip, so with Environment, Collection and Global all writable the button painted 21px outside the `w-72` card.

**Plan.** The overflow is the symptom; the cause is that the row was composed from nothing the rest of the app uses. The chips were three `<button role="radio">` with per-chip active classes and a hand-written arrow handler, one file over from `toggle-group.tsx`, whose header names that exact shape as the defect the primitive exists to end, and the primary action shared a row with a control, which no other branch of this component does. So both halves become primitives: the choice is a `ToggleGroup` (Radix owns the `radiogroup`/`radio` roles, the single tab stop and the arrow keys) and Create moves into the same `flex justify-end gap-2` footer the edit branch uses one state over, which is what takes the widest line back inside the card. The rejected alternative is the issue's first version - `flex-wrap` and a second row: it stops the overflow while leaving the hand-rolled segmented control in place, so the next reader still inherits a copy of a primitive that never receives the primitive's fixes.

Two behaviours from #1380 are wired to the primitive rather than kept by hand:

- `onValueChange` still returns focus to the value field, ignoring the empty string Radix emits when the active segment is pressed again (a scope has no "off"), the same guard every other `ToggleGroup` call site carries.
- **Selection follows focus.** Radix moves the roving tab stop and leaves selecting to a press, so an arrow would otherwise light one segment while Enter created in another; each segment picks itself `onFocus`, which is the radio group's own pattern and what the arrows did before.

**A second defect the measurements turned up.** With the label beside the segments they get 266px less its own width, and the two widest UI faces the appearance settings offer spend more than that on "Environment Collection Global": System overflowed by 5px and JetBrains Mono by 24, so the last segment would paint outside the card the Create button had just stopped painting outside of. The label now stacks above the control, the way `LoadTestConfigDialog` labels its own segmented control; the widest face then measures 230px inside 266.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- `pnpm test variable-popover` - **54 passed** (51 before; the 3 new cases below).
- Full app suite: **659 files, 7454 tests, all passing**. (One run showed a single failure in `tooltip-delay.test.tsx`, which was my own temporary screenshot harness under `src/` tripping its "no nested `TooltipProvider`" source scan - the harness is not part of this diff and the guard passes with it removed.)
- New cases, each mutation-checked: the scope choice is the shared control (`data-slot="toggle-group"`, one class list across segments, one `data-state="on"`); Create is not in the choice's row (its parent is the `justify-end` footer and does not contain the group); the file writes no `role="radio"` of its own (the issue's `rg` acceptance check, as a guard).
- Four existing cases asserted the hand-rolled attributes and now assert the same behaviour through the primitive: the group is still one Tab stop, held on the group and handed to the chosen segment rather than on a roving `tabIndex` of ours; the arrows still move selection and focus together, one task later, because Radix focuses on a `setTimeout`; both Enter-to-create paths now enter the group the way Tab does.

## Visual evidence

Measurements and before/after screenshots are in a comment below (Chromium mount of the component against the app's own `index.css`, stated as such): 21px outside the card before at the default face, 61px at the Mono face, 11px inside at all four faces after, identical in light and dark.

- The issue's test spec says to assert a `role="group"` root. Radix's single-select root is `role="radiogroup"`, which is what the existing accessible-name test already queried, so that is what is asserted.
- It also names `data-orientation` on the root. The app's `ToggleGroup` passes no `orientation`, so Radix stamps none - and that is worth keeping: with no orientation, ArrowUp/ArrowDown move too, exactly as the hand-rolled chips did. A test covers the vertical pair so a later `orientation="horizontal"` cannot silently kill it.
- `docs/design-system.md` has no segmented-control section to add a call site to. It does have an Accessibility paragraph that cited this row as the "arrow handler on each `role=\"radio\"` button" example; since that implementation is gone, the paragraph is rewritten around the primitive, keeping the hand-rolled guidance for the groups that still need it (`ProfilePicker.tsx`, which keeps its suppression - untouched here).
- The segments are neutral rather than carrying `VARIABLE_SCOPE_CONFIG[scope].tint`, per the issue's fix pathway ("the segments are neutral, as every segmented control in the app is"). The scope colour still appears on the badge and the token once the variable exists, and no doc names this row as a scope-colour call site.
- Reviewed and not acted on: `size="xs"` is left implicit (the primitive's default, as `CodeSection.tsx` also relies on), and the card containment has no automated pixel guard - jsdom has no layout, and the repo has no visual-regression harness, so the numbers above are the evidence.

## Checklist
- [x] I have updated documentation (`docs/app/COMPONENTS.md`, `docs/design-system.md`, in the same commits)

## Related Issues
Closes #1391


athrvk:
## Visual evidence

Captured by mounting `VariablePopover` in Chromium against the app's own `index.css`, with Environment, Collection and Global all writable - the same method the issue's measurements used. This is a component mount, not the running Electron app: the change is renderer-only and touches nothing native, and no engine build or window chrome is involved in it. Measurements come from `getBoundingClientRect()` on the card and the Create button in the same run that took each shot.

| | Create button vs card edge | choice row overflow |
|---|---|---|
| before, default face (Space Grotesk) | **+21px outside** | 32px |
| before, JetBrains Mono UI face | **+61px outside** | 72px |
| after, all four UI faces (Grotesk, Inter, System, Mono) | **11px inside** | 0px |

Identical numbers in light and dark; geometry does not depend on the theme.

**Before, default face** - the Create button paints past the card's right border:
https://raw.githubusercontent.com/athrvk/vayu/deba5eb574df62c358efda8d7b711ebc07ccb7c0/.pr-assets/1391/before-light.png

**Before, with the JetBrains Mono UI face** - the same defect at its worst: "Global" sits on the border and only a sliver of Create is on screen. This is what made me stack the label rather than leave it beside the segments (second commit).
https://raw.githubusercontent.com/athrvk/vayu/deba5eb574df62c358efda8d7b711ebc07ccb7c0/.pr-assets/1391/before-mono.png

**After, light** - the segmented control on its own line under the label, Create in the footer row, everything inside the card:
https://raw.githubusercontent.com/athrvk/vayu/deba5eb574df62c358efda8d7b711ebc07ccb7c0/.pr-assets/1391/after-light.png

**After, dark** - included because the change swaps the per-scope chip tint for the primitive's neutral segments, which is a colour change; the active segment reads by `bg-accent` plus weight in both themes:
https://raw.githubusercontent.com/athrvk/vayu/deba5eb574df62c358efda8d7b711ebc07ccb7c0/.pr-assets/1391/after-dark.png

The PNGs are committed at `deba5eb` and deleted again in `2ad0134`, so they render permanently from that blob while "Files changed" stays code-only - they intentionally do not appear in the diff.

---

Issue #1358, opened 2026-09-04
app(notifications): opt-in system notifications for the events that happen while the user is not looking - run finished, engine lost, update ready, sign-in completed


athrvk:
## Problem

Every notification Vayu shows is an in-app toast. The Settings panel called "Notifications" (`NotificationsPanel.tsx`) configures the toast stack's position, duration, size and severity threshold, and nothing else; `rg "new Notification" app/` is empty in both the renderer and the main process on master `711f359f`. So a user who starts a thirty-minute load test and switches to their editor finds out it finished by switching back. The same holds for a collection run, for the engine dropping out from under an open workspace, for an update that has finished downloading, and for the OAuth loopback flow, where the user is by design in another application (the system browser) when the token lands.

## What the owner asked for

An opt-in. Off by default. When on, the events that are relevant surface as OS notifications; the ones that are not stay as toasts. The list below is the classification, drawn from every `showToast` call site (68 across 30 files) and the terminal events the two run services observe.

### Relevant: asynchronous, terminal, and the user may not be looking

| Event | Where it is known today | Notification body |
|---|---|---|
| Load test completed, stopped, or failed | `LoadTestService` SSE `complete` handler and `handleError` (`load-test-service.ts:160-195`, `:152`) | "Load test finished - 12,400 requests, p95 210 ms, 0.3% errors" / "Load test stopped" / "Load test failed - <reason>" |
| Collection run reached a terminal status | `ScenarioRunService` `complete` (`scenario-run-service.ts:84-96`) | "Collection run finished - 41 passed, 2 failed" |
| Engine disconnected outside a user-requested restart | `queries/health.ts` transitions to `disconnected` (the classifier from #1227 already separates a restart window) | "Vayu's engine stopped responding" |
| Engine restart failed | `useEngineRestart.ts` error toast | "The engine could not be restarted" |
| Update downloaded (Windows silent, Linux AppImage) or available (macOS notify) | `useAppUpdate.ts:42-53` via `onUpdateAvailable` / `onUpdateDownloaded` | "Vayu 0.26.0 is ready - restart to update" |
| OAuth token obtained through the loopback flow | `TokenStatusRow.tsx:105` "Authorized" toast, and `authorize.ts` loopback mode where the user is in the system browser | "Signed in - back to Vayu to continue" |
[...]

PR #1395, opened 2026-09-04
feat(app): an inbox can say when it captured something while you were away - #1388

### LINKED ISSUE #1388: app(inbox): notify when an inbox captures a request while Vayu is in the background - a per-inbox toggle, off even when system notifications are on
## Problem

#1358 built the system-notification channel and classified every event the app raises into relevant and not. One row of its Relevant table was marked **optional** and was deliberately left out of the PR that closed it: an inbox capturing a request while the window is not in front.

It was left out because it is the only event in that table with no natural rate. Every other one is terminal and rare - a run ends once, an update lands once, a sign-in completes once. A webhook source can deliver hundreds of requests a minute, and a notification per capture would train the user to turn the whole feature off, taking the run-finished notification with it. That is why #1358 specified it as a **per-inbox** toggle that is off even when system notifications are on, rather than as another entry in the shared list.

Nothing about it is blocked: `services/notify.ts` holds the kinds and the opt-in read, `electron/notify.ts` answers focus and support, and the click route through `useNotificationActivation` already opens a tab by id.

## Fix pathway

1. **A kind and a target.** Add `inboxCaptured` to `NOTIFY_KINDS` and a target the activation hook can open - the inbox tab for that inbox (`openTab({ type: "inbox", entityId })`), which is the surface the capture landed on.
2. **The toggle belongs to the inbox, not to Settings.** It is a property of one inbox the way its canned response is, so it is stored with the inbox rather than in `client-settings-store`, and it defaults to off. The Notifications settings panel says nothing about it; the inbox's own controls do.
3. **Two gates, not one.** A capture notifies only when the global opt-in is on *and* that inbox's own toggle is on. `services/notify.ts` reads the first; the call site reads the second, so a busy inbox cannot be made loud by a setting the user turned on for run results.
4. **A rate the stream cannot beat.** Unlike every current caller, this one fires on a stream the app does not control. Coalesce: at most one notification per inbox per interval, with the body saying how many arrived ("3 requests captured") rather than posting three. The interval belongs beside the capture stream's existing batching, not as a new setting.
5. Body carries the method and path of the most recent capture, as the issue's table states: "Inbox received POST /webhook".

## Tests

- The call site: a capture with the global opt-in off posts nothing; with it on and the inbox's toggle off, nothing; with both on, one post carrying the method and path. Mutation check: drop the per-inbox gate and the second case fails.
- Coalescing: N captures inside one interval produce one notification whose body names N; the next interval posts again. Mutation check: remove the coalescer and the first case posts N times.
- The toggle persists with the inbox and defaults off - the same shape as the store test #1358 added for `systemNotifications`, which is the defect this class of field keeps hitting.
- `notify.callers.test.ts` pins which modules may post one, so the inbox module has to be added to that list deliberately. Update it in the same commit.
- `pnpm test notify inbox`, `pnpm type-check`, `pnpm lint`.

## Docs, same commit

- `docs/app/COMPONENTS.md`: the inbox entry gains the toggle and states the two gates.
- `docs/architecture.md`: the system-notification paragraph names the inbox as the one event with a per-source opt-in and why.

## Sequencing

After #1358, which is merged as the PR that built the channel this rides on. Independent of everything else open. Touches `services/notify.ts` (one kind), the inbox module and its store, `useNotificationActivation.ts` (one target), and `notify.callers.test.ts`.

## Acceptance criteria

- With system notifications on and an inbox's own toggle off - the default - a capture produces nothing outside the app.
- With both on and Vayu in the background, a capture produces one notification naming the method and path; clicking it opens that inbox.
- Twenty captures in a burst produce one notification saying twenty, not twenty notifications.
- With Vayu in front, a capture produces nothing outside the app whatever the toggles say.

athrvk:
## Description

#1358 built the system-notification channel and left one row of its Relevant table out on purpose: an inbox capturing a request. It is the only event there with no natural rate, so it arrives with two gates and a window.

**The plan.** The channel already had everything but a kind and a target, so the work is a call site rather than a mechanism: one kind (`inbox-captured`), one target (`{ view: "inbox", inboxId }`) carried through main's `readTarget`, preload and `useNotificationActivation`, and a notifier fed from the live capture stream - the only place the app hears about a capture at all. A capture posts only when the global opt-in is on **and** that inbox's own toggle is, the second read at the capture and again at the post, so a burst the user just silenced stays silent. One window of captures is one notification naming how many arrived; the window is trailing, not leading, because a leading edge posts once for the first capture and once for the rest, which is two notifications where the issue allows one - the cost is that a lone capture is announced a window late, immaterial for a notification whose reader is by definition in another application.

**The alternative I rejected.** The issue says the toggle is "stored with the inbox... the way its canned response is", and the canned response lives on the engine's record - so the literal reading is a new field on `InboxInfo`, its merge-patch parser and its wire shape. I did not take it: the engine's inbox is in-memory state that never outlives the process that opened it (`database.cpp`: captures are cleared at startup because "no inbox survives the process that opened it"), the engine never acts on this flag, and an engine field would put a desktop-UI preference into an API that MCP clients and the CLI also read - for exactly the same observable lifetime a client-side map keyed by inbox id gives. `host-sleep-store` is the existing precedent for client state keyed by an engine entity id whose record has no field for it. The issue's own Sequencing section and its `component:app` label scope the change to the app, and its Tests section asks for a *store* test, so this is the reading I took.

**Deliberate deviations, all noted here rather than in the code:**

- *Where the toggle is stored*, as above: `stores/inbox-notify-store.ts`, keyed by inbox id, off by default, not `client-settings-store`.
- *"The interval belongs beside the capture stream's existing batching"* - the capture stream has no batching. `useInboxLive`'s `onmessage` merges each SSE event straight into the query cache; the only intervals near it are the reconnect backoff and the engine's live-poll cadence. So the window is a named constant in the notifier (`INBOX_CAPTURE_NOTIFY_WINDOW_MS`, 10s) with its reasoning written down, and not a setting.
- *One reach the issue does not state*: only the inbox being watched has a live stream (the engine allows one per inbox, each holding a pool thread, so the app opens one for the addressed inbox). A capture notifies for that inbox - which is the case the feature is for: open the inbox, go elsewhere to trigger the webhook, hear about it. An inbox nothing is watching still records; it does not notify.
- *`createThrottledBatcher` grew two options* (`leading`, `intervalMs`) rather than the notifier hand-rolling a second buffer-and-timer. Both default to today's behaviour, and the two streaming callers are unchanged.

**Not in the issue:** the `retainInboxes` prune is the one thing here it does not ask for. It is not scope creep but the cost of the storage choice above - an inbox id belongs to the engine process that minted it, so without it the persisted map grows one unreachable entry per inbox ever started.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`pnpm test`, `pnpm type-check` and `pnpm lint` all green. The full suite ran green on the base commit first (665 files, 7519 tests) and again on this branch (668 files, 7547 tests), so nothing here stands on a pre-existing failure. The one case added after that run - the focus check named for the capture kind - was run on its own (`electron/notify.test.ts`, 24 passing) and CI has since covered the whole tree.

New and extended cases, 28 in total across six files:

- `modules/inbox/capture-notifier.test.ts` (10, new) - driven through the real `systemNotify` and the real stores with a stubbed preload bridge, so both gates are genuinely exercised rather than spied past: nothing with the inbox's toggle off, nothing with system notifications off, one notification naming the method and path with both on, nothing until the window elapses, twenty captures in one window answered by one notification saying twenty and naming the latest, the next window posting again, a window silenced by turning the toggle off during it, captures from before it was turned on never announced, a pending post dropped on teardown, and one window per inbox.
- `stores/inbox-notify-store.test.ts` (7, new) - off by default, one inbox at a time, off stored as absence, the prune, no re-render when a prune drops nothing, and a hand-edited or wrong-shaped payload degrading to nothing.
- `modules/inbox/InboxView.notify.test.tsx` (4, new) - the header toggle reads and writes the store, and a preference is forgotten only on a list read the engine answered.
- `modules/inbox/useInboxLive.test.tsx` (2 added) - a streamed capture reaches the notifier, and a torn-down stream drops its pending post.
- `services/throttled-batcher.test.ts` (2 added) - the trailing-only edge and the explicit cadence.
- `electron/notify.test.ts` (2 added), `hooks/useNotificationActivation.test.tsx` (1 added) - an inbox target survives the round trip and is refused without an id, a click opens the inbox tab, and the focus check is pinned for the capture kind by name.

Every guard is mutation-checked, and the five that matter were run rather than asserted: removing the record-time gate, removing the commit-time gate, restoring the leading edge, and removing the prune each redden exactly the case that names them.

**In the real app** (Linux, this branch's engine built from source, Electron 44 on Xvfb with a notification daemon): both gates on and Vayu behind another window, five captures 300 ms apart produced one notification reading "Inbox received 5 requests / 5 requests captured - latest POST /webhook"; the same five with the inbox's toggle off produced nothing, with `compare -metric AE` reporting 0 against the idle frame while the engine's `captureCount` still rose from 5 to 10. Screenshots and method: https://github.com/athrvk/vayu/pull/1395#issuecomment-5541747581

## Checklist

## Related Issues

Closes #1388


athrvk:
Driven in the real app on Linux - this branch's engine built from source, the packaged renderer under Electron 44 on an Xvfb display with a `dunst` notification daemon, an inbox started from the Services drawer, and the webhooks sent with `curl`. Not a stubbed render.

**The control, in the inbox header**

the Notify toggle, off and on

Off is the default, beside the Live badge. Nothing about it appears in the Notifications settings panel, which still speaks only of runs, the engine, updates and sign-ins.

**Both gates on, Vayu behind another window, five captures 300 ms apart**

one notification saying five requests

One notification, ten seconds after the first capture: "Inbox received 5 requests / 5 requests captured - latest POST /webhook", under the app's own icon. Not five notifications.

**The same five captures with the inbox's toggle off**

nothing posted with the toggle off

The negative proved rather than asserted: `compare -metric AE` between the corner before and fourteen seconds after the burst reports **0**, and the engine's `captureCount` went from 5 to 10, so the five requests were recorded and silently. The global opt-in stayed on throughout, which is the point - it is not the gate that was closed.

**Platforms.** Linux only. macOS and Windows notification behaviour is what `electron/notify.ts` already owned before this PR (the AppUserModelId, the bundle-authorization latch) and is covered by its platform-stubbed unit tests; this change adds a kind and a target to that channel rather than a platform path, so there is nothing platform-specific here to photograph.

The three PNGs live at the commit the links name and were removed again in the next one, so they render permanently while the reviewable diff stays code and docs only - they do not appear under "Files changed".

---

Issue #1150, opened 2026-08-30
app+engine(idle): the runs list is polled every 5s for the app's whole life from the App root - three processes wake per tick with History closed and nothing running


athrvk:
Sub-issue of #1143. Verified at master `e1c578a`, adversarially re-verified. The sweep's one confirmed idle finding - everything else at idle checked out clean (see #1143).

## Problem

`app/src/App.tsx:55` mounts a bare `useRunsQuery()` at the root, and every observer of that key gets a 5s `refetchInterval` (`queries/runs.ts:32,77`, gated only to the unpaged first page at `:43-45`; `staleTime` does not gate `refetchInterval`, `lib/query-client.ts:36-40`). So the poll runs for the app's entire visible life - History closed, no run active, app untouched. Each tick:

- **Renderer:** a real `GET /runs?limit=50` (`config/network.ts:54`), JSON.parse of a ~50-row payload, structural-sharing walk, objects discarded (steady GC churn).
- **Engine:** three SQLite queries (`runs.cpp:591-592,605`) and re-parsing up to 50 immutable `config_snapshot` JSON blobs to rebuild the same nine-key summary (`runs.cpp:565` -> `:99-117`), plus a `log_info` line (`:1333-1334`).
- **Main process:** at production verbosity (`--verbose 1`, `sidecar.ts:507-508`) that info line is flushed to stdout, piped, split, filtered and `console.log`ged (`sidecar.ts:514-527`) - a main-process wake every 5s at idle.

The 5s cadence and page-1 gating are documented deliberate (`runs.ts` docstring, `timing.ts:106-109`, `state-management.md:1213-1218`). **The always-mounted root observer is not** - `COMPONENTS.md:55` calls the root mount "prefetching" (which `useRunsQuery` exceeds by installing a lifetime poll) and `state-management.md:1472` attributes the polling to "the history list" (mounted only while the drawer is open). Every surface that actually renders runs already observes the same key while visible: `HistoryList.tsx:105`, `WelcomeScreen.tsx:50`, `palette/sources/useEntityItems.ts:68`. The root observer adds polling precisely when nothing displays the data.

## Fix / solution pathway

[...]

Issue #506, opened 2026-08-11
app(inbox): a transient SSE disconnect can permanently demote Live to Running - the auto-reconnect races the engine's dead-socket detection and eats a fatal 409


athrvk:
From the 2026-08-11 closure verification of #480 (verified at master `3a7c937`). Narrow race, degraded-not-broken, but the degradation is silent and permanent until the user switches tabs: low severity, real bug.

## Problem

The inbox's Live stream dies unrecoverably on an unlucky reconnect:

1. A transient disconnect drops the EventSource; the browser auto-reconnects immediately.
2. The engine's one-stream-per-inbox claim (`engine/src/http/routes/inbox.cpp:792-796`) is still held - the poll loop only notices the dead socket within `inboxLivePollIntervalMs` (250ms default; the keep-alive cadence at `:839-845` is the detection bound).
3. The reconnect therefore lands a **409 `inbox_live_in_use`** - and EventSource treats any non-200 as fatal: no further retries, ever.
4. `useInboxLive.ts:101-127`: `onerror` keeps the closed source and the effect's deps never re-run, so nothing re-subscribes. The capture list is deliberately not polled (`app/src/queries/inbox.ts:13-15`), so new captures stop appearing entirely; the badge silently reads `Running` instead of `Live`, and the only recovery is switching away and back.

So the failure window is "disconnect + reconnect inside ~250ms" - narrow, but it is exactly what a laptop sleep/wake or a flaky localhost proxy produces, and the user gets no signal beyond a badge word.

## Fix / solution pathway

Two halves, both small; do both:

1. **App: own the retry instead of trusting EventSource.** In `onerror`, close the source and schedule a re-subscribe with a short backoff that clears the 409 window (e.g. retry at `inboxLivePollIntervalMs + jitter`, capped attempts, then surface a "Live updates stopped - resume" affordance instead of silently showing Running - the ScenarioRunView's "Live updates stopped" callout is the wording precedent). The effect must tear down any scheduled retry on unmount/inbox-switch.
[...]

PR #1401, opened 2026-09-04
fix(app): hold inbox capture streams app-wide, not in the inbox tab - #1400

### LINKED ISSUE #1400: app(inbox): the "notify on capture" toggle only works while that inbox's tab is the active tab - the view owns the only stream, so switching to a request tab before leaving silences it
## Problem

#1388 (PR #1395, master `c41ea644`) added a per-inbox toggle whose tooltip promises: "Notify through the system when this inbox captures a request while Vayu is in the background" (`app/src/modules/inbox/index.tsx:100`). `docs/app/COMPONENTS.md:666`, `docs/app/state-management.md:925-927` and `docs/architecture.md:191-198` make the same promise, with no caveat about which tab is on screen.

What ships is narrower, and silent about it. The capture stream and the notifier live inside the inbox view:

- `app/src/components/layout/Shell.tsx:92-112` renders one tab surface at a time (`case "inbox": return <InboxView />` at `:109`, mounted at `:467`; the comment at `:53-56` says "One tab's surface is mounted at a time"). No keep-alive, no hidden panel - verified: no `hidden`, `display:none` or `keepMounted` under `app/src/components/layout/`.
- `app/src/modules/inbox/index.tsx:219` is the only non-test caller of `useInboxLive` (verified: zero other callers of `useInboxLive` or `new EventSource` under `app/src`), and `useInboxLive.ts:200` is where the one `EventSource` on `GET /inbox/:id/live` is built. `createCaptureNotifier` (`capture-notifier.ts:84`) is created inside that hook and disposed with it (`useInboxLive.ts:251-256` calls `notifier.dispose()`, which discards the pending batch, `capture-notifier.ts:111-116`).
- The main process knows nothing about captures; `app/electron/notify.ts:181-183` only relays a target the renderer posts.

So: turn the toggle on for inbox A, click a request tab to keep working, then switch to another application. Captures on A are recorded (the engine does that regardless) but never announced. The same is true for a second running inbox while inbox B's tab is open, and for any inbox while Settings or a dashboard is in front. The notifier's own dispose comment (`capture-notifier.ts:112-114`) lists three reasons a stream ends - switched inbox, closed the tab, quit - and misses the fourth, switching to a different tab type, which is the one that ordinarily precedes leaving the window.

PR #1395's body records "only the inbox being watched has a live stream", which is broader than the truth: it is only the inbox whose tab is *active*. The toggle state persists across sessions (`stores/inbox-notify-store.ts:63-101`) while the stream that honours it lives only as long as the tab is in front.

Severity: low to moderate. Silent partial no-op of an opt-in feature the owner asked for by name; the toggle reads as on and nothing on screen says it is not in effect. No data at risk - captures are stored, `captureCount` still rises, the list reconciles when the user returns.

## Why the stream is view-owned today, and why that constraint does not block the fix

The engine allows exactly one live stream per inbox (`engine/src/http/routes/inbox.cpp:1105-1110`, 409 `inbox_live_in_use`) and each stream parks a cpp-httplib pool thread for its life (`:1188-1194`). `useInboxLive.ts:9-15` cites that as the reason the hook opens and closes with the selected inbox. Measured against the engine at `c41ea644`:

- The API server uses httplib's default task queue (`engine/include/vayu/http/server.hpp:110`, no `new_task_queue` in `server.cpp`): base `max(8, hardware_concurrency - 1)` threads, elastic up to 4x that, so parked streams consume headroom rather than blocking the next request.
- `GET /runs/:id/live` and `GET /runs/:id/events` already park threads the same way while a dashboard is open.
- Every *running* inbox is its own `httplib::Server` (`engine/src/http/managed_listener.cpp:58,85-86`) and already spawns a full worker pool the moment it starts. One more main-pool thread per notify-enabled inbox is a small marginal cost on top of what that inbox already spends.
- There is no cap on inbox count anywhere (verified: no `MAX_INBOXES` in engine or app); `managed_listener.cpp:41-43` assumes "a handful".

So one stream per running inbox with the toggle on is affordable. The one-stream rule does mean the fix must *share* a stream between the notifier and the view, never open a second one.

## Fix pathway

Hoist stream ownership out of the view into a service that outlives every tab, which is the pattern the app already uses for runs:

1. **New `app/src/services/inbox-watch-service.ts`**, modelled on `app/src/services/load-test-service.ts:8-14` ("This service runs independently of React components, ensuring the SSE connection stays alive regardless of navigation"). It owns the `EventSource` per inbox, reconciled against the set (running inboxes with the toggle on) union (the inbox the active tab addresses), refcounted per inbox, with a hard cap on concurrent streams written down against the pool numbers above. It creates the `CaptureNotifier` per stream; `capture-notifier.ts` moves nowhere and is called as today, so `services/notify.callers.test.ts:40-51`'s allowlist needs no new entry.
2. **A `useInboxWatchers()` hook mounted once in `app/src/App.tsx`**, beside `useNotificationActivation()` (`:91`) and `useHostSleepRecorder()` (`:101`). The latter's comment is the precedent: mounted at the app level because "the dashboard may not be the open tab".
3. **`useInboxLive` becomes a consumer.** It subscribes to the service's per-inbox `{ watching, stopped, resume }` instead of constructing its own `EventSource` at `useInboxLive.ts:200`; the query-cache merge (`mergeCapture`) and the retry policy (#506) move into the service with the socket. The view's unmount releases its reference; the socket closes only when nothing else holds the inbox.
4. **`retainInboxes` moves with it.** It is called only from `InboxView` (`index.tsx:183-187`), so the persisted toggle map is pruned only while the inbox tab is open; the app-level hook is where it belongs. PR-body note, not separately tracked.
5. **Reconciliation, not a channel:** `GET /inbox` already carries `captureCount` (`engine/src/http/routes/inbox.cpp:430`, `app/src/types/api.ts:1049`) on a 10 s poll (`queries/inbox.ts:131-136`). It cannot replace the stream - TanStack's `refetchIntervalInBackground` is unset in `lib/query-client.ts:33-47`, so the poll pauses in a hidden window, and a count delta has no method or path to put in the notification. It can catch a stream that died while the window was hidden; optional.

Rejected: rewriting the tooltip to "while this tab is open". That makes the toggle honest and leaves the feature not doing what #1388 was filed for; CLAUDE.md's stopgap rule applies.

## App changes

This is the app change. Engine untouched: the one-stream contract and the 409 stay as they are. MCP: no tool reads inbox captures (verified: `rg -n "inbox" app/electron/mcp/tools.ts` finds the start/stop/list tools only).

## Tests

- **The regression, first:** a test under `app/src/modules/inbox/` reusing the Shell scaffolding in `app/src/components/layout/shell-tab-identity.test.tsx:28-60`: open an inbox tab, turn the toggle on, switch to a request tab, push a capture through a stubbed `EventSource`, advance the 10 s window, assert exactly one `systemNotify.post`. Mutation check is built in: this test fails on today's master and must fail again with the service removed.
- `app/src/services/inbox-watch-service.test.ts` (new): the reconciler opens a stream for a running, toggle-on inbox no tab addresses; closes it when the toggle goes off or the inbox stops; never exceeds the cap; a second consumer for the same inbox shares the socket. Mutation: make the reconciler skip non-addressed inboxes and the first case reddens.
- `app/src/modules/inbox/useInboxLive.test.tsx` (extend): unmounting `InboxView` while the service still wants that inbox does not close the socket; the last holder leaving does. Mutation: drop the refcount, the first case reddens.
- `app/src/modules/inbox/capture-notifier.test.ts` passes **unchanged** - the evidence this is a wiring change.
- `pnpm test inbox-watch-service useInboxLive capture-notifier InboxView.notify shell-tab-identity`, `pnpm type-check`, `pnpm lint`.

## Docs, same commit

- `docs/app/COMPONENTS.md:666-680` (the `capture-notifier.ts` and `useInboxLive.ts` bullets): where the stream is owned.
- `docs/app/state-management.md:925-955`: the `inbox-notify-store` section names `InboxView` as `retainInboxes`' caller and "two readers at the moment a capture arrives"; both change.
- `docs/architecture.md:191-198`: the per-inbox paragraph gains the sentence that the stream is held app-wide for every running inbox with the toggle on.
- `docs/app/README.md` or `app/src/modules/README.md:73-78` if the services list is enumerated there.

## Sequencing

Touches `useInboxLive.ts`, `modules/inbox/index.tsx`, `App.tsx`, and adds a file under `services/`. Parallel-safe with #1362 (PR #1397, new `services/run-progress.ts`, no inbox files) at the logic level; rebase-level only on `App.tsx` if both land. #1364 item 2 (unread badge for inbox captures) needs this first - a badge that counts captures on any running inbox has no source until the stream is app-level; its body says so. Independent of #1363 and #1347.

## Acceptance criteria

- With system notifications on and inbox A's toggle on: open A, switch to a request tab, put Vayu in the background, send a webhook to A - one OS notification naming the method and path arrives after the batching window.
- Two running inboxes with the toggle on, neither tab active, window unfocused: a capture on either notifies, naming that inbox.
- The inbox view still shows captures live while it is open, and stopping the inbox or turning the toggle off closes its stream (the engine's `try_claim_live` count returns to zero - one `GET /inbox/:id/live` per inbox, never two, verified by the service test).
- The toggle's tooltip, `docs/app/COMPONENTS.md` and `docs/architecture.md` describe the same scope as the code.

---

athrvk:
## Description

The per-inbox `Notify` toggle promises a system notification when that inbox captures a request "while Vayu is in the background". The stream that hears those captures was owned by the inbox view, and `Shell` mounts one tab's surface at a time - so the click onto a request tab that ordinarily precedes leaving the window unmounted the view, closed the `EventSource` and disposed the `CaptureNotifier`. The toggle read as on, persisted across sessions, and announced nothing.

**Root cause and approach.** The stream was view-owned by design (`useInboxLive.ts:9-15`): the engine allows exactly one live stream per inbox and each parks a pool thread, so the hook opened and closed with the *selected* inbox. That constraint rules out a second socket, not a longer-lived one. Ownership therefore moves to `services/inbox-watch-service.ts`, beside `load-test-service.ts` and for the reason that file already states - "this service runs independently of React components, ensuring the SSE connection stays alive regardless of navigation". It holds one socket per watched inbox: the union of the inboxes a mounted view has retained (refcounted) and the running inboxes whose toggle is on, capped at `MAX_INBOX_WATCH_STREAMS` (8, sized against the engine's `max(8, hardware_concurrency - 1)` API pool) with the inbox on screen always given a slot. The socket, the #506 reconnect, the #554 listener-is-gone check and the `mergeCapture` write move with it; `useInboxLive` becomes a consumer that retains, subscribes and renders. `hooks/useInboxWatchers.ts`, mounted once in `App.tsx` beside `useHostSleepRecorder`, computes the standing want and takes over `retainInboxes` pruning - which was also bound to the tab being open.

**Rejected:** rewriting the tooltip to "while this tab is open". That makes the toggle honest and leaves the feature not doing what #1388 was filed for; CLAUDE.md's stopgap rule applies.

`useInboxesQuery` gains an `enabled` option so the app-level watcher observes the 10s inbox-list poll only while at least one inbox may notify - a root observer nobody reads is what #1150 removed. Every surface that shows the list omits it and observes unconditionally.

Engine untouched: the one-stream contract and its `409 inbox_live_in_use` stay as they are. MCP is unaffected - no tool reads inbox captures through the renderer (`get_inbox_captures` polls the engine directly and says so).

**No user-visible change** beyond the toggle now working from another tab: no component renders differently, so there is nothing to screenshot. The badge (`Live` / `Running`), the "Live updates stopped" callout and its Resume are unchanged in markup and behaviour.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Run locally from `app/`, all green:

- `pnpm test` (full suite) - 672 files, 7574 tests.
- `pnpm exec vitest run inbox notify mcp-invalidation reduced-motion` - 18 files, 168 tests.
- `pnpm type-check` (renderer, electron, electron-tests), `pnpm lint`, `pnpm format:check` - clean.

No pre-existing failures reproduced on the base commit. Engine and C++ tests not run: the diff touches no engine file.

New and changed tests:

- `src/services/inbox-watch-service.test.ts` (new, 17 cases): the reconciler opens a stream for a notifying inbox no view addresses and closes it when the want goes; one socket shared between the want and the view, kept on the view's release and closed on the last holder's; the cap is never exceeded and the on-screen inbox always gets a slot; a capture with no view mounted reaches both the query cache and the notification; the reconnect, the bounded give-up, the resume, and the no-retry-after-a-deliberate-stop path.
- `src/hooks/useInboxWatchers.test.tsx` (new, 7 cases): the regression itself - toggle on, switch to a request tab, push a capture, exactly one notification naming the method and path; a second notifying inbox no tab ever addressed; the stream closing when the toggle goes off or the engine reports the inbox stopped; pruning, and keeping the map when the list read failed; and no inbox-list poll while nothing may notify.
- `src/modules/inbox/useInboxLive.test.tsx` (rewritten, 8 cases): the hook is now a consumer, so it pins holding, releasing without closing a stream somebody else wants, adopting a stream the service already holds *on the first render*, and moving its reference when the addressed inbox changes. The socket-level cases moved to the service test.
- `src/modules/inbox/InboxView.notify.test.tsx`: the two pruning cases moved to `useInboxWatchers.test.tsx`; the view now asserts it does *not* prune.
- `src/modules/inbox/capture-notifier.test.ts` passes **unchanged** - the evidence this is a wiring change.

Mutation-checked, by actually reverting and re-running: making `wantedInboxIds()` ignore the standing want (i.e. back to view-owned streams) reddens 6 cases across the two new files, including the end-to-end regression; reading the service through state-plus-effect instead of `useSyncExternalStore` reddens the first-render case.

Not run: `mkdocs build --strict` (mkdocs is not installed in this container). The docs edits add no page, link or heading, only prose inside existing sections.

## Review round

Two read-only reviewers went over the diff (one against the issue's acceptance criteria, one against CLAUDE.md conventions). Acted on:

- **`useSyncExternalStore` instead of state-plus-effect** in `useInboxLive` (second commit). The service already exposes the `subscribe` + `getState` pair, and `usePrefersReducedMotion` records why that is the right primitive; the effect version rendered one frame of `Running` before catching up to `Live` on a tab reopened over a stream the service already held. The retain stays an effect of its own - it is a want, not a read.

Discarded, with reasons:

- *"`resume()` and a toggle flipped off-and-on reconnect with no backoff and can meet the engine's still-held claim (409)."* True, and deliberate: a Resume that waited would look like a control that did nothing, and both paths are answered by the same retry ladder within one step. The old hook behaved identically (its resume closed and reconnected in the same tick), so this is not a regression. Now written down at `resume()`.
- *"No test proves the engine's claim is actually released before a reconnect."* Correct, and out of reach of a renderer test - the mock `EventSource` is the boundary. The engine's own claim/409 behaviour is covered engine-side.

## Deferred

- **#1403** - a background stream that spends its reconnect budget goes quiet with nothing on screen saying so. New with this PR: before it, a closed tab meant no background stream at all. Uncommon (six consecutive refusals against a live engine), and the fix is a cadence decision worth its own issue rather than a guess made here.

## Checklist

Docs updated in the same commit as the code: `docs/app/COMPONENTS.md` (the `useInboxLive` bullet, plus new bullets for the service and the app-level hook), `docs/app/state-management.md` (the `inbox-notify-store` readers and `retainInboxes`' caller), `docs/architecture.md` (the per-inbox notification paragraph), and the two cross-references that cited `useInboxLive` as the reconnect precedent (`docs/app/api-integration.md`, `useExecutionEvents.ts`).

## Related Issues
Closes #1400


athrvk:
Two red intermediate heads in this PR's check history, both Windows shards, both the 5s-default timeout in a file this diff does not touch:

- `11d1558`, shard 2/2: `src/modules/palette/CommandPalette.test.tsx > the empty query > puts the most recently focused tab first` - "Test timed out in 5000ms".
- `3595ef2`, shard 1/2: `src/modules/request-builder/components/LoadTestConfigDialog/LoadTestConfigDialog.test.tsx > load profile → fields > shows rate and duration for Constant RPS, and no connection count` - same message.

Neither is this change's: the diff touches the inbox stream, `App.tsx` and `queries/inbox.ts`, and neither of those files renders the palette or that dialog. The work this PR adds at the app level is gated on the notify map being non-empty, which it is not in any of these tests. Both are the failure mode `app/CLAUDE.md` records under #846 / #1280 - a render-heavy case spending most of the 5s default idle and crossing it under Windows-shard load - and both passed on other heads of this same branch, including the current one.

`LoadTestConfigDialog` also passes locally on the merged tree, and re-running that job once (the one re-run spent on this PR) turned it green. The current head `685fd1d` is green end to end and mergeable; nothing here is left to fix.

---

PR #1402, opened 2026-09-04
feat(app): a close that stops a running service names it first - #1363

### LINKED ISSUE #1363: app(lifecycle): closing the window on Windows and Linux quits the app and stops every running inbox, mock server and issuer - decide whether Services outlive the window, and make the close honest either way
## Problem

`app.on("window-all-closed")` quits on every platform but macOS (`app/electron/main.ts:1277-1281` on master `c41ea644`), and the quit path shuts the engine down. Everything the Services panel starts runs inside the engine: inboxes (`api.listInboxes`, `services/api.ts:566`), mock servers (`MockServerControl.tsx`), mock issuers (`NewIssuerDialog.tsx`). So a Windows or Linux user who has pointed another tool at a Vayu mock server or inbox and then closes Vayu's window loses that service with no warning. On macOS the same click leaves the app running with no window, the Dock icon brings it back, and the services keep serving. The platform difference is Electron's default, not a decision anyone made for Vayu; nothing in the code or docs says whether a running service is meant to survive its window (`docs/architecture.md` has no sentence about it, verified).

`app/src/modules/services/useRunningServices.ts` already knows what is running, which is the one input any resolution needs.

## Decision to make

Two coherent answers; the issue asks for one of them, not both.

**A. Services are window-scoped.** Closing the window ends them, on every platform. Then the close must say so when it matters: if `useRunningServices` reports anything active, the close is intercepted (the `close` handler at `main.ts:435` already exists for the save flush, and `before-quit-flushed` at `:220` is the flush's completion signal) and a dialog names what will stop ("Closing Vayu stops the inbox on port 9880 and the mock server for Orders API") with Close anyway and Cancel. macOS gets the same behaviour on Quit, and the window-close on macOS keeps its hide-not-quit default since nothing stops there.

**B. Services outlive the window.** Closing the last window on Windows and Linux hides it and leaves the app in a tray icon while any service is running; with none running, it quits as today. The tray menu lists the running services with their ports, Open Vayu, and Quit. `Tray` is not imported anywhere yet (verified on `c41ea644`); the icon is `shared/icon_png/vayu_icon_256x256.png` (already an `extraResource`) scaled per platform, with a template image on macOS if a tray is added there too. `window-all-closed` then quits only when nothing is running and the tray is not showing.

Recommendation: **A** unless there is a real use for headless services. Vayu is an API client; its mock servers and inboxes exist for a session of work, and a tray icon is a permanent piece of UI to maintain for a case most users never hit. A is one dialog. If B is chosen, the tray belongs in `app/electron/tray.ts` with its own test and a `Tray`-free build on platforms where the setting is off.

## Tests

For A: main-side test with stubbed `dialog` and a stubbed running-services answer: close with services running shows the message naming each service and cancels on Cancel, proceeds on Close; close with none running does not prompt; the save-flush path (`before-quit-flushed`) still runs after the prompt is accepted. Renderer: the IPC that reports running services answers from `useRunningServices`' source. Mutation check: remove the intercept and the "does not quit on Cancel" assertion fails.

For B: tray created only when a service is running at close; menu lists services by label and port; Open shows and focuses the window; Quit runs the normal quit path; tray destroyed when the last service stops with no window open. Platform stubbed both ways for the macOS template-image branch.

Either: `pnpm test startup-order lifecycle useRunningServices`, `pnpm type-check`, `pnpm lint`.

## Docs, same commit

- `docs/architecture.md` lifecycle section: what happens to running services when the window closes, per platform, stated once.
- `docs/app/COMPONENTS.md` Services panel entry: the same sentence from the user's side.

## Sequencing

Touches `main.ts` (close and window-all-closed handlers), `preload.ts`, `useRunningServices.ts`. Independent of the other shell issues from this sweep; #1347 touches `createWindow` only. The notification work from #1358 and the progress work from #1362 (PR #1397) touch `main.ts` elsewhere; rebase, no logic overlap.

## Acceptance criteria

- The chosen behaviour is written in `docs/architecture.md` and is the same on Windows and Linux; macOS differs only where the OS convention requires it.
- Under A: closing the window with an inbox running prompts, names the inbox and its port, and Cancel leaves everything running; with nothing running the window closes without a prompt.
- Under B: closing the window with a mock server running leaves it reachable from another tool, a tray icon lists it, and Quit from the tray stops it.

---

athrvk:
## Description

Everything the Services drawer starts - webhook inboxes, mock servers, mock issuers - runs inside the engine, and the engine dies with the app. `window-all-closed` quits on every platform but macOS, so on Windows and Linux the X button stopped every running service with nothing said; on macOS the same click left them serving. The split was Electron's default rather than a decision anyone made for Vayu, and neither the code nor the docs said which one was meant.

**The plan.** Root cause is the un-decided lifecycle, not a missing dialog: `main.ts` quits on last-window-closed, the quit path stops the engine (`quitShutdown` -> `stopMcp`/`stopEngine`), and every Services-drawer service lives in that process. Resolution is the issue's option A - services stay window-scoped - with the close made honest: the renderer publishes what the engine is holding for it on a new one-way `services:running` channel, main keeps that snapshot (dropping it with the renderer that published it), and a close or quit that would stop something asks first, naming each service. Cancel leaves the window, the renderer and every service exactly as they were. The rejected alternative was asking the renderer at close time through the `before-quit`/`before-quit-flushed` round trip `save-flush.ts` already has: it puts a second 2s ceiling on the one gesture the user is waiting on, reads the same TanStack cache the snapshot is published from (no fresher), and cannot answer at all when the renderer is gone - which is exactly the case where the dialog must not name dead services.

**What the user sees**

- Windows/Linux: closing the window with services running asks *"Close Vayu and stop the 3 services it is running?"* and lists them (`the inbox on port 9880`, `the mock server for Orders API, on port 9881`, `the mock issuer on port 9882`), with **Close anyway** / **Cancel**. With nothing running, nothing changes.
- macOS: the window-close asks nothing, because it stops nothing - the app keeps serving with no window. Quit asks, on every platform.
- A quit from a signal never asks: `install.sh` replaces a running AppImage that way, and a dialog would hold the upgrade open waiting for an answer nobody is there to give.
- The question comes **before** the save flush. A flush run first would have told the renderer its work was ending, and Cancel has to leave that untouched; confirm and the flush runs on the way out as it always did.
- A dialog that cannot be shown answers no rather than rejecting into nobody's hands: an unanswered question is not consent to stop the user's services.

**Drift found while verifying the issue.** `useRunningServices.ts` exported only `useRunningServiceCount()` - a number, not the descriptors a dialog can name - so the issue's "already knows what is running" held only as a count. It now exports the list, and the count is that list's length rather than a second sum over the same three queries: the Dock saying 2 while the dialog names 3 is a disagreement no test that reads one of them can see.

Kept out of scope deliberately: the updater's own quit paths (`update:restartToInstall`, `update:quitForUpdate`) are left asking like any other quit - a human is present for both - and no tray, no `window-all-closed` change, since option B was not chosen.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- `pnpm test` - **675 files, 7654 tests, all passing** (7648 on the base, plus the 6 this adds beyond its first commit). 39 cases in `electron/service-stop-guard.test.ts` and 7 new in `src/modules/services/useRunningServices.test.tsx`.
- `pnpm type-check`, `pnpm lint`, `pnpm format:check` - clean. Prettier was run on the touched app files only.
- No engine change, so no engine build or ctest run: nothing in `engine/` could go from green to red. (The engine *was* built, to run the real app for the screenshots below.)
- Platform branches are stubbed, never asserted from the host: the macOS cases pass a `platform` of `"darwin"` and the Windows/Linux cases `"linux"`, from the same test file.
- **The decision is behavioural, not a source scan.** A first pass tested the `main.ts` wiring only as text, and two one-line mutations passed every assertion: dropping the `!` in the guard check, and closing the window whether or not the user said yes. The decision now lives in `holdForConfirmation`, driven through a fake event and a fake dialog - a "no" runs nothing at all, a "yes" retries the gesture, and the retry finds the guard cleared, which is what carries a confirmed close into the save flush. Mutation-checked: ignoring the answer reddens "does nothing at all on Cancel". What remains text-asserted is only that `main.ts` calls it in each handler ahead of the flush, the same bound `save-flush.test.ts` works to.
- Other mutation checks, stated in the test file header: make `closeStopsServices` true on every platform and the macOS case reddens; drop the `confirmed` latch and "asks once for the close and the quit it becomes" reddens; return the snapshot without the `rendererGone` gate and the crashed-renderer case reddens.

**Visual evidence** is in a comment on this PR: the real app on Linux under Xvfb, with a real engine and three real services, showing the dialog naming all three, Cancel leaving them serving (`curl` to the inbox still 200), and the same three taken silently by one click on the base build.

## Checklist

Docs updated in the same commit as the code: `docs/architecture.md` (the ownership model now states that local services are window-scoped and what each platform's close does), `docs/app/architecture.md` (main-process lifecycle bullet and the preload bridge entry for `setRunningServices`), `docs/app/state-management.md` (the question precedes both flush paths), `docs/app/COMPONENTS.md` and `app/src/modules/README.md` (the Services entry, from the user's side).

## Related Issues

Closes #1363


athrvk:
## Visual evidence - the real app, not a mock

Captured on Linux under Xvfb 1440x900 with openbox, running the real dev stack: this branch's compiled main process, the Vite renderer, and a real `vayu-engine` built from this tree. Three services were started through the engine's own routes and picked up by the Services drawer - an inbox on 9880, a mock server for a collection named "Orders API" on 9881, and a mock issuer on 9882 - and then the window's X was clicked with `xdotool`. Linux only; macOS and Windows behaviour is covered by the platform-stubbed unit tests, not by a screenshot.

**After (this branch): the close names all three, and the dialog is a real native message box.**

The Vayu window with the Services drawer listing an inbox on port 9880, a mock server for Orders API on 9881 and a mock issuer on 9882, and a native dialog reading "Close Vayu and stop the 3 services it is running?" with those three named and Close anyway / Cancel buttons

**Cancel leaves everything as it was.**

The same window after clicking Cancel: no dialog, the Services drawer still lists all three services and the Dock still reads 3 services

Verified from outside the app, immediately after Cancel: `curl http://127.0.0.1:9880/anything` -> **200** (the inbox is still listening) and `curl http://127.0.0.1:9881/orders` -> **501** (the mock is still listening, answering the no-example route as it should). Clicking X again and choosing **Close anyway** then quit the app, and both ports went to **000** - connection refused - along with the engine's `/health`.

**Before (the PR base, master `a346428`): the same three services, one click, nothing said.**

Two frames side by side: the base build with the Dock reading 3 services, and the same screen after one click on the X, showing an empty desktop with the app gone

The base build was produced by checking out `a346428`, recompiling the main process and relaunching against the same engine and the same three services. One click on the X and the app was gone with no prompt; `curl` to the inbox went to **000** in the same breath. That is the defect this PR closes.

The PNGs are served from commit `0c912ff`, which introduced them; the next commit removed them, so they do not appear under "Files changed" and the reviewable diff stays code-only.

---

Issue #1362, opened 2026-09-04
app(runs): show a load run's progress on the taskbar button and the Dock icon - one `setProgressBar` call the OS paints while the user is elsewhere


athrvk:
## Problem

No file under `app/electron` calls `BrowserWindow.setProgressBar` on master `c41ea644` (re-verified; first recorded at `711f359f`). While a load test runs, the only place its progress exists is the dashboard inside the window. A user who has switched to another application, the case #1357 (wake lock) and #1358 (system notifications) also address, has no way to glance at how far along it is. Windows paints a progress fill on the taskbar button and macOS draws a bar on the Dock icon from the same Electron call. Linux gets nothing: Electron 44 removed Unity launcher support and `setProgressBar` is documented as Windows and macOS only (see the comment below).

## Fix pathway

1. **Renderer computes, main paints.** The dashboard store already holds the run's duration or iteration target (`dashboard-store.ts:36-39`, `duration` / `iterations`) and receives live metrics through `LoadTestService`. Add a derived `progress: number | null` (0 to 1; null when the run is open-ended and has no denominator) and send it to main through `power`-style IPC (`power:hold` / `power:release` in `preload.ts:275-290` are the model), `runs:progress { value }`, throttled to the metrics batcher's cadence (`createThrottledBatcher` in `services/throttled-batcher.ts`, which reads `liveRefreshMs` per push) rather than a second timer.
2. **Main paints.** `mainWindow.setProgressBar(value)` while a run is active; `setProgressBar(-1)` on complete, stop, error, or renderer teardown, so a crashed renderer cannot leave a frozen bar. An open-ended run (no duration, no iteration cap) uses mode `indeterminate` on Windows (`setProgressBar(2, { mode: "indeterminate" })`) and nothing on macOS, where an indeterminate Dock bar is not a thing. On `linux` the module returns without calling.
3. **Failure shows as failure.** On a run that ends failed, paint mode `error` for the two seconds before clearing on Windows; macOS has no error mode, so it clears. The notification from #1358 (`NOTIFY_KINDS.loadRunFailed`, `load-test-service.ts:230`) is what says why.
4. Collection runs (`ScenarioRunService`) have a step count and get the same treatment; a single design-mode Send does not, it is too short to draw.

## Tests

[...]

PR #1404, opened 2026-09-04
feat(runs): put a collection run's resolved plan size on its stream

### LINKED ISSUE #1398: engine/app(runs): a collection run's taskbar bar is indeterminate because nothing publishes the resolved plan length - put it on the run's stream
## Problem

Filed from #1362, which mirrored a run's progress onto the Windows taskbar button and the macOS Dock icon. A load run gets a real fraction there; a collection run gets an indeterminate bar, and this issue is the reason why, recorded rather than left as a "later".

The client cannot compute a collection run's denominator, and deliberately does not try. `RunCollectionDialog.tsx` says so at the point where it would:

```
// Omitted, not sent as the row count: the engine resolves the
// default, and a client that computed its own would be a
// second copy of a rule only one side can enforce.
```

Two engine paths are involved, and only one of them is short of a total (corrected at master `c41ea644`; the first version of this issue cited the wrong path):

- **A scenario *load* run** (`RunManager::start_run` with a scenario, `run_manager.cpp:1031-1044`, executed by `execute_scenario_load` in `engine/src/core/scenario_load.cpp`) already publishes its total: `context->requests_expected = max_iterations * step_count` (`scenario_load.cpp:478`), carried as `requestsExpected` on every `metrics` tick (`run_manager.cpp:2004`, `metrics_collector.cpp:1057`) and read by the dashboard's `deriveRunProgress`. That path is fine.
- **A sequential collection run** (`RunManager::start_scenario_run`, `run_manager.cpp:1072-1084`, executed by `execute_scenario_run` in `engine/src/core/scenario_runner.cpp:864`) is the one `ScenarioRunService` follows. It knows its size at start - `asked.iterations` (`scenario_runner.cpp:875`, the outer loop at `:972`) times `plan.steps.size()` (`:816`) - and publishes only per-step `step` frames (`:405`, `:790`). No metrics thread runs for it (`start_scenario_run` spawns none), and `ScenarioRunService.connect` says so (`scenario-run-service.ts:88-92`). It reports a `stepCount` afterwards on the `GET /runs` list row (`engine/src/http/routes/runs.cpp:89`) and `iterations` in the summary (`:424`), neither of which the live path reads:
  - `ScenarioStepEvent` (`app/src/types/domain.ts:860-876`) carries `iteration` and `stepIndex` per step and no total.
  - `useScenarioRunStore` holds the steps seen so far and a fold of their outcomes; no plan length, no iteration count.
  - `RunScenarioSummary.stepCount` (`domain.ts:1247`) is a list-row field the run's live path never reads.

So `services/run-progress.ts` reports `null` for `RUN_PROGRESS_KEYS.collectionRun`, which paints indeterminate on Windows and nothing on macOS. That is honest, and it is less than the OS could show.

## Fix pathway

1. Publish the resolved plan size on the sequential run's own stream, once, where the client already listens: an opening frame on the same SSE topic (`build_sse_frame` in `scenario_runner.cpp:405` is the shape) carrying `stepsPerIteration`, `iterations` and the product, emitted before the first `step` frame in `execute_scenario_run`. The engine is the only side that knows it, which is the whole point. `requestsExpected` on the load path is the precedent for the field's meaning: an upper bound, since a run can stop early on failure or the `maxStepsPerIteration` cap (`:966`).
2. `sse-client.ts` gains a handler for the new frame beside `onStep` (`:249-256`); `ScenarioRunService` reads it, holds it beside `activeRunId`, and reports `stepsCompleted / total` to `runProgress.report(RUN_PROGRESS_KEYS.collectionRun, …)` on each committed batch, the way `LoadTestService` reports its fraction from `deriveRunProgress`.
3. Keep `null` as the fallback: a run whose total never arrives (an older engine, a replay from an offset past the opening frame) must go on showing indeterminate rather than a fraction of a number nobody sent.

## App changes

`app/src/services/sse-client.ts`, `scenario-run-service.ts`, `types/domain.ts` (the new frame's type). MCP: the `run_collection` tool does not follow the stream (verified: no `EventSource` under `app/electron/mcp/`), so nothing there.

## Tests

- Engine (`engine/tests/scenario_runner_test.cpp`): the topic's first frame carries the resolved total for a sequential run, with an explicit `iterations` and with a data file bound and `iterations` absent (one pass per row).
- Renderer: `scenario-run-service.test.ts` asserts the fraction reported on a batch, and that a run whose total never arrived reports `null` on every batch. Mutation check: drop the fallback and the second case reports `0`, painting an empty determinate bar for a run that is running fine.
- `pnpm test scenario-run-service run-progress sse-client`, `pnpm type-check`, `pnpm lint`; `python build.py -t && ctest --preset linux-dev -R scenario_runner`.

## Docs, same commit

- `docs/app/COMPONENTS.md`: the dashboard's OS-chrome paragraph states the collection run shows only an indeterminate bar - correct it when it no longer does.
- `docs/engine/api-reference.md`, where the run stream's `step` frame is described, for the new opening frame; `docs/architecture.md` if the SSE event shapes are listed there.

## Sequencing

Blocked on nothing, but pointless before #1362 lands (PR #1397), which is what reads the value. Touches `scenario_runner.cpp`, the engine test, `sse-client.ts` and `scenario-run-service.ts`; parallel-safe with everything else open (#1400 touches the inbox stream, not the run stream).

## Acceptance criteria

- A collection run's taskbar button on Windows fills as its steps complete and reaches the end as the run finishes.
- A run whose plan size never reaches the client still shows indeterminate progress, not a fraction.
- No step count is computed in the renderer.

---

athrvk:
## Description

A collection run's taskbar button was indeterminate because nothing published its denominator. `execute_scenario_run` resolves its own size before the first step - `asked.iterations` passes over a plan of `plan.steps.size ()` - but the sequential path spawns no metrics thread, so unlike the load path it has no `requestsExpected` riding a tick, and `ScenarioRunService` reported `null` for the whole run. The fix publishes one opening `plan` frame on the run's existing tick topic (`{"stepsPerIteration","iterations","stepsExpected"}`, framed by `RunContext::append_event` so its id is assigned under the ring's lock), and the renderer holds it and reports `stepsCompleted / stepsExpected` on each committed batch - the cadence `LoadTestService` already reports its own fraction on. The alternative considered and rejected was republishing the total on every `step` frame, the shape `requestsExpected` has on the load path: `build_step_payload` is deliberately constant-size to protect the fixed-size ring (the reason the assertion *list* rides the stored row and only a tally rides the frame), and a value fixed for the whole run does not earn a per-step key. The cost of one frame is that a watcher attaching after it has been evicted never sees it, which is exactly what the `null` fallback is for - such a run stays indeterminate rather than showing a fraction of a number nobody sent.

No step count is computed in the renderer: `RunCollectionDialog` still sends none, and the client's only arithmetic is the division and a clamp.

**Rebuilt on #1406.** Reviewing this diff turned up that `runProgress` claimed the OS indicator by run *kind*, so a superseded run's stranded flush could repaint another run's bar; that was filed as #1405 and has since landed on `master` as #1406. The merge here takes that ownership model whole rather than working around it: the run's start claims the indicator by id - which is also what shows the indeterminate state until the `plan` frame arrives - and every fraction names the run it is for. `reportProgress` keeps no guards of its own: "a run that has ended" and "a run whose failure already flashed" are the same rule the claim now enforces, and `run-progress.test.ts` pins both on the real module.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

All green locally on Linux at `c02e14b` (the tree with `master` merged in):

- `python build.py -e -t` then `ctest --preset linux-dev`: **2908 passed, 0 failed**. Three new engine cases in `scenario_runner_test.cpp` - the stream's first frame carries the resolved total, the same for a run whose iterations come from a bound data set's row count, and the existing per-step frame test now expecting the opening frame ahead of the four steps. Tests that reach for step frames read them through a new `step_frames` fixture helper rather than indexing past the opening one.
- `pnpm test`: **7681 passed across 677 files**. New renderer cases: the fraction reported per committed batch (each naming its run), a run whose plan frame never arrives reporting `null` on every batch (mutation check: drop the fallback in `runFraction` and the division by a total of nothing reports `1`, the clamp of `Infinity` - a bar that reads finished on a run's first step), the clamp when a `setNextRequest` run outruns its plan (drop it and it reports `1.5`), and a flush that lands after the stream closed naming no run, so the claim check can refuse it. `parsePlanEvent` has its own narrowing cases in `sse-client.test.ts`.
- `pnpm type-check`, `pnpm lint`, `pnpm format:check` clean; `clang-format --dry-run -Werror` clean on the engine files.
- `mkdocs build --strict` was not run here (mkdocs is not installed in this environment); CI's docs `Build` job covers it and has been green on every push.

**One thing this branch got wrong twice and the record should show:** `clang-format -i` here reindented six `if (auto x = ...; cond)` continuation lines the change never touched, because this machine's clang-format 19 and CI's disagree about them - the patch-level divergence `engine/CLAUDE.md` warns to read the diff for. `c6b5289` put those lines back, and `1c91be5` then fixed the one line that genuinely was mine, whose wrapping a rename had pushed past the column limit.

## Screenshots

None, deliberately. The only surface this change paints is native OS chrome - the Windows taskbar button and the macOS Dock icon - and Linux paints nothing at all since Electron 44 dropped Unity launcher support (`electron/run-progress.ts` covers Windows and macOS only). A screenshot from this Linux container would show the absence that platform already has, so what stands in for it is the renderer-side unit tests above, which assert the exact values handed to `runProgress.report` on every path.

## Checklist

- [x] I have added tests

Docs updated in the same commits as the code they describe: `docs/engine/api-reference.md` (the `plan` frame under `GET /runs/:runId/live`, its upper-bound rule and the eviction caveat) with the engine commit; `docs/app/COMPONENTS.md` and `docs/architecture.md` (both of which stated a collection run can only ever show an indeterminate bar) with the app commit. Both of those paragraphs now sit beside #1406's own sentences about the claim, and read together.

## Related Issues

Closes #1398


athrvk:
Review notes on my own diff, so the reasoning is on the record rather than only in a session:

**Fixed in `c6b5289`.** `clang-format -i` here reindented six `if (auto x = ...; cond)` continuation lines this change never touched, and CI's clang-format 19 disagrees with this machine's about them - the patch-level difference `engine/CLAUDE.md` warns to read the diff for. Those lines are back to what the tree had, and the file's only remaining change is the fourteen added ones. `build_plan_data` became `build_plan_payload`, since every JSON-returning frame builder in the engine carries that suffix. One mutation-check comment was wrong about its own numbers: dropping the null fallback divides by a `stepsExpected` of nothing and reports `1` (the clamp of `Infinity`), not `0` - a bar that reads finished rather than empty. The test failed under either reading; the comment now names the one a reader would see.

**Filed as #1405, not fixed here.** `runProgress.report` claims the single OS indicator by `RunProgressKey` - a kind of run - and takes it over unconditionally, where `fail` and `clear` are guarded by what the indicator is currently showing. `sseClient` is a singleton whose `disconnect()` tells the displaced stream nothing, so a collection run whose last batch is still buffered when a load run starts will report a stale fraction over the load run's bar when the timer fires. It self-corrects on the load run's next flush, so the visible effect is at most one refresh window, and the honest fix is to key the claim on the run rather than on its kind - that is #1362's ownership model, `run-progress.ts`'s signature and both services, which is a wider diff than this issue asked for.

**Considered and left alone.** `handleClose` nulls `activeRunId` before the final `stepBatcher.flush()`, so the last partial window's steps reach the store but their fraction is never reported. That is deliberate and commented at `reportProgress`: `runProgress.clear` runs synchronously in the same tick, so nothing could paint in between, and reporting on the way out would be one more IPC message for a bar that is about to be cleared. It does mean the reported series can end just short of `1` where `LoadTestService`'s ends on its last tick; nothing renders the difference.

---

Issue #1335, opened 2026-09-03
app(deps): September 2026 dependency sweep - Electron 44, the safe minors, and the majors that need a migration pass


athrvk:
## What this is

`pnpm outdated` on master `7f377096` (app/ on pnpm 10.33, Node 22) lists 46 packages behind. This issue sorts them by what each upgrade changes for the app and what it costs, so the work lands as three commits in order rather than one bump that reddens CI for a reason nobody can name. Peer ranges were checked against the registry on 2026-09-03.

## 1. Electron 42 to 44 - changes what the app ships on

Electron 44.1.1 carries Chromium 152 and Node 24.19 (42.3.3 is Chromium 148, Node 24.15). This is the one upgrade users feel: renderer engine, and the Node the main process and the MCP server run on.

Breaking changes that touch this app, checked against `electron/` and `electron-builder.json`:

- **macOS 13 or later required.** `electron-builder.json` builds a `universal` mac target; no `minimumSystemVersion` is set, so the DMG's stated floor comes from Electron. Say so in the release notes and on the download page (`docs/`), since a macOS 12 user's update will fail to launch.
- **32-bit Windows and Linux ARMv7 binaries are gone.** The builder config targets x64 and universal only; nothing to change, but confirm `release.yml` publishes no ia32 artefact (it does not today).
- **`clipboard` module removed from the renderer, and the main-process API is now Promise-based.** `grep -rn clipboard electron/` finds nothing; the renderer uses `navigator.clipboard` and `clipboardData` (`UrlInput.tsx:53`). No change needed; the grep is the acceptance check.
- **`select-client-certificate` may fire with a null `webContents`.** Not handled anywhere in `electron/`; nothing to change.
- 43's changes (frameless windows default to rounded corners on Linux, dialogs default to Downloads, `NativeImage.toBitmap()` sRGB) do not touch code here, but the Linux rounded-corner default is visible: check the window on Linux once, and set `roundedCorners` explicitly if the frame changes.

Verify: `pnpm electron:dev` on each OS, the updater test suite (`updater.test.ts` stubs `process.platform`), and one packaged build per OS from `release.yml` on a branch.

## 2. Safe minors and patches - one commit, no code

[...]

PR #1408, opened 2026-09-04
build(app): eslint 9.39 to 10.9 - part 3 of the September sweep (#1335)

athrvk:
## Description

The **eslint group** of #1335 part 3, in the shape that issue's Sequencing asks for ("one PR each for `eslint`, `vitest`+`jest-dom`+`jsdom`, and the rest together"): `eslint` 9.39.5 to **10.9.1**, `@eslint/js` to **10.0.1**, `eslint-config-prettier` to **10.1.8**.

**Why now rather than at the next sweep.** ESLint 9.x reached end of life on 2026-08-06, so the gate CI enforces at zero warnings has been running an unmaintained line for a month. 10.9.1 (published 2026-08-24) is the newest release the seven-day `minimumReleaseAge` floor admits; 10.10.0 shipped today and is inside the window, so it is deliberately not taken.

**The plan, and the alternative rejected.** Three rules joined `eslint:recommended` in 10.0 and they report nine findings here. All nine are real defects rather than rule noise, so the approach is to fix the code, not to configure the rules off - which was the alternative, and it would have meant carrying three permanent `off` entries whose reason is "the code predates the rule". Seven are `preserve-caught-error`: six of them are two identical error ladders in `http-client.ts` (a hand-rolled copy of the same translation, which is the thing `CLAUDE.md` warns about), so they collapse into one `asTransportError` helper that both the buffered and the streaming path call; the seventh is the sidecar's give-up message. Two are `no-useless-assignment`: dead initialisers in `sidecar.ts` and `xml-format.ts`. `no-unassigned-vars` reports nothing.

Adopting the rule needed one type: `ErrorOptions` is not in the ES2020 lib, so `new Error(msg, { cause })` did not compile. `ES2022.Error` joins the `lib` list in all three tsconfigs - the runtime has had `cause` since Chromium 93 and Node 16 (`updater.ts` already reads it off a caught error), so the entry describes what the app runs on and the emit target stays ES2020.

### What changed

| Area | Change |
|---|---|
| `app/package.json`, lockfile | eslint 10.9.1, @eslint/js 10.0.1, eslint-config-prettier 10.1.8; `--ext ts,tsx` dropped from the `lint` script |
| `app/src/services/http-client.ts` | the two transport-error ladders become one `asTransportError`, every branch attaching the caught error as `cause` |
| `app/electron/sidecar.ts` | the restart give-up carries the failed attempt as `cause`; `lastError` becomes a `const` inside the catch that assigns it |
| `app/electron/main.ts` | `restartEngine` logs the error rather than its text, so that cause has a reader |
| `app/src/lib/xml-format.ts` | `end` declared without the `-1` every branch immediately overwrites |
| three tsconfigs | `ES2022.Error` added to `lib`; `tsconfig.node.json` states its list rather than inheriting it |
| `docs/app/building.md` | the ESLint line, and the `lib` entry per config with the reason |

**`--ext ts,tsx` is gone, measured rather than assumed.** It has been inert since flat config; run with and without it, `eslint .` reports the same **1396** files, because selection comes from the config's own `files` keys.

**Stating `lib` on the main-process config narrows it, and the commit says so.** `target: ES2020` alone implied `lib.es2020.full`, which also carries `ScriptHost`, `WebWorker.ImportScripts` and `DOM.AsyncIterable`; the four-entry list drops those three. The first two describe environments the main process is not, the third types `for await` over a DOM stream `electron/` never opens, and neither sibling config carried them - the three configs now agree. (An earlier commit message here called the rest of the list "what `target: ES2020` already implied"; `7a074fa` corrects that and records the probe.)

**Not bumped, with reasons.** `eslint-plugin-jsx-a11y` 6.10.2 is its own latest release and its peer range stops at eslint 9, so `pnpm install` prints one peer warning; there is no newer version to take. Its rules do still run - probed directly, an `` with no `alt` fails at the line - and `react-hooks`, `react-refresh`, `typescript-eslint` and `prettier/prettier` were probed the same way, since a plugin that silently stops running is the failure mode that would otherwise pass as green. `eslint-plugin-react-refresh` stays at 0.4.26: 0.5.6 is the release that stops flagging the Radix re-exports (see `9240ec8f`) and it matures under the seven-day floor on 2026-09-09. `typescript-eslint` 8.68 and `eslint-plugin-react-hooks` 7.1.1 already declare `^10`.

## Type of Change
- [x] Bug fix (seven `catch` sites that dropped the original error)
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Run locally on the head of this branch, all green:

- `pnpm lint` - clean (9 errors before the fixes, 0 after)
- the repo-root run over the JavaScript outside `app/` (`scripts/perf/*.mjs|cjs`) with `--config app/eslint.repo-js.config.mjs` - clean
- `pnpm type-check` (all three projects, tsc7) and `pnpm electron:compile` (tsc 5.9) - clean, which is what proves the `lib` entry satisfies both compilers
- `pnpm format:check` - clean
- `pnpm test` - **7658 tests in 675 files pass**, no pre-existing failures on this tree. (The first commit's message says 7654: that run predated the four cases below. 7658 is the number for the tree as it stands.)

Four assertions added and each mutation-checked (remove the `cause`, the case reddens):

- `http-client.error-shape.test.ts` - three cases, one per branch of the ladder: an abort surfaces as "Request timeout" with the abort on `cause`, a `TypeError` as "Network error: ..." with the rejection on `cause`, and a thrown non-error as "Unknown error occurred" with the thrown value on `cause`.
- `http-client.stream.test.ts` - the streaming path's idle timeout carries its abort, so the shared helper is locked on both call sites.
- `sidecar.test.ts` - the existing "restarted engine never becomes healthy" case now also asserts the thrown error's `cause`.

Every message a caller can see is byte-identical to before ("Request timeout", `Network error: ${message}`, "Unknown error occurred"), so the existing assertions on those strings across nine test files pass untouched.

**No user-visible change.** Nothing renders differently; the diff is lint tooling plus error plumbing behind messages that keep their exact text. No screenshots, considered and not applicable.

## Checklist

## Related Issues

Part of #1335 - part 3's eslint row. Deliberately **not** `Closes`: the remaining part 3 rows (`concurrently` 10, `jest-dom`+`jsdom`, `graphql` 17, `zod` 4, `lucide-react` 1, the `@types/node` decision and moving the emit scripts to `tsc7`) stay open on that issue.

Two notes that are tooling rather than tracked work, recorded here per `CLAUDE.md` rather than filed:

- `docs/app/building.md`'s Production Dependencies list still says "Electron 28"; it is 44 since #1373. Adjacent to this diff, not part of it.
- `vitest` no longer appears in `pnpm outdated` - the 5.0 row #1335 lists is not on the registry's latest line today, so whoever takes that group should re-derive it rather than trust the table.


athrvk:
Merged master in (`8d1e587`) because #1401, #1406/#1407 and #1404 landed while this was open and brought ~2000 lines of new `app/src` code that ESLint 10 had never seen - a bump whose whole risk is "the new rules find something" is only proven against the tree it will actually gate.

It finds nothing: on the merged tree `pnpm lint` is clean, `pnpm type-check` and `pnpm format:check` clean, and **7686 tests in 677 files** pass (up from 7658/675, which is the tests those three PRs brought with them). The three new `eslint:recommended` rules report zero findings in the merged-in code, so nothing there needed touching.

The Testing section above still describes this branch's own commits; the numbers for the current head are the ones in this comment.

---

Issue #1403, opened 2026-09-04
app(inbox): a background capture stream that spends its reconnect budget goes quiet with nothing on screen saying so


athrvk:
## Problem

#1400 (PR #1401) gave every running inbox with the `Notify` toggle on a live capture stream held app-wide by `app/src/services/inbox-watch-service.ts`, so a webhook that lands while the user is in another tab or another application raises its notification. Two ways that stream can stop without any surface saying so, verified at master `75c7277b`:

**1. The reconnect budget runs out.** The stream carries #506's bounded reconnect: after `INBOX_LIVE_MAX_RETRIES` (5, `inbox-watch-service.ts:71`) refused attempts it sets `stopped` (`:350-357`) and waits for `resume()` (`:228`), which nothing in the background calls. The ladder is `400/800/1600/3200/5000` ms with up to 50% jitter (`:101-107`), so six consecutive failures span roughly 11 to 17 seconds of an engine that is unreachable or refusing - a short outage, not a long one. The listener-is-gone check (`:364`) ends the stream without spending a retry when the inbox was stopped, so this needs an engine that is up and refusing, or briefly unreachable.

That give-up is right for a stream a surface is watching: the inbox view adopts the service's `stopped` state and renders "Live updates stopped" with a Resume button (`modules/inbox/index.tsx:386-401`). It is silent for a stream nothing is watching, which is the case the app-level watcher exists for. A background stream that has given up holds no socket (`disconnect()` `:294-299`), raises no notification (the notifier fires only from `receive()` `:336`), and the toggle still reads as on.

**2. The cap evicts.** `MAX_INBOX_WATCH_STREAMS = 8` (`:88`, reason at `:76-87`); `wantedInboxIds()` (`:248-257`) slices the union to the cap. The ninth notify-enabled running inbox is silently unwatched: no surface, no log line, the toggle still on. Rare, but the same shape.

Neither case is unrecoverable, and the user is simply never told: turning the toggle off and on, or stopping and restarting the inbox, drops the id from `streams` so the next reconcile opens a fresh one; opening the inbox tab shows the existing callout and Resume. Note that `captureCount` on the 10 s list poll does **not** keep rising while the user is in another application - `refetchIntervalInBackground` is unset in `lib/query-client.ts:33-47`, so the poll pauses in a hidden window.

## Fix pathway

Two candidates; the choice is the substance of this issue, and both have a constraint the first version of this issue missed.

[...]

PR #1413, opened 2026-09-04
feat(app): own the Dock and taskbar icon - menu, unread badge, run marks, drop-to-import

### LINKED ISSUE #1364: app(shell): OS icon integrations the app has none of - Dock menu and Windows Jump List, unread badge for inbox captures, failed-run overlay and flash, files dropped on the icon
## Problem

On master `75c7277b` (re-verified; first recorded at `711f359f`), `rg "app.dock|setUserTasks|setJumpList|setBadgeCount|setOverlayIcon|flashFrame|open-file|addRecentDocument" app/electron` returns nothing. The app's icon in the Dock and on the taskbar is inert: right-clicking it shows only what the OS puts there, it never carries a count or a state, and dropping a file on it does nothing. None of these is a defect, and none has a strong case on its own for an API client, which is why they are filed together, low priority, as one decision about how much of the OS icon Vayu wants to own. Each is a few lines against an existing signal.

## The five, with the signal each would show

1. **Dock menu (macOS) and Jump List (Windows).** `app.dock.setMenu` and `app.setUserTasks`. Items: New Request, and the three most recently opened collections, each opening the app on that collection. The recent list comes from `tabs-store`'s history (#1245) or the collections query; a Jump List task launches the app with an argument the main process must parse on `second-instance` (`main.ts:1220-1222` already handles focus; it would gain an argument reader). Linux has no equivalent since Unity support was removed in Electron 44.
2. **Unread badge for inbox captures.** `app.setBadgeCount(n)` on macOS (Linux support went with Unity in Electron 44; see the comment below), `setOverlayIcon` with a rendered count on Windows. `n` is captures arrived while the window was unfocused and not yet viewed; clears when the Inbox view is shown. Pairs with the per-inbox capture notification from #1388: `app/src/modules/inbox/capture-notifier.ts` (`createCaptureNotifier`) is where a capture is recorded while unfocused. Since #1400 (PR #1401) the stream is held app-wide: `app/src/services/inbox-watch-service.ts` owns one `CaptureNotifier` per stream (created at `:276`), and the badge increments in `InboxWatchService.receive` where `stream.notifier.record(capture)` is called (`:336`), which fires for every running inbox whose `Notify` toggle is on, whether or not its tab is open. The source the badge needed exists. Note the cap of 8 streams (`:88`) and #1403's stopped-stream case; a badge that says "0" for an evicted or stopped stream is the same silence.
3. **Failed-run state.** After a load or collection run ends failed while the window is unfocused: `setOverlayIcon` with a red mark on Windows, `app.dock.bounce("critical")` once on macOS, cleared when the window regains focus. #1358's notification already tells the user why (`NOTIFY_KINDS.loadRunFailed`, `load-test-service.ts:265`); this is the glance version. #1362's `runProgress.fail` (`services/run-progress.ts`) fires at the same point and already paints Windows' error mode on the taskbar for two seconds; the overlay is the persistent form of that.
4. **Flash frame on completion.** `mainWindow.flashFrame(true)` when a run finishes unfocused on Windows and Linux, `false` on focus. Overlaps with 3 and with #1358; include only if the notification setting is off and the user wants a quieter cue.
5. **Files dropped on the icon.** `app.on("open-file")` on macOS and the file argument on Windows (`process.argv` on first launch, `second-instance` argv afterwards) route to the existing import path (`ImportModal`'s `readSpecFile` / collection importers, `electron/spec-file.ts`, `electron/data-file.ts`), so a Postman export or an OpenAPI document dropped on the Dock or taskbar icon opens the import dialog with the file preselected. This is the one of the five that adds a capability rather than a cue.

## Fix pathway

One module, `app/electron/os-icon.ts`, exporting small functions each guarded by `process.platform` and by `Notification`-style support checks, with a `clearAll()` called on focus and on quit. Renderer signals arrive through the IPC channels the run and inbox features already have: `electron/power-save.ts` and `services/wake-lock.ts` (`power:hold` / `power:release`, from #1357), `electron/notify.ts` and `services/notify.ts` (`notify:show`, `notify:availability`, `notify:activated`, `notify:test`, from #1358), `electron/run-progress.ts` and `services/run-progress.ts` (`runs:progress`, from #1362), and the terminal handlers `notifyTerminal` in `load-test-service.ts:202` and its `scenario-run-service.ts:207` counterpart, which already call `runProgress.clear`/`fail` (`load-test-service.ts:166`, `:263`). Copy `electron/run-progress.ts`'s shape: platform injected as a dependency, never read from the host, one focus decision in main, a renderer service that posts, and a platform test that drives all three branches. No new store. Take 5 first if only one lands; take 1 second.

## Tests

- Per function, platform stubbed all three ways: the macOS branch calls `app.dock.*`, the Windows branch calls `setOverlayIcon`/`setUserTasks`, Linux calls nothing for the badge; unsupported platforms call nothing.
- `open-file` and argv routing: a `.json` or `.yaml` path reaches the import handler with the file, a non-file argument does not; `second-instance` with a path focuses and imports.
- Focus clears badge, overlay and flash.
- `pnpm test os-icon startup-order`, `pnpm type-check`, `pnpm lint`.

## Docs, same commit

- `docs/architecture.md`: the OS icon behaviours per platform, in the same table style as the menu, notification and window notes (the wake-lock paragraphs are at `:151-173`, notifications at `:175-210`, the OS progress indicator at `:212-230`).
- `docs/app/COMPONENTS.md` Inbox entry (`:638-770`; `capture-notifier.ts` is the bullet at `:675-687`, `services/inbox-watch-service.ts` at `:697-713`) for the badge, and the import entry for drop-on-icon.

## Sequencing

#1357, #1358, #1362 (PRs #1397, #1406, #1407) and #1400 (PR #1401) are all merged, so every signal this reads exists. Nothing here is blocked. #1403 (stopped or evicted background streams) touches `inbox-watch-service.ts` where item 2 hooks in; land either first, the overlap is one call site. Independent of the context-menu and menu-bar issues.

## Acceptance criteria

- Right-clicking the Dock icon on macOS or the taskbar icon on Windows lists New Request and the recent collections, and each opens the app there.
- Two inbox captures arriving while the window is unfocused show "2" on the Dock icon and a count overlay on Windows, cleared on opening the Inbox.
- A load run that fails while unfocused marks the icon and the mark clears on focus.
- Dropping a Postman collection on the Dock or taskbar icon opens Vayu with the Import dialog on that file.

---

### LINKED ISSUE #1414: app(shell): flash the taskbar when a run finishes unfocused - the quieter cue #1364 deferred
## What this is

Item 4 of #1364, the one of that issue's five it scoped conditionally rather than outright: `mainWindow.flashFrame(true)` when a run finishes while the window is unfocused, `false` on focus. The other four landed in PR #1413; this did not, and this issue is that deferral's record rather than a comment saying "later".

The issue's own wording is the reason it was held: *"Overlaps with 3 and with #1358; include only if the notification setting is off and the user wants a quieter cue."* That is a settings question, and PR #1413 answers no settings questions - it wires signals the app already had to surfaces the app did not.

## Why it is not simply the fifth line of the same module

Three things have to be decided that the other four items did not need:

- **It is conditional on a preference that does not exist.** "Include only if the notification setting is off" means reading `systemNotifications` from `client-settings-store` and flashing only when the user has declined the louder cue. That store is the renderer's (localStorage-backed, and main cannot see it), so the condition belongs on the renderer side of `icon:signal` while every other decision in `electron/os-icon.ts` is deliberately main's. Either the renderer sends a signal main paints unconditionally - which breaks that module's split - or main gains a second input.
- **It overlaps item 3 rather than adding to it.** On Windows a failed run already marks the taskbar overlay (`electron/os-icon.ts`, `failedOverlayBitmap`), and #1362's `runProgress.fail` already paints the error state on the taskbar button for two seconds. A flash on top of both is a third thing saying the same thing.
- **macOS flashes until told to stop.** Since Electron 31, `flashFrame(true)` on macOS bounces continuously until `flashFrame(false)` - noted on #1364 - so the clear is not optional and is the same window-focus clear item 3 already owns. That part is cheap: `main.ts` now has a `focus` handler on the window (added in #1413) and `OsIconPainter.focused()` behind it.

Also worth stating: unlike the badge and the overlay, **this one does work on Linux.** `flashFrame` was not removed with Unity support, so this is the only one of #1364's five with something to show on all three platforms, which is a point in favour of eventually taking it.

## Fix / solution pathway

`electron/os-icon.ts` gains one signal (`{ kind: "runFinished" }` or a flag on the existing `runFailed`) and one branch calling `window().flashFrame(true)`, cleared in the existing `focused()` alongside the failed mark. The condition - is the notification setting off - is read in `services/os-icon.ts` before sending, the way `services/notify.ts` reads the opt-in before posting, so main keeps its single question. Decide first whether a *finished* run flashes or only a *failed* one; #1364's wording says "when a run finishes", which is the broader of the two and the noisier.

Rejected before it is proposed: flashing in addition to the Windows overlay mark with no condition at all. That is three cues for one event on the platform that already has two.

## App changes (required)

`app/electron/os-icon.ts` (the branch and the clear), `app/src/services/os-icon.ts` (the setting check and the send), and one call site in `load-test-service.ts` / `scenario-run-service.ts` beside the `osIcon.runFailed()` calls #1413 added.

## Tests

`app/electron/os-icon.test.ts` - platform-stubbed all three ways, and unlike every other case in that file **Linux is not a no-op here**, which is worth an explicit case. Focus clears the flash. `app/src/services/os-icon.test.ts` - the setting off sends, the setting on does not. Mutation check: remove the clear and the macOS case keeps flashing.

## Docs (same commit)

`docs/architecture.md`, the OS-icon paragraphs added by #1413 - they currently state what each platform shows and say Linux gets nothing, which this makes untrue for one row.

## Sequencing

After #1413 merges; it is the PR that creates every file named above. Nothing is blocked on this. Parallel-safe against everything else open.

## Acceptance criteria

- [ ] A run finishing while the window is unfocused flashes the taskbar button on Windows and Linux and bounces the Dock on macOS, and only when the system-notification setting is off.
- [ ] The flash clears when the window regains focus, on macOS in particular, where it otherwise continues indefinitely.
- [ ] The three-platform test covers Linux as a platform that *does* paint, not as a no-op.
- [ ] `docs/architecture.md`'s per-platform statement is corrected for the Linux row.

Deferred from #1364, whose other four items landed in #1413.

### LINKED ISSUE #1415: app(runs): a failed run reports itself as finished - the client discards the status the engine sends, so nothing that reacts to a failure fires
## Problem

Verified at master `75c7277b`, and found while wiring #1364's failed-run icon mark, which is the third feature to depend on this and the third not to work.

**The renderer cannot currently tell a failed run from a finished one, and always says finished.** Every terminal state of a live run - completed, stopped, failed - reaches `LoadTestService.handleClose`, and it reports one kind unconditionally:

```ts
// app/src/services/load-test-service.ts:328-332
this.notifyTerminal(
    runId,
    NOTIFY_KINDS.loadRunFinished,
    summary ?? "The run ended, but its report could not be read."
);
```

It has just awaited the run's report immediately above, and that report carries `metadata.status` (`app/src/types/domain.ts:1757`), which is `"Failed"` for a failed run. The field is never read. `scenario-run-service.ts` has the same shape with `collectionRunFinished`.

`handleError` - the only place that produces `NOTIFY_KINDS.loadRunFailed`, calls `runProgress.fail` and (since #1364) marks the OS icon - is reachable only from the `try/catch` around the synchronous `new EventSource(url)` constructor (`sse-client.ts:345`). A run that fails on the engine does not go through it.

## Why the client cannot tell, and where the answer already is

`sse-client.ts` registers one `complete` listener and ignores the frame's body:

```ts
// app/src/services/sse-client.ts:311-316
this.eventSource.addEventListener("complete", () => {
    onMessage({ ...this.currentMetrics });
    this.disconnect();
    onClose();
});
```

The engine emits `complete` from two places, and **one of them already carries the status**:

- `engine/src/http/routes/metrics.cpp:309-314` - the run was already terminal when the client connected. This frame sets `completion_event["status"] = to_string (run->status)`, and `RunStatus::Failed` is one of the values it is emitted for (`:309`).
- `engine/src/http/routes/metrics.cpp:434-438` - the live stream reaching its end, which is the ordinary case. This frame carries `event` and `runId` and **no status at all**.

So the wire is inconsistent with itself, and the client discards the half that is populated. Both halves need fixing, and the second is the one that matters: the common path is a run that fails while the client is watching it.

The native `error` listener is not a second door either - it also converges on `onClose()` (`sse-client.ts:331-341`), deliberately, because the canonical recovery is the stored report.

## What is broken downstream

Three shipped features are wired correctly to a signal that never arrives:

1. **#1358** - the system notification for a failed run. A failed run notifies "Load test finished", with the failure's own summary line, if any.
2. **#1362** - `runProgress.fail`, the Windows taskbar error state. Never painted for a real failure.
3. **#1364** - `osIcon.runFailed()`, the Dock bounce and the taskbar overlay mark (PR #1413). Same.

None of the three is wrong in itself; all three are gated on `NOTIFY_KINDS.loadRunFailed` / `collectionRunFailed` reaching `notifyTerminal`, which the paragraph above explains it does not.

## Fix / solution pathway

1. **Engine**: give `metrics.cpp:434-438`'s completion frame the same `status` field `:309-314`'s already has, so the two `complete` emitters agree on their shape. One line, and it makes the wire self-describing rather than requiring the client to go and ask.
2. **App**: parse the `complete` frame in `sse-client.ts` and hand the status to `onClose`, then have each service pick the terminal kind from it. Do not read it from the report instead: the report fetch can fail (`handleClose` already handles that, and falls back to a summary-less notification), and a failure that could not be read must still report a failure rather than silently reporting success.
3. Belt and braces where it costs nothing: when the frame carried no status and the report *did* load, `metadata.status` is a correct second source. Prefer the frame.

Rejected: inferring failure from the absence of metrics, or from a non-empty error list in the report. Both are guesses about what the engine already knows and is willing to say.

## App changes (required)

`app/src/services/sse-client.ts` (parse the frame, widen `onClose`), `app/src/services/load-test-service.ts` and `app/src/services/scenario-run-service.ts` (choose the kind). `app/src/services/run-progress.ts` and `app/src/services/os-icon.ts` need no change - they are already correct and will start firing.

## Engine changes

`engine/src/http/routes/metrics.cpp` - the second completion frame gains `status`.

## Tests

- `sse-client` - a `complete` frame carrying `status: "Failed"` reaches `onClose` as a failure; one carrying no status does not claim a failure.
- `load-test-service` / `scenario-run-service` - a failed terminal close posts `loadRunFailed` / `collectionRunFailed`, calls `runProgress.fail` and `osIcon.runFailed`; a clean one posts finished and does none of those. Mutation check: revert the kind selection and the failed case reddens on all three.
- Engine: the live-stream completion frame carries the status, asserted on the stream rather than on the stored row.

## Docs (same commit)

`docs/engine/api-reference.md`, the run stream's `complete` event - it should state that the frame carries the run's status, since one of the two emitters already does and the doc describes one event.

## Sequencing

Independent of everything currently open. PR #1413 (#1364) does not need to wait for this and this does not need to wait for it: #1413's icon mark is already wired to the signal this makes reachable, so the two compose with no shared line. Land either first.

## Acceptance criteria

- [ ] Both of the engine's `complete` frames carry the run's status.
- [ ] A run that fails while the client is watching it produces `loadRunFailed` / `collectionRunFailed`, not the finished kind.
- [ ] The system notification, the taskbar error state and the OS icon mark all fire for that run, with a test for each proving it through the SSE path rather than by calling `handleError` directly.
- [ ] A terminal close whose report fetch failed still reports a failure when the frame said so.

athrvk:
## Description

The Dock and taskbar icon was inert. On master `75c7277b`, `rg "app.dock|setUserTasks|setBadgeCount|setOverlayIcon|flashFrame|open-file" app/electron` returned nothing: right-clicking the icon showed only what the OS puts there, it never carried a count or a state, and dropping a file on it did nothing.

**Root cause and approach.** There is no defect in #1364 itself - the surfaces were never wired - so the question it really asks is *how much of the OS icon Vayu should own*, and the answer that stays maintainable is that all five items are one split, not five features. Every one is main answering something the renderer cannot: whether the user is looking (the window's focus, which `document.hasFocus()` cannot tell from minimized or behind another app) and what this platform can show (three surfaces, one of which Electron 44 emptied). That is `run-progress.ts`'s split, so this copies its shape rather than inventing one - a `createX(deps)` factory with every Electron surface injected, a `parseX` that drops a malformed message rather than throwing, `registerXIpc` with `createRendererWatch` so the marks die with the renderer that asked for them, and a platform-injected three-branch test.

**The alternative I rejected**, and why the diff is shaped as it is: #1364 proposes one module. Two of the five *hand the app something* rather than *show it something*, and they share no state, no platform guard and no test harness with the other three - but they share everything with each other, since a Jump List task and a dropped document both arrive as a command line through the same three doors. So they are `electron/open-intent.ts`, one reader for all of it, and the Dock menu's own clicks go back out through it too rather than on a second channel: a click on **Payments** in the Dock menu and the Jump List task beside it are one request, and the buffering a click with no window yet needs is the buffering a cold launch's file needs.

I also rejected shipping the Windows overlay as PNG assets. `setOverlayIcon` takes a `NativeImage`, so a count is a picture the app must produce; eleven binaries under `shared/` would be regenerated by hand and reviewed as blobs. `electron/os-icon-overlay.ts` draws them into a BGRA bitmap instead - no Electron import, no file system, no state - which is the only form a test can read back. [What it produces is in this comment.](https://github.com/athrvk/vayu/pull/1413#issuecomment-5545775663)

### What the icon carries

| | macOS | Windows | Linux |
|---|---|---|---|
| unread captures | `app.setBadgeCount` | `setOverlayIcon`, a drawn count | - |
| a failed run | `dock.bounce("critical")` | `setOverlayIcon`, a drawn mark | - |
| the icon's menu | `dock.setMenu` | `app.setUserTasks` | - |
| a run that ended | `dock.bounce("informational")` | `flashFrame` | `flashFrame` |

1. **Dock menu and Jump List.** New Request plus the three most recently opened collections, from `tabs-store`'s `navHistory`, deduped, newest first, with any id no longer in the collections query dropped so a deleted collection cannot appear. Both entries carry an argument (`--vayu-new-request`, `--vayu-open-collection=<id>`), because a task with none is an ordinary launch that opens the app rather than opening anything in it.
2. **Unread badge for inbox captures.** Counted only while the window is unfocused; cleared when the Inbox is on screen, not merely when the window is focused, because those captures are still unread. `setBadgeCount` is macOS-only since Electron 44, per the correction on the issue.
3. **Failed-run state.** Only while the user is elsewhere, cleared on the window's `focus`. Hooked into each run service's `notifyTerminal` rather than beside `runProgress.fail`: that method is guarded by `notifiedRunId`, so it fires exactly once per run, and every terminal path reaches it.
4. **Flash on completion** (folded in from #1414). Only when the system notifications are off - it substitutes for the toast the user declined rather than joining it - so the opt-in is read renderer-side, where it lives. **This is the one row Linux still has**, so it is answered ahead of the gate the other three share.
5. **Files dropped on the icon.** `open-file` on macOS, `process.argv` and `second-instance` argv elsewhere. Main carries the path and never reads the file; the renderer reads it back through the gated `specFile:read` channel the `$ref` bundler already uses, and feeds one `BatchDocument` into the ledger a picked file goes into.

### Three bugs the review phase caught, all fixed here

- **A failed run said it finished, so nothing reacted to a failure** (#1415, folded in). The renderer could not tell the two apart: `handleClose` reported `loadRunFinished` unconditionally, having just awaited a report whose `metadata.status` can be `Failed` and never reading it, while `sse-client` discarded the `complete` frame's body. `handleError` - the only producer of the failed kinds - is reachable only from the `try/catch` around the `EventSource` constructor. **Three shipped features were wired to a signal that never arrived**: #1358's failed-run notification, #1362's taskbar error state, and this branch's icon mark. The engine already knew - `write_stats_frame`'s completion event has always carried the status; `stream_live_metrics`' has not, and that is the frame an ordinary live run ends on. It now carries the same field, omitted rather than guessed when the row is not terminal yet. The client reads it: the load service prefers the frame and falls back to the stored row (its fetch can fail, and a failure that could not be read must still report a failure); the collection service has only the frame, because it notifies before it fetches and waiting would trade that promptness away.
- **New Request off the menu did nothing.** It went through `commandById("new-request").perform(baseCommandContext())`, copying what `useNotificationActivation` does for Settings. But that command declares `available: (ctx) => ctx.surfaces !== undefined` - the flow can need a collection picker, and a picker needs a mounted host - and `baseCommandContext()` carries no `surfaces` by design. `registry.test.ts:109` already pins exactly that. So `perform` was `ctx.surfaces?.newRequest()`: a silent no-op. It now re-dispatches the chord to the one window handler, which is what `editor-chords.ts` exists for, and the test asserts the keydown reaching `window` matches `NEW_REQUEST_CHORD` rather than mocking the call.
- **Stale mocks in the run-service suites** hid the new call site until the full suite ran. Extended, and both suites now *assert* the cues rather than only stubbing them.

### Deliberate deviations from the issue specs

- **Two modules rather than one**, for the reason above.
- **macOS gets an informational bounce, not `flashFrame`** for item 4, though it has `flashFrame`: there a flash bounces the Dock continuously until something turns it off, which is the weight `bounce("critical")` already gives a *failed* run, and reusing it would make the two indistinguishable. The informational bounce ends by itself, so it is the one cue here needing no clear. (#1414's own acceptance text said "bounces the Dock on macOS" without saying which bounce; this is that decision made.)
- **Windows' one overlay carries both marks**, the failure winning while it stands and the count returning underneath it on focus. #1364 names both surfaces without saying which wins; this is the rule, and it is asserted.
- **The badge counts every capture a held stream sees**, a superset of what notifies: an inbox whose per-inbox Notify toggle is off still counts while its tab is open. That is the hook point the issue names (`inbox-watch-service.ts:336`), and it is right - a badge is a number on an icon, not an interruption, so declining the interruption is not declining the count.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Run here on `fd76752`:

- `cd app && pnpm test` - **688 files, 7844 tests, all passing** (7825 on master).
- `cd engine && ctest --preset linux-dev` - **2908 / 2908 passing**, after `python build.py -e -t`.
- `pnpm type-check`, `pnpm lint` (`--max-warnings 0`), `pnpm format:check` - clean.

New suites: `electron/os-icon.test.ts` (44), `electron/os-icon-overlay.test.ts`, `electron/open-intent.test.ts`, `src/services/os-icon.test.ts` (8), `src/hooks/useOsIcon.test.tsx`, `src/hooks/useOpenIntent.test.tsx` (5), `src/modules/collections/ImportModal.pending-path.test.tsx`, plus cases added to `electron/startup-order.test.ts`, `src/services/sse-client.test.ts` and both run-service suites.

Every platform-dependent case names its platform rather than reading the host's, per `app/CLAUDE.md`: a Linux CI runner cannot pass a macOS assertion by accident. Behavioural cases carry their mutation check as a comment - including the `startup-order` guard, which reddens if `app.on("open-file")` is moved inside `whenReady` (a no-op everywhere but macOS, where it loses exactly the case the event exists for).

**One coverage gap, stated rather than filed** (`app/CLAUDE.md` puts test shape in a PR body, not the tracker): the engine's new `status` field on the *live* completion frame has no engine-side assertion. No fixture drives `GET /runs/:id/live` today - `stats_route_test.cpp` calls handlers directly with no HTTP server, and no test file registers the metrics routes - so asserting it needs a streaming fixture that does not exist. The eight lines mirror `write_stats_frame`'s already-tested emitter, and the contract that bit users is asserted end to end on the client (`parseTerminalStatus`, plus both services' failed-close cases). A follow-up that builds that fixture would be welcome; it is more work than this diff.

**No screenshot of the feature, deliberately.** Every surface added is native OS chrome on macOS and Windows; Linux paints nothing for three of the four rows by design, and this container is Linux, so there is no visible surface here to capture. The drawn overlay images - the one part that *can* be shown honestly - are in [this comment](https://github.com/athrvk/vayu/pull/1413#issuecomment-5545775663), rendered by importing the shipped module rather than screenshotted. `pnpm electron:dev` on macOS and Windows is what would exercise the rest.

## Checklist

## Related Issues

Closes #1364
Closes #1414
Closes #1415


athrvk:
## The Windows overlay images, rendered from the shipped code path

`BrowserWindow.setOverlayIcon` takes a `NativeImage`, so the unread count and the failed mark are pictures this app has to produce. `electron/os-icon-overlay.ts` draws them into a BGRA bitmap rather than shipping eleven PNGs under `shared/`. Below is what that code actually produces, so the choice can be judged on the result rather than on the argument for it.

The five overlay states: discs carrying 1, 2, 9, 9+ and an exclamation mark

Left to right: one, two and nine captures; twelve captures, which is where `OVERLAY_MAX_COUNT` gives up on being exact and says `9+`; and the failed-run mark, on its own darker disc so the two states are one glance apart rather than one colour apart.

**What this is and is not.** These are the exact bitmaps `countOverlayBitmap()` and `failedOverlayBitmap()` return, produced by importing the shipped module, writing its `Uint8Array` out and converting BGRA to PNG - nothing was drawn by hand for this comment. They are shown at 4x nearest-neighbour so the glyph shapes are visible; Windows draws them at 16x16 in the corner of the taskbar button.

**They are not a screenshot of Windows, and there is no screenshot in this PR.** Every surface this PR adds is native OS chrome on macOS and Windows only - the Dock badge and menu, the taskbar overlay and Jump List - and the Linux branch paints nothing at all by design, since Electron 44 removed Unity launcher support. This container is Linux, so the feature has no visible surface here to capture. What stands in for it is the platform-stubbed tests: `electron/os-icon.test.ts` drives all three branches by name rather than reading the host's, which is what `run-progress.test.ts` established and the reason a Linux CI runner cannot pass a macOS assertion by accident.

The one renderer-side surface, the Import dialog opening on a dropped document, is reachable only by dropping a file on a Dock or taskbar icon, so it has the same constraint. `ImportModal`'s new path ends in the same `runBatch` call the File tab already used, which is what its existing suite covers.

Verification run here on `e591b00`: `pnpm test` 688 files / 7825 tests passed, `pnpm type-check`, `pnpm lint` and `pnpm format:check` clean, `mkdocs build --strict` clean.

The images are committed at `7e26bdb` and deleted in the commit after it, so they render permanently while staying out of **Files changed**.

---

athrvk:
## Re-verified after merging master, and a note on CI

**Merged `origin/master` (`515a3fc`) in.** #1411 landed on `inbox-watch-service.ts` and its `COMPONENTS.md` entry while this was open. The code merged cleanly; the one conflict was in `docs/app/COMPONENTS.md`, where both sides extended the same bullet - master rewrote its closing sentence into the background-resume paragraph (#1403/#1412), and this branch appended the `services/os-icon.ts` bullet after it. Resolved by keeping master's rewrite in full, with this branch's bullet following it; neither side lost a line. The badge call site survived the auto-merge intact and sits after `stream.notifier.record(capture)` as before.

Re-run on the merge commit `119b121`:

- `cd app && pnpm test` - **689 files, 7859 tests, all passing**.
- `pnpm type-check`, `pnpm lint` (`--max-warnings 0`), `pnpm format:check` - clean.
- `cd engine && ctest --preset linux-dev` - **2908 / 2908**, run on `fd76752`; the merge touched no engine source, so it stands.

(The PR body's figures are from `fd76752`, before the merge - these supersede them.)

## CI has not run since the first commit, and I could not make it

`PR CI - run tests` ran once, on `d0898678`, and failed on `Lint frontend` - the Prettier reflow of two files I authored, fixed in the very next commit. Every push since then has produced **no run at all**: the PR's combined status is `pending` with zero checks on `119b121`, and only `Apply path labels` appears.

The likely cause is the trigger. `labeler.yml` uses `pull_request_target`, which fires for these pushes; `pr-tests.yml` uses plain `pull_request`, whose `synchronize` events GitHub suppresses for pushes made with an app token - the same guard that stops workflows recursing. The `opened` event got through when the PR was created, which is why exactly one run exists.

I have deliberately not worked around it: an empty commit or a close-and-reopen to kick CI are both off the table. **Any push you make yourself, or a re-run from the Actions tab, should produce a real run** - and `CI gate` is a required check, so it needs one before this can merge. Everything that gate runs is green locally on this exact commit, per the numbers above, but that is my word rather than the runner's and the two are not the same thing.

---

athrvk:
## Correcting my last comment: CI was not suppressed, and the title check was failing

My previous comment said pushes here were producing no CI runs and blamed the `pull_request` vs `pull_request_target` trigger. **That was wrong**, and the run list says so: `PR CI - run tests`, `PR title`, `CodeQL` and `Docs site` all fired on `119b121` the moment it landed. The quiet stretch was the PR sitting as a **draft** - only `Labeler`, which is `pull_request_target`, ran during it. Nothing about the token.

The real failure I had missed underneath that: **`PR title` has been red since the first commit**, and it was a genuine one.

```
The subject "OS icon integrations - Dock menu and Jump List, unread capture badge,
drop-on-icon import" should start with a lowercase letter
```

`pr-title.yml` runs `amannn/action-semantic-pull-request` with `subjectPattern: ^(?[A-Z]).+$`, and my subject opened on "OS". Retitled to `feat(app): own the Dock and taskbar icon - menu, unread badge, run marks, drop-to-import`, which is the same shape the recently merged PRs use - one scope, lowercase subject. Editing the title re-runs that check on its own; no push was needed and none was made.

`Docs site` is already green on this head. `PR CI - run tests` and `CodeQL` are running now, and I will act on whatever they report rather than leaving them.

Apologies for the noise in the previous comment - the local runs it cites are still accurate, but its explanation of CI was not.

---

athrvk:
## CI is green on `119b121`

`CI gate` **passed**, and the PR is `clean` - no conflict, nothing outstanding.

| | ubuntu | macOS | windows |
|---|---|---|---|
| App | pass | pass | pass (2 shards) |
| Engine build + C++ tests | pass | pass | pass |
| clang-tidy on the changed engine source | pass | - | pass |

Plus `App quality checks` (lint, repo-JS lint, format, type-check), `Engine formatting`, `Docs site`, `CodeQL`, and `Conventional title` after the retitle.

That is the runner confirming what I had only run locally, on all three platforms rather than this container's Linux alone - including the C++ change on macOS and Windows, which I could not build here.

The one red mark still visible in the checks list is the superseded `Conventional title` run from before the retitle; its replacement above it is green. Nothing to do about it - a check run is not editable in place - and it does not gate the merge, as `CI gate` passing shows.

Ready for review.

---

PR #1424, opened 2026-09-05
chore(release): 0.26.0

athrvk:
## Description

Cuts **0.26.0**. Two files' worth of change and nothing else: the version bump (`VERSION`, `engine/CMakeLists.txt`, `engine/include/vayu/version.hpp`, `app/package.json`) and the curated release notes at `.github/release-notes/v0.26.0.md`, which `release.yml` reads as the GitHub Release body on tag push.

`v0.26.0.md` already existed on master as a stub started by the Electron 44 PR (#1335), which had to state the new macOS floor somewhere a user reads before updating. This commit does what its own header comment asked for: dates it, and extends it from `git log v0.25.0..master` (258 commits).

The release is themed as **desktop integration** - system notifications, the taskbar and Dock indicator, a wake lock, right-click menus, global Back/Forward and OS document handover - on top of the Electron 44 platform floor. The macOS-13 requirement and the application-identifier change are both kept as leading `### Changed` bullets, since they are the two entries an existing user has to read before updating.

> [!IMPORTANT]
> **Merge athrvk/vayu#1423 first.** That is this release window's vcpkg baseline bump (cpp-httplib 0.53.1 -> 0.54.1, curl 8.21.0 -> 8.22.0), deliberately kept out of this PR so a dependency change and a version bump cannot break a platform together and be indistinguishable. Once it lands I will rebase this branch onto master so CI here runs against the baseline the release will actually build with.

### Tagging, after this merges

Per the `releasing` skill, and **not** done on this branch - a squash/rebase merge rewrites the hash, so a tag here would point at a commit that never reaches master:

1. Wait for the merge commit's **Warm build cache** run to go green. `VERSION` is in that workflow's `push` paths, so merging this fires a warm run that compiles the engine into the master sccache scope; the tag sits on the same commit, and a tag build started before the warm run finishes reads an empty scope and compiles cold (#700). ~15-25 min per platform.
2. `git checkout master && git pull && git tag v0.26.0 && git push origin --tags`.
3. Read the sccache hit rate in the release run's log and expect more than zero engine hits (#659 item 5, #700).

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update (release notes; the rest is a version string)

## Testing

A release commit needs no test run - every edit is a version string or Markdown. The one check the skill calls for was made:

```
$ ./engine/build/vayu-engine --help
Vayu Engine 0.26.0
```

## Checklist

- [ ] I have added tests (if applicable) - n/a

## Related Issues

n/a - release chore. Notes reference the issues the release closes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CdzN6CbBVwkWViZer9Lmde)_


athrvk:
#1423 has merged, so the blocker in the description is cleared. Rebased this branch onto master (`8c69e0a`) and pushed - CI here now runs against the baseline the release will actually build with.

While rebasing I added one bullet to the notes: the release ships cpp-httplib 0.54.1 and curl 8.22.0, and the cpp-httplib move has a user-facing consequence worth stating rather than folding into internal churn. 0.54.0 wrapped the per-connection serving loop in a guard that stops an exception thrown by a handler escaping the worker thread; before it, only `routing()` was inside a try/catch, so a throw from a content provider or a logging handler reached a task queue with no catch of its own and terminated the process. The webhook inbox overrides that loop, so it takes the guard too - and it is the one listener that may bind past loopback.

Nothing else changed: still the version bump plus the notes file.

---

athrvk:
Review of `v0.26.0.md` at the rebased head against master `90c2fff3`. The notes are the user-facing record, so three sentences that describe behaviour the code does not have need changing before the tag:

1. **Wake lock threshold.** The Added bullet says "a load run declaring five minutes or more asks once as it starts ... Five minutes because it is under the shortest sleep timer a laptop ships with". The Changed section then says "Vayu asks about a long-running run at ten minutes, not five" (#1371). The code is ten: `LONG_RUN_SECONDS = 600` at `app/src/modules/dashboard/utils/keepAwake.ts:40`. The Added bullet should say ten and carry #1371's reason, and the one-line Changed bullet can go, since a first release of the feature has nothing to change from.

2. **Progress indicator hand-back.** The Dock and taskbar bullet says "the most recently started owns it and the other takes it back when that one ends". It does not, deliberately: `run-progress.test.ts:203` pins "leaves no bar behind when the run that took over ends", with the reason in the test (the superseded run's stream is already closed, so a fraction painted from it could never advance). Say the newer run owns the indicator and the bar clears when it ends.

3. **Linux window manager.** The internal-work bullet says the weekly performance harness "gained a window manager on the Linux leg". It did not: `.github/workflows/perf-measure.yml:310-311` records that running openbox "changed nothing, so none is carried here", and #1347's fix (#1422) bounds the wait instead. Say the harness measures the packaged app's cold start and that the Linux leg reports why it cannot produce a figure, or drop the window-manager clause.

Everything else I checked against the verifications of this window holds: the notification defaults and focus rule, the inbox capture batching and resume, the icon badge and marks, the close prompt, the installer fix, the httplib and curl bump, and the supersession hand-off.

---

---

Issue #1377, opened 2026-09-04
app(deps): take the Electron 44 patch line once the release-age floor clears


athrvk:
## What this is

#1373 landed Electron **44.0.0**, which is the newest release the seven-day release-age floor admitted at the time: 44.1.0, 44.1.1 and 44.2.0 were all inside the window. The floor now lives in `app/pnpm-workspace.yaml` as `minimumReleaseAge: 10080` (it was `.npmrc`'s `minimum-release-age` when #1373 merged; pnpm 11, which CI pins since master `c41ea644`, reads pnpm settings only from the workspace file). The lockfile pins what it resolved, so a later patch is a deliberate bump, not something a routine `pnpm install` picks up.

Measured against the floor on 2026-09-04:

| Version | Released | Mature at |
|---|---|---|
| 44.0.0 | 2026-08-25 00:45Z | 2026-09-01 00:45Z (shipped in #1373) |
| 44.1.0 | 2026-08-31 13:54Z | 2026-09-07 13:54Z |
| 44.1.1 | 2026-09-01 18:25Z | 2026-09-08 18:25Z |
| 44.2.0 | 2026-09-04 03:15Z | 2026-09-11 03:15Z |

## Why it is worth a bump rather than waiting for the next sweep

No CVE is named in 44.1.0, 44.1.1 or 44.2.0. What they carry is crash fixes, and three of them reach this app - checked against what `electron/` actually calls:

- **44.1.1: an intermittent Linux startup crash** - a race between Pango and the main thread initialising fontconfig. This is the one that needs no app code to hit it, and Vayu ships an AppImage and a deb.
- **44.1.0: crashes when a `webContents` or parent window is destroyed while still referenced.** `oauth.ts` creates and destroys a sign-in window per flow.
- **44.1.0: region-qualified `--lang` values such as `de-DE` loading no locale resources on macOS.**

The Chromium bumps they carry (152.0.7977.54 to .65 to .76) are upstream backports with no advisory attached.

[...]

Issue #467, opened 2026-08-10
app(tooling): adopt TypeScript 7 (native tsc) in stages - the type-check gate can take the speedup now, the full swap waits for 7.1 to unblock typescript-eslint


athrvk:
Owner prompt, 2026-08-10: TypeScript 7 shipped (GA 2026-07-08) with the Go-native compiler as a drop-in `tsc`, reporting 8-12x faster type-checks. Assessment verified at master `054059a`; ecosystem claims are per the 7.0 release guidance and should be re-checked against the official TypeScript release notes at implementation time.

**Status refresh (verified at master `90c2fff3`, 2026-09-05): Stage 0 shipped in PR #614. Stage 1 shipped in two parts - PR #977 moved the `type-check` gate to `tsc7` (31.8 s to 4.9 s, diagnostics parity recorded in the 2026-08-25 comment), and PR #1427 moved the emit scripts too (`build`, `electron:compile`, `electron:watch`, `app/package.json:15,17-18`) after a measured comparison: 48 byte-identical `.js` files between the two compilers and one `.d.ts` quote-style line. `typescript` `^5.9.3` (`:110`) is now invoked by no script and stays only as the peer of `@typescript-eslint/*` `^8.68.0` (`:91-92`, peer `typescript >=4.8.4 <6.1.0`) and for `script-typedefs.docs-compile.test.ts`. Only Stage 2 remains open, gated on typescript-eslint.**

## What it buys this repo, measured

`pnpm type-check` runs three sequential `tsc --noEmit` invocations (`app/package.json` scripts) and took **27.8s wall** on this container at `054059a` (noisy cloud host - treat as order-of-magnitude, not a benchmark). At the reported speedups that lands around 3-5s. The same `tsc` gates `pnpm build` (`tsc && vite build`), and CI runs type-check on every app-touching PR. Contributors also get the rebuilt native LSP in editors (8x project-load per the release notes) - though that is a per-developer opt-in, not a repo change.

**What it does NOT touch, verified**: the in-app script editor's TypeScript (completions, the #443 strictNullChecks question) runs Monaco's own bundled language service (`monaco-editor ^0.56.0`), independent of the repo's `typescript` package. Vitest transforms via its own pipeline and does not consume the TS API here (no typecheck-mode usage, verified: `rg typecheck app/vitest.config*` empty).

## The blocker for a clean swap

[...]

PR #1427, opened 2026-09-05
build(app): graphql 17, zod 4, lucide-react 1 and the TypeScript 7 emit - the rest of the September sweep (#1335)

athrvk:
## Description

#1335 part 3's **"the rest together"** group: `concurrently` 10, `graphql` 17, `zod` 4, `lucide-react` 1, the `@types/node` decision, and the `typescript` emit swap.

**Why this group and not the one the table orders first.** `vitest`+`jest-dom`+`jsdom` is not takeable today. `vitest` 5.0.0 published 2026-09-03 12:24Z and `app/pnpm-workspace.yaml`'s `minimumReleaseAge: 10080` admits no 5.x release until 2026-09-10; the group's own note bumps `jest-dom` *after* vitest. Adding `vitest` to `minimumReleaseAgeExclude` is the thing #1377 establishes we do not do. The two groups are disjoint packages, so the table's order costs nothing by being taken second. Detail in the claiming comment on the issue.

**Plan.** Root cause of the work, in one line per package: a major moves the API under code that reads it, and only two of the five moved anything here. `graphql` 17 changed where a default value lives and how the package is shipped; `zod` 4 changed `z.record`'s arity and what `ZodRawShape` names. The other three (`lucide-react`, `concurrently`, `@types/node`) are inert against this tree, and the `typescript` swap is a script change gated on a measured emit comparison. The alternative I rejected for `graphql`: keeping `arg.defaultValue` and reading `arg.default` only as a fallback. It would work today and rot on schedule - `defaultValue` is deprecated for removal in graphql 18, and the code would still carry the string-versus-enum guess the literal makes unnecessary.

### graphql 16.14.2 → 17.0.2

Two changes reach this app.

1. **A default value is now a literal.** `arg.default` is `{ literal: ConstValueNode }` for anything built from SDL or an endpoint's introspection, and the deprecated runtime `arg.defaultValue` is left `undefined` there. Reading the old field alone, the schema explorer printed *no* defaults (`search(term: String!, ranking: Ranking)`), and `insert-skeleton.ts` counted a defaulted non-null argument as required. Defaults now print through graphql's own `print()`, which also retires the guess the old formatter had to make: it took every string bare so an enum default read `RELEVANCE` rather than `"RELEVANCE"`, which meant a genuine `String` default printed as something the user could not type back. The three required-argument predicates are now `isRequiredArgument` from graphql itself, which reads both spellings - a hand-rolled copy of a primitive does not receive the primitive's fixes.
2. **graphql 17 ships CommonJS and ESM side by side**, and its development build refuses a type built by the other copy (`Cannot use GraphQLScalarType "ID" from another module or realm`). `vitest.config.ts` aliased `graphql` to `index.js` while the inlined `graphql-language-service` resolved `index.mjs`, so every schema object crossing between them failed that check - 22 test failures across the variables schema, the variables diagnostics and the language providers. The alias names the ESM entry now, with the reason in the file. `dedupe` alone does not settle it: one package in two builds is not a duplicate it can collapse. `vite.config.ts` needs nothing - it has no graphql alias, and the production build resolves one copy.

`graphql-language-service` 5.6.0 already declares `graphql ^15.5 || ^16 || ^17`; nothing there moves.

### zod 3.25.76 → 4.5.2

- `z.record(valueSchema)` is removed; the 15 call sites in `electron/mcp/tools.ts` take the two-argument `z.record(z.string(), valueSchema)`.
- `McpTool.inputSchema` and `McpPromptDef.argsSchema` are spelled `Record<string, z.ZodType>` rather than `z.ZodRawShape`. In zod 4 that alias resolves to the *core* `$ZodType`, which has no `.description` - the per-field text a client renders beside an input, and which `tools.test.ts` asserts on. Same reason for `outputSchema`, and `z.ZodTypeAny` (a deprecated v4 compat alias) becomes `z.ZodType` in the test's casts.
- `@modelcontextprotocol/sdk` 1.30 already declares `zod: ^3.25 || ^4.0` and ships a v3/v4 shim (`zod-compat`) at the `registerTool` boundary, so the SDK does not move. Nothing in this tree touches the rest of zod 4's breaking list: no `.email()`, `.merge()`, `errorMap`, `.errors`/`.issues` or `z.infer`.
- The generated JSON Schema was read back, not assumed: `run_request`'s `headers` still comes out `{"type":"object","additionalProperties":{"type":"string"}}` with its description, `required` still lists only `url`, and `protocol.integration.test.ts` drives a real `initialize` → `tools/list` → `tools/call` round trip through the SDK.

### lucide-react 0.562 → 1.35

Inert here, checked rather than assumed: none of the 14 removed brand icons is imported (140 distinct icons are, all statically, so a rename would be a compile error rather than a runtime one), no UMD or deep import, no icon resolved from a string at runtime. The issue expected v1 to set `aria-hidden` unconditionally; **it does not** - 1.35's `Icon` keeps the `hasA11yProp` exception, so `FilePartCell`'s `aria-label`led warning triangle stays in the accessibility tree. Icon geometry was compared name by name between the two majors: 135 of 140 identical, 5 redrawn upstream - see the [comparison image](https://github.com/athrvk/vayu/pull/1427#issuecomment-5549662692) in the comments. 1.35.0 rather than the 1.41.0 the issue names is the newest release the age floor admits.

### concurrently 9 → 10, and the Node floor

v10 declares `node >= 22`. The only two invocations are `type-check`'s `--group --names` and `electron-dev.mjs`'s `-k -r`; both were run against v10. The documented floor was Node 20, which the 10 bump makes false - and which was already stale, since the Node 20 line went end-of-life on 2026-04-30 while `.nvmrc` and every CI job say 22. `CLAUDE.md`, `CONTRIBUTING.md`, `docs/building.md` and `docs/app/building.md` now say 22 (and `CONTRIBUTING.md`'s `pnpm ≥ 8` becomes 11, which is what CI pins).

### @types/node: held at 24, the row's "decide"

Moved *down* from 25 to 24, the major Electron 44's bundled Node is (24.19/24.20), rather than up to 26. Types ahead of the runtime describe APIs `electron/` can call and the shipped main process does not have; the main process is what consumes them. No post-Node-24 built-in is used anywhere in `electron/` or `scripts/`, and `vite`/`vitest`'s own `@types/node` peer ranges both admit 24. The rule for the next reader is recorded in `docs/app/building.md`: bump it when Electron's Node major moves, not when DefinitelyTyped's does.

### The TypeScript emit swap

`build`, `electron:compile` and `electron:watch` now run `tsc7`, which every `type-check` project already ran. The swap was gated on a comparison rather than trust: `tsconfig.node.json` built with 5.9 and with 7.0.2 in turn produces **47 byte-identical `.js` files** and one `.d.ts` line that differs in quote style (7 single-quotes a string containing `"` instead of escaping it). `typescript` 5.9 stays installed and is now invoked by nothing: `typescript-eslint`'s peer range is `>=4.8.4 <6.1.0`, and `script-typedefs.docs-compile.test.ts` drives the Compiler API by package name. Stage 2 of #467 - one `typescript@7`, alias gone - still waits for 7.1. `typescript-toolchain.test.ts`'s bare-`tsc` guard passes unchanged; every script still names its compiler by path.

## Type of Change
- [x] Bug fix (the graphql 17 default-value and required-argument reads)
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Full app suite, type-check, lint, format, a production build and a strict docs build, all on this branch.

- `pnpm test` - **689 files, 7860 tests, all passing** (322.96s). Baseline on `d84bbab` before any edit: 689 files / 7859 tests green, so the delta is exactly the one test added below and nothing here is a pre-existing failure.
- `pnpm type-check` - green (renderer, main process, electron tests)
- `pnpm lint` - green, zero warnings
- `pnpm format:check` - green (`tools.ts` and `insert-skeleton.ts` reformatted after the edits; nothing else touched, and nothing outside `app/`)
- `pnpm build` - green through `tsc7` + vite
- `mkdocs build --strict -f .github/mkdocs.yml` - clean, including the renamed `#one-compiler-runs-two-are-installed` anchor and its two inbound links from `docs/architecture.md` and `docs/app/architecture.md`
- `pnpm exec concurrently -k -r ...` and `pnpm type-check` on v10 - the four flags this repo uses

One test added, mutation-checked: `schema-tree.test.ts` pins an input field's default to `: Ranking = RELEVANCE`, the spelling the argument row already had and the input-field row did not (it printed `= "RELEVANCE"` through `JSON.stringify`). Reverting `formatDefault` to the old form reddens it plus five existing cases.

## Checklist

## Related Issues

Refs #1335 - part 3's last group but one; `vitest`+`jest-dom`+`jsdom` stays open there and is takeable from 2026-09-10.

Noted rather than filed, per the tracker rule (neither changes what the app does): `vitest.config.ts`'s `__dirname` prints a Vite 8 config-loader deprecation warning on every run, and `app/package.json` states no `engines` field, so the Node floor this PR documents is prose rather than something `pnpm install` enforces.


athrvk:
### Visual evidence: what the lucide major actually changes on screen

The upgrade's user-visible surface is icon geometry, so it was measured rather than assumed. Every icon name this app imports (140 distinct, across 155 files) was read out of both packages' own icon node data and compared, following v1's alias re-exports - `History` is one of those: in 1.x it re-exports `rotate-ccw-clock`, whose geometry is identical to 0.562's `history`.

**135 of 140 are identical. Five changed upstream:**

Five lucide icons compared across 0.562.0 and 1.35.0 - BookmarkPlus, Calendar, Clock, Waves and Zap - shown before, after, and overlaid

All five are the same icon redrawn on the same grid at the same stroke width - `Calendar`'s body moves up a unit and loses the tick height, `Clock`'s hands shorten, `Waves` shifts, `Zap` and `BookmarkPlus` are re-cut slightly. Nothing changes size, weight or meaning, so no layout or contrast rule in `docs/design-system.md` is touched.

**What this image is and is not.** It is rendered from each package's icon node data at the app's own SVG settings (`stroke-width: 2`, `linecap`/`linejoin` round, `viewBox 0 0 24 24`) - the exact input the `Icon` component turns into DOM - not a screenshot of the running app. The app itself is unchanged in how it draws them: `lucide-react` 1.35's `Icon` still merges classes the same way and still keeps the `hasA11yProp` exception, so an `aria-label`led icon is still left out of `aria-hidden` (`FilePartCell`'s warning triangle is the one place that matters), and both characterization snapshots - which embed live lucide SVG markup - pass untouched.

The other user-visible line in this PR is the GraphQL schema explorer's default values. Reaching it in the running app needs a live introspectable endpoint, and the change is one token wide, so it is pinned by assertion instead: an input field's default now reads `: Ranking = RELEVANCE` where it read `: Ranking = "RELEVANCE"`, matching the argument row beside it, and reverting the fix reddens six cases in `schema-tree.test.ts`.

The two commits that carry this PNG add it and then delete it, so it does not appear under "Files changed"; GitHub serves the blob from the commit that introduced it, which is what the link above points at.

---

athrvk:
### Review pass, and what it changed

Two adversarial reads of the diff (one against #1335 part 3's rows, one against `CLAUDE.md`'s conventions) found four things worth acting on. All four are now in `6408675` and `cd44db7`; the suite is **689 files / 7862 tests green**, up from the 7860 the body quotes, which is the three tests below.

1. **The two halves of the graphql fix disagreed.** `insert-skeleton.ts` asks graphql's `isRequiredArgument`, which reads *both* `default` and the pre-17 `defaultValue`; `formatDefault` read only the first. A schema built in code rather than from SDL or introspection can still set the old spelling, and the explorer would then print no default for exactly the argument the insertion path is leaving out as already answered. Both read both now, pinned by a test that builds such a schema (`new GraphQLObjectType` with `defaultValue: 7`).
2. **The `isRequiredArgument` swap had no test that could fail.** Its answer differs from the old predicate only for an argument that is non-null *and* defaulted, and the shared fixture writes no such argument - so reverting that hunk reddened nothing. `insert-skeleton.test.ts` now builds `report(range: Int! = 7, label: String!)` and asserts `range` stays out of the declared variables while the document still validates. Mutation-checked: restoring `a.defaultValue === undefined` reddens it.
3. **Two spellings zod 4 only tolerates.** The `.refine` on `thresholds` passed `message:`, which v4 keeps as a deprecated alias for `error:` - and a deprecated alias is what the next major takes away. Three `z.ZodRawShape` casts also survived in `tools.test.ts`, naming the core type the interface deliberately stopped naming. Both corrected.
4. **`docs/app/building.md` still said pnpm ≥ 8** one line under the Node floor this branch was correcting, while `CONTRIBUTING.md` moved to 11 in the same commit. It says 11 now, with why: that is what all five `pnpm/action-setup` steps pin, and 11 is the version that reads `minimumReleaseAge` from `pnpm-workspace.yaml` at all.

Everything else both reviews checked came back clean, including the parts most worth doubting: `z.string()` is the right key schema at all 15 `z.record` sites (every one is a string-keyed map - header names, variable names, column names, claim names, none of which had a narrower key before), an explicit `= null` default still prints as `null` and stays distinguishable from "no default", every one of the icon identifiers in use resolves in lucide v1 (its deprecated names survive as aliases), and Electron 44's own `package.json` depends on `@types/node ^24.9.0`, which is the independent check on this PR holding that package at 24.

Two earlier "CI gate" failures on this PR, at `58a9633` and `6408675`, were the workflow's concurrency group cancelling those runs when the next commit landed - every job reads `cancelled`, none ran a test body. The run on `cd44db7` is the one that counts.

---

PR #1448, opened 2026-09-05
ci: make the sccache error log say when it has nothing to say

athrvk:
## Description

athrvk/vayu@f7b9b90b added `SCCACHE_ERROR_LOG` and a step to print it, because v0.17.1's Windows leg reported 18 sccache "Cache errors" that `--show-stats` could only count. **Three weeks and four releases later that step has never printed a line** - and v0.26.0's Windows leg reported the same 18.

They are not a race. Both Windows jobs on the same commit:

| job | requests | hits | misses | **errors** |
|---|---|---|---|---|
| Warm cache (07:59) | 319 | 261 | 12 | **18** |
| Release (08:05) | 319 | 273 | 0 | **18** |

Identical 18 errors and identical 7 `c [msvc]` hits in both. The same C translation units fail to cache on every run. macOS shows the same shape (52 errors, all `c [clang]`, plus 60 cache write errors); Linux/gcc has zero. All of the engine's C is vendored - quickjs-ng and hdrhistogram, ~22 files, matching Windows' 18 errors + 7 C hits almost exactly.

### Why the diagnostic was mute

Three reasons, all fixed here:

1. **It could not tell "no errors" from "no logging".** The step branched on whether the log file was empty and printed `No sccache errors logged.` either way. Stats saying 18 while the log said nothing was rendered as success. The check now keys off **the stats**, which sccache records whether or not it wrote a word, leaving three honest outcomes: no errors; errors with a log to read; errors with no log, which now says the diagnostic is broken.
2. **`SCCACHE_LOG` was `warn`,** and whatever sccache says about these errors, it does not say at warn. Now `sccache=debug`, scoped so its dependencies do not drown the file. The log is never printed whole - the action extracts WARN/ERROR lines plus a bounded tail, because burying the cause in per-compile chatter would repeat the original defect in a new shape.
3. **It lived only in `release.yml`, which runs on tag push alone.** It could not be exercised without cutting a release, which is why it went unchased. `cache-warm.yml` produces the identical errors, runs on master pushes *and* `workflow_dispatch`, and had **no** sccache logging env at all. It has both now.

### Where it fails, and where it only warns

`cache-warm.yml` passes `fail-on-blind-spot: true` - it is cheap to re-run and blocks nobody, so "sccache errored and would not say why" is worth a red run. `release.yml` warns instead: an sccache error falls through to the real compiler, so it costs build time and nothing else, and a mute diagnostic must not stop a build that is otherwise correct.

It is a composite action rather than a step in each workflow for the reason `winget-preflight` is: the copy not being debugged is the copy that drifts.

## Type of Change

- [x] Bug fix (a diagnostic that reported success while blind)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

The action's script was extracted from its own YAML (not paraphrased) and driven through all four states against a stub `sccache`:

| State | Result |
|---|---|
| 0 errors | `Nothing to report.`, exit 0 |
| 18 errors + populated log | prints WARN/ERROR lines and tail, notice, exit 0 |
| 18 errors + empty log, `fail-on-blind-spot: false` | **warning** naming the count, exit 0 |
| 18 errors + empty log, `fail-on-blind-spot: true` | **error** naming the count, **exit 1** |

Mutation check on the state v0.26.0's Windows leg was actually in (18 errors, empty log):

```
OLD step: "No sccache errors logged."   exit=0
NEW action: ::error::sccache recorded 18 cache error(s) but wrote no log …   exit=1
```

Also verified: all three YAML files parse, and `cache-warm.yml`'s **Cache key agreement** guard still matches its `SCCACHE_GHA_ENABLED` and compiler-launcher patterns in both workflows after the env insertion.

## Checklist

- [x] I have added tests (if applicable) - the four-state drive above; a workflow step has no unit-test harness in this repo
- [x] I have updated documentation - the action carries the rationale; no doc described the old step

## Related Issues

Follow-up to athrvk/vayu@f7b9b90b. **This does not yet say why the 18 error - it makes the next run able to.** Once a run carries a real log, the cause is a separate fix.

https://claude.ai/code/session_01CdzN6CbBVwkWViZer9Lmde

---
_Generated by [Claude Code](https://claude.ai/code/session_01CdzN6CbBVwkWViZer9Lmde)_

### DIFF
--- a/.github/actions/sccache-diagnostics/action.yml
+++ b/.github/actions/sccache-diagnostics/action.yml
@@ -1,0 +1,121 @@
+name: sccache diagnostics
+description: >
+  Print sccache's error log when the run recorded cache errors, and fail the
+  blind spot where it recorded errors but logged nothing.
+
+# f7b9b90b added SCCACHE_ERROR_LOG and a step that prints it, because v0.17.1's
+# Windows leg reported 18 "Cache errors" that `--show-stats` could only count.
+# Three weeks and several releases later that step had never printed a line,
+# while every Windows run went on reporting the same 18. The diagnostic was not
+# reporting "no errors"; it was reporting nothing at all, and nothing could tell
+# the two apart, because the step branched on whether the log file was empty and
+# said "No sccache errors logged." either way.
+#
+# So the check here keys off the **stats**, not off the log. sccache knows how
+# many compilations errored whether or not it wrote a word about them, and the
+# only honest states are:
+#
+#   errors == 0                -> nothing to report.
+#   errors > 0, log has lines  -> print them; this is the case f7b9b90b wanted.
+#   errors > 0, log is empty   -> the diagnostic is broken, and saying so is the
+#                                 whole point. This is the state that hid for
+#                                 three weeks.
+#
+# It is an action rather than a step in each workflow for the reason
+# winget-preflight is: release.yml runs on tag push alone, so a diagnostic that
+# lives only there cannot be exercised without cutting a release - which is
+# exactly why nobody noticed it was mute. cache-warm.yml runs on master pushes
+# and on workflow_dispatch, so the same check there can be run on demand.
+
+inputs:
+  error-log:
+    description: >
+      Path sccache was told to write its log to (SCCACHE_ERROR_LOG). Read from
+      the environment when not given.
+    required: false
+    default: ''
+  fail-on-blind-spot:
+    description: >
+      Fail the job when sccache recorded cache errors but wrote no log. True on
+      the warm workflow, which is cheap to re-run and blocks nothing; false on a
+      release, where a broken diagnostic must not stop a build that is otherwise
+      fine - an sccache error falls through to the real compiler, so it costs
+      time and nothing else.
+    required: false
+    default: 'false'
+
+runs:
+  using: composite
+  steps:
+    - name: Report sccache errors
+      if: always()
+      shell: bash
+      run: |
+        set -uo pipefail
+
+        log="${{ inputs.error-log }}"
+        if [ -z "$log" ]; then
+          log="${SCCACHE_ERROR_LOG:-}"
+        fi
+
+        # The text form is parsed rather than the JSON, matching the existing
+        # non-empty-scan check in cache-warm.yml: a stats-format change on a
+        # future sccache then breaks the parse loudly instead of silently
+        # answering zero.
+        stats="$(sccache --show-stats 2>/dev/null || true)"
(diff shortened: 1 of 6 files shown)


athrvk:
Extended after review feedback: the first commit was under-scoped.

**Five workflows compile the engine through sccache.** It covered two - `release.yml` (where f7b9b90b's version lived) and `cache-warm.yml` (dispatchable, so testable). That left three as blind as `release.yml` had been, which is this branch's own defect in a different place.

Now covered:

| Workflow | Runs on | Mode |
|---|---|---|
| `cache-warm.yml` | master push, dispatch, weekly | **fails** on blind spot |
| `pr-tests.yml` | every PR, ubuntu + windows + macos | warn |
| `release.yml` | tag push | warn |
| `engine-tidy-scan.yml` | weekly | warn |
| `perf-measure.yml` | weekly | warn |

`pr-tests.yml` is the one that matters most and the one I had missed - it runs these compilations on every pull request across all three platforms, far more often than a release does, and it is where a regression would show up in the PR that caused it rather than at tag time.

**`codeql.yml` stays out deliberately.** It sets no compiler launcher on purpose: CodeQL builds its database from observed compiler invocations, and an sccache hit skips the compiler the tracer needs to see. Excluding it is correct, not an oversight.

**Warn-only everywhere except `cache-warm.yml`.** A cache diagnostic must never be what turns a pull request red - these errors fall through to the real compiler, so they cost build time and nothing else. `cache-warm.yml` keeps `fail-on-blind-spot` because it is cheap to re-run and blocks nobody.

**Expect a warning annotation on engine pull requests** until the cause is fixed. That is the intent: 18 compilations on Windows and 52 on macOS have gone uncached since v0.17.1 and nothing said so.

### Unrelated lead found while checking this

`pr-tests.yml` already carries a comment reasoning about the MSVC hazard:

> The launcher is safe on MSVC here because windows-prod is a Release build - CMake passes no `/Zi`, so there is no separate PDB for sccache to choke on.

That holds for sources the engine's own CMake controls - there is no `/Zi` anywhere in `engine/CMakeLists.txt`. But every failing compilation is **C, never C++**, and all the engine's C is vendored: `quickjs-ng` comes in through `add_subdirectory` with its own `CMakeLists.txt` (which the existing comment at line 46 notes resets `cmake_minimum_required`), and `hdrhistogram` is built as a separate target. A vendored build controlling flags for exactly the files that fail is a plausible cause.

Recording it here rather than acting on it - the whole point of this PR is to stop guessing. The log will say.

---

PR #1456, opened 2026-09-05
fix(engine): the bound OpenAPI export adds to a document, and removes nothing (#1442)

### LINKED ISSUE #1442: engine(export): the bound OpenAPI export erases a response's named `examples` and writes `content` beside a `$ref` - a spec-origin collection's export rewrites the user's contract silently
## Problem

For a collection imported from an OpenAPI document, "Export as OpenAPI" patches the stored bytes (`engine/src/core/openapi_export.cpp`, `patch_bound_document` around `:609`): it writes row values into declared parameters' `example` (`patch_parameters:411`) and response examples into the operations (`patch_operation:449`), and `docs/app/openapi.md:410` promises everything else is "left exactly as it was". Measured at master `c41ea644` (files unchanged at `90c2fff3`) by importing an OpenAPI 3.0.3 document with `$ref` parameters, bodies and responses, five security schemes, two servers, tags and named examples, exporting it, and diffing source against export:

1. **Named `examples` collapsed to one `example`.** `listPets`'s 200 response had `examples: {two: ..., none: ...}`. The import took the first named example (`openapi_drafts.cpp:551`, `first_named_example`, `:747-753`), and the export wrote it back as `media["example"] = ...; media.erase("examples")` (`openapi_export.cpp:292-311`). The `none` example and both summaries are gone from the user's document. Nothing in `notes` says so.
2. **`content` written beside `$ref`.** Two responses were `404: {$ref: "#/components/responses/NotFound"}`. The export added `content` as a sibling of `$ref`. A Reference Object admits no siblings in 3.0; spec-conformant tools ignore the addition and validators reject the document.
3. **An example the document never had.** `/owners/{ownerId}/pets`'s 200 gained an `example` synthesised at import from the `Pet` schema's `example` (`status: available`), written back as if the document had documented it.
4. **Edits the export cannot express, uncounted.** A request body edited in Vayu is never written back (`patch_operation` writes parameters and examples only); new header or param rows the document does not declare, and method or URL edits, are ignored. Users who "edit the imported stub and export the updated document" get the original body. The notes count `operationsRemoved`, `requestsWithoutOperation`, `sharedParametersLeft`, `examplesWithoutMediaType`, but not this.
5. **Serialisation.** A YAML source is re-serialised from `ordered_json` (`serialize` around `:900`): comments, anchors, key quoting and number formatting are lost, against `openapi.md:410`'s "left exactly as it was". A documentation fix, since a comment-preserving YAML writer is not in scope.

Swagger 2.0 (presence pass only, `vocabularyNotWritten`) and 3.1 (webhooks, `pathItems`, `mutualTLS` all preserved) exported cleanly apart from the same `examples` behaviour.

Severity: common for real specs (named `examples` and `$ref` responses are the norm), silent, destructive to the user's source of truth - the exported document is what they check in.

## Fix pathway

1. **Never erase what the import did not read.** In `openapi_export.cpp:292-311`, when the media object already has `examples`, write Vayu's value into the entry it was imported from (the import can record the example key it took, beside `specOperation`) or add a new named entry, and never `erase("examples")`. When the media has a single `example`, replacing it is fine.
2. **Respect Reference Objects.** In `patch_operation`, when a response (or parameter, or request body) is a `$ref`, do not write into it; either skip with a `notes` counter (`referencedResponsesLeft`, matching the existing `sharedParametersLeft`) or resolve the reference to its component and patch there once. Skipping is the safe default; the components are shared.
3. **Mark synthesised examples.** An example the import sampled from a schema rather than read from the document should not be written back; record its origin on the example (a `synthesised` flag in the stored example, or simply do not persist schema-sampled examples as `origin: import`) and skip them in `writable_examples` (`:269`).
4. **Count what cannot be written.** `notes` gains `bodiesNotWritten`, `rowsNotDeclared` and `operationsEdited` for the edits `patch_operation` cannot express, so the dialog says "3 edited request bodies are not in the document; the bound export writes parameters and examples only". The docs already explain why; the count makes it visible per export.
5. **Docs.** `openapi.md:410`'s "left exactly as it was" becomes "structurally unchanged; comments and formatting are not preserved", with the list of what the bound export writes.

## App changes

`app/src/modules/collections/ExportSpecDialog.tsx` renders the new note rows; `electron/mcp/tools.ts` `export_spec` forwards them. `types/api.ts` for the notes shape.

## Tests

- `engine/tests/openapi_export_test.cpp`: import a fixture whose 200 has `examples: {a, b}` and whose 404 is `$ref`; export bound; assert `examples.b` still present with its summary, the 404 still a bare `$ref`, and the notes count the referenced response. Mutation checks: restore `erase("examples")` and the first assertion reds; write into the `$ref` and the second reds.
- A case with a schema-sampled example asserting it is not written back.
- `python build.py -e -t && cd engine && ctest --preset linux-dev -R OpenApiExport --output-on-failure`.

## Docs, same commit

`docs/app/openapi.md` bound-export section (`:405-440`), `docs/engine/api-reference.md` for the new notes fields.

## Sequencing

Wave 1 of the interop cluster: touches `openapi_export.cpp`'s bound path and the notes type. #1441 (the skeleton export) touches the same file's other half and the same notes type, so land this first and rebase #1441 onto it. Parallel-safe with #1443 (Postman import), #1444 (OpenAPI `security`), #1445 (curl) and the two-writer cluster (#1436 to #1440).

## Acceptance criteria

- Exporting an unmodified spec-origin collection produces a document that a 3.0 validator accepts and that differs from the source only in whitespace and key order, pinned by a structural diff in the engine test.
- A response with several named examples keeps all of them after export.
- Every edit the bound export cannot write is counted in the notes and shown in the dialog.

---

athrvk:
## Description

The bound direction of "Export as OpenAPI" patches the user's own stored document, and three of its writes were edits to that contract rather than updates to it: it erased a response's named `examples` map to write one unnamed `example`, wrote `content` as a sibling of a `$ref` response, and wrote back an example the *import* had sampled off a schema as though the document had declared it.

**Root cause, one mechanism:** the write path treated the media object under a response as its own to rewrite. `write_media_examples` ended in `media.erase ("examples")`; `write_response_examples` reached for `content` through `child_record` without asking whether the response was a Reference Object; and `writable_examples` filtered on truncation and media type but not on where the stored example came from - `ExportExample` carried no origin at all, while the row it is read from has carried one since #588 (`request_examples.origin`).

**Approach:** one rule, applied in the bound direction only - *this export may add what Vayu learned and update what it read, and may remove nothing the document declares.* Its consequences:

- **Named examples survive.** A media object already answering with an `examples` map keeps every entry and gains one for a response somebody kept; a single `example` is replaced by the value that would replace it, and moves into the map under the key `example` when a second example has to go somewhere.
- **A `$ref` response is left bare** and counted as `referencedResponsesLeft` (the sibling of `sharedParametersLeft`): a Reference Object admits no siblings in 3.0, and the component it names is shared with every operation that references it.
- **Provenance decides the rest.** An example whose value the document already declares is counted (`examplesAlreadyDeclared`), never rewritten. An *imported* example for a media object that declares no example at all was sampled off that response's schema at import - so it is counted (`examplesSampledAtImport`) instead of written, and where the document declares no such response either, the status is not documented from it at all: putting back a response the contract dropped is not this export's to do.

That makes the headline case true: exporting a spec-origin collection nobody edited gives back the document it was bound to, pinned by a whole-document structural assertion rather than field spot-checks.

The edits this direction cannot express are counted too, so the dialog states them per export instead of the docs stating them once: `bodiesNotWritten`, `rowsNotDeclared` and `operationsEdited`. A row has a home whether the operation declares its parameter inline, by `$ref` (followed one hop for the *name* only) or through the Path Item's own `parameters`; `Authorization` and `Content-Type` are excepted, since OpenAPI states them as `security` and as the body's media type. `operationsEdited` compares through `operation_match`'s own path-shape primitives rather than a second copy of them. `derive_schema` becomes an `ExportDirection`, because every one of these rules is a consequence of who owns the document underneath and a bool at four call sites said nothing about that.

**Rejected:** recording the imported example's map key beside `specOperation` so the export could update that exact entry (the issue's first option for part 1). It changes the stored shape of every imported example to buy write-back into an entry the user did not ask to change; comparing values at export time answers the same question - "is this already in the document?" - from data both sides already have, and needs no migration. Its one remaining gap - an *edited* imported example is added beside the entry it came from rather than replacing it - is filed as #1457 rather than left as a comment.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` - **2916 engine tests, all passing.** (The issue's `-R OpenApiExport` matches nothing: these suites are `BoundExport`, `SkeletonExport`, `ExportSerialization` and `SpecExportRouteTest` - 45 tests together, all green.)
- `cd app && pnpm test` - **7926 tests in 694 files, all passing**; `pnpm type-check`, `pnpm lint` and `pnpm format:check` clean.
- clang-format 19 and clang-tidy 19 clean on every touched translation unit (the pre-commit hook ran both).
- No pre-existing failures were observed on the base commit.

Seven new `BoundExport` cases and one new `SpecExportRouteTest` case, on a new fixture (`petstore-refs-bound.json`) shaped like a real document - a response with named examples, a `$ref` response, a response that declares a schema and no example, and a path-level parameter:

- `WritesNothingIntoADocumentNobodyEdited` - the whole exported document equals the fixture, key order included.
- `AddsAKeptResponseBesideTheDocumentsExamplesRatherThanOverThem`
- `LeavesARefdResponseBareAndLeavesTheComponentItNamesAlone`
- `WritesBackAResponseSomebodyKeptButNotOneTheImportSampled` - the same body, written or not by where it came from.
- `DocumentsAStatusFromAKeptResponseButNotFromASampledOne`
- `CountsNoUndeclaredRowForAParameterTheDocumentDeclaresByReference`
- `CountsTheEditsItHasNoWayToWriteIntoTheDocument`
- `SpecExportRouteTest.LeavesAnImportedExampleTheDocumentDeclaresNoneForAlone` - the wiring test: it pins that the row's `origin` actually reaches the exporter.

Mutation-checked, each reverted separately against a rebuilt binary:

| reverted | reds |
|---|---|
| `media.erase ("examples")` restored | `AddsAKeptResponseBesideTheDocumentsExamplesRatherThanOverThem` |
| the `$ref` response guard removed | `WritesNothingIntoADocumentNobodyEdited`, `LeavesARefdResponseBareAndLeavesTheComponentItNamesAlone` |
| `example.from_import` dropped from the condition, so every undeclared example reads as sampled | `WritesBackAResponseSomebodyKeptButNotOneTheImportSampled`, `WritesOneStoredExampleAsExampleAndSeveralAsANamedMap`, `NeverWritesATruncatedBodyAndCountsTheOnesItLeftOut`, `SpecExportRouteTest.WritesAStoredExampleIntoTheBoundDocument` |
| the `$ref` parameter's name lookup dropped | `CountsNoUndeclaredRowForAParameterTheDocumentDeclaresByReference` |
| the plan-before-create guard removed | `DocumentsAStatusFromAKeptResponseButNotFromASampledOne` |
| positional (`ordered_json`) example equality restored | `WritesNothingIntoADocumentNobodyEdited` |

`SpecExportRouteTest.WritesAStoredExampleIntoTheBoundDocument` now passes `origin: "user"`, which is what the app sends for a response somebody kept (`save-as-example.ts`); a row that states no origin is an `import` row by the schema's own default, and the spec sync already treats it as one.

No 3.0 validator runs in this repo, so the acceptance criterion's "a validator accepts it" is carried by the stronger assertion that an unedited export **is** the input document, structure for structure, plus the `$ref` case that was the only thing making the output invalid.

The last three rows came out of an adversarial review pass over this branch's own diff; [that comment](https://github.com/athrvk/vayu/pull/1456#issuecomment-5551854687) has the reasoning.

## Visual evidence

The export dialog's summary is user-visible, so it was captured from the **real app** - `electron .` against a live engine, driven on an Xvfb display - with the workspace built by a real `POST /import` of the new fixture. Before/after shots and the notes the engine answered with are in [this comment](https://github.com/athrvk/vayu/pull/1456#issuecomment-5551749154); every new counter is non-zero there. Linux only; the change is text rows in an existing list with no platform-conditional code.

## Checklist

## Related Issues

Closes #1442

Follow-up filed: #1457 (an edited imported example is added beside the entry it came from, because the import records no example key).


athrvk:
### The export dialog, before and after

Captured from the **real app** on this machine - `electron .` against a live engine on an Xvfb display, not a stub render. The workspace behind both shots is one real `POST /import` of an OpenAPI 3.0.3 document with a named `examples` map, a `$ref` response and a response that declares a schema and no example; then, through the API as a user would through the UI: a body and two rows added to one request, the other request's path edited after it was matched, and one response kept from a send.

Same engine, same collection, same counts - only the dialog differs. The "before" is `ExportSpecDialog.tsx` at the PR's base commit `228f113`, which ignores the new keys the engine sends.

The export dialog on master: eight count rows, none of them explaining the examples the export left alone, the $ref response it did not write into, or the body and rows it could not carry

The export dialog on this branch: fourteen count rows, adding the $ref response left as it is, the request whose body is not written, the two rows the operation declares no parameter for, the request no longer matching its stamped operation, the example already declared in the document, and the example the import sampled from a schema

The notes the engine answered with for that collection, which is the same run the "after" shot renders:

```json
{"requestsExported": 2, "examplesWritten": 1, "examplesAlreadyDeclared": 1,
 "examplesSampledAtImport": 1, "referencedResponsesLeft": 1,
 "bodiesNotWritten": 1, "rowsNotDeclared": 2, "operationsEdited": 1}
```

Every new counter is non-zero there, produced by a real import rather than by a fixture handed to the exporter - which is the end-to-end evidence that the row's origin reaches the export and that each count means what its sentence says.

**Platforms:** Linux only. The change is text rows in an existing list with no platform-conditional code, and the rendered sentences are pinned by `ExportSpecDialog.test.tsx`.

The two PNGs are committed at `c4a128f` and deleted again in the next commit, so they render here permanently while the reviewable diff stays code and docs only - they intentionally do not appear under "Files changed".

---

athrvk:
### Three defects found reviewing this branch, fixed in `d9c286b`

An adversarial pass over my own diff turned up three cases where the export still treated the document as flat. Each is fixed and mutation-checked.

**1. A row the document declares by `$ref` was reported as a row it declares nowhere.** `patch_parameters` skipped a `$ref` parameter before learning its name, so the new `rowsNotDeclared` fired for a parameter shared through `components.parameters` - how essentially every real spec declares a pagination or tenant parameter - and the same edit was then reported twice: once correctly as `sharedParametersLeft`, once wrongly as a row with no home. The reference is now followed one hop to read the *name* only; nothing is written through it. A Path Item's own `parameters` are read the same way, since OpenAPI makes them the operation's parameters too. New test: `CountsNoUndeclaredRowForAParameterTheDocumentDeclaresByReference`.

**2. A response the document had dropped could be put back.** The response object was created - `description` and all - before anything knew whether an example would be written into it. For an imported example of a status the document no longer declares (a spec re-fetched after the row was stored), that wrote a response the contract had deliberately removed: exactly the invention this PR exists to stop. The writes are now planned before the node is created. A status left bodiless for a reason of *this* export's - a truncated body, a media type nobody recorded - still lands, because there the response is known and only its body is missing. New test: `DocumentsAStatusFromAKeptResponseButNotFromASampledOne`.

**3. Key order made two identical examples look different.** `ordered_json` compares objects positionally, so a document reformatted between the import and the export would have had its own declared example written back beside itself as a second copy. Comparison now goes through `nlohmann::json`, which orders keys.

| mutation | reds |
|---|---|
| the `$ref` name lookup dropped | `CountsNoUndeclaredRowForAParameterTheDocumentDeclaresByReference` |
| the plan-before-create guard removed | `DocumentsAStatusFromAKeptResponseButNotFromASampledOne` |
| positional equality restored | `WritesNothingIntoADocumentNobodyEdited` |

The last one works because the fixture's stored 200 body now carries its members in the other order from the document, so the round-trip case pins the comparison too.

Engine suite: **2916 tests, all passing** (45 across the four export suites). `docs/engine/api-reference.md` and `docs/app/openapi.md` state all three rules.

One correction to the table in the PR description: the `from_import` row means dropping `example.from_import` from the condition, so that *every* undeclared example is treated as sampled - that is the mutation that reds those four. Disabling the branch outright (never sampled) reds three instead: the two named plus the new route case.

---

PR #1458, opened 2026-09-05
fix(app): the variables editor merges onto the freshest map instead of its own copy (#1439)

### LINKED ISSUE #1439: app(variables): the variables editor saves its whole map, so an agent's `update_environment` or `update_globals` is overwritten by the next blur while the table is dirty
## Problem

`app/src/modules/variables/main/VariableTableEditor.tsx` holds a copy of one scope's whole variable map. Its seed effect refuses to reseed while the table is dirty (`hasPendingChangesRef.current`, set at `:313`, cleared at `:331`), which is right for the user's rows. Its `performSave` (`:340` onward) then PUTs `{ id, name, variables: <the editor's whole map> }`, and the engine replaces the blob (`apply_json_field`). So:

1. The table is open with an unsaved row.
2. An agent calls `update_environment` adding or changing a variable (`electron/mcp/tools.ts`, the tool's own read-merge-write across two round trips, "Fetch the current env so we merge (the PUT replaces the whole blob)").
3. The `environment` event invalidates `environments.all` (`mcp-invalidation.ts:105`); `dataVariables` changes; the seed effect returns early because the table is dirty.
4. The user's next blur or Cmd+S sends the editor's map. The agent's variable is gone.

The context bar does this right: `useVariableCommit` (`components/layout/context-bar/variable-commit.ts:178-215`) reads the map out of the query cache at commit time and merges one key. The editor is the surface that holds a copy. Verified at master `90c2fff3`; the file changed since the sweep at `c41ea644` but the guard and the save shape did not.

Severity: less common (the Variables tab must be open and dirty), silent, destroys the agent's write. `name` survives because it is read from the freshened prop.

## Fix pathway

Save changed rows, merged onto the freshest map, the way the context bar does:

1. Keep the seeded map beside the edited one. On save, compute the row-level diff (added, changed, removed keys) and apply it to the **current cached** map from `useEnvironmentsQuery` / `useGlobalsQuery`, not to the seeded one; send that. An agent's untouched keys survive; the user's edits land.
2. A key both sides changed is a conflict: keep the user's value and show the same conflict callout #1436 and #1437 introduce, naming the variable.
3. While dirty, when the cached map changes, adopt rows the user has not touched into the table (they appear as new rows) so the user sees the agent's work rather than a stale table; a touched row is left alone.
4. `useVariableCommit`'s merge helper is the primitive to share; do not write a second merge in the editor.

## App changes

`app/src/modules/variables/main/VariableTableEditor.tsx`, a shared merge helper lifted out of `variable-commit.ts` (or imported from it), the conflict callout. No engine change: the PUT still carries the whole blob, now computed from fresh data. MCP unchanged.

## Tests

`VariableTableEditor` tests (find the existing file with `rg -l VariableTableEditor app/src --glob '*.test.*'`): mount with `{A}`, edit a row to dirty it, rerender with the query returning `{A, B}`, blur; assert the mutation payload contains `B` and the user's edit. Today it does not contain `B`, which is the mutation check. Add: both-changed key shows the callout; a clean table reseeds as today. `pnpm test VariableTableEditor variable-commit VariablesSection`, `pnpm type-check`, `pnpm lint`.

## Docs, same commit

`docs/app/state-management.md`, the variables editor and context-bar commit sections (they should now name one merge rule).

## Sequencing

Wave 1 of the two-writer cluster, parallel-safe with #1437, #1438 and #1440; shares the conflict callout with #1436 and #1437. Disjoint from everything else open.

## Acceptance criteria

- With the Variables tab open and one row edited, an agent adding a variable to the same environment results in a saved environment containing both the agent's variable and the user's edit.
- An agent changing the variable the user is editing produces a visible conflict naming it; the user's value is kept until they choose.
- The editor and the context bar merge through one helper, pinned by a test that both import it.

---

athrvk:
## Description

`VariableTableEditor` (the Variables tab's table editor for globals,
environments and collections) held a local copy of one scope's whole
variable map and saved that whole copy on every blur. If an MCP agent's
`update_environment` / `update_globals` / `update_collection` landed while
the table had an unsaved edit, the next save sent the editor's stale copy
of the map back out - silently overwriting the agent's write, since the
row-init effect deliberately refuses to reseed over a dirty table (the
guard that protects the user's own keystrokes).

**Root cause, one mechanism:** `performSave` built its payload entirely
from the editor's own local rows rather than reading the freshest known
state at save time. The context bar's `useVariableCommit` already solved
this for a single-key commit - it re-reads the query cache at commit time
and merges just the one changed key onto it - but that merge was inlined
there rather than shared, so `VariableTableEditor` never got it.

**Approach:** extract the merge into `app/src/lib/variable-merge.ts`
(`mergeVariableChanges` / `findVariableConflicts`) and have both writers use
it. `performSave` now diffs the user's edited rows against a `baselineRef`
(the map the table last agreed with the server on) to get only the keys
the user actually touched, then merges that diff onto `dataVariables` - the
live, query-cache-backed prop - rather than replacing it wholesale. A key
neither side touched is untouched by the write; a key only the user touched
saves as edited; a key only an agent touched survives because it was never
in the diff to begin with.

A key **both** sides touched is a conflict rather than a coin flip: while
the table is dirty, the reseed effect recomputes the same diff on every
`dataVariables` change purely to call `findVariableConflicts`, and the
table shows a `Callout` naming the key with a "Take theirs" action. The
user's edit is what saves by default (it's already the value in the diff);
nothing is resolved silently in either direction until the user chooses.

**Rejected:** live-adopting an untouched external addition into the visible
table while the user is still editing (part of the issue's suggested fix
pathway). `performSave`'s merge already guarantees the *saved* result is
correct regardless of what the table currently displays, so the only thing
live adoption would add is letting the user see the agent's write before
saving - a UX nicety outside the issue's acceptance criteria. Building it
turned out to introduce a second, largely-redundant code path (the table's
local rows would just end up mirroring what the merge already computes) and
made an early version of the merge-at-save mutation-check pass for the
wrong reason - reconciliation was masking the very regression the test was
supposed to catch. The simpler design is also the more robust one here.

**Also fixed in this PR:** an adversarial review round caught that deleting
a row with an open conflict left the conflict's `Callout` on screen for a
key no longer in the table - "Take theirs" then found no row to update, and
the following save still read the key's absence as the user's own deletion
regardless, silently discarding the value the button claimed to take.
`removeVariable` now drops the conflict for the key it deletes. This is
squarely inside acceptance criterion 2 (conflict handling), so it's fixed
here rather than filed separately.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- `cd app && pnpm test` - **7953+ tests, all passing**; `pnpm type-check`,
  `pnpm lint` (zero warnings) and `pnpm format:check` clean.
- No engine changes, so no engine build/test run.
- `mkdocs build --strict` (from `.github/`) clean - validates the new
  cross-reference anchor between `COMPONENTS.md` and `state-management.md`.

New tests, mutation-checked (reverted the fix, confirmed the case reds,
restored):

- `app/src/lib/variable-merge.test.ts` - unit tests for `mergeVariableChanges`
  (carries an untouched key through, adds, deletes, does not mutate its
  input) and `findVariableConflicts` (no conflict when only one side moved a
  key or both land on the same value; a conflict when both move a key to
  different values, including a deletion colliding with a concurrent
  update). Two more tests read `variable-commit.ts` and
  `VariableTableEditor.tsx`'s source directly to pin that both actually
  import `mergeVariableChanges` from the shared module - a hand-rolled copy
  in either file would not fail any behavioural test above, only this
  import guard.
- `app/src/modules/variables/main/agent-write-reconciliation.test.tsx` -
  behavioural, on the real component: an untouched agent addition survives
  the user's next save beside the user's own edit; an untouched agent
  deletion is not resurrected by the save; a same-key conflict shows the
  `Callout`, keeps the user's value through a save, and clears once that
  save lands; "Take theirs" applies the fresh value and saves it; deleting
  a conflicting row drops its callout and the key stays deleted.
- All pre-existing variables-editor tests still pass unchanged:
  `concurrent-edit-clobber`, `VariableTableEditor.status-stomp`,
  `created-at-ordering`, `secret-reveal-identity`, `key-value-parity`,
  `CollectionVariablesSection`, `VariableRow`.

This PR went through an independent two-pass review (acceptance criteria,
line by line, and repo conventions/scope) before the second commit; the
acceptance-criteria pass is what caught the delete/conflict bug above. The
conventions pass found one cosmetic nitpick (a few redundant `as VariableMap`
casts where the value was already that type) - cleaned up in the same
follow-up commit.

## Visual evidence

Engine-only surface has no visual change here since there is no engine
change at all, but this **is** a user-visible app change (a new `Callout`
can appear in the Variables tab) and I was not able to drive the full
Electron app under Xvfb in this run to capture it - the conflict UI is
exercised end-to-end in `agent-write-reconciliation.test.tsx` (React
Testing Library against the real component tree) rather than in a
screenshot. Noting this explicitly rather than omitting the step silently.

## Checklist
- [x] I have added tests

## Related Issues

Closes #1439


athrvk:
The "CI gate" failure on `e6a5340` is not a real failure: every App job on that commit passed except `App on macos-latest`, which shows `cancelled` - it was still running when I pushed `f077ec9` about a minute later, and GitHub's concurrency cancellation killed it mid-run. `e6a5340` is no longer the PR's head, so this doesn't need a fix; I'm watching CI on the current head (`f077ec9`) instead.

---

Issue #1460, opened 2026-09-05
engine(import): the Postman importer drops oauth2 `state`/`tokenName`/a present `accessToken`, and re-encodes an already-invalid `%` in a query value


athrvk:
## Problem

Measured during #1443 (the Postman importer's silent-drop counters) and deliberately left out of that PR: both are real, but neither is in #1443's Fix pathway or Acceptance criteria, so fixing them there would have widened that PR beyond its own scope.

**1. `map_postman_oauth2` drops `state`, `tokenName`, and a present `accessToken`.** (`engine/src/core/import_document.cpp:400-454`, verified at master `755a651`.) `state` is never read at all. `tokenName` is never read at all. `accessToken` is read only inside the narrow bearer-shortcut guard (`:406-408`: `!stated("grant_type") && !stated("accessTokenUrl") && !stated("authUrl") && stated("accessToken")`); once any of `grant_type`/`accessTokenUrl`/`authUrl` is present, a pre-fetched `accessToken` sitting alongside them is silently dropped even though `default_oauth2_config`/`oauth2_auth` never carries it forward on the general path. None of the three is counted anywhere - not in `ImportTally`, not in `nonExecutableAuth`.

**2. `join_params_into_urls` re-encodes an already-invalid percent escape.** (`engine/src/core/import_document.cpp:2348-2363`, mechanism traced during #1443's review.) `safe_decode` (`:841-858`) returns text unchanged when a `%` is not followed by two hex digits (e.g. a literal `%ZZ` in a query value), which is the right per-field answer. But `join_params_into_urls`'s later re-encoding pass (`append_params` -> `query_string` -> `encode_uri_component`) then percent-encodes that literal `%`, turning `%ZZ` into `%25ZZ` on the stored URL - a value change the structural (non-wire) conformance comparison used by `import_parse_test.cpp` cannot see, since it compares parsed params, not the rejoined URL string.

Severity: both rare (a Postman export using OAuth2's minimal `accessToken`-only shape alongside a full grant config, or a query value containing a literal un-escaped `%`) and silent.

## Fix pathway

[...]

Issue #1436, opened 2026-09-05
app(request-builder): an open request never adopts an external change except its name - the next autosave writes the stale record back over an agent's edit, a move wipes the user's draft, a delete flushes it into a 404


athrvk:
## Problem

The MCP server and the UI write the same requests. After an MCP write, `mcp:data-changed` invalidates the TanStack cache and the request refetches (`app/src/lib/mcp-invalidation.ts:58-77`); that part works, and the tab strip shows the new title. The request builder does not read it. Six experiments against the real provider, query and save-manager stack at master `c41ea644` (files unchanged at `90c2fff3`):

| Case | Observed | Outcome |
|---|---|---|
| Clean tab, agent changes the URL | builder still shows the old URL; only `name` updates | stale; a Send sends the stale request |
| Dirty body, agent changes the URL | autosave PUTs the user's body **with the old URL**; engine row reverts to the pre-agent URL | silent loss of the agent's write |
| Dirty URL, agent changes the URL | user's URL wins, no conflict signal anywhere | silent loss of the agent's write |
| Agent deletes the request under a dirty tab | "This request no longer exists" pane is right, but the provider's unmount flush PUTs the draft to the deleted id, gets a 404, toasts "Request not found" as an error | silent loss of the user's edit, misleading toast |
| Agent moves the request (`move_item`) under a dirty tab | `collectionId` changes, the reset gate fires, the draft is replaced and the pending save cancelled | silent loss of the user's edit |
| Agent's write lands while the user's save is in flight | the #1381 generation guard passes (it only tracks local edits); the save's response overwrites the fresher cache | silent loss of the agent's write |

Root cause, one mechanism with two faces:

- `RequestBuilderProvider.tsx:604-608`: the draft is seeded once and re-seeded only when `id`, `collectionId` or `initialResponse` change. A refetch of the same id with new content changes none of them. Only `name` has an adopt path (`:653-665`, added by the fix for the exact same defect on that one field; `RequestBuilderProvider.name-sync.test.tsx:13-20` describes it).
[...]

PR #1461, opened 2026-09-05
fix(engine): count and map the Postman importer's silent field drops

### LINKED ISSUE #1443: engine(import): the Postman importer drops custom verbs, hawk and oauth1 auth, path-variable values, `strictSSL` and variable descriptions with no counter - only file bodies are reported
## Problem

`docs/index.md:195-196` promises that what the Postman importer cannot keep is "dropped and reported with a count rather than silently". Measured at master `c41ea644` (`engine/src/core/import_document.cpp` unchanged at `90c2fff3`) by importing a Postman v2.1 collection built to hit every branch; the parse meta reported `skipped [{file_body: 2}]` and `nonExecutableAuth 3`, and nothing else, while these were lost:

| Field | What happened | Anchor |
|---|---|---|
| `request.method: "PROPFIND"` (also `PURGE`, `LINK`, `COPY`, any custom verb Postman allows) | silently became `GET` | `to_method:642-653` allows seven methods; the `unsupported_method` skip kind exists and is never counted on this path |
| `auth.type: "hawk"`, `"oauth1"`, `"edgegrid"`, a non-string `type` | became `{mode: none}`, not in `skipped`, not in `nonExecutableAuth` | `map_postman_auth:600`, `:633`; `postman.md:197` documents the drop as silent, which is the defect |
| `url.variable[]` (Postman path variables with values, `/users/:userId` + `userId=42`) | dropped; `:userId` stays literal in the URL | `pm_url:884-905` never reads `variable`; undocumented |
| a URL given only as `host[]` and `path[]` with no `raw` (schema-legal) | imports as `""` | `pm_url:893` reads only `raw` |
| `protocolProfileBehavior.strictSSL: false` | dropped; `verifySSL` stays `true`, so the request fails on send | `pm_redirects:944-961` reads two keys; `requests.cpp:236` accepts `verifySSL` |
| collection `variable[].description`, `variable[].type` other than `secret` | dropped | `to_var_record:128-160` |
| oauth2 `state`, `tokenName`, an existing `accessToken` | dropped | `map_postman_oauth2:400-455` |
| a raw URL with an invalid escape (`x=%ZZ`) | rewritten to `x=%25ZZ` | `join_params_into_urls:2348` re-encodes |

Everything else in the fixture survived: nesting, names, descriptions, disabled rows with descriptions, every executable auth mode including oauth2 grant details, all body modes with file parts flagged, scripts at all three levels, variables with `secret`, redirects, examples.

Severity: uncommon per field, silent, and two of them change what is sent (`PROPFIND` as `GET`, `strictSSL` ignored). Low priority, cheap to fix: counters and three mappings.

## Fix pathway

1. **Count every drop.** Add skip kinds where missing and count them in the paths above: `unsupported_method` (with the verb in the item's note), `unsupported_auth` (hawk, oauth1, edgegrid, unknown), `path_variables` (per item that carried `url.variable`), `url_without_raw`, `variable_metadata` (description or non-secret type). `nonExecutableAuth` stays for digest, awsv4 and ntlm, which import as data.
2. **Map what Vayu has a home for.** `strictSSL: false` to `verifySSL: false`; `url.variable[]` values to a request-level param or, better, to `{{userId}}` in the URL plus a collection variable when the key is unique (say which in the PR); a `host[]`/`path[]`-only URL assembled from its parts the way Postman does.
3. **Docs.** `docs/app/import-collections/postman.md` lists each counter and each mapping; `:197` stops describing a silent drop.

The MCP `import_document` path and the dialog share `parse_import`, so both get the counts.

## App changes

`app/src/modules/collections/ImportModal.tsx` renders any new skip kinds it does not know (check the summary's kind-to-label map); `electron/mcp/tools.ts` `import_document` forwards `meta.skipped` already. `types/api.ts` if skip kinds are an enum.

## Tests

- `engine/tests/fixtures/import-conformance.json` and the Postman conformance test: add items for each row above and assert the counter and, where mapped, the field (`verifySSL === false`, the URL carries the path variable's value). Mutation check: remove one counter, its case reds.
- `pnpm test ImportModal` for the label map.
- `python build.py -e -t && cd engine && ctest --preset linux-dev -R Import --output-on-failure`.

## Docs, same commit

`docs/app/import-collections/postman.md`, `docs/index.md:195-196` if the sentence's list grows.

## Sequencing

Wave 1 of the interop cluster, parallel-safe with #1441, #1442 (the two export issues) and #1445 (curl). Shares `import_document.cpp` with #1444 (OpenAPI `security`) in different functions (`parse_postman` region vs `parse_openapi`); rebase-level only. Disjoint from the two-writer cluster (#1436 to #1440).

## Acceptance criteria

- Importing a Postman v2.1 item with `method: "PROPFIND"`, `auth.type: "hawk"`, `url.variable`, `strictSSL: false` and a variable description produces a parse summary that names each of the five, and the request carries `verifySSL: false` and the path variable's value.
- No Postman field the importer reads and discards is absent from the summary, pinned by the conformance fixture.

---

athrvk:
## Description

The Postman importer silently dropped several fields instead of counting them, breaking the promise `docs/index.md` makes: "what the importer cannot keep is dropped and reported with a count rather than silently." Root cause: `to_method`, `map_postman_auth`, `pm_url`, `pm_redirects` and `to_var_record` each had a fallback branch that discarded a field with no counter and, in two cases, a field Vayu could actually store but never read.

Approach: extend the existing `ImportTally`/`PostmanCounts` mechanism (the same one `file_body`/`malformed_item` already use) with five new skip kinds - `unsupported_method`, `unsupported_auth`, `path_variables`, `url_without_raw`, `variable_metadata` - and map the two fields Vayu has a home for: `protocolProfileBehavior.strictSSL` to the request's `verifySSL`, and `url.variable[]` path segments to a `{{key}}` template plus a merged collection variable (falling back to a `host[]`/`path[]`-assembled URL when `raw` is absent). `path_variables` and `url_without_raw` are marked informational (like `default_response`) since the value survives as a mapping rather than being lost.

Rejected alternative: mapping `url.variable[]` onto a request-level query param instead of a collection variable, which the issue allowed as a fallback. A collection variable reuses the existing `to_var_record` primitive and lets multiple requests share one `{{userId}}` template the way a hand-built collection would, rather than duplicating the value per request.

One "written but never read" bug caught by this PR's own test: `verifySSL` was set on the parsed draft but `apply_request`'s optional-field carry-list didn't know about it, so the value would have been silently dropped between parse and persisted apply - the same defect class issue #1436 is about. Fixed by adding it to the carry-list; `PostmanImport.MapsStrictSSLFalseToVerifySSLThroughToTheApplyPayload` pins both halves.

Also fixed in this PR: the environment/globals variable parser (`parse_postman_variables`) shared `to_var_record` but hardcoded `meta.skipped: []`, so it never surfaced its own metadata drops. Wired through the same tally - it was a one-line gap next to the function being changed anyway, not separate scope.

## Deferred (not in this PR)

Two related findings measured during this fix are explicitly outside issue #1443's Fix pathway and Acceptance criteria (see the issue for the exact reasoning): `map_postman_oauth2` still drops `state`/`tokenName`/a present `accessToken` in some paths, and `join_params_into_urls` still re-encodes an already-invalid `%` escape. Filed as #1460 with the measured detail and a fix pathway.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` - all 2915 tests pass, including 7 new `PostmanImport.*` cases (one per fixed field, each mutation-checked: reverting the corresponding line reds the case) and the full `ImportConformance`/`ImportApplyPayload` fixture suite, whose `postman-environment`/`postman-globals` cases now assert `variable_metadata` counts.
- `cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check` - all pass (7939 app tests, including a new case in `ImportModal.notices.test.tsx` pinning the two new kinds as informational vs. destructive).
- No user-visible change beyond two new label strings in the import preview's drop/notice list (`ImportModal.tsx`'s `SKIPPED_LABELS`), covered by the component test above rather than a screenshot - a two-line text addition to an existing bulleted list didn't seem to warrant spinning up a full Electron/Xvfb session.

## Checklist

## Related Issues
Closes #1443
Deferred follow-up: #1460

---
_Generated by [Claude Code](https://claude.ai/code/session_01QZtUCoUwS4AQnAWGXrLByU)_


athrvk:
CI's `Build` check (the `mkdocs build --strict` docs job) failed on `bbf4aeb`: a link I added in `docs/app/import-collections/README.md` guessed the wrong anchor slug for `postman.md`'s "Variables & environments" heading (`#variables-environments` instead of `#variables--environments`). `.github/mkdocs.yml` pins GitHub-compatible slugs, which keep the double dash a removed `&` leaves behind rather than collapsing it. Fixed in `4f9df10` and verified locally with `mkdocs build --strict -f .github/mkdocs.yml` (clean build, no warnings).

---

PR #1471, opened 2026-09-05
fix(engine): read an OpenAPI operation's own security into its imported auth

### LINKED ISSUE #1444: engine(import): the OpenAPI importer ignores per-operation `security`, so an operation declared with `security: []` inherits the collection's bearer token and sends credentials the spec says it must not
## Problem

The OpenAPI importer maps the document's one primary security scheme onto the collection's auth (`engine/src/core/import_document.cpp`, `parse_openapi` around `:2203`, with `openapi_drafts.cpp` building the requests). Per-operation `security` is read nowhere. Measured at master `c41ea644` (files unchanged at `90c2fff3`) with a 3.0.3 document declaring five schemes: `listPets` (`apiKeyQuery` or `oauth`), `deletePet` (`basicAuth`), `placeOrder` (`oauth`), `getThing` (`mutualTLS`) all imported as `auth: inherit`, taking the collection's bearer scheme; and `getPet`, declared `security: []` - explicitly unauthenticated - **also** imported as `inherit`, so Vayu sends the bearer token to an endpoint the contract says takes none.

`docs/app/import-collections/openapi-v3.md:283` documents "per-operation `security`, scopes" under "Dropped / not represented", so the omission is recorded. What is not recorded, and is the defect, is that the drop is not neutral: an operation that opted out of security is given security. For an API that rejects unexpected credentials, or logs them, that is a wrong request the user did not write.

Also dropped silently on the import side, measured: the second `servers` entry (no environment created), `requestBody` named `examples` in favour of a schema-level `example`, `multipart/form-data` skipped when a second media type exists, a `deprecated: true` flag, array parameter examples (`["available"]` became `""`). And 3.1 documents are reported as `format: "OpenAPI 3.0"` in the parse meta while the export dialect says 3.1.0.

Severity: less common (specs that mix schemes per operation), silent, and the `security: []` case sends credentials where none were declared. Not data loss; a wrong request.

## Fix pathway

1. **Read per-operation `security`.** In the draft builder, when an operation carries `security`, resolve it: `[]` means `auth: none` (the explicit opt-out, and the one case that must not be `inherit`); a single requirement naming the same scheme the collection took means `inherit`; a requirement naming a different scheme Vayu can express (`http basic`, `http bearer`, `apiKey`, `oauth2`) means that mode on the request, with the same `{{clientId}}`-style placeholders the collection mapping uses; anything else (`mutualTLS`, `openIdConnect`, alternatives Vayu cannot pick between) stays `inherit` and is counted (`security_unmapped`) with the scheme name in the note.
2. **Keep the exports consistent.** #1442 (the bound export) does not write `security`; nothing changes there. #1441 (the skeleton export) part 2 writes `securitySchemes`, and a request-level mode must round-trip into a per-operation `security` entry there. Coordinate with #1441.
3. **Small ones in passing:** report the 3.1 format correctly in the meta; count the dropped second server (`servers_dropped`) so the user knows an environment was not created.
4. **Docs.** `openapi-v3.md:283` moves per-operation `security` from "dropped" to "mapped", with the table of what each shape becomes.

## App changes

None beyond the summary's label map for any new skip kind (`ImportModal.tsx`). MCP `import_document` forwards the meta already.

## Tests

- OpenAPI conformance fixture and test: an operation with `security: []` imports as `auth.mode === "none"`; one naming `basicAuth` imports as `basic`; one naming the collection's scheme imports as `inherit`; one naming `mutualTLS` stays `inherit` and is counted. Mutation check: remove the `[]` branch and the first case reds with `inherit`.
- `python build.py -e -t && cd engine && ctest --preset linux-dev -R Import --output-on-failure`.

## Docs, same commit

`docs/app/import-collections/openapi-v3.md`, `docs/app/openapi.md` where the collection-level scheme mapping is described.

## Sequencing

Wave 2 of the interop cluster: lands after #1441, whose `securitySchemes` export is what a request-level mode needs to round-trip. Shares `import_document.cpp` with #1443 (Postman import) in a different function; rebase-level only. Parallel-safe with #1442 and #1445; disjoint from the two-writer cluster (#1436 to #1440).

## Acceptance criteria

- Importing a document whose operation declares `security: []` produces a request with no auth that sends no `Authorization` header, verified through `POST /compose`.
- An operation declaring a scheme Vayu can express gets that mode; one it cannot is counted in the summary by scheme name.
- The parse meta names 3.1 documents as 3.1.

---

athrvk:
## Description

**Root cause.** The OpenAPI importer mapped only the document's one primary security scheme onto the collection's auth (`primary_scheme` + `scheme_to_auth_v3`/`v2` in `engine/src/core/import_document.cpp`) and never read an operation's own `security`. `draft_request` hardcoded every imported request's auth to `{"mode":"inherit"}` unconditionally. So an operation declared `security: []` - OpenAPI's explicit "this endpoint takes no auth" - imported as `inherit` and sent the collection's bearer token to an endpoint the contract says takes none.

**Approach.** `SpecRequestDraft` (not `DraftRequest` - see below) now carries the operation's raw `security` array when the operation declares the key at all (`openapi_drafts.cpp`'s `build_drafts`). `parse_openapi` resolves it through a new `operation_auth_override`, which reuses the existing `scheme_to_auth_v3`/`v2` mappers rather than reimplementing scheme mapping:
- absent `security` key → keeps today's `inherit`.
- `[]` or `[{}]` (OpenAPI's "security is optional here" shape) → `{mode: "none"}`.
- one requirement naming the same scheme the collection already took → `inherit` (identical either way, and it keeps following the collection if its auth is edited later).
- one requirement naming a different scheme Vayu can express → that scheme's mapped mode.
- anything Vayu cannot resolve to one concrete mode - an OR of alternatives, an AND of multiple schemes, or a scheme type with no Vayu mode (`mutualTLS`, `openIdConnect`) - stays on `inherit` and is counted as a new `security_unmapped` skip kind rather than guessed at. Same rule for Swagger 2.0's `securityDefinitions`.

**Two related gaps caught while in this code** (both named in the issue's Problem section, not folded into a wider fix):
- `servers_dropped`: an OpenAPI 3.x document's `servers` entries past the first name no environment an import creates - now counted instead of silently ignored.
- The parse meta's `format` always reported `"OpenAPI 3.0"` for every 3.x document, because `walk::Dialect` deliberately collapses 3.0/3.1 into one `V3` for readers that only need "is this 3.x". Added `walk::dialect_format_name`, which re-reads the document's own `openapi` string, and used it in both `POST /import/parse` (`import_document.cpp`) and `POST /specs/describe` (`openapi_document.cpp`), which shared the same collapsing bug.

**Rejected alternative:** putting the new `security` field on `DraftRequest` instead of `SpecRequestDraft`. `DraftRequest` is exactly what the sync diff (`spec_request_drafts_of` / `spec_diff.hpp`) compares, under a deliberate "one auth value for every operation" contract its own doc comment states (a difference there is always the user's). Per-operation security is import-only; the sync diff's behavior is untouched by this PR.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` - **2950 engine tests, all passing** (fresh full-suite run on this branch, no pre-existing failures).
- 9 new cases in `engine/tests/import_parse_test.cpp` (`ImportParse.OperationSecurity*`, `OperationWithNoSecurityKeyInheritsAndCountsNothing`, `CountsEveryServerPastTheFirstAsDropped`, `ReportsA31DocumentApartFrom30`) covering every branch above, for both OpenAPI 3.x and Swagger 2.0.
- Mutation-checked by hand: disabled the `security: []` → `none` branch, rebuilt, and confirmed exactly the two tests exercising it (`OperationSecurityEmptyArrayImportsWithNoAuth`, `OperationSecurityIsReadForSwagger20Too`) turned red while the other 2948 tests, including every other new case, stayed green; restored, full suite green again.
- `cd app && pnpm type-check && pnpm lint && pnpm format:check && pnpm test` - all clean, **7984 app tests passing** (no app behavior change beyond the new `SkippedItem` kinds' labels; `pnpm test ImportModal` - 66 tests - specifically green).
- `clang-format-19 --dry-run -Werror` and `clang-tidy-19` (via the repo's pre-commit hook) clean on every touched engine file.

**Pre-existing, unrelated finding, not fixed here:** `clang-tidy` on `engine/tests/import_parse_test.cpp` surfaces a `clang-analyzer-optin.performance.Padding` finding on `struct vayu::Response` (`engine/include/vayu/types.hpp:445`), reached because the test file transitively includes it via `vayu/db/database.hpp`. Verified this reproduces identically on the base commit with none of this PR's changes applied (`git stash` + re-run) - it is the same finding open PR #1469 already fixes with a `NOLINTNEXTLINE`, just not yet merged to `master`. Not this PR's to fix.

**No user-visible change beyond three new import-summary lines.** The renderer's Body/Auth panels, MCP tools, and `POST /compose` are unaffected: a request's `auth` field already has a renderer, and the new `security_unmapped`/`servers_dropped` skip kinds ride the existing `ImportModal` skipped-items list (`SKIPPED_LABELS` entries added) the same way `unsupported_auth` already does. No screenshot: the change is additional rows in an existing list-of-strings UI, already covered by `ImportModal` tests.

## Docs, same commit

`docs/app/import-collections/openapi-v3.md` (per-operation security moved from "dropped" to "mapped", with the full mode table; `servers_dropped`/`security_unmapped` added to the `SkippedItem.kind` table; the `meta.format` line states the 3.1 read), `docs/app/import-collections/openapi-v2.md` (same auth override rule, cross-referenced rather than duplicated; the shared-kinds sentence updated). `app/src/types/api.ts`'s `SpecDescribeResponse.format` doc comment gains the 3.1 value.

## Checklist
- [x] I have added tests

## Related Issues
Closes #1444

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gky8UmTWFJfHJZD74Ys3ud)_


athrvk:
The "CI gate" failure on `fe5fba9` (the pre-merge commit) is not a real failure - its `app` and `engine` jobs show `cancelled`, not a test failure, because I pushed `2f180a9` (merging `master` in to clear a "behind" state) while that run was still in progress. A fresh run is already underway on `2f180a9` with no failures so far; that's the one that counts.

---

athrvk:
Root-caused and fixed the real "Engine on ubuntu-latest" failure on `2f180a9`: the release preset builds with `-O3 -Werror -Wconversion`, which my local debug-preset testing didn't exercise. `operation_auth_override` returned a bare `json{...}` from a function returning `std::optional<json>` in three places - copy-initializing an `optional<json>` from a `json` puts nlohmann's templated `operator ValueType()` up against `optional`'s own converting constructor, which GCC's release build reports as an ambiguous-conversion warning. `scalar_stub` in `openapi_drafts.cpp` already has a comment documenting this exact gotcha and its fix (`std::make_optional`); `operation_auth_override` now follows the same pattern.

Verified in `9bf0f74`: `cmake --preset linux-prod && cmake --build --preset linux-prod` builds clean, and `ctest --preset linux-prod` - the same preset CI's "Engine on ubuntu-latest" job uses - passes all 2950 tests.

---

PR #1511, opened 2026-09-06
fix(engine): tally a load replay's mixed pass/fail pm.test results

### LINKED ISSUE #1502: engine(runs): one failing `pm.test` in a load replay collapses the whole sample to "Script error: " with an empty message and discards the tests that passed
## Problem

Measured at master `800426d8`. A load run whose `tests` script has two `pm.test` calls, one passing (`status is 200`) and one failing (`body has a unicorn`):

- `POST /execute`, same script: `[{"name":"status is 200","passed":true},{"name":"body has a unicorn","passed":false,"error":"AssertionError: Expected undefined to equal 'yes'"}]`.
- The load run's report: `testValidation { samplesTested 1, testsPassed 0, testsFailed 1 }` and the failure row `{"failures":["Script error: "],"totalFailed":1,"totalPassed":0}`.

Six further runs confirmed the rule: any script with at least one failing test reports `0 passed / 1 failed / "Script error: "`; scripts whose tests all pass tally correctly.

Mechanism:

- `engine/src/runtime/script_engine.cpp:8755-8760`: after the script runs, `result.success = false` is set when any `TestResult` has `passed == false`, with no `error_message`.
- `engine/src/core/run_manager.cpp:173-191` (`run_replay`): the `if (result.success)` branch counts tests individually; the `else` branch counts one failure and records `"Script error: " + result.error_message`, which is empty. `result.tests` is never read on that branch.
- The design path (`execution.cpp:698-718`, `append_tests`) reads `result.tests` regardless of `success`, which is why Send reports both.

Every replay test in `scenario_load_test.cpp:725-760` uses one `pm.test` per script, so the mixed case is untested, and no test pins the failure text (`load_identity_test.cpp:204-216` prints it).

Severity: medium. Loud but wrong: the report names no assertion, hides the ones that passed, and misstates the tally the assertion-rate threshold in #1494 will be built on.

## Fix pathway

1. In `run_replay`, read `result.tests` first: for each test, count passed or failed and record `name + ": " + error_message` for the failed ones, exactly as the design path does. Only a script that threw (`result.error_message` non-empty and `tests` empty, or a caught exception) takes the "Script error" branch, and that branch carries the message.
2. `script_engine.cpp:8755-8760` keeps `success = false` for the caller's sake but the replay no longer keys on it; alternatively split `success` into `ran` and `all_tests_passed`, which is clearer for every caller.
3. Failure messages are still capped at `MAX_FAILURE_MESSAGES` (10, `constants.hpp:691`), but each is a real assertion name.

## Tests

`scenario_load_test.cpp` or `load_identity_test.cpp`: a script with one passing and one failing test yields `passed 1, failed 1` and a failure message that names the failing test and carries the assertion text; a script that throws before any test yields one failure with the thrown message. Mutation check: restore the `else`-branch shape, the mixed case reds. `python build.py -t && cd engine && ctest --preset linux-dev -R "ScenarioLoad|LoadIdentity"`.

## Docs, same commit

`docs/engine/scripting.md` `#load-test-scripts`: what the failure list carries per sample.

## Sequencing

Lands first, before #1494's children (the assertion-rate child reads this tally). Independent of everything else open.

## Acceptance criteria

- The mixed script above reports `testsPassed 1, testsFailed 1` per sample and the failure text `body has a unicorn: AssertionError: Expected undefined to equal 'yes'`.
- A throwing script reports the thrown message, never an empty `Script error: `.

---

athrvk:
## Description

A load run's deferred script replay (`run_replay` in `engine/src/core/run_manager.cpp`) branched on `ScriptResult::success`. `script_engine.cpp` sets that flag to `false` whenever *any* `pm.test` assertion in the script fails, with no `error_message` set for that reason alone (an empty message is reserved for a real script-level error). Because `run_replay`'s `else` branch never read `result.tests`, any script with even one failing assertion reported `testsPassed 0, testsFailed 1` and the failure text `"Script error: "` with an empty message - discarding every assertion that actually passed, and hiding which assertion failed.

**Plan.** Three research passes: one verified the issue's file:line anchors against current code (`run_manager.cpp:175-192`, `script_engine.cpp:8757-8763`, and the non-load `append_tests` in `engine/src/http/routes/execution.cpp:701-715` it should mirror - the issue cited `engine/src/core/execution.cpp`, which has moved), one traced every consumer of the load report's `testValidation`/failure-message shape (the app's `TestValidationSummary.tsx`, `types/domain.ts`, the MCP `get_run_report` tool) to confirm the fix changes only the counting logic and not the JSON shape, and one read `docs/engine/scripting.md`'s Load Test Scripts section for the sentence that goes stale under the real fix. Root cause: `result.success` conflates "the script threw" with "an assertion failed" into one flag, and the load path's counting branched on that flag instead of on whether `result.tests` (which is always populated, whether or not the script went on to throw) is empty.

**Fix.** `run_replay` now reads `result.tests` first whenever it is non-empty and tallies each test individually - exactly what the design (non-load) path already does via `append_tests`. Only a script with *no* tests at all falls back to `result.success`: a clean no-assertion script still counts as passed, and a script that threw before ever reaching a `pm.test` call still reports its real thrown message under `"Script error: "`, never an empty one.

**A gap an adversarial review pass caught and this PR also closes**: a script that runs a test to completion and *then* throws (or fails the `pm.request` writeback) has a non-empty `result.tests`, so the first version of this fix tallied the completed test(s) but silently dropped the throw - worse than the pre-fix behaviour for that one sub-case, which at least reported it as one opaque failure. `run_replay` now records that throw as an additional failure alongside whatever tests already ran, carrying its real message. The tally logic was extracted into its own `tally_replay_result` helper in the same commit: adding this branch inline pushed `run_replay`'s cognitive complexity past clang-tidy's gate (26 vs. threshold 25), and the tally reads better as one self-contained decision than as more nesting in an already-long loop body.

**Alternative considered and rejected.** Making the script-level error message additive in the *tests-empty* branches too (there is only one such branch, so this was never actually a fork) - not applicable here; the one real design choice was the additive-vs-dropped question above, resolved as described.

## Type of Change
- [x] Bug fix

## Testing

- `engine/tests/scenario_load_test.cpp`: three new cases next to the existing `AFailingAssertionIsAttributedToItsOwnStep`.
  - `AMixedPassAndFailScriptReportsBothTalliesAndNamesTheFailure` - the issue's own repro (one passing `pm.test`, one failing) - asserts `passed 1, failed 1` and that the stored failure text is the exact `body has a unicorn: AssertionError: Expected undefined to equal 'yes'`, not `"Script error: "`.
  - `AThrowingScriptWithNoTestsNamesTheThrownMessage` - a script with no `pm.test` calls that throws - asserts the stored failure text carries the real thrown message (`boom`), not an empty one.
  - `AScriptThatThrowsAfterAPassingTestStillReportsTheThrow` - a script whose single `pm.test` passes and which then throws - asserts `passed 1, failed 1` and that the thrown message (`late boom`) is recorded, not dropped.
- Mutation-checked by hand, twice: reverting `run_manager.cpp` to the pre-fix branch reddens exactly the mixed-tally test (1 of 2965 engine tests); separately reverting just the new throw-after-tests branch reddens exactly the third test above (the failure row goes missing entirely). Everything else is unaffected either time.
- `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`: **2965/2965 passing** on the final state, including all 44 `ScenarioLoadTest` cases and the `ScriptSendRequestCap*`/`LoadIdentityTest` suites the issue named.
- No app files touched, so `pnpm test` / `pnpm type-check` were not run (nothing in the diff could affect them).

## Docs, same commit

`docs/engine/scripting.md`'s Load Test Scripts section: the "aggregated and reported" bullet now states that `testsPassed`/`testsFailed` tally every `pm.test` call across sampled responses (mixed pass/fail included) and that only a script that threw before any test ran gets a single script-level message.

## No user-visible change

Engine-only: the load report's JSON shape (`testValidation`, the stored failure row's `failures`/`totalFailed`/`totalPassed`) is unchanged, only the values populating it are corrected. No app files were touched.

## Checklist
- [x] My code follows the style guidelines (clang-format 19 clean, clang-tidy clean including the cognitive-complexity gate, on the two changed engine files)
- [x] I have added tests

## Related Issues
Closes #1502

### DIFF
--- a/docs/engine/scripting.md
+++ b/docs/engine/scripting.md
@@ -2360,7 +2360,13 @@ The **language** is current; what is missing is the **host environment**:
 - `samplesTested` in the report (`TestsSampled`) is the **size of that sample**,
   not the run's request count, and `sampling.responseSamplesDropped` beside it
   says how many responses the bound thinned away
-- Results are aggregated and reported in the final report
+- Results are aggregated and reported in the final report: `testsPassed` /
+  `testsFailed` tally every `pm.test` call across every sampled response, so a
+  script whose assertions are a mix of passing and failing reports both counts
+  and names each failing assertion (`name: message`) rather than collapsing
+  the sample to one opaque failure. Only a script that threw before any
+  `pm.test` ran reports a single script-level failure, carrying the thrown
+  message (issue #1502)
 - `pm.info` reports the same identity a Send does: `eventName` is `"test"`, and
   `requestId` / `requestName` are the run's linked request when it has one. It
   also reports `iteration` - the index the sampled submission claimed before it
--- a/engine/src/core/run_manager.cpp
+++ b/engine/src/core/run_manager.cpp
@@ -172,8 +172,11 @@ std::vector<std::string>& failure_messages) {
             }
             auto result = engine.execute (*replay.script, script_ctx);
 
-            if (result.success) {
-                // Check individual test results
+            // Tally every test the script actually ran before falling back to
+            // a script-level verdict - `result.tests` is populated whether or
+            // not the script went on to throw (issue #1502), so a mixed
+            // pass/fail script must not be collapsed to one opaque failure.
+            if (!result.tests.empty ()) {
                 for (const auto& test : result.tests) {
                     if (test.passed) {
                         totals.passed++;
@@ -182,10 +185,9 @@ std::vector<std::string>& failure_messages) {
                         record_failure (test.name + ": " + test.error_message);
                     }
                 }
-                if (result.tests.empty ()) {
-                    // Script ran but had no pm.test() calls - count as passed
-                    totals.passed++;
-                }
+            } else if (result.success) {
+                // Script ran but had no pm.test() calls - count as passed
+                totals.passed++;
             } else {
                 totals.failed++;
                 record_failure ("Script error: " + result.error_message);
--- a/engine/src/core/run_manager.cpp
+++ b/engine/src/core/run_manager.cpp
@@ -92,6 +92,42 @@ struct ScriptReplay {
 };
 
 /**
+ * @brief Tally one replayed script's verdict into @p totals.
+ *
+ * `result.tests` is populated whether or not the script went on to throw, so
+ * it is read first: a mixed pass/fail script must not collapse into one
+ * opaque failure (issue #1502). A throw after tests already ran (or a failed
+ * `pm.request` writeback) is a second, distinct failure and is recorded
+ * alongside them, never dropped just because the tests that did run are
+ * counted - the same distinction `script_errored` draws in the design path.
+ */
+template <typename RecordFailure>
+void tally_replay_result (const vayu::ScriptResult& result,
+ScriptValidationTotals& totals,
+RecordFailure&& record_failure) {
+    if (!result.tests.empty ()) {
+        for (const auto& test : result.tests) {
+            if (test.passed) {
+                totals.passed++;
+            } else {
+                totals.failed++;
+                record_failure (test.name + ": " + test.error_message);
+            }
+        }
+        if (!result.success && !result.error_message.empty ()) {
+            totals.failed++;
+            record_failure ("Script error: " + result.error_message);
+        }
+    } else if (result.success) {
+        // Script ran but had no pm.test() calls - count as passed
+        totals.passed++;
+    } else {
+        totals.failed++;
+        record_failure ("Script error: " + result.error_message);
+    }
+}
+
+/**
  * @brief Replay one script against one set of samples, tallying the results.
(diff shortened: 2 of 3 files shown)


athrvk:
The "CI gate" failure reported on `28f4178` (this branch's first commit) is not a real failure: its `engine` job log shows `engine: cancelled`, because I pushed a follow-up commit (`15c18ca`) while that run was still in progress, which GitHub Actions cancelled in favor of the new run. The current head (`15c18ca`) has its own fresh CI run in progress; watching that one through to green.

---

Issue #1491, opened 2026-09-06
engine(db): the header-strip predicate deletes a user's own bare-UUID X-Request-ID and any header named x-vayu-version regardless of value, trash rows included - it matches on shape and name, not on provenance


athrvk:
## Problem

Found by the upgrade sweep (0.25 workspace opened on `cb829ad3`), re-anchored at `228f1131`. Measured straight out of SQLite before and after the first head start:

```
none-none    before [X-Vayu-Version: 0.24.0][X-Request-ID: 550e8400-…-446655440000][User-Agent: Vayu/0.24.0][X-Custom: none-none]
             after  [X-Custom: none-none]
none-inherit before [X-Request-ID: my-own-fixed-id][User-Agent: MyLoadTester/9][x-vayu-version: hand-written][X-Custom: keep-me]
             after  [X-Request-ID: my-own-fixed-id][User-Agent: MyLoadTester/9]                             [X-Custom: keep-me]
```

The predicate is `legacy_managed_row` in `engine/src/http/default_headers.cpp:106-118`: a row is deleted when the folded name is `x-vayu-version` (any value), or `x-request-id` whose value is a bare RFC 4122 UUID (`is_bare_uuid`, `:66`), or `user-agent` whose value starts with `vayu/`. Two consequences:

1. **A UUID the user pinned is indistinguishable from the one the renderer wrote.** A request whose `X-Request-ID` was set deliberately to a fixed UUID (a replay case, a Postman or Insomnia import carrying one) loses it on upgrade. The first row above is the renderer's own; the second row shows that a non-UUID value survives, which is the only reason the rule looks narrow.
2. **`x-vayu-version` is matched on name alone**, so `x-vayu-version: hand-written` went too. A user who named a header that way did so on purpose.
3. **Trash rows are rewritten.** `none-inherit` had `deleted_at` set and was rewritten anyway; a request restored after upgrading comes back without those rows.

The comment above the predicate (`:105`) says each rule is as narrow as it is for a reason; the measured cases are the ones the narrowness does not reach.

Severity: medium. Uncommon, silent, destructive in a small way, and unrecoverable once the `.bak` has been refreshed (#1487).

## Fix pathway

[...]

Issue #984, opened 2026-08-25
engine(db): recovery restores an unvalidated backup and deletes the corrupt original - validate the .bak, quarantine the evidence


athrvk:
Found in the 2026-08-25 operational sweep, verified at master `96385b2`.

## Problem

The corruption-recovery path in `Database`'s constructor (`engine/src/db/database.cpp:614-720`) makes two risky moves:

1. **It restores `<db>.bak` without validating it.** The comment at `database.cpp:685` says it outright: "we assume the backup itself is valid". The main DB gets an open + `sync_schema` probe before being trusted; the backup it falls back to gets nothing. A corrupt backup (torn copy, disk fault at the same time as the main-file fault) is restored and then either crashes the same probe path or, worse, half-works.
2. **It deletes the corrupt original.** When no backup exists (or after restoring), the corrupt `<db>`/`-wal`/`-shm` files are deleted so the engine can start fresh. Recovery-marker aside (`engine/src/db/recovery.cpp` writes `<db>.recovery`, surfaced on `/health` - that part is good), the user's actual data is destroyed at the moment it was most needed for salvage: SQLite's `.recover` tooling can often extract most rows from a corrupt file, but only if the file still exists.

Combined with the facts that the `.bak` is overwritten on every clean start (`database.cpp:713`) and there is no user-facing backup at all (separate issue, filed alongside), the current worst case is total, unrecoverable data loss from one bad file.

## Fix / solution pathway

All inside `database.cpp`'s recovery block plus `recovery.cpp`:

1. **Validate before restoring**: subject the `.bak` to the same probe the main DB gets (open + `sync_schema`, or `PRAGMA integrity_check` capped with `integrity_check(1)` for speed - state the choice). Only a passing backup is restored. A failing backup is left in place (evidence) and the flow proceeds as if no backup existed.
2. **Quarantine instead of delete**: rename the corrupt `<db>` to `<db>.corrupt-<UTC timestamp>` (and its `-wal`/`-shm` alongside) rather than deleting. Keep at most 2 quarantined sets - prune older ones at the same point, so the directory cannot grow unbounded.
3. **Record the richer outcome**: extend the `<db>.recovery` marker vocabulary (currently `restored_from_backup` | `deleted_corrupt`, `docs/engine/api-reference.md:449-474`) - e.g. `restored_from_backup`, `backup_also_corrupt`, `started_fresh_quarantined`. The health payload shape stays a string, so this is additive.

## App changes

[...]

PR #1520, opened 2026-09-06
fix(engine): batch the header-strip startup pass and back it up before it runs

### LINKED ISSUE #1487: engine(db): the header-strip startup pass rewrites every request row unbatched before health answers, on every start, and the pre-upgrade backup is taken after the probe has already migrated the live file
## Problem

Found by the upgrade sweep, which built engines at `v0.24.0`, `v0.25.0` and `cb829ad3` and upgraded populated workspaces through each path; re-anchored at `800426d8`. Everything below is measured.

0.26.0 added `Database::strip_stored_managed_headers` (`engine/src/db/database.cpp:2031-2048`, called from `init()` at `:1178`), which removes the version, request-id and user-agent headers earlier releases wrote into saved requests. Three defects in how it runs:

1. **Unbatched, per row, on the startup path.** The pass does `get_all<Request> ()` and one `replace (request)` per matching row with no `transaction_guard` - the exact shape #838 fixed for `seed_default_config` (`:3382`, "sixty-odd commits became one"), against a table that is thousands of rows. It runs inside `db.init ()` before the listener starts (`engine/src/daemon.cpp:158-160`, `:217`), so `/health` does not answer until it finishes.

   | workspace | 0.25 `init()` | head first start | of which the strip | head second start |
   |---|---|---|---|---|
   | 3,059 requests, 15 MB | 41 ms | 3,991 ms | 3,958 ms | 155 ms |
   | 12,059 requests | - | 15,556 ms | 15,509 ms | 315 ms |
   | 3,000 requests, `dbSynchronous=2` | - | 7,719 ms | 7,689 ms | - |

   About 1.3 ms per request with the shipped `synchronous=OFF`. `ENGINE_HEALTH_POLL_BUDGET_MS` is 45 s (`app/electron/constants.ts:73`), so a workspace around 35,000 requests crosses the app's readiness budget on this hardware, sooner on slow storage.

2. **It runs forever.** The pass is idempotent and never marked done, so every start materialises every request (bodies and scripts included) through `get_all` to find nothing. The second-start column above is that cost.

3. **The backup does not hold the pre-upgrade rows.** `Database::Database` (`:1043-1062`) probes the real file with `sync_schema ()` first (`:1046`), then `copy_db_files` (`:1062`) refreshes `<db>.bak` from the already-migrated file, and `init ()` then runs the strip. So after head's first start `.bak` still holds the old rows, and after the second start it holds the stripped ones. Verified by reading `requests.headers` out of `vayu.db.bak` after each start. The only user-owned copy is a manual `POST /workspace/backup`.

4. **Undocumented.** `docs/engine/db-schema.md` documents the other two startup repair passes (spec-binding stamping, inbox clearing); `rg -i "strip_stored|managed-header" docs .github/release-notes` finds nothing, and the v0.26.0 notes say the headers "are no longer stored" without saying existing rows are rewritten on first launch.

The predicate's precision (a user's own bare-UUID `X-Request-ID` is deleted) is #1491.

## Fix pathway

1. **One transaction, one pass.** Wrap the loop in `transaction_guard` like `seed_default_config`. Select only candidate rows with a SQL `LIKE` over `headers` for the three folded names (`x-vayu-version`, `x-request-id`, `user-agent`) instead of `get_all`, so a clean workspace costs one scan of the text column rather than a full materialisation.
2. **Mark it done.** Record the pass in `config_entries` (or a `schema_meta` row - see #1492) and skip it when the marker is present. A repair that runs once is a migration and gets a migration's bookkeeping.
3. **Back up before, not after.** In the constructor, take `<db>.bak` from the file **before** the probe's `sync_schema ()`, or keep a second `<db>.pre-upgrade.bak` written only when a repair pass is about to change rows, retained until the next successful start. The user must have one restart of grace **after** the rewrite, not one before it.
4. **Document.** `db-schema.md`, the `requests.headers` row and the startup section: what the pass removes, when it runs, where the pre-rewrite copy is.

## App changes

`engine/src/db/database.cpp` (constructor and the pass). No renderer change.

## Tests

`engine/tests/database_test.cpp` (or the temp-database fixture): seed 1,000 requests carrying `X-Vayu-Version`, open, assert one transaction (count `BEGIN` via the sqlite trace hook, or assert the wall clock stays under a generous bound with `synchronous=FULL`); a second open does not call the strip (spy on the marker). Mutation checks: remove the guard, the transaction count reds; remove the marker, the second-open case reds. A backup-order test: `.bak` after the first open contains the unstripped headers. `python build.py -t && cd engine && ctest --preset linux-dev -R database`.

## Docs, same commit

`docs/engine/db-schema.md` (startup repairs, `requests.headers`, backup semantics), `docs/engine/architecture.md` if it describes the `.bak` refresh.

## Sequencing

Wave E of the upgrade sweep. Land this before #1491 (same function) and before #1492 (which gives the marker a home; until then `config_entries` is fine).

## Acceptance criteria

- Opening a 12,000-request 0.25 workspace on head answers `/health` within the app's 45 s budget on the CI runner, and the strip's log line reports the count from one transaction.
- The second open of the same workspace does not run the strip.
- After the first open, a copy of the pre-rewrite database exists on disk and its `requests.headers` still carries the old rows.
- `db-schema.md` names the pass.

---

athrvk:
## Description

`Database::strip_stored_managed_headers` (`engine/src/db/database.cpp`) rewrote every stored request one row at a time with no transaction, on every engine start, before `/health` could answer - and it never recorded that it had already run, so a clean workspace paid the cost of materialising every request forever. Separately, the constructor's `.bak` refresh happens after `probe_database`'s own `sync_schema()` has already touched the live file, so a schema migration that goes wrong leaves no pre-migration copy to recover from.

**Plan.** Three research passes verified the issue's file:line anchors still held at master `d1fe93d` (the strip loop at `database.cpp:2031-2048`, the constructor's probe-then-backup order at `:1027-1078`, `seed_default_config`'s `transaction_guard` pattern at `:3382`), traced the strip predicate (`strip_legacy_managed_headers`, owned by a separate issue, #1491) and the `.bak`/backup consumers to confirm nothing else depends on the current ordering, and read `engine/CLAUDE.md` plus `docs/engine/db-schema.md` for the documented conventions a fix here has to follow. Root cause: the pass conflated "select every row" with "rewrite the ones that changed", and the constructor's backup is refreshed from whatever the live file holds *after* migration rather than before. Fix: a SQL `LIKE` filter narrows candidates before any JSON parsing, one transaction wraps the actual rewrite, a `config_entries` marker (`managedHeadersStripped`, `advanced`) skips the scan entirely once nothing is left to strip, and a `<db>.pre-upgrade.bak` snapshot - written once, before the first row is ever rewritten - is added alongside the existing `.bak` rather than reordering the constructor, since the existing `.bak`/recovery logic is delicate and untouched by this issue's own fix pathway's first option.

**Alternative considered and rejected:** reordering the constructor to take `.bak` before `probe_database`'s `sync_schema()` (the issue's other suggested option). Rejected because `copy_db_files` there is only trusted after `probe_database` validates the file (issue #984's fix - "the backup is only ever taken from a database that validated"), and moving it earlier would back up an unvalidated file, risking overwriting a good backup with a bad one. The second-file approach the issue also offers gets the same guarantee (pre-rewrite rows survive more than one restart) without touching that validated-backup invariant.

## Type of Change
- [x] Bug fix
- [x] Documentation update

## Testing

- `engine/tests/default_headers_test.cpp`: the existing `StartupStripsTheStoredRowsAndLeavesTheUsersAlone` now resets the new marker first (its own `SetUp` already runs `init()` on an empty database, which now marks the migration done before the test seeds its legacy-header rows). Two new cases: `ASecondStartupSkipsTheScanOnceEverythingIsMarkedStripped` (a row added *after* the marker is set is never touched - proves the skip is unconditional, not just "nothing new to find") and `ThePreUpgradeBackupPreservesTheRowsBeforeTheFirstRewrite` (opens `<db>.pre-upgrade.bak` as a second `Database` and asserts it still holds the unstripped row).
- `engine/tests/config_route_test.cpp`: `AdvancedFlagsExactlyTheRecordedInternals` enumerates every seeded `advanced` config key; the new marker is one more, so it's added to the expected set.
- `engine/tests/temp_database.hpp`: `remove_database_files` now also cleans up `<path>.pre-upgrade.bak` and its `-wal`/`-shm` sidecars, so a test that provokes it doesn't leak the file into the next scratch directory.
- Mutation-checked by hand: removing the `get_config_bool` marker guard makes the second-startup test find and strip the "late" row (fails as expected); removing the `copy_db_files` call makes the backup-existence assertion fail outright.
- `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`: **2965/2965 passing**, including the two new cases and every existing `DefaultHeaderConfigTest`/`ConfigRouteTest`/`DatabaseTest`/`ScratchDatabaseCleanup` case.

**Known test gap, stated rather than hidden:** the issue's suggested test for "one transaction" is a SQL trace-hook counting `BEGIN` statements, or a wall-clock bound under `synchronous=FULL`. Neither exists as infrastructure anywhere in this codebase (confirmed by grep for `sqlite3_trace`/`sqlite3_trace_v2`), and a wall-clock assertion in a shared CI runner risks exactly the flakiness this repo's own testing conventions warn against. The transaction-wrap itself is not independently pinned by a test; the marker-skip and pre-upgrade-backup behaviors are, and both would regress visibly if the transaction wrap were removed carelessly alongside them.

## Docs, same commit

`docs/engine/db-schema.md`: a new `.pre-upgrade.bak` note beside the existing `.bak` semantics, and a new "The header-strip pass (issue #1487)" subsection describing the LIKE filter, the transaction, and the one-time marker.

## No user-visible change

Engine-only: `engine/src/db/database.cpp`, its tests, and `docs/engine/db-schema.md`. No `app/` files touched, so `pnpm test`/`pnpm type-check` were not run (nothing in the diff could affect them).

## Checklist
- [x] My code follows the style guidelines (clang-format 19 clean, clang-tidy clean on the changed engine files)
- [x] I have added tests

## Related Issues
Closes #1487

### DIFF
--- a/docs/engine/db-schema.md
+++ b/docs/engine/db-schema.md
@@ -41,6 +41,12 @@ up looking like a fresh install.
 > `POST /workspace/backup` / `vayu-cli backup`, described in
 > [architecture.md](architecture.md#workspace-backups).
 
+> **`<db>.pre-upgrade.bak` is different: it is written once.** A one-time
+> repair pass (below) takes it immediately before its first rewrite, and
+> nothing after that touches it again - unlike `<db>.bak`, which the *next*
+> clean start would refresh from a file the repair has already rewritten. It
+> is the one place the pre-repair rows survive more than a single restart.
+
 ### The quarantine (issue #984)
 
 A database that fails validation is **moved, not deleted**: `<db>` becomes
@@ -141,6 +147,29 @@ rewrite slow enough to matter is one this line explains.
 Best-effort like the sweeps before it: a failure is a warning, never a daemon that will not start.
 `auto_vacuum` was rejected as the alternative - switching an existing database's mode requires a
 full `VACUUM` anyway, and it changes page bookkeeping for every write thereafter.
+
+### The header-strip pass (issue #1487)
+
+`Database::strip_stored_managed_headers()`, called from `init()`, removes the `X-Vayu-Version`,
+`X-Request-ID` and `User-Agent` rows a pre-#1229 renderer wrote into a saved request's `headers`
+column on the app's own behalf - default headers are applied per transfer now and never stored.
+Like the passes above it runs before `/health` can answer, but it is scoped so a workspace with
+nothing left to strip pays close to nothing for it:
+
+- **A SQL `LIKE` over `headers` first.** Only rows whose header text could contain one of the
+  three names are loaded into a `Request` and JSON-parsed; `LIKE` is ASCII case-insensitive by
+  default, the same fold the exact predicate applies to a matched key. A workspace already
+  stripped costs one scan of the column, not a materialisation of every request's body and
+  script.
+- **One transaction for the rewrite**, not one implicit commit per candidate row - the same shape
+  `seed_default_config` uses.
+- **A one-time marker.** Once a pass finds nothing left to strip, it records a
+  `managedHeadersStripped` config entry (`advanced`, no everyday user story) and every later start
+  skips the scan outright rather than re-running it to confirm nothing changed. #1492 will give
+  this kind of migration bookkeeping a proper home; until then a config entry is where this
+  file's other internal-only flags already live.
+- **`<db>.pre-upgrade.bak`** is written once, immediately before the first row this pass ever
+  rewrites - see the backup note above.
 
 ---
 
--- a/docs/engine/db-schema.md
+++ b/docs/engine/db-schema.md
@@ -324,7 +324,9 @@ Stores individual HTTP request definitions.
 [{"key":"Content-Type","value":"application/json","enabled":true,"description":""}]
 ```
 Disabled rows (`"enabled":false`) are preserved in storage and filtered at HTTP-execution time only.
-Duplicate keys are allowed.
+Duplicate keys are allowed. A pre-#1229 renderer also wrote its own `X-Vayu-Version`,
+`X-Request-ID` and `User-Agent` rows here; [the header-strip pass](#the-header-strip-pass-issue-1487)
+removes them once, at startup.
 
 **body** - discriminated union:
 ```json
--- a/engine/src/db/database.cpp
+++ b/engine/src/db/database.cpp
@@ -2028,10 +2028,81 @@ int64_t Database::stamp_hashless_spec_bindings () {
     return stamped;
 }
 
+namespace {
+// A repair that runs once is a migration and gets a migration's bookkeeping
+// (issue #1487). #1492 will give this kind of marker a proper home; until
+// then a `config_entries` row - `advanced`, no everyday user story - is where
+// this file's other internal-only flags already live.
+constexpr std::string_view MANAGED_HEADERS_STRIPPED_KEY =
+"managedHeadersStripped";
+} // namespace
+
 int64_t Database::strip_stored_managed_headers () {
     std::lock_guard<std::recursive_mutex> lock (impl_->mutex);
+
+    if (get_config_bool (std::string (MANAGED_HEADERS_STRIPPED_KEY), false)) {
+        return 0;
+    }
+
+    // Only rows whose `headers` text could contain one of the three legacy
+    // names are worth loading and JSON-parsing. `LIKE` is ASCII
+    // case-insensitive by default, the same fold `legacy_managed_row` applies
+    // to a matched key, so a workspace with nothing left to strip costs one
+    // scan of this column rather than a materialisation of every request's
+    // body and script (issue #1487).
+    auto candidates = impl_->storage.get_all<Request> (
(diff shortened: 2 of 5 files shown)


athrvk:
The "CI gate" failure reported on `1f36a19` is not a real failure: its job log shows `app: cancelled` and `engine: cancelled`, because I pushed a follow-up commit (`fbaeaa1`) while that run was still in progress, which GitHub Actions cancelled in favor of the new run. The current head (`fbaeaa1`) has its own fresh CI run in progress; watching that one through to green.

---

Issue #1517, opened 2026-09-06
mcp(elements): `elements` on the request and collection tools, run overrides on the run tools, the `vayu://elements/kinds` resource, and the registry conformance test


athrvk:
Part of #1512. After #1513; independent of the app child.

## Scope

Anchors at master `d1fe93d1`, `app/electron/mcp/tools.ts`.

1. **Request and collection tools.** `create_request` (`:5132-5133`), `update_request` (`:5203-5257`, merge-patch list `:5222-5226`), `create_collection` and `update_collection` (`:4408-4409`, `:4457-4484`) take `elements` (the whole list; merge-patch semantics stay at the field level, the engine validates the contents and the tool relays the engine's 400 verbatim). `preRequestScript` and `postRequestScript` stay as **sugar** on these tools only: the tool translates them into `script.pre` / `script.post` elements before calling the engine, so an agent that only knows scripts keeps working; the tool descriptions say so and point at the resource. `readValidationScript` (`:1593-1600`) retires.
2. **Run tools.** `run_request` (`:4230-4257`) and `start_load_run` (`:6657-6690`; `composeLoadRunRequest` `:1630-1690` deletes `preRequestScripts` at `:1671-1674` and replaces `tests` at `:1680-1686`) take ad-hoc `elements` instead of `tests`; `run_collection` and the scenario branch of `start_load_run` (`startScenarioLoadRun`, `:2826-2905`; refusals `:2766-2767`) take the override block (`timers`, `scripts`). The refusal rows are reworded for the element vocabulary.
3. **Resource.** `vayu://elements/kinds` serves the engine catalogue (`GET /elements/kinds`), the `vayu://scripting/completions` pattern (`scripting.cpp:29`). Tool descriptions for `elements` say to read it rather than guess kinds.
4. **Schema.** The zod schema for an element is `{ id?, kind: string, enabled?, name?, config: record }`; the kind list is **not** enumerated in zod, because the engine is the authority and a new kind must not require a tool change. The conformance test below is what keeps the two honest.
5. **Conformance.** `element-kinds.conformance.test.ts` reads `engine/tests/fixtures/element-kinds.json` and asserts the resource handler returns the same kinds and that every kind named in a tool description exists; mutation check: rename a kind in the fixture, reds.
6. **Data-changed.** `mcp:data-changed` for request and collection writes already invalidates the request and collection queries (#1438); no new event family.

## Tests

[...]

Issue #1516, opened 2026-09-06
app(elements): the ElementList primitive with schema-driven forms, an Elements tab on requests and collections replacing the two script tabs, the inheritance view, and run-dialog overrides


athrvk:
Part of #1512. Starts after #1513 (the wire shape), completes after #1514 (outcomes to show).

## Scope

Anchors at master `d1fe93d1`.

1. **`components/shared/ElementList/`.** An ordered list of rows: kind badge, optional name, origin chip for inherited rows, enable switch, drag handle, delete, and the kind's form below the row. Built from the conventions `KeyValueEditor` (`components/shared/KeyValueEditor/index.tsx:53-67`) and the settings row primitives (`NumberSettingRow`, `SelectSettingRow`, `ToggleRow`, reused by `SettingsPanel.tsx:56-60`) already set. **The default form is generated from the kind's JSON Schema** served by `GET /elements/kinds` (#1513): string, integer, number, boolean, enum and string-array properties map to the row primitives, the way `SettingsMain.tsx:471-660` renders config entries by type. A bespoke form is an optional override in one map keyed by kind (`elementForms.ts`); phase 0 ships bespoke forms only for `script.*` (the `ScriptPanel` editor body, `ScriptPanel.tsx:293-303`) and `extract.json` (a path field with the response-body picker). An unknown kind renders the generic form.
2. **Add menu.** Grouped by the catalogue's `category`, labels and descriptions from the catalogue, never a hand-written list.
3. **Request builder.** One **Elements** tab replaces Pre-request and Tests (`RequestTabs/index.tsx:43-104`, badge rule `:77-85`); badge = enabled element count. `RequestState.preRequestScript` and `testScript` (`request-builder/types.ts:168`) become `elements`; `script-variants.tsx:67-106` folds into the kind-form map; `buildUpdatePayload` (`index.tsx:101-159`) sends `elements` when changed; the transformer (`request-transformer.ts:90-140`), `useTreeCrud.ts:380-398` (duplicate copies `elements`), `types/domain.ts:591-640`.
4. **Collection detail.** One Elements tab replaces Pre-request and Post-request (`CollectionDetail/index.tsx:30-40`), drafted through `useEntityDraft` keyed `${id}:elements` with the `ExternalChangeCallout` #1437 landed.
[...]

Issue #1514, opened 2026-09-06
engine(elements): the pipeline in the design send and the sequential collection run, the phase-0 kinds (extract.*, assert.*, timer.think, script.*), and per-element outcomes in the step trace


athrvk:
Part of #1512. Second; this is the collection-run parity bar.

## Scope

Anchors at master `d1fe93d1`. Item 0 was added at master `cc8b01c6` after PR #1528 landed #1513 additively; its anchors are at `cc8b01c6`.

0. **Cut-over inherited from #1513.** PR #1528 kept `pre_request_script` / `post_request_script` mapped (`engine/src/db/database.cpp:216-217, 254-255`) and backfilled `elements` beside them with a startup repair pass, `Database::fold_scripts_into_elements` (`:2191-2263`, marker `scriptsFoldedIntoElements`, snapshot `<db>.pre-elements-fold.bak` at `:2240-2244`), because nothing ran a `script.*` element before this issue's pipeline exists: dropping the columns there turned about twenty `ScenarioRunnerTest` cases red. Once this pipeline runs `script.pre` / `script.post`, the two columns are dead data, so this issue lands the cut-over #1513's body described, as a clean cut. The owner's decision: no release ships until the whole of #1512 is built, so master is allowed to break the app's script editor between this issue and #1516; no transitional alias.
   - `migrate_before_sync (path)` in the `Database` constructor ahead of the probe's `sync_schema ()` (`:1043-1062` region): open with raw sqlite3, read `PRAGMA user_version`; newer than the engine's `SCHEMA_VERSION` refuses to start naming both; 0 copies the file to `<db>.pre-migration.bak` (kept until the next successful start, the #1487 rule), then in one transaction folds any row whose script column is non-blank and whose `elements` lacks the matching `script.pre` / `script.post` entry (the json1 SQL from #1513's body; a row the repair pass already folded is left alone), then sets `user_version = 1`. Only then does `sync_schema ()` see the new mapping and `DROP COLUMN` the two script columns (`sqlite_orm.h:23181-23193`). This is where #1492's schema version lands. A released 0.26 database never ran the fold, so the unfolded case is the one every upgrading user hits.
   - The repair pass, its marker and `.pre-elements-fold.bak` retire: the version is the marker.
[...]

Issue #1512, opened 2026-09-06
engine+app: per-step element model - extractors, assertions, timers, controllers, scripts and metrics as pluggable elements on requests and collections, run by one pipeline in every execution path (parent)


athrvk:
## Why

The owner's product decision: the collection run is to be at par with a JMeter test plan, and the parity is to be delivered the way JMeter delivers it - as **elements** a user attaches to a request or a folder (extract this field, assert that, wait here, loop these, record this), added and removed independently, inherited down the tree, and run by the engine at fixed phases of a step. Not as a feature per capability. Backward compatibility of the database, the engine API and the app is explicitly not required.

Analysed at master `d1fe93d1` (element-model sweep, static, every anchor below verified). What Vayu has today, per JMeter element, in the sequential collection run:

| JMeter | Collection run today | Gap |
|---|---|---|
| Threads, ramp-up, loop count | one user; `scenario.iterations` (`scenario_plan.cpp:190-211`); load runs have concurrency and ramp | no per-user loop count, no action-after-error |
| Loop / If / While / Switch / Once Only | script-only, through `pm.execution.setNextRequest` and `skipRequest` (`scenario_runner.cpp:660-723`) | no declarative form; throws under load |
| Transaction controller | none; per-step rows only | missing |
| Throughput / Random Order | none | missing |
| CSV Data Set | `scenario.data`, row `i % rows`, `{{data.col}}`, `pm.iterationData` | none |
| Cookie manager | environment jar (`scenario_runner.cpp:918-928`); per-user cookies under load | none |
| JSON / Regex / Boundary extractor | script only (`pm.response.json()` + `set()`) | no declarative form; the load path never re-resolves what a script set (#1495) |
| JSR223 pre / post | `pre_request_script` / `post_request_script`, inline through `execute_exchange` (`request_exchange.cpp:545-599`) | under load: post deferred to a replay, pre never |
| Response / JSON / Duration / Size assertion | `pm.test` only; never touches a verdict | no declarative form (#1497) |
| Timers | none in any mode (#1498) | missing |
| setUp / tearDown | none (#1499) | missing |

Two of those columns are behaviours attached to a request at a phase: the scripts. Everything else on a request is sampler configuration (method, URL, params, headers, body, auth, execution settings). The element model generalises the two script columns into an ordered list of typed behaviours and gives every kind the same storage, validation, inheritance, execution, reporting and editing.

## The extensibility contract (the owner's constraint, and what the tests pin)

[...]

PR #1528, opened 2026-09-06
feat(engine): element registry, storage, catalogue and an additive script-to-elements fold

### LINKED ISSUE #1513: engine(elements): the `elements` column, the kind registry, `GET /elements/kinds`, route validation against the registry, `PRAGMA user_version`, and the one-shot migration of the two script columns
Part of #1512. Lands first; nothing executes an element yet.

## Scope

The storage, the registry and the wire shape of the element model, plus the migration that folds the two script columns into it. Anchors at master `d1fe93d1`.

1. **Registry.** `engine/include/vayu/core/elements.hpp` and `engine/src/core/elements/registry.cpp`: `struct ElementKind { std::string kind; int version; std::vector<Phase> phases; nlohmann::json config_schema; std::string label, description, category; HotPathClass hot_path; std::function<std::unique_ptr<Element>(const nlohmann::json&)> compile; }`; `class Element { virtual Phase phase(); virtual void apply(ElementContext&); }`. `Registry::instance().register_kind(...)` from a static initialiser in each kind's own file (`core/elements/<kind>.cpp`), the one registration point. `Registry::validate(const nlohmann::json& elements) -> std::optional<std::string>` runs the shape check (array of objects with `id`, `kind`, `enabled`, optional `name`, `config`), the kind lookup, and valijson against `config_schema` (`schema_validation.cpp:432-451` is the in-tree use), returning messages of the form `elements[2] (kind 'assert.sttus') is not a known element kind - expected one of …` and `elements[2] (assert.status): 'in' must be an array of status codes`. Phase 0 registers no behaviour kinds; it registers `inherit.disable` (validated, no `apply`) and, in the test binary only, `test.echo`.
2. **Storage.** `elements TEXT NOT NULL DEFAULT '[]'` on `requests` and `collections` (`database.cpp:236-277`, `:210-235`; `types.hpp:1044-1113`, `:1009-1042`). `pre_request_script` and `post_request_script` leave the mapping on both tables. Engine assigns `el_` ids when absent (`utils/id.hpp`), accepts well-formed client ids, refuses duplicates within one list.
3. **Routes.** `apply_elements_field` beside `apply_json_field` (`routes.hpp:295`) used by `apply_request_fields` (`requests.cpp:180-267`) and `apply_collection_fields` (`collections.cpp:213-297`), hence also by `POST /import/apply` (`import.cpp:648,776`). Both serializers gain the field: `serialize(db::Request)` (`json.cpp:440`) and `serialize_to_stream` (`:1200-1244`, through `write_json_column`). `preRequestScript`, `postRequestScript` and `tests` are refused with a 400 naming `elements` on every route that took them (`read_pre_request_script` / `read_post_request_script`, `script_parts.cpp:74-78,495-501`, retire). `POST /compose` emits the resolved `elements` list (`payload_from_stored`, `request_composer.cpp:939-991`) in place of the joined script strings; `compose_elements(chain, request)` replaces `compose_script_parts` (`:876-886`): root to leaf collection elements, then the request's own, minus disabled and `inherit.disable` targets, each stamped `origin: { kind: collection | request, id, name }`.
4. **Catalogue.** `GET /elements/kinds` returns the registry: for each kind its name, version, phases, schema, label, description, category and hot-path class. This is what #1516 (app), #1517 (MCP) and `docs/engine/elements.md` read. The engine test suite writes the same JSON to `engine/tests/fixtures/element-kinds.json` on every run so the vitest conformance test has it without a running engine.
5. **Schema version and migration.** `Database::Database` gains `migrate_before_sync(path)` called before the probe's `sync_schema()` (`database.cpp:1046`): open the file with raw sqlite3, read `PRAGMA user_version`; if it is newer than the engine's `SCHEMA_VERSION`, refuse to start with a message naming both; if 0, copy the file to `<db>.pre-migration.bak` (kept until the next successful start, the #1487 rule), then in one transaction run the json1 SQL that builds `elements` from the two script columns (`json_array(json_object('id', 'el_' || lower(hex(randomblob(8))), 'kind', 'script.pre', 'enabled', 1, 'config', json_object('script', pre_request_script)))`, appending the post element the same way, for rows where either column is non-blank), then `PRAGMA user_version = 1`. Only then does `sync_schema()` see the new mapping and `DROP COLUMN` the two script columns (`sqlite_orm.h:23181-23193`, SQLite 3.35 and later; bundled 3.53). No repair pass, no `config_entries` marker: the version is the marker. `db-schema.md:10-14` is rewritten: destructive changes are versioned migrations now.

## Tests

- `engine/tests/elements_registry_test.cpp`: `test.echo` appears in the catalogue with its schema; validation refuses an unknown kind, a bad field and a duplicate id with the messages above; mutation check: skip the schema step, the bad-field case reds.
- Route tests: `PUT /requests/:id` with `elements` round-trips; with `preRequestScript` returns 400 naming `elements`; both serializers carry the field (extend the "standing trap" test, `json.cpp:275-283`).
- Migration test on the temp-database fixture: a 0.26-shaped file with scripts opens, `elements` holds `script.pre`/`script.post` with the exact text, both columns are gone, `user_version` is 1, `.pre-migration.bak` exists and still has the columns; a second open does nothing; a file with `user_version` 99 is refused. Mutation check: run the migration after `sync_schema`, the exact-text case reds (the columns are gone before it reads them).
- `python build.py -t && cd engine && ctest --preset linux-dev -R "Elements|Migration|Requests|Collections|Compose"`.

## Docs, same commit

`docs/engine/db-schema.md` (the column, the version, the migration section), `docs/engine/api-reference.md` (request and collection fields; `POST /compose` output; the refused keys; `GET /elements/kinds`), a new `docs/engine/elements.md` skeleton with the shape, the phases and the empty kind table, added to the nav per the `docs-site` skill.

## Acceptance criteria

- `GET /elements/kinds` lists `inherit.disable` (and `test.echo` in the test build) with schemas.
- `PUT /requests/:id` with an unknown kind returns the 400 above; with a valid list stores and lists it through both serializers.
- A 0.26 database migrates once as described; the app and MCP do not need to change to keep working on requests without elements (they do change in #1516 and #1517 to edit them).

---

athrvk:
## Description

Lands the storage, the registry and the wire shape of the element model (issue #1512's per-step behaviour system), the first of that epic's children.

**Plan, root cause and the deviation from the issue's literal text:** #1513 as written asks for `pre_request_script` / `post_request_script` to be *dropped* from the schema, gated by a new `PRAGMA user_version` migration. Implementing that literally breaks the engine today: those two columns still drive every real script execution path (design send, sequential run), and the pipeline that would run a `script.*` *element* instead doesn't exist until #1514. Verified empirically - a first pass that dropped the columns turned ~20 currently-passing `ScenarioRunnerTest` cases red, because a migrated request's post-request assertion script silently stopped running with no replacement.

The alternative implemented here: keep both script columns mapped and add `elements` **additively**. `Database::fold_scripts_into_elements()` (a startup repair pass, modeled directly on the existing `strip_stored_managed_headers` precedent - candidate scan, one transaction, `<db>.pre-elements-fold.bak` snapshot, a `scriptsFoldedIntoElements` config-entry marker) backfills a `script.pre` / `script.post` entry per non-blank script, without touching the columns. Nothing in the app, MCP, or any existing route breaks: `preRequestScript` / `postRequestScript` keep working exactly as before, `elements` sits beside them. The real column drop and `PRAGMA user_version` gate are deferred to #1514, where the execution pipeline that replaces the two columns actually exists. `engine/CLAUDE.md`'s two #1513-referencing lines are corrected to describe this.

**What's in scope and landed:**
- **Registry** (`engine/include/vayu/core/elements.hpp`, `engine/src/core/elements/registry.cpp`): `ElementKind`, `Element`, `Phase`, `HotPathClass`, and `Registry::instance()` - a function-local-static singleton with an explicit call list (not a namespace-scope self-registering static initialiser, which `engine/CLAUDE.md`'s "nothing at namespace scope is built at run time" rule declines; the explicit list keeps the same "one new file, one new line" cost on the same "explicit list, never a glob" precedent `engine/tests/CMakeLists.txt` already uses). `Registry::validate` runs the shape check, the kind lookup and valijson against each kind's `config_schema`.
- **Storage**: `elements TEXT NOT NULL DEFAULT '[]'` on `requests` and `collections`. The engine assigns an `el_` id when a client omits one, rather than refusing the write.
- **Kinds registered**: `inherit.disable` (validated, never run - consumed directly by `POST /compose`'s chain resolution) and `script.pre` / `script.post` (validate-only, no `compile` - registered so the fold's own output round-trips through the same validator every write route runs, not because anything executes them). `test.echo` is registered in the test binary only.
- **Routes**: `apply_elements_field` beside `apply_json_field`, wired into `apply_request_fields` / `apply_collection_fields` (hence also `POST /import/apply`). Both serializers (`serialize`, `serialize_to_stream`) gain the field.
- **Catalogue**: `GET /elements/kinds`. The test suite writes it to `engine/tests/fixtures/element-kinds.json` on every run (a registered `::testing::Environment`, not a namespace-scope initializer, called explicitly from `main()` - `cert-err58-cpp` again), for #1516/#1517's future conformance tests.
- **Compose**: `compose_elements(chain, request)` resolves the collection-chain-then-request element list, minus anything disabled or named by an `inherit.disable` entry, each stamped with its origin. Added to `POST /compose`'s output *alongside* the existing `preRequestScripts` / `postRequestScripts` lists (additive, matching the rest of this PR).
- **Docs**: `docs/engine/db-schema.md` (both tables, the new fold section), `docs/engine/api-reference.md` (the PUT field rule, `POST /compose` output, `GET /elements/kinds`), a new `docs/engine/elements.md` (nav'd under Engine) with the shape, phases, registry contract and kind table.

## Type of Change
- [x] New feature

## Testing

`engine/tests/elements_registry_test.cpp` (new): catalogue contents (all four registered kinds with schemas), `Registry::validate` rejecting an unknown kind / a bad config field / a duplicate id and accepting a well-formed list, `PUT` round-tripping `elements` through both serializers, `PUT` with an unknown kind returning 400 naming it, an element with no id getting one assigned, legacy script fields still saving beside `elements`, and `fold_scripts_into_elements` (folds once, keeps scripts untouched, idempotent, marker-skips a later start).

- `python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`: **2980/2980 passed** (2969 pre-existing + 11 new).
- `pip install -r requirements-docs.txt && mkdocs build -f .github/mkdocs.yml --strict`: clean.
- Two independent read-only reviews (acceptance-criteria-by-line, repo-conventions) ran against the full diff; both rounds of findings were fixed in this branch (`script.pre`/`script.post` registration, id-assignment, the pre-fold backup, the stale `CLAUDE.md` lines, the missing fixture file). Two low-severity nitpicks were left as-is with reasons: no HTTP-level route test for `GET /elements/kinds` (its handler is a two-line wrapper over the directly-tested `elements_catalogue()`); `compose_elements`'s `inherit.disable` resolution treats disable targets as global across the whole chain rather than per-level, which is harmless today given random `el_` ids and is noted in a code comment for #1516/#1517 to revisit once ids are hand-editable.

## CI fix: Windows ASan leg

This PR touches `engine/CMakeLists.txt`, so it runs `sanitizers.yml`'s five
sanitizer legs on top of the usual gate - the only legs that build the engine
at `/Od` on MSVC. `first_request` (added to `engine/tests/import_parse_test.cpp`
by #1441, on master since before this branch) descended into a collection's
children with a range-for whose body always returns on the first iteration,
so its increment is never reached. MSVC reports that at `/Od` as C4702
(unreachable code), fatal under `/WX`, which is unconditional on that
compiler - `windows-prod` never sees it because `/O2` folds the loop away,
and no other Windows leg builds Debug. Same defect class, and the same fix,
as #1037's dead `return` in `read_monitor_block`: replace the unreachable
construct with an equivalent that has none, rather than suppress the
warning.

Reproduced against the failing job's own `cl.exe` command line for that one
translation unit: C2220/C4702 at the same line before the fix, clean after.
The six test files ninja never reached before the original failure
(`elements_registry_test.cpp` among them) and every one of this PR's own
modified or new engine sources (`elements.cpp`, `collections.cpp`,
`requests.cpp`, `request_composer.cpp`, `server.cpp`, `database.cpp`,
`json.cpp`) had already compiled clean in that same failing run, before
ninja stopped - so this was the only Windows-only warning in the diff.

\#1037 left open whether `sanitizers.yml`'s Windows leg should pass MSVC
`/WX-` so a warning-as-errors gate that fires only on a `pull_request` diff
touching `engine/CMakeLists.txt` (rather than every push) stops being a
second, weaker copy of the real gate. Still unresolved, still an owner call,
still not folded into this fix - noted here because it recurred.

## Checklist

## Related Issues
Closes #1513

## Visual changes

No user-visible change - engine-only (registry, storage, routes, compose, a startup repair pass). The app and MCP are untouched in this PR; #1516 and #1517 give them an editor for `elements`.


athrvk:
`asan on windows-latest` is failing on `engine/tests/import_parse_test.cpp:102` - MSVC's `/WX` promotes a `C4702: unreachable code` warning inside `first_request()` to an error under `/fsanitize=address`:

```
D:\a\vayu\vayu\engine\tests\import_parse_test.cpp(102) : error C2220: the following warning is treated as an error
D:\a\vayu\vayu\engine\tests\import_parse_test.cpp(102) : warning C4702: unreachable code
```

This is not this PR's: `import_parse_test.cpp` is untouched by this diff (it's an unrelated import-conformance test helper), and `first_request()` already carries a `VAYU_IGNORE_FALSE_NULL_DEREFERENCE` annotation for a different, previously-diagnosed false positive from a different compiler in this exact function - suggesting this file has hit MSVC static-analysis false positives before. The last scheduled `Sanitizers` run on `master` (2026-08-31, `c654ec1`) was fully green, so this reads as a toolchain drift on the `windows-latest` ASan image (the job log shows `VCToolsVersion 14.51.36231` / Visual Studio 18) rather than something in this branch's diff, but I have no Windows/MSVC toolchain in this container to reproduce or verify a fix against.

Proposed patch, for whoever picks this up with Windows access to verify: restructure the tail of `first_request()` so the compiler cannot conclude the `for` loop always returns - e.g. hoist the "does `children` have at least one entry" check the same way the `requests` branch above it does, rather than relying on the loop body's single iteration to always `return`. I'm leaving this to a human/Windows-capable session rather than guessing blind at an MSVC-specific fix I can't compile-check.

I don't have the means to re-run this job from here. Keeping this PR's other checks watched.

---

PR #1534, opened 2026-09-06
fix(engine): report the configured worker count from /health, not the core count

### LINKED ISSUE #1508: engine(health): `/health` reports `workers` as the machine's core count, not the configured value the runs actually use - and the app declares the field and reads it nowhere
## Problem

Owner-measured on an installed 0.26.0 with `workers` set to 8 in the engine config: `/health` answered `"workers": 12`, and the run used 8 (thread count went from 14 to 24 for the run's duration).

Verified at master `800426d8`:

- `engine/src/http/routes/health.cpp:38` `response["workers"] = std::thread::hardware_concurrency ();` - unconditionally the core count.
- `engine/src/core/run_manager.cpp:1474` `db.get_config_int ("workers", 0)` is what a run actually uses, with 0 meaning auto-detect. `docs/engine/api-reference.md:612-620` describes exactly this: "the engine reads it at the start of every run".
- `app/src/types/domain.ts:2291` declares `workers: number` on `EngineHealth`; `rg -n "\.workers" app/src --glob '!*.test.*'` finds no reader. Written by the engine, typed by the app, displayed nowhere - the written-but-never-read shape.

Severity: low. The one place the number appears is the raw health JSON, and there it is wrong whenever the user has set the knob.

## Fix pathway

1. `build_health_response` reports the resolved value: the configured `workers` when non-zero, else `hardware_concurrency()`, through the same helper the run manager should use so the two cannot drift (extract `resolve_worker_count (db)` beside `:1474` and call it from both).
2. Give the field a reader or drop it. Recommended: the Dock's engine tooltip (the one that shows the version) adds "N workers", read from the health query the Dock already observes. If that is judged noise, remove `workers` from `EngineHealth` and from the health response, with the docs example updated; a field nobody renders is a field that drifts.
3. `docs/engine/api-reference.md:480-500` health example: the value is the configured worker count, auto-detected when unset.

## Tests

Engine: a health route test with `workers` set to 3 in the temp database asserts `"workers": 3`; with it unset, asserts the hardware count; mutation check: restore the unconditional read, the first reds. App: if rendered, a Dock test asserts the tooltip text. `ctest -R Health`, `pnpm test Dock`.

## Sequencing

Independent.

## Acceptance criteria

- With `workers` configured to N, `/health` reports N; unset reports the auto-detected count, and both equal what the next run's thread count shows.
- The field is either shown in the app or absent from the type and the wire.

---

athrvk:
## Description

`/health` reported `std::thread::hardware_concurrency()` unconditionally, ignoring the `workers` config setting a run actually uses. Owner-measured on an installed 0.26.0 with `workers` set to 8: `/health` answered `"workers": 12` (the machine's core count), while the run used 8. The app's `EngineHealth.workers` field was declared and typed but read nowhere - a "written but never read" case.

**Root cause and plan:** `build_health_response` (`engine/src/http/routes/health.cpp`) read the core count directly instead of the resolved `workers` setting `execute_load_test` (`engine/src/core/run_manager.cpp`) already reads via `db.get_config_int("workers", 0)`. Fixed by extracting `vayu::core::resolve_worker_count(db)` (new `engine/{include,src}/core/worker_count.{hpp,cpp}`) - the configured value when non-zero, else the detected core count - and calling it from both `/health` and `execute_load_test`, so the two numbers cannot drift apart again. `execute_load_test`'s debug log line, which used to print `"auto"` for an unconfigured value, now prints the resolved number instead, which is strictly more informative and needed no test change (nothing asserted the log text).

For the app side (acceptance criterion: "the field is either shown in the app or absent from the type and the wire"), the field already had no reader anywhere. Rather than remove it from the wire - the acceptance criteria also require `/health` to report the *correct* value, which only makes sense if the field stays - I gave it a reader: `engine-store` now carries the last-polled `workers` value (mirroring how `recovery` is already handled), and the Dock's version string (`v0.26.0`) is a focusable tooltip showing the worker-thread count once the first health poll answers, following the exact pattern the engine-status indicator's own error tooltip already uses.

**Also fixed in this PR:** the `jsx-a11y/no-noninteractive-tabindex` suppression ceiling in `docs/design-system.md` and `a11y-suppressions.test.ts` (16 → 17) - a direct, mechanical consequence of the new tooltip trigger following the documented process for adding one, not separate scope.

## Type of Change
- [x] Bug fix

## Testing

- `engine/tests/health_route_test.cpp` (new): `GET /health`'s `workers` matches a configured value (`apply_config_update`, `{"workers":"3"}`) and falls back to the detected core count when unset. Mutation check performed by hand: reverting `build_health_response` to the unconditional `hardware_concurrency()` read reds `WithAConfiguredValueHealthReportsIt` (confirmed, then reverted).
- `app/src/queries/health.test.ts`: a new case asserts a successful poll writes `workers` into `engine-store`.
- `app/src/components/layout/Dock.worker-count.test.tsx` (new): no tooltip affordance before the first poll answers; the count renders on hover, with correct singular/plural wording.

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`: **2976/2976 passed** (2974 pre-existing + 2 new).
`cd app && pnpm test`: **8087/8087 passed** (8084 pre-existing + 3 new). `pnpm type-check`, `pnpm lint`, `pnpm format:check`: all clean.
`mkdocs build -f .github/mkdocs.yml --strict`: clean.

## Checklist

## Related Issues
Closes #1508

## Visual changes

Engine-only fix plus one small, additive app affordance (a tooltip on existing text, no layout change). Not captured with a screenshot: the change is a hover tooltip on the Dock's existing version string, verified instead through `Dock.worker-count.test.tsx`'s rendered-and-focused assertions, which is the more precise evidence for a focus/hover affordance than a static image would be.

Docs updated in this commit: `docs/engine/api-reference.md` (`/health`'s `workers` semantics), `docs/app/state-management.md` (`engine-store`'s new `workers` field), `docs/app/COMPONENTS.md` (the Dock's version-tooltip bullet), `docs/design-system.md` (the a11y suppression list/ceiling).


athrvk:
`asan on windows-latest` failed on the first push, but not from this PR's diff: `engine/tests/import_parse_test.cpp:102`'s `first_request` used a range-for that always returns on its first iteration, so MSVC's `/Od` build (the Windows ASan leg is the only one that builds at `/Od`; `/O2` folds the loop away elsewhere) reports the unreached increment as C4702, fatal under the unconditional `/WX`. This is the same pre-existing defect already fixed in open PR #1528 ("CI fix: Windows ASan leg"), not yet merged to `master`.

Ported that exact fix into this branch (commit 1e1d4eb) rather than wait on #1528 to merge - it replaces the range-for with an emptiness-checked `children.at (0)` descent, which MSVC does not flag. Verified locally: the `ImportParse` and `ImportParseRoute` suites that exercise `first_request` (28 cases) pass, and `clang-format`/`clang-tidy` are clean on the changed file.

---

Issue #1229, opened 2026-09-01
app, engine(http): the default headers Vayu adds are stored into the request, differ by send path, cannot be removed, repurpose a public header name, and no compression is negotiated - make the engine own one visible, optional, namespaced set


athrvk:
Owner-requested review of the headers Vayu sends by default and the User-Agent. Verified at master `5677ed5`.

## What goes out today

| Header | Who adds it | Design send (app) | Load, scenario, collection runner, MCP send |
|---|---|---|---|
| `User-Agent: Vayu/<version>` | engine, only when the request has none (`engine/src/http/event_loop/curl_utils.cpp:369-370`; constant `engine/src/core/user_agent.cpp:21`) | yes | yes |
| `X-Vayu-Version: <version>` | renderer, at send (`app/src/modules/request-builder/index.tsx:210-212`) | yes | no |
| `X-Request-ID: <uuid>` | renderer, fresh per send (same lines) | yes | no, unless a stale one is stored (below) |
| `Accept: */*` | libcurl default | yes | yes |
| `Accept-Encoding` | nobody - `CURLOPT_ACCEPT_ENCODING` is set nowhere in `engine/src` (verified) | absent | absent |
| body-implied `Content-Type` | engine | yes | yes |

The three managed rows sit at the top of the Headers tab, cannot be disabled or removed, the version row cannot be edited, the other two can (`app/src/modules/request-builder/utils/system-headers.ts:143-190`, enforced through `useHeadersManager.ts:82-90`).

## Problem

1. **The managed rows are persisted into the stored request.** Save writes `toKeyValueEntries(request.headers)` with no filter (`index.tsx:509`); that helper strips the `system` flag and keeps the row (`components/shared/KeyValueEditor/key-value.ts:88-98`). Every saved request therefore carries a frozen `X-Request-ID` UUID and the version strings of the day it was saved. The history view's own write path knows this is wrong and filters them (`modules/history/main/save-run-to-request.ts:141-148,396-399`); the builder does not. Consequences: a load test, a scenario, a collection run or an MCP `send_request` takes the stored headers and sends the **same** `X-Request-ID` on every request of the run, plus a `User-Agent` and `X-Vayu-Version` pinned to an old release until the request is re-saved from the builder. Exports and generated code carry them too (no consumer of `SYSTEM_HEADER_KEYS` exists outside the builder and the history view, verified).
[...]

PR #1538, opened 2026-09-07
fix(engine,app): disable rather than delete a legacy-shaped header row, and match its exact value shape

### LINKED ISSUE #1491: engine(db): the header-strip predicate deletes a user's own bare-UUID X-Request-ID and any header named x-vayu-version regardless of value, trash rows included - it matches on shape and name, not on provenance

athrvk:
## Description

Owner-measured on the upgrade sweep (issue #1491): `Database::strip_stored_managed_headers()`'s predicate, `legacy_managed_row` (`engine/src/http/default_headers.cpp`), matched `X-Vayu-Version` on **name alone** regardless of value, and `X-Request-ID` on **any bare UUID** regardless of case or version - not on the exact shape the pre-#1229 renderer actually wrote. Worse, the pass **deleted** a matching row outright, and it rewrote rows already sitting in the trash, so a false match (a hand-typed `x-vayu-version: hand-written`, or a user's own lowercase-v4-shaped `X-Request-ID`) was silently and unrecoverably lost, trash included.

**Root cause and plan:** tighten `legacy_managed_row` to the renderer's *exact* historical value shape - a plain `MAJOR.MINOR.PATCH` for `X-Vayu-Version`/the `User-Agent` suffix (new `is_plain_semver`), and a lowercase RFC 4122 v4 UUID for `X-Request-ID` (`is_bare_uuid` now checks the version nibble `4` and variant nibble `8`/`9`/`a`/`b`, lowercase hex only - `generateUUID()`'s exact output). Then change `strip_legacy_managed_headers` to **disable a matched row in place** (`enabled: false`, `source: "legacy-default"`) instead of deleting it - the wire is exactly as clean (a disabled row is never sent, per the existing `enabled` convention `request_composer.cpp`'s `flatten_stored_headers` already honors), but a false match is reversible from the header table rather than destructive. The startup scan now skips trashed rows (`is_null(&Request::deleted_at)`); `restore_request_locked`/`restore_collection_locked` run the same disable-and-mark step on a row's way back out of the trash, so a request sitting there stays byte-identical until it is actually restored.

**Also fixed while implementing (necessary, not scope creep):** `app/src/modules/request-builder/utils/system-headers.ts`'s `isLegacyManagedHeader` is the app's own copy of this exact rule (its doc comment says the two "must stay identical"), consumed by `toHeaderItems` on every request load (`request-builder/index.tsx:337`). Since the engine's fix leaves a matched row's key/value untouched and only flips `enabled`/`source`, the *old*, broader app-side filter would still match it by shape and silently strip it from the editor's own state on every load - and then delete it for real on the next save, defeating the whole point of disabling instead of deleting. `isLegacyManagedHeader` now carries the identical tightened value-shape rules, and `toHeaderItems` passes a row already marked `source: "legacy-default"` straight through instead of re-filtering it, so it renders as an ordinary disabled row the user can find and re-enable. `KeyValueEntry.source`'s type is widened to `HeaderRowSource = AutoHeaderSource | "legacy-default"` to carry the new marker value.

**Alternative considered and rejected:** matching `User-Agent` only on the `vayu/` name prefix (the pre-fix behavior) rather than requiring the full `vayu/<plain version>` shape. Rejected because it's exactly the kind of name/family match the issue is about - a `User-Agent: Vayu/<garbage>` a caller wrote by hand would still be caught.

## Type of Change
- [x] Bug fix

## Testing

- `engine/tests/default_headers_test.cpp`: rewrote the startup-pass tests to assert disable-and-mark (not deletion), added `MatchesTheExactValueShapeNotJustTheHeadersNameOrFamily` (a hand-typed `X-Vayu-Version`, an uppercase UUID, and a v1 UUID all survive **enabled**), `IsIdempotentOnARowItHasAlreadyDisabledAndMarked`, and `StartupLeavesATrashedRequestByteIdenticalUntilItIsRestored` (0 rows touched by the startup scan while trashed; disabled-and-marked only after `restore_deleted`). Mutation-checked by hand: reverting the `is_bare_uuid` version/variant check or the `is_plain_semver` check reds the new shape test; reverting disable-back-to-delete reds `DisablesAndMarksTheRowsTheRendererUsedToSaveRatherThanDeletingThem`; reverting the `is_null(deleted_at)` filter reds the trash test.
- `app/src/modules/request-builder/utils/system-headers.test.ts`: added cases mirroring the engine's tightened shapes, plus `does not filter a row the engine's repair pass already disabled and marked` (mutation check: without the `source === "legacy-default"` passthrough, this reds).

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`: **2974/2974 passed**.
`cd app && pnpm test`: **8086/8086 passed**. `pnpm type-check`, `pnpm lint`, `pnpm format:check`: all clean.
`mkdocs build -f .github/mkdocs.yml --strict`: clean.

Two independent read-only reviews ran against the full diff (acceptance-criteria-by-line, repo-conventions). One real finding survived: a startup-pass log line (`database.cpp`) still said "Removed Vayu's own headers..." after the behavior changed to disabling - fixed to "Disabled Vayu's own headers...". Two minor points were raised and left as-is with reasons: `is_plain_semver` allows a leading-zero group like `01.2.3` (harmless - `VAYU_VERSION_STRING` has never had one, and tightening further adds complexity the issue doesn't ask for), and a row a user manually disabled that happens to hold an exact-shape value gets `source: "legacy-default"` stamped on it too (an accepted consequence of shape-only matching, per the issue's own design - the safe direction, since it only ever affects `enabled: false` rows and never deletes).

## Checklist

## Related Issues
Closes #1491

## Visual changes

No new UI surface. The app-side change reuses the existing disabled-row rendering in `KeyValueEditor` - a `legacy-default`-marked row was previously either invisible (old broad match, silently dropped) or would have vanished on the next save; it now renders and behaves exactly like any other disabled header row, unchanged in appearance. Proven through `system-headers.test.ts`'s rendered-row assertions rather than a screenshot, since no new visual state or component is introduced.

Docs updated in this commit: `docs/engine/db-schema.md` (the header-strip pass section: exact-shape matching, disable-not-delete, trash handling), `docs/app/COMPONENTS.md` (`system-headers` module bullet), `docs/app/state-management.md` (`KeyValueEntry.source`'s widened type).


athrvk:
fix merge conflict

athrvk:
Merged `master` in and resolved the one real conflict (a doc paragraph in `docs/engine/db-schema.md` where this PR's and #1539's edits landed on the same lines). Verified after merging: full engine suite green, full app suite (8123 tests) green, lint/format/type-check clean, `mkdocs --strict` clean. Pushed as `1e896cf`.

---

PR #1539, opened 2026-09-07
fix(engine,app): cap a design run's stored config snapshot body

### LINKED ISSUE #1486: engine(runs): a design run's config snapshot keeps the whole request body, and the compact runs list parses every row's snapshot - one 50 MB send makes the history list take seconds
## Problem

Found by the persistence sweep at master `cb829ad3` (engine 0.25.0), re-anchored at `228f1131`. Measured against a real engine on a loaded box, so absolute numbers are inflated; the ratios are not.

| Call | With one 50 MB design run stored | After deleting it |
|---|---|---|
| `GET /runs?limit=50` cold | 6.4 s | 126 ms |
| `GET /runs?limit=50` warm | ~430 ms | - |
| `GET /runs?limit=50&type=load` | 3 ms | - |
| bare `GET /runs` (legacy no-param) | 52 MB payload | - |

Mechanism:

- `engine/src/http/routes/execution.cpp:1211` stores `sanitize_config_snapshot (req.body)` for a design send. `engine/src/utils/json.cpp:1246-1265` strips credentials from `auth` and keeps everything else, the request body included. The trace body is capped at 5 MiB (`maxTraceBodyBytes`, with `bodyTruncated`); the snapshot has no cap.
- #89 replaced `configSnapshot` in list rows with a six-key `summary`, but the summary is built by parsing each row's `config_snapshot` (`runs.cpp`, the summary helper), and the `q` filter and `json_valid` guard at `engine/src/db/database.cpp:2469-2476` run over the column. One 50 MB row is 50 MB of JSON parse per page load. The renderer polls the first page every 5 s while History is mounted.

Severity: medium. Any large send (a file upload, a big JSON fixture) leaves a row that taxes every history refresh until it is pruned.

## Fix pathway

1. **Cap the snapshot's body.** In `sanitize_config_snapshot`, replace a `body.content` (and form file parts' inline content) over the trace cap with `{"truncated": true, "bytes": N}`, the shape the trace already uses. The snapshot exists for the report's request panel and for `q` search; neither needs 50 MB of body, and the report already shows the truncated trace body.
2. **Stop parsing the snapshot for the list.** Store the six summary keys as their own column (`runs.summary`, JSON, written at run creation), so the list reads a small column and the `q` filter runs over it. `sync_schema` adds the column; existing rows backfill on first read or in a one-time startup pass that is marked done.
3. Bare `GET /runs` (legacy, slated for removal per #89) is unchanged.

## App changes

`engine/src/utils/json.cpp`, `engine/src/db/database.cpp` (the `Run` table and the list query), `engine/src/http/routes/execution.cpp`, `runs.cpp`. No renderer change; the list shape is the same.

## Tests

`engine/tests/json_test.cpp`: `sanitize_config_snapshot` on a body over the cap yields the truncated marker and keeps `auth.mode`; mutation check: remove the cap. A runs-route test: after one run with a large body, the list row's `summary` is complete and the response is under a fixed byte budget; a timing assertion is not needed, the byte budget is the proxy. `python build.py -t && cd engine && ctest --preset linux-dev -R "json|runs"`.

## Docs, same commit

`docs/engine/db-schema.md`, the `runs` table (the new column, the body cap in the snapshot). `docs/engine/api-reference.md`, `GET /runs/:id` (the snapshot's body may be truncated).

## Sequencing

Wave D of the persistence sweep. Independent of every open issue; the `runs` table is touched by nothing else open.

## Acceptance criteria

- After a design send with a 50 MB body, `GET /runs/:id` returns a snapshot whose body is the truncated marker, and `GET /runs?limit=50` answers within the same order of magnitude as with no such run.
- The report's request panel still shows method, URL, headers and the truncated body.

---

athrvk:
## Description

Owner-measured (issue #1486): a design run's `config_snapshot` stored the raw request body with no cap, so one 50 MB send left a permanently bloated row behind - `sanitize_config_snapshot` (`engine/src/utils/json.cpp`) redacted `auth` but never bounded `body.content`, unlike the trace body it sits beside, which is already capped at `maxTraceBodyBytes` with a `bodyTruncated`/`bodyBytes` marker.

**Root cause and plan - and a deviation from the issue's literal fix pathway, with reasoning.** Three parallel research passes (issue-verification, consumer-tracing, docs/conventions) found that the issue's *second* half - "stop the compact list from re-parsing every row's full snapshot", proposed as a new `runs.summary` column backfilled at run creation - is already solved, differently and more cheaply, by issue #1150's `RunSummaryCache` (merged in PR #1374): a run's list-row summary is derived from `config_snapshot` **once per run id** and held in a bounded in-memory cache for the process's life, because `config_snapshot` is write-once. `docs/engine/db-schema.md` documents this explicitly and warns that persisting a summary column would fight that design. Worse, the issue's proposed column name, `runs.summary`, is **already a mapped column** holding a completely different thing (the whole-run terminal results payload) - adding a second, differently-timed `summary` would collide with it outright.

So this PR implements only the half of the issue that is still open: capping `config_snapshot`'s `body.content` at the same `maxTraceBodyBytes` limit the trace already uses, with the identical `bodyTruncated`/`bodyBytes` sibling-key marker `cap_node_body` uses for a trace node (one level down here, since `content` lives inside `body` rather than being it). This still satisfies the issue's substantive complaint: the `q` filter's SQL `LIKE` and `collectionId`'s `json_extract` (`run_filter_where`, `engine/src/db/database.cpp`, run unconditionally on every list query) scan `config_snapshot`'s full stored length on every call, and the one-time cache-population parse also pays for it - all now bounded regardless of how large a client's original body was.

**Alternative considered and rejected:** adding the `runs.summary` column anyway, under a different name (e.g. `list_summary`) to dodge the collision. Rejected because it would re-litigate a settled, documented design (#1150) for no measured benefit - the cache already makes the per-poll cost O(1) after the first parse, and the remaining costs (initial parse, `q`/`collectionId` scan) are exactly what capping the body fixes directly, without a schema change or a backfill pass.

**A consumer the issue didn't mention, found and fixed:** capping `config_snapshot`'s body broke an explicit assumption in `design-run-seed.ts` - its doc comment said the snapshot's body "is not capped" and is why the editable copy showed the full body even when the trace was truncated. `requestBodyTruncated` (which gates whether "Save this run to the request" is allowed to write the body back) now folds in the snapshot's own `bodyTruncated` flag alongside the trace's, so a run truncated only on the config-snapshot side can no longer have a partial body silently written back as if complete.

## Type of Change
- [x] Bug fix

## Testing

- `engine/tests/request_builder_test.cpp` (new cases): an oversized `body.content` is cut to `maxTraceBodyBytes` and marked `bodyTruncated`/`bodyBytes`; a body within the cap is untouched (no marker keys added); a body shape with no `content` field (`mode: "none"`) is left alone rather than crashing. All existing `SanitizeConfigSnapshot` cases updated for the new required `max_body_bytes` parameter.
- `app/.../design-run-seed.test.ts` (new case): a run truncated only on the config-snapshot side (trace untouched) still sets `requestBodyTruncated`, proving `save-run-to-request`'s guard fires for either source.

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure`: **2977/2977 passed** (2971 pre-existing + 6 new `SanitizeConfigSnapshot` cases; the 3 existing cases were updated for the new parameter, not added).
`cd app && pnpm test design-run-seed save-run-to-request`: **47/47 passed**. `pnpm lint`: clean (fixed one prettier wrap CI caught on the first push).

## Checklist

## Related Issues
Closes #1486

## Visual changes

No user-visible change to any screen: the fix is a storage-side cap plus a truncation marker. The one behavioral surface - "Save this run to the request" refusing to write back a truncated body - already had a notice for the trace-truncated case (`buildChangeset`'s `kept` row, in `save-run-to-request.ts`); this PR only widens when that existing notice fires, so no new UI was added.

Docs updated in this commit: `docs/engine/db-schema.md` (a new `config_snapshot`-body-cap paragraph beside the existing redaction note, and why a persisted summary column was rejected), `docs/engine/api-reference.md` (`configSnapshot.body`'s cap documented beside the existing trace-truncation note on `GET /runs/:runId`).

### DIFF
--- a/app/src/modules/history/main/design-run-seed.test.ts
+++ b/app/src/modules/history/main/design-run-seed.test.ts
@@ -363,5 +363,19 @@ describe("seedFromRun", () => {
 		it("leaves it undefined for an untruncated run", () => {
 			expect(seedFromRun(run(), liveRequest).requestBodyTruncated).toBeUndefined();
 		});
+
+		it("sets requestBodyTruncated from the config snapshot's own cap, with no trace cut", () => {
+			// sanitize_config_snapshot caps body.content independently of the
+			// trace, so a run can be truncated there with the trace untouched.
+			const truncatedSnapshotOnly = run({
+				configSnapshot: {
+					method: "POST",
+					url: "https://api.example.test/users?page=2",
+					body: { mode: "json", content: "SLICE", bodyTruncated: true, bodyBytes: 5_242_880 },
+				},
+			} as Partial<Run>);
+
+			expect(seedFromRun(truncatedSnapshotOnly, liveRequest).requestBodyTruncated).toBe(true);
+		});
 	});
 });
--- a/app/src/modules/history/main/design-run-seed.test.ts
+++ b/app/src/modules/history/main/design-run-seed.test.ts
@@ -371,7 +371,12 @@ describe("seedFromRun", () => {
 				configSnapshot: {
 					method: "POST",
 					url: "https://api.example.test/users?page=2",
-					body: { mode: "json", content: "SLICE", bodyTruncated: true, bodyBytes: 5_242_880 },
+					body: {
+						mode: "json",
+						content: "SLICE",
+						bodyTruncated: true,
+						bodyBytes: 5_242_880,
+					},
 				},
 			} as Partial<Run>);
 
--- a/app/src/modules/history/main/design-run-seed.ts
+++ b/app/src/modules/history/main/design-run-seed.ts
@@ -38,7 +38,15 @@ interface DesignSnapshot {
 	method?: string;
 	url?: string;
 	headers?: Record<string, string>;
-	body?: { mode?: string; content?: string; fields?: KeyValueEntry[] };
+	body?: {
+		mode?: string;
+		content?: string;
+		fields?: KeyValueEntry[];
+		/** Set by `sanitize_config_snapshot` when `content` exceeded `maxTraceBodyBytes`. */
+		bodyTruncated?: boolean;
+		/** The body's original byte length, present only when truncated. */
+		bodyBytes?: number;
+	};
 	auth?: { mode?: string };
 	preRequestScripts?: ScriptPart[];
 	postRequestScripts?: ScriptPart[];
@@ -78,10 +86,9 @@ export interface DesignRunSeed {
 	recordedAuthMode?: string;
 	/**
 	 * True when the engine truncated this run's stored request body
-	 * (`maxTraceBodyBytes`). The editable copy still shows the full body from the
-	 * config snapshot (which is not capped), but "Save this run to the request"
-	 * must not write a possibly-incomplete body back - see
-	 * {@link applyRunToRequest}. Read from `trace.request.bodyTruncated`.
+	 * (`maxTraceBodyBytes`), on the trace or on the config snapshot itself -
+	 * both are capped at the same limit. "Save this run to the request" must
+	 * not write a possibly-incomplete body back - see {@link applyRunToRequest}.
 	 */
 	requestBodyTruncated?: boolean;
 }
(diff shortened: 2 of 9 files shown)


athrvk:
Is the max body byte constant user configurable ?

athrvk:
Yes. `maxTraceBodyBytes` is an existing, user-configurable engine setting - this PR doesn't add a new constant, it reuses the same one the trace body already capped at.

It's seeded as a regular `ConfigEntry` (`engine/src/db/database.cpp:4165`):
- Category `data_retention`, label "Max Stored Trace Body Size" - shows up in Settings under Data Retention (not in the collapsed Advanced section, since `advanced` isn't set on it).
- Default 5 MiB, valid range 1 KiB – 100 MiB.
- Editable via the Settings UI or directly through `PUT /config` / the MCP `update_config` tool.

Because both the trace body and `config_snapshot`'s body now read the same `maxTraceBodyBytes` value, raising or lowering it in Settings changes both caps together - there's no separate knob for the snapshot side, which keeps them from drifting out of sync with each other.

---

PR #1551, opened 2026-09-07
fix(engine): centralize the HTTP request log and retire the legacy verbose bool

### LINKED ISSUE #1510: engine(logging): one request log line per HTTP call, at one level, from one place - today 48 hand-written lines cover about half the routes at inconsistent levels, `GET /inbox` and the mock routes log nothing, and a second `verbose` bool runs beside the logger's own level
## Problem

Owner-requested after the 0.26.0 CPU and memory measurement, where the engine log could not say whether the app's three service polls (`GET /inbox`, the mock-server and mock-issuer lists) ever ran: those read paths write no log line at any verbosity, so a `curl` of them leaves the log untouched while `GET /runs` and `GET /health` log at DEBUG. Verified at master `800426d8`.

The engine already has the right centre. `vayu::utils::Logger` (`engine/include/vayu/utils/logger.hpp`) is a singleton with a console verbosity (`--verbose [LEVEL]`, `0` warn and error, `1` info, `2` debug, default `1`; `daemon.cpp:75-77,104`) and a separate file level (`logLevel` config, default `debug`, `daemon.cpp:174-175`, `constants.hpp:64`), rotation (`maxLogFileBytes`) and pruning. What is not central is the request log: each route writes its own line by hand, or does not.

1. **Coverage is uneven.** 48 hand-written `"METHOD /path - …"` lines across `src/http/routes/*.cpp` against 46 registered route patterns, but distributed by author rather than by route: `runs.cpp` has 17, `inbox.cpp` 0 of 8 routes, `mock_server.cpp` 0 of 4, `examples.cpp` 0 of 4, `scripting.cpp` 0 of 2, `events.cpp` 0 of 1, `specs.cpp` 1 of 4, `collections.cpp` 2 of 4. (`rg -cE 'log_(info|debug) \("(GET|POST|PUT|DELETE)' src/http/routes/*.cpp` per file, against `rg -cE 'server\.(Get|Post|Put|Delete) \('`.)
2. **Levels are inconsistent.** `GET /runs/:id` logs at INFO (`runs.cpp:1400`), `GET /runs` at DEBUG (`:1385`), `GET /health` at DEBUG (`health.cpp:68`), `GET /requests/:id` at INFO. So `-v 1` shows some reads and hides others with no rule behind which.
3. **No status or duration anywhere.** The lines record that a handler was entered, never what it answered or how long it took, so the log cannot answer "which call was slow" or "which call 500'd", which is the question a bug report needs.
4. **A second switch runs beside the logger.** `Server` still takes a legacy `bool verbose` (`server.cpp:28-29`, `daemon.cpp:197-199`, `verbose_legacy = verbosity >= 1`), threads it into `RouteContext` (`:174`) and from there into `execute_exchange` (`execution.cpp:1538`, `request_exchange.cpp:506,579`) and `start_run` (`:1918-1919`, `run_manager.cpp:299`), where it gates the curl transfer frames (`client.cpp:552-556`, `TransferDebug.verbose`) and a replay progress line. The logger's own level is not consulted on that path: with the default `-v 1`, every design send's transfer frames are handed to `log_debug` (`client.cpp:95`) and land in the file, whose default level is `debug`. Two switches for one intent is the drift CLAUDE.md names; the daemon comment at `:197` already calls it legacy.

## Fix pathway

1. **One request line, from the server.** cpp-httplib's `set_logger` (a post-response hook with `Request` and `Response`) is the single place: one DEBUG line per call in one format, `GET /inbox 200 1.3ms 412B`, with the start time captured in a `set_pre_routing_handler` stamp (the inbox server already uses pre-routing at `inbox.cpp:123`; the management server has none). Method, path, status, duration and response bytes only - never the query string, headers or body, so no token from `/oauth2/*` or a `?q=` search can reach the log. Non-2xx at INFO, 5xx at WARNING, so `-v 1` shows what went wrong and `-v 2` shows everything.
2. **Delete the hand-written entry lines** that only say "handler entered" (the 48 above). Lines that carry information the server hook cannot know stay and are re-levelled by one rule: state changes the user asked for (a run started or stopped, an inbox or mock started, a config write, a shutdown) at INFO; internal detail at DEBUG; refusals the caller can fix at DEBUG (they are in the request line's status already); anything the engine could not do at WARNING or ERROR. Write the rule down once in `logger.hpp`'s header comment.
3. **Retire the legacy bool.** `Server`, `RouteContext`, `execute_exchange` and `start_run` stop taking `verbose`; the curl transfer frames are gated on `Logger::instance ().get_verbosity () >= 2` (or a `Logger::enabled (Level::DEBUG)` helper that also considers the file level), and the replay progress line at `run_manager.cpp:299` becomes an unconditional `log_debug`. The per-run payload key `verbose` (`run_manager.cpp:1156`) is a different thing (a caller asking for transfer frames on one run) and stays.
4. **The inbox listener** (`inbox.cpp`, its own `httplib::Server`) gets the same hook so a capture's arrival is one DEBUG line; it is the other server in the process.
5. **Docs.** `docs/engine/architecture.md` gains a "Logging" section: the two levels the console offers above quiet, the file level, the request-line format and the level rule; `docs/engine/api-reference.md`'s `logLevel` entry links to it; `docs/engine/building.md`'s "Run with verbose output" example shows what `-v 2` prints for one call.

## App changes

None required. Optional: the Settings entry for `logLevel` (`database.cpp:4033-4043`) already explains the console and file split; its description gains one sentence naming the request line.

## Tests

- `engine/tests/server_test.cpp` (or the health route test): with the logger captured (a test sink, or the file at a temp path), one `GET /health` yields exactly one request line matching `^GET /health 200 \d+(\.\d+)?ms \d+B$` at DEBUG, and a `GET /runs/does-not-exist` yields a 404 line at INFO; mutation check: remove the hook, the count case reds.
- A source-scanning guard over `src/http/routes/*.cpp` asserting no `log_(info|debug) ("(GET|POST|PUT|DELETE)` line remains, and that it scanned a non-empty file set.
- `client` test: the transfer frames reach the logger only when verbosity is 2; mutation check: restore the bool gate.
- `python build.py -t && cd engine && ctest --preset linux-dev -R "Server|Health|Client"`, then clang-tidy on the touched files.

## Sequencing

Independent of everything open. Touches every route file lightly (line deletions), so land it between other engine PRs rather than alongside one; #1503 (the load run's warnings array) and #1502 add log lines of their own and should follow this one so they use the rule.

## Acceptance criteria

- Every route on the management server, and every capture on an inbox listener, produces exactly one request line at `-v 2`, in one format, with status and duration; `-v 1` shows the non-2xx ones; `-v 0` shows none.
- `rg "log_(info|debug) \(\"(GET|POST|PUT|DELETE)" engine/src/http/routes` is empty and the guard test pins it.
- `Server`, `RouteContext`, `execute_exchange` and `start_run` no longer carry a `verbose` parameter; the curl transfer frames follow the logger's level.
- `docs/engine/architecture.md` has the Logging section and the level rule.

---

athrvk:
## Description

Before this, a route's request line was hand-written per handler: 48 lines across 24 files, roughly half the routes had none at all (`GET /inbox` and the mock listings among them), levels disagreed with no stated rule, and none of them carried a status code or duration - so the log couldn't answer "which call was slow" or "which call 500'd", and an absent line was never evidence a route didn't run.

**Approach.** `vayu::http::install_request_logger` (`http/request_log.hpp`) is the one hook: a `set_pre_routing_handler` stamps the start time onto `Response::user_data` (cpp-httplib's own mechanism for a pre-routing handler to hand a route handler data - a post-response `set_logger` reads it back, since both run on the same connection thread), and emits one line - `GET /inbox 200 1.3ms 412B` - method, path, status, duration and response bytes, never the query string, headers or body (any of the three can carry a token or a credential). The level follows the status: 2xx at DEBUG, 3xx/4xx at INFO, 5xx at WARNING (visible even at `-v 0`, since an engine failure isn't something a quiet run should hide). It's installed on both `httplib::Server` instances the engine owns - the management API (`Server::setup_routes`) and the webhook inbox listener (`routes/inbox.cpp`) - but deliberately not on the mock server's listener, which already installs its own pre-routing handler for routing an arbitrarily long mocked path (#1139) and whose return value decides whether cpp-httplib routes the request at all.

Of the 48 original lines, 27 were pure "handler entered" announcements and are deleted outright. The remaining 21 carry information the hook can't know (a run starting or stopping, a count a handler computed) - those stay, but no longer repeat the method and path the centralized line already carries, since that was the actual redundancy: a hand-written `"POST /reorder - wrote " + n + " collections"` becomes `"wrote " + n + " collections"`. A source-scan test (`request_log_test.cpp`) fails the build if any route file's `log_info`/`log_debug` call starts with a bare HTTP method name again.

**Also retired in the same issue** (its own fix-pathway item, not scope creep): the daemon's legacy `bool verbose` - threaded through `Server`, `RouteContext`, `execute_exchange`, and every load-run worker function down to `client.cpp`'s curl transfer-debug frames - already had a dead-but-present replacement sitting in `Logger` (`set_verbose`/`is_verbose`, marked "Legacy support", zero callers). Rather than wire those in, every former `if (verbose)` gate on an internal-detail log line becomes an unconditional DEBUG line (the level itself is now what decides visibility, which is the whole point of centralizing on `Logger`), and the curl transfer frames gate on `Logger::instance().get_verbosity() >= 2` directly, per the issue's own fix pathway. The per-run JSON `"verbose"` key (`run_manager.cpp`'s `configure_event_loop`, a caller opting one load run into curl debug frames on the event-loop path) is a fully independent switch, confirmed by tracing every read site, and is untouched. Removing the parameter eliminated it from 13 function signatures and ~28 test call sites - all mechanical arity fixups, since no test asserted on the bool's behavior (verified: every call site passed a literal `false`).

**One pre-existing test removed, in scope.** `RunsRouteLogging.TheListingLineIsDebugNotInfo` (#1150) was a source-scan guard asserting one specific route's listing line stayed DEBUG rather than INFO, to avoid waking the Electron main process every 5s of polling. That hand-written line is one of the 27 deleted above; the property it guarded is now structural (every route's 2xx response logs at DEBUG, not one route's own discipline), and `request_log_test.cpp`'s `RequestLogLevelTest` covers the level rule directly.

**What's safe to skim:** the per-file mechanical diffs in `engine/src/http/routes/*.cpp` (each is either a straight deletion or a one-line prefix strip) and the ~28 test call-site fixups (`, false)` dropped from an argument list). What's worth a closer look: `request_log.cpp`'s use of `Response::user_data` for the timestamp handoff, and the level-selection logic in `run_manager.cpp`/`scenario_runner.cpp` (each former `if (verbose) log_info(...)` block became an unconditional `log_debug(...)`, not a behavior-preserving no-op - these lines now always reach the file sink, gated only by the console's `-v` level and the file's own `logLevel` floor).

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

`python build.py -e -t && cd engine && ctest --preset linux-dev --output-on-failure` - **3011/3011 passed**, 0 failed. New coverage: `request_log_test.cpp` - `format_request_log_line` matches the documented example exactly and rounds duration to one decimal place; two level-boundary tests (`TwoXxIsDebugAndFourXxIsInfo`, `FourXxIsInfoAndFiveXxIsWarning`) verified through the file sink's own configured level floor, the same mechanism `logger_test.cpp`'s `LoggerLevelTest` uses (a status's level is otherwise unobservable from outside the `Logger` singleton); the source-scan guard plus its planted-positive companion test (a longer function name, a `log_warning` line, and a comment are all confirmed *not* to trip it, so an over-eager matcher couldn't pass on a tree that had reverted every deletion). `server_bind_test.cpp` gained an integration case hitting a real `Server` with a real `httplib::Client`, sending a header and a query param with recognizable secrets, and asserting the resulting log line matches the format *and* that neither secret appears in it - proving the hook actually fires through `Server::setup_routes` rather than only existing as a callable function. `app/`: no changes; `pnpm type-check`/`pnpm lint` not run since nothing outside `engine/` was touched.

Mutation-checked by hand: reverting `install_request_logger`'s level branch to always call `log_debug` reds `FourXxIsInfoAndFiveXxIsWarning`; restoring any of the 27 deleted lines reds the source-scan guard.

## Checklist
- [x] My code follows the style guidelines (clang-format 19 / clang-tidy 19 via the repo's pre-commit hook)
- [x] I have updated documentation (`engine/CLAUDE.md`'s logging convention bullet, `docs/engine/architecture.md`'s new "Request logging" section, `docs/engine/cli.md`'s `-v 2` example)

## No screenshot
Engine-only change with no app-visible surface - the daemon's `-v`/`--verbose` flag semantics (0/1/2) are unchanged; only their internal plumbing moved.

## Related Issues
Closes #1510

---
_Generated by [Claude Code](https://claude.ai/code/session_019MPgyfKyGniJbRuqTtHS76)_


athrvk:
CI failure on `C++ Tests (windows-latest)` and `asan on windows-latest`: `RequestLogLevelTest.FourXxIsInfoAndFiveXxIsWarning` failed on both, same commit (`998fc89`).

Root cause, not a flake: `ScratchLogDir` in `request_log_test.cpp` and `server_bind_test.cpp` (two new helper copies this PR added) named their scratch log directory from a bare per-process `atomic<int>` counter starting at 0. ctest runs each test as its own process invocation, so `TwoXxIsDebugAndFourXxIsInfo` and `FourXxIsInfoAndFiveXxIsWarning` each started that counter at 0 independently and collided on the identical temp path when scheduled concurrently - the exact race `logger_test.cpp`'s original `ScratchLogDir` avoids by also hashing the test's own name into the path. The race exists on every platform; it only surfaced on Windows this run.

Fixed in b7ade57: both copies now hash in `current_test_info()->name()` the same way. Verified locally with 5 stress-repeats of the affected tests under `-j8` plus a full clean suite run (3039/3039 passed) before pushing.

---

athrvk:
`tsan on macos-latest` failed on `b7ade57` - not this PR's: vcpkg died fetching OpenSSL's source archive (`curl operation failed with response code 504`) before any build or test step ran, a transient GitHub-archive timeout unrelated to any file this PR touches. No fix to port; this is infrastructure, not code. Re-run attempted but blocked ("workflow is already running") since other jobs in the same run were still in progress - will re-run once the run finishes.

---

Issue #1553, opened 2026-09-07
app(elements): script.pre/script.post forms lost the "Names mentioned" variable-reference chips


athrvk:
## Problem

Before issue #1512's migration, both script editors painted a **"Names
mentioned:"** row above their Monaco editor (issue #659, extended by #604 and
#1075): `referencedVariables` split `pm.*` reads from `{{template}}`
mentions, and `describeDataToken` / `describeColumnReference` /
`describeScopedRead` painted each name against `DATA_TOKEN_TONE_CLASS` - the
same tone table the URL bar and the Data tab's column audit use, so a column
one surface calls declared is the one every surface calls declared.

#1516 retired both editors that carried this row - the request builder's
`ScriptPanel` and the collection's `ScriptTab` - replacing them with one
`ScriptElementForm` (`components/shared/ElementList/`) used as the bespoke
form for a `script.pre`/`script.post` element on both the request builder's
and the collection's Elements tab. `ScriptElementForm` is a primitive under
`components/shared/`, so it cannot depend on either host's context
(`useRequestBuilderContext`, or the collection's `useVariableResolver`/
`useDataContract` pair) the way the two retired editors independently did -
and porting the row properly means each *host* wrapping its own
`script.*` rows with the row, not `ScriptElementForm` growing a context
dependency it structurally cannot have.

This was a disclosed, deliberate scope reduction in #1516's PR body, not an
oversight discovered after the fact - but it is a real, user-visible
regression (documented in `docs/app/COMPONENTS.md`'s "Shared Response
Viewer" section... actually the "Names mentioned" paragraph, search
`DATA_TOKEN_TONE_CLASS`) and belongs on the tracker rather than only in a PR
description.

## Fix pathway

1. `ElementsPanel.tsx` (request builder): for each `script.pre`/`script.post`
   element in `request.elements`, render the row above (or beside) that
   element's row in `ElementList`, reading `useRequestBuilderContext()`'s
   `getAllVariables`/`getVariableOrigins`/`dataColumns` exactly as the old
   `ScriptPanel` did - the functions are unchanged, only the row's caller
   moves.
2. `ElementsTab.tsx` (collection detail): same row, reading
   `useVariableResolver({ collectionId })` and `useDataContract(collectionId)`
   exactly as the old `ScriptTab` did.
3. Consider a shared row component (`ScriptReferencesRow` or similar) taking
   `script: string`, `allVariables`, `getVariableOrigins`, `dataColumns` as
[...]

Issue #1552, opened 2026-09-07
app(elements): RunCollectionDialog/LoadTestConfigDialog need the Timers/Scripts override block #1495 defines


athrvk:
## Problem

Issue #1516's scope named a Run-dialog item: `RunCollectionDialog` gains
`Timers: as configured | off` and `Scripts: as marked | all inline | all
deferred`, sent as "the strictly validated `elements` override block #1495
defines"; the load-test section of the same dialog and `LoadTestConfigDialog`
get the same controls.

At the time #1516 landed, **#1495 (the load-path elements pipeline) had not
landed**, and it is what defines that override block - no engine route
accepts an `elements` override on `POST /runs` today (confirmed by grepping
the engine source for `elements.*override`/`timersOverride`/`scriptsOverride`
at #1516's base commit: zero matches). Building the UI controls without a
contract to send them against would mean inventing a payload shape the
engine cannot validate or use - exactly the "config knob the issue did not
ask for" CLAUDE.md's implementer rules refuse. #1516 therefore shipped
without this piece, disclosed in its PR body as a deliberate scope reduction
(it is not part of #1516's own Acceptance criteria, which name only the
Elements tab, the generated/bespoke forms and per-step outcomes).

What #1516 *did* ship, so this issue is scoped to exactly the gap: the
load-test warning ("Pre-request script will not run") is already
elements-aware - `LoadTestConfigDialog`'s `hasPreRequestScript` prop is now
derived from `scriptTextFor(request.elements, "script.pre")` rather than a
raw string field, and `RunCollectionDialog`'s "Scripts do not run" line is
unchanged prose (it never read a field to begin with). Neither dialog has a
`Timers`/`Scripts` control of any kind.

## Fix pathway

1. Once #1495 lands, read its final `POST /runs` contract for the override
   block (the shape `elements.md`/`api-reference.md` documents after that
   PR - do not assume #1516's guess of `{timers, scripts, includeScriptTime}`
   is exactly what shipped; verify against the merged route).
2. `RunCollectionDialog.tsx`: add the `Timers: as configured | off` and
   `Scripts: as marked | all inline | all deferred` controls (state block is
   `RunCollectionDialog.tsx`'s existing `useState` group, payload block is
   `handleRun`'s `startRun.mutate({...})`), sent as the override block on the
   `scenario` object. Same two controls on the load-test section of the same
   dialog.
3. `LoadTestConfigDialog/index.tsx`: add the `Scripts` control only (this
[...]

Issue #1495, opened 2026-09-06
engine(elements): the pipeline on the load paths - per-user scope overlay, the residual-token pass, inline declarative kinds, `ready_at` timers, the `script.*` inline opt-in, and a replay that skips inline steps (correlation under load)


athrvk:
Part of #1512. Third, after #1513 and #1514. Absorbs #1496 (pre-request scripts under load are `script.pre` elements with `inline` here).

## Problem

Measured at master `800426d8`, unchanged at `d1fe93d1`: a collection with `login` (a post script doing `pm.environment.set("token", …)`) and `me` (`Authorization: Bearer {{token}}`). As a sequential run, step 2 sends `Bearer abc123`. As a scenario load run (`scenario` + `mode`), all 225 step-2 requests sent `Bearer {{token}}` literally, the environment stayed empty, and the run reported zero errors. A five-iteration single-request load run with a pre script that adds a header and sets a URL variable sent `GET /pre/{{suffix}}` with no such header, five times.

Why: `submit_one` (`scenario_load.cpp:272-297`) binds only the data-row and identity templates (`bind_step_iteration`, `:277`) and submits; the residual-token pass has no caller on either load path (`resolve_residual_tokens` is called from `request_exchange.cpp:566` and `execution.cpp:1376` only); the post script runs only in the deferred replay (`run_manager.cpp:273-333`), where its `set()` lands in a post-run scope nobody sends; the pre script is never read (`run_manager.cpp:701-708`; `scenario_load.hpp:599-608`). `docs/engine/architecture.md:759-766` records "inline scripts on the load path" as a non-goal with a bounded-pool escape hatch; the owner has taken the escape hatch.

## Scope

1. **Per-user state.** `VirtualUser` (`scenario_load.hpp:663-671` is the `busy` acquire/release pair that already serialises a user's submit and completion) gains a `ScopeOverlay` (names written by this user; reads fall through to the run scopes; about 200 B per written variable, so 1,000 users cost kilobytes, not the 30 MB a full scope copy would) and `ready_at`. The single-request load path has one implicit user per submission and gets the overlay on `Submission`.
[...]

PR #1554, opened 2026-09-07
feat(app): the ElementList primitive with schema-driven forms, an Elements tab replacing the script tabs

### LINKED ISSUE #1516: app(elements): the ElementList primitive with schema-driven forms, an Elements tab on requests and collections replacing the two script tabs, the inheritance view, and run-dialog overrides

athrvk:
## Description

Issue #1514 (already merged) cut the engine over to the element model: `preRequestScript`/`postRequestScript` are gone from `requests`/`collections`, replaced by a generic `elements` column, and the engine now refuses the two old fields outright (400 naming `elements`) on every write route, while no longer serializing them on read at all. That cutover shipped **before** this PR, which means at `master`'s current tip the app is silently broken: every `PUT` that touches a script 400s, and every `GET` of a request or collection reads back an empty script regardless of what the engine actually holds (the transformers kept reading the field names the engine stopped sending). This PR is the fix - the app-side half of #1512's parity bar, closing #1516.

**Approach.** `RequestState`/`Request`/`Collection` gain `elements: ElementDef[]` (`{id, kind, enabled, name?, config}`); a script is now a `script.pre`/`script.post` element like any other kind. A new shared primitive, `ElementList` (`components/shared/ElementList/`), renders the list: a generic form is derived from each kind's JSON Schema (served by `GET /elements/kinds`, wired through a new `useElementKindsQuery`), and a bespoke override map replaces only `script.pre`/`script.post` with the existing Monaco editor + snippets, so a script keeps "the same editor" the issue's acceptance criteria asks for. The request builder's Pre-request/Tests tabs and the collection's Pre-request/Post-request tabs both collapse into one Elements tab bound to this primitive. Per-element outcomes (extract/assert/timer, non-script kinds) now render in the sequential run's step card and the response pane's Tests tab, beside the unchanged script `testResults`.

A new request or collection also seeds an empty `script.pre`/`script.post` pair rather than starting `elements: []` - the two tabs this feature replaced were always present and ran only once given a non-blank body, so a fresh entity now looks the same way in the Elements tab rather than hiding that slot behind the Add menu. Seeded at every creation path (`useTreeCrud.ts`'s New Collection/Folder/Request, `useNewRequest.ts`'s command-palette flow); imports already carry a source document's real scripts through unchanged (the engine's parser writes `script.pre`/`script.post` elements directly from Postman's `event[]` array and Insomnia's script fields, so nothing here affects that path).

**Alternative rejected:** keeping the old two-string-field UI and translating to/from a synthetic single-element array at the API boundary only, so none of the tab/panel code would need to change. Rejected because it doesn't touch the actual acceptance bar - "add `extract.json` and `assert.status` from the catalogue, edit them through generated forms" needs a real list UI, not a translation shim over a fixed two-slot model, and it would have meant building the real `ElementList` primitive later anyway once #1517 (MCP) or the next element kind needed it.

**Scope reductions, disclosed rather than silently dropped** (none are in #1516's acceptance criteria):
- The Run dialogs' `Timers`/`Scripts` override controls (named in #1516's own scope section) are not in this PR - the engine contract they'd send (`#1495`, the load-path elements pipeline) hasn't landed, so there is nothing to validate the payload against yet. Filed as #1552, sequenced after #1495. The load-test warning that *did* exist (pre-request scripts don't run under load) is still here, just recomputed from the elements list instead of a raw string.
- The old script editors' "Names mentioned" variable-reference chip row (pm/template/data-column painting) isn't ported to the new bespoke script form. `ScriptElementForm` is a primitive under `components/shared/` and structurally cannot depend on either host's context the way the two retired editors independently did - porting the row means each host wraps its own script rows with it, which is real, disclosed follow-up work, not a "no editor at all" regression. Filed as #1553, unblocked, can land whenever.
- `extract.json`'s bespoke form (a path field with a response-body picker) is generic-form only, per #1516's actual acceptance criteria, which names generated forms for `extract.json` and `assert.status` explicitly.
- The MCP `create_request`/`create_collection` tools do **not** get the same default script-pair seeding - that's a deliberate exclusion, not a deferral: MCP is a programmatic surface an agent drives explicitly, and auto-injecting elements nobody asked for would surprise a caller that expects `elements: []` when it passed none.

**Also fixed in this PR** (found while tracing every consumer of the two retired fields, per this repo's own "written but never read" rule):
- `app/electron/mcp/tools.ts`'s `run_request` tool sent `preRequestScript`/`postRequestScript` straight through `POST /compose`'s inline path to `POST /execute`, which now 400s on those keys - every MCP-driven send carrying a script was failing outright.
- `composeLoadRunRequest` (also MCP) composed `elements` for `/compose` but never folded the resulting `script.post` text back into `tests`, the field `POST /runs` actually reads (that route is unchanged by #1514) - a saved test script was silently dropped from every load run started through MCP.
- `DesignRunView.tsx` was still passing the retired `inheritedPreScripts`/`inheritedPostScripts` prop names to `RequestBuilderProvider`, which only reads `inheritedElements` now - History's "runs before your own" notice was silently rendering nothing for every past run.
- `GenericElementForm` crashed on `assert.jsonpath`'s `expected` property, whose JSON Schema is a literal `null` (unconstrained - it accepts any JSON value) rather than an object; the new conformance test caught this before it could crash the real catalogue.

## Visual changes

Screenshots are from a real running build (engine + Electron), not a mock.

**Generated forms from the live `GET /elements/kinds` catalogue**, including a nested-object property (`assert.status`'s `range.min`/`range.max`) rendered as a nested fieldset:

Generic schema-driven forms

**The bespoke override**: `script.pre`/`script.post` still render the real Monaco editor and snippet panel, with the engine's own live validation surfacing as a toast:

Bespoke script editor with live validation

**`InheritedElementsNotice`** on a request whose collection has a saved `script.pre` element, with the per-entry Disable control:

Inherited elements notice

**The collection's own Elements tab**, saved successfully (single tab replacing Pre-request/Post-request):

Collection Elements tab

**A brand-new, never-touched request** inside a brand-new collection: both already carry their own empty `script.pre`/`script.post` pair (tab badge "Elements 2"), and the collection's pair shows up above as "Runs before your own":

A fresh request and collection both default to a script pair

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- `cd app && pnpm test` - **715 test files, 8091 passed, 1 skipped** (the element-kinds conformance test's "fixture missing" branch, which correctly doesn't run since `engine/tests/fixtures/element-kinds.json` is present in this checkout), rerun clean after merging `master`'s concurrent `methodSource`/`AutoMethod` change into this branch.
- `pnpm type-check` - clean across the renderer, electron and electron-test projects.
- `pnpm lint` / `pnpm format:check` - clean.
- New tests: `ElementList.test.tsx` (generic-vs-bespoke form dispatch, Add menu grouped by category with `inherit.disable` excluded, toggle/delete/reorder), `element-kinds.conformance.test.ts` (reads the real engine fixture, ran for real - not skipped), `ElementsTab.test.tsx`/`ElementsPanel.test.tsx`/`InheritedElementsNotice.test.tsx`, `elements-parts.test.ts` (chain walk, `inherit.disable` suppression, `scriptTextFor` joining), `useTreeCrud.default-elements.test.tsx` (every creation path sends a `script.pre`/`script.post` pair - mutation-checked: removing the seed from one call site reddens exactly that test's case). Mutation-checked per this repo's convention throughout (e.g. the conformance test's "unknown kind falls to generic form" case reds if the schema-driven branch is removed; the variable-invalidation gate's source scan reds if either send path reverts to gating on a script-part list alone).
- MCP: `pnpm test tools` - 441/441 (3 new), full MCP suite 829/829; three mutation checks (skip the stored-elements read, drop the composed-tests fold, re-send a retired field name) each reddened the intended tests.
- Engine: untouched by this PR (all changes are app-side); not rebuilt/retested here since #1514 already covers the engine-side contract this PR consumes.
- Manually exercised against a real built engine under Electron (screenshots above): added `extract.json`, `assert.status` and `script.pre` elements from the live catalogue, hit and read a real validation-error toast for a missing required property, saved a 3-element request and a collection-level script element, confirmed the request-side inherited-elements notice picks up the saved collection element, and confirmed a freshly created collection and request both default to a populated script pair.

**Not covered by an automated test, and why:** the two disclosed scope reductions above (#1552, #1553) have no code in this PR to test.

## Checklist
- [x] My code follows the style guidelines (ESLint / Prettier, both clean)
- [x] I have updated documentation (`docs/app/COMPONENTS.md`, `docs/app/state-management.md`, `docs/app/api-integration.md`)

## Related Issues
Closes #1516

Also fixes two functional regressions from #1514's engine-side cutover that predate this PR (see "Also fixed in this PR" above): MCP's `run_request`/load-run script handling, and `DesignRunView`'s inherited-elements notice.

Follow-ups filed, sequenced and out of this PR's acceptance criteria:
- #1552 - Run-dialog `Timers`/`Scripts` override controls, blocked on #1495.
- #1553 - Port the "Names mentioned" variable-reference chip row to the new Elements tabs, unblocked.


athrvk:
`Engine on macos-latest` failed on `adffbdf` - not this PR's: the build itself never runs, it dies during vcpkg's dependency fetch with `curl operation failed with response code 504/502` downloading `automake-1.17.tar.gz` from `ftpmirror.gnu.org` and the `nlohmann-json` v3.12.0 archive from `github.com`, both external mirrors this PR's diff has no connection to (this PR touches no engine or vcpkg files at all). Re-run kicked off for the failed job only.

---

Issue #1559, opened 2026-09-07
mcp(elements): run_collection and start_load_run's scenario branch need the Timers/Scripts override block #1495 defines


athrvk:
## Problem

Issue #1517's scope named a run-tools item: `run_request` and `start_load_run`'s single-target path "take ad-hoc `elements` instead of `tests`", and "`run_collection` and the scenario branch of `start_load_run` (`startScenarioLoadRun`, `app/electron/mcp/tools.ts:2970`; refusal rows in `SINGLE_TARGET_LOAD_FIELDS`, `tools.ts:2885`) take the override block (`timers`, `scripts`)".

At the time #1517 landed, **#1495 (the load-path elements pipeline) had not landed**, and it is what defines that override block and what teaches `POST /runs` to read a top-level `elements` override at all - confirmed by reading `engine/CLAUDE.md`'s own elements paragraph at #1517's base commit: "a **load** run (`POST /runs` at scale) runs no element yet - that is #1495's job, so a scenario load run still only inspects a step's compiled `elements` to decide what to warn about and what to sample, and its own deferred `tests` script keeps the pre-#1514 joined-parts shape untouched." Building the override block or ad-hoc `elements` argument without an engine contract to send them against would mean inventing a payload shape the engine cannot validate or use - exactly the "config knob the issue did not ask for" CLAUDE.md's implementer rules refuse. #1517 therefore shipped without this piece, disclosed in its PR body as a deliberate scope reduction (it is not part of #1517's own Acceptance criteria).

What #1517 *did* ship, so this issue is scoped to exactly the gap:
- `create_request` / `update_request` / `create_collection` / `update_collection` take `elements` (the whole list), with `preRequestScript` / `postRequestScript` sugar folding into `script.pre` / `script.post` entries.
- `run_request` takes an ad-hoc `elements` argument, appended to the composed request's own elements after `POST /compose` - this half of the "ad-hoc elements instead of tests" ask *is* done, since `/execute` already accepts `elements` in its payload.
- `start_load_run`'s single-target (non-scenario) path still only accepts `postRequestScript` / `tests`, unchanged - `POST /runs` has no top-level `elements` argument to send ad-hoc extractors/assertions against yet.
- `run_collection` and `start_load_run`'s scenario branch have no `timers` / `scripts` override argument of any kind.
[...]

PR #1560, opened 2026-09-07
feat(mcp): elements on request/collection tools, ad-hoc elements on run_request, and the elements/kinds resource

### LINKED ISSUE #1517: mcp(elements): `elements` on the request and collection tools, run overrides on the run tools, the `vayu://elements/kinds` resource, and the registry conformance test

athrvk:
## Description

Issue #1514 already cut the engine over: `POST /requests`, `PUT /requests/:id`, `POST /collections`, `PUT /collections/:id` and `POST /execute` all refuse `preRequestScript`/`postRequestScript`/`tests` outright now that `elements` is the only script source. `app/electron/mcp/tools.ts` had zero `elements` handling, so `create_request`, `update_request`, `create_collection`, `update_collection` and any `run_request` call carrying a non-empty script were already 400ing against current master - this PR isn't adding a feature on top of working tools, it's what makes them work again.

**Approach.** The four CRUD tools gain an `elements` argument (whole-list semantics, matching how `headers`/`auth` already work) with `preRequestScript`/`postRequestScript` kept as sugar: a non-empty string folds into a `script.pre`/`script.post` entry, an empty one removes it. `elements` and the sugar can't both be given on one call - stating the whole list already says what belongs in it. The tricky part is `update_request`/`update_collection`: `PUT` replaces `elements` wholesale (no per-entry patch over a JSON column), so sugar-only calls now do a `GET` first and fold the script into the *existing* stored list - without that read, clearing a script would silently drop every other element the row already had. `run_request` gets the same script→element translation plus a genuinely new ad-hoc `elements` argument, both appended *after* `POST /compose` rather than laid over the inline `request` (which would replace, not add to, whatever the composed request already carries) - this preserves the pre-#1514 behavior where an ad-hoc script ran *alongside* a saved request's own scripts, not instead of them.

Also added `vayu://elements/kinds` (re-serving `GET /elements/kinds`) and `element-kinds.conformance.test.ts`, which pins the resource, the shared `elementSchema`'s `kind` examples, and every kind literal a tool description names to the engine's own fixture - mutation-checked by hand (see Testing).

**Alternative rejected:** retiring `readValidationScript` entirely, as the issue's body suggested. `POST /runs` never cut over to `elements` (engine/CLAUDE.md: a load run "runs no element yet" - that's #1495's job), so `start_load_run`'s ad-hoc `postRequestScript`/`tests` still has to reach the engine as a raw `tests` field, exactly as before. Retiring the function would have broken that working path with nothing to replace it. Kept the function, updated its doc comment to explain the two callers now do different things with its answer.

**Deliberate deviation from the issue, filed as #1559:** `start_load_run`'s single-target path (ad-hoc `elements` instead of `tests`) and the `timers`/`scripts` override block on `run_collection`/`start_load_run`'s scenario branch are deferred - `POST /runs` doesn't accept a generic `elements` override yet, and #1495 (not yet landed) is what defines that contract. Building it now would mean inventing a payload shape the engine can't validate, which is exactly the "config knob the issue didn't ask for" this repo's rules refuse. This mirrors #1552's precedent for the identical gap on the app side.

**Also fixed in this PR:** `composeLoadRunRequest`'s `droppedPreRequestScripts` counter was silently broken by #1514 - it counted a `payload.preRequestScripts` array that compose stopped emitting, so it always reported 0 regardless of whether a saved request actually had an unrunnable pre-request script. Now counts `script.pre` entries in the composed `elements` instead. Found while tracing what compose actually returns post-#1514 for this same code path; not fixing it would have left the PR's own tests asserting against a shape that no longer reflects reality.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

`cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check` - all green. Full suite: 715 files / 8151 tests passed. New/updated coverage: rewrote the `create_request`/`update_request`/`create_collection`/`update_collection` script tests to assert `elements` instead of the now-refused raw fields (including the fetch-merge-on-clear behavior, and refusal when both `elements` and sugar are given); rewrote the `run_request` ad-hoc script tests and the whole "Load-testing a saved request" block in `tools.test.ts` (it was asserting against a pre-#1514 compose shape that no longer matches the engine); added `resources.test.ts` and `protocol.integration.test.ts` coverage for the new resource.

Mutation-checked by hand: renamed `script.pre` to `script.presend` in `engine/tests/fixtures/element-kinds.json`, ran `pnpm test element-kinds` - 6 of 9 tests reddened (the ones scanning tool schemas and the shared element schema for kind literals); reverted, back to 9/9 green, `git status` confirmed the fixture was byte-identical to what's committed.

No engine files changed, so no engine build/test run.

## Checklist
- [x] My code follows the style guidelines (`pnpm lint` / `pnpm format:check` zero errors)
- [x] I have updated documentation (`docs/engine/mcp.md`'s Tools table, the two script-alias sections, "Collection-level state", "Storing a request's scripts" → "elements", "Load-testing a saved request", the Resources table, and a new "The element catalogue" section)

## No screenshot
Engine-side surface with no app-visible change: only the MCP tool schemas and a new MCP resource moved. `app/` itself is untouched.

## Related Issues
Closes #1517

Follow-up filed: #1559 (the `timers`/`scripts` override block and `start_load_run`'s single-target `elements`, blocked on #1495).


athrvk:
`Engine on macos-latest` failed, but this PR touches no engine files (only `app/electron/mcp/`, `docs/`, and `.github/workflows/pr-tests.yml`) - the job only ran at all because editing the workflow file itself triggers the full matrix.

Root cause from the job log: vcpkg's `c4core` build failed fetching `https://github.com/fastfloat/fast_float/archive/...zip` - two consecutive `curl operation failed with response code 504` (gateway timeout), a GitHub-side archive-serving flake with nothing to do with this diff. It died in the dependency-install phase before any engine test ran. Every other job on this PR (App on all three platforms, App quality checks, Engine formatting, CodeQL, installer scripts) is green.

Re-running the failed job once the run finishes (can't re-run failed jobs while the workflow is still in progress on the other legs).

---

athrvk:
Re-run confirmed the flake: `Engine on macos-latest` and `CI gate` are both green now. All checks passing, no merge conflict, no open review threads. Ready for review.

---

athrvk:
fix merge conflict

athrvk:
Merge conflict resolved and pushed (`96c9c18`).

Master had advanced with #1554 (issue #1516's app-side `ElementList` work), which included a commit fixing the same root MCP script-translation bug this PR does, using a different internal implementation. Rather than pick one side mechanically, I adopted their `scriptElement`/`upsertScriptElement`/`readScriptEdits`/`scriptTextFromElements` machinery wholesale (it correctly folds composed elements into `tests` for the load-run path, which my original version didn't) and layered this PR's actual new capability - the `elements` argument on the four CRUD tools, ad-hoc `elements` on `run_request`, and the `vayu://elements/kinds` resource - on top via the existing mutual-exclusion check (`elements` and script sugar can't both be given on one call).

Post-merge validation, all green: `pnpm type-check`, `pnpm lint`, `pnpm format:check`, and the full `pnpm test` suite (716 files / 8114 tests passed, 1 skipped). No conflicts remain in `tools.ts` or `tools.test.ts`; `docs/engine/mcp.md` wasn't touched by master's concurrent changes, so it needed no reconciliation.

---

Issue #1564, opened 2026-09-07
engine+app(runs): wire pass/fail budgets into the collection (sequential) run path


athrvk:
## Problem

Anchored at master `0c7f22ce62c5563a709ec971c959bbcec2da52df` (after #1497's
`maxAssertionFailureRatePct` metric and `thresholds.failRun` land).

`POST /runs`' `thresholds` block validates identically for every run mode -
`vayu::core::validate_thresholds` is called once, unconditionally, from
`validate_run_config` (`engine/src/http/routes/execution.cpp:1063-1069`) - but
only the **load** path ever judges a run against it:
`vayu::core::evaluate_thresholds` has exactly one call site in the whole
engine, `run_manager.cpp`'s `collect_summary_inputs`, called from
`finish_load_test` (both the single-request and the scenario **load** run
finish through it). `execute_scenario_run`
(`engine/src/core/scenario_runner.cpp:909-1075`), the **sequential**
(design-mode collection) run's own finisher, never calls
`evaluate_thresholds` at all and has no `RunSummaryInputs`-shaped input to
call it with - it builds its own `ScenarioSummaryInputs`/summary payload
(`build_scenario_summary_payload`, `scenario_runner.cpp:434`).

So a collection run's config today can carry a `thresholds` block that
validates cleanly and is then silently ignored: the stored summary has no
`thresholds` key and the report has no `thresholdValidation` section, exactly
as if the block had never been sent - the "written but never read" pattern
`CLAUDE.md` names as this codebase's most repeated defect, just not yet
user-visible because `RunCollectionDialog.tsx` has never offered the field
(no thresholds UI exists there today).

Issue #1497 (this parent's wave-4 sibling) explicitly scoped this in: "the
sequential run's step outcome is already `Failed` on a failed assertion
(#1514), and the same metric evaluates over its per-step tallies, so a
collection run can fail on assertions too", and named
`RunCollectionDialog` under its App changes as "thresholds for a collection
run, if not already offered". It was deliberately deferred out of that PR:
wiring `evaluate_thresholds` into `execute_scenario_run` needs its own
design decision (how `ScenarioSummaryInputs` maps onto the `RunSummaryInputs`
shape `evaluate_thresholds` takes - convert, overload, or give the sequential
path its own narrower evaluator) rather than a one-line call, and building
new `RunCollectionDialog` UI on top of an unproven wiring choice would be the
"config knob the issue didn't ask for" complexity rule. #1497's own PR ships
[...]

PR #1565, opened 2026-09-07
feat(engine,app): maxAssertionFailureRatePct threshold and thresholds.failRun

### LINKED ISSUE #1497: engine(runs): assertion outcomes never touch a run's verdict - add a `maxAssertionFailureRatePct` threshold over element and script assertions, and let a failed budget fail the run
Part of #1512. After #1495 (the tallies it reads) and #1484 (the zero-sample rule it reuses).

## Problem

Measured at master `800426d8`, unchanged at `d1fe93d1`: a load run whose only `pm.test` fails on every sampled response, with `thresholds: { "maxErrorRatePct": 0 }`, ends `completed` with `errorRate 0.0`, `testValidation { testsPassed 0, testsFailed 1 }` and `thresholdValidation { verdict: "passed" }`. The final status is Completed or Stopped only (`run_manager.cpp:1427-1428`; the sequential runner marks a failed step and keeps the run Completed, `scenario_runner.cpp:1026-1036`), the error rate counts non-success status codes only (`threshold_eval.cpp:46-60`), and the threshold table is five fixed metrics (`:66-85`). With the element model, `assert.*` outcomes join `pm.test` results as the run's assertion tally; today neither reaches a verdict.

In JMeter an assertion failure marks the sample failed and counts toward the error percentage; in k6 `checks` is a rate you can threshold and fail the run with.

## Scope

1. **Metric.** `maxAssertionFailureRatePct` in the `ThresholdMetric` table, measured as `failed / (passed + failed)` over the run's assertion tally: `assert.*` element outcomes (per-step atomics from #1495 under load; step traces in the sequential run) plus script test results, inline and replayed. `inputs.tests` (`run_manager.cpp:1310-1316`) gains the element counts. Not evaluated when nothing asserted (the #1484 rule), so a run with no assertions is unaffected.
2. **Verdict.** A failed threshold already yields `verdict: "failed"`; the run's terminal status gains `Failed` when any threshold fails and the run asked for it (`thresholds.failRun: true`, default false to keep today's history semantics), so the run list and the taskbar cue (#1415's path) show it.
3. **Per step.** The report's per-step element tallies (#1495) already say where; the sequential run's step outcome is `Failed` on a failed assertion (#1514), and the same metric evaluates over its per-step tallies, so a collection run can fail on assertions too.

## App changes

`LoadTestConfigDialog/budgets.ts` (the metric row and unit), `ThresholdVerdict.tsx` label table, `RunCollectionDialog` (thresholds for a collection run, if not already offered), `types/domain.ts`.

## Tests

`threshold_eval_test.cpp`: the metric reads the combined tally, passes and fails at the limit, is unevaluated with nothing asserted; mutation check: remove the metric, reds. `run_manager` test: `failRun: true` with a failed budget stores `Failed`. App: the budget row renders and validates; `pnpm test budgets ThresholdVerdict`.

## Docs, same commit

`docs/engine/api-reference.md` (`#the-thresholds-block-passfail-budgets`), `docs/engine/elements.md` (assertion outcomes feed the verdict), `docs/compare/vayu-vs-jmeter.md`.

## Acceptance criteria

- A load run whose `assert.status` fails on every response, with `maxAssertionFailureRatePct: 0`, reports the threshold failed and, with `failRun: true`, ends in `Failed`; the same with a failing inline `pm.test`.
- A run with no assertions and the metric set reports it as not evaluated.
- Existing runs and runs without the metric are unchanged.

---

athrvk:
## Description

Issue #1497: assertion outcomes (`assert.*` elements, `pm.test` calls) never touched a run's verdict. Two research passes against current master (`0c7f22ce`) found #1495 and #1514 - the issue's own prerequisites - already merged, and turned up a scope question the issue's acceptance criteria settle: they test only **load** runs. The sequential/collection run path (`scenario_runner.cpp`) never evaluates thresholds at all today - `evaluate_thresholds` has exactly one call site in the whole engine - so wiring it in would mean inventing how `ScenarioSummaryInputs` maps onto `evaluate_thresholds`'s input shape, a design decision the issue doesn't settle either. Filed as a follow-up, #1564, and linked from the docs section this PR touches, mirroring PR #1560's precedent for the identical class of gap.

**What's new.**
- `maxAssertionFailureRatePct`, a sixth `ThresholdMetric` (`threshold_eval.cpp`), measured against a new `RunSummaryInputs::assertions` (`AssertionTotals{passed,failed}`).
- `thresholds.failRun: true` (default false): when a load run's config sets it and a declared budget failed, the run's terminal status becomes `Failed` instead of `Completed`. Decided before `finish_load_test`'s single locked status write, never a second read (`engine/CLAUDE.md`'s "reads, decides and writes holds one lock" rule) - `RunStatus::Failed` already existed (used for crashes), so this is new *wiring* to it, not a new enum value or DB migration.
- The combined assertion tally itself needed real plumbing, not just a read: `StepElementTallies` already tracks `assert.*` pass/fail correctly (`assert_kinds.cpp` pushes into the same `tests` list a `pm.test` call does), but a `script.pre`/`script.post` element's own per-run outcome collapses every `pm.test` call it made into one pass/fail bit (`ScriptResult::success` already folds a failed test in) - so a script with one passing and one failing assertion is indistinguishable, at the element-tally level, from one where both failed. `scenario_load.cpp`'s `run_step_before`/`run_step_after` now tally `pre_result.tests` / `post_result.tests` individually (`ScenarioLoadState::inline_script_tests_passed/failed`), for *inline* load-run scripts; a deferred (non-inline) script's tests still come from the existing `ScriptValidationTotals` replay. `assertion_totals` reads which elements are assertions through the registry's own `category` field (`Registry::instance().find(kind)->category == "assert"`), never a kind-name string comparison, per #1512's extensibility contract.

**Also fixed in this PR:** the app compared SSE terminal-status frames against capitalized `"Completed"|"Stopped"|"Failed"` (`sse-client.ts`) while the engine has only ever emitted lowercase (`to_string(RunStatus)`) - so `parseTerminalStatus` always returned `null` in production, and the taskbar/Dock failure cue (`os-icon.ts`'s `runFailed`, gated on this same comparison in `load-test-service.ts`/`scenario-run-service.ts`) never fired for a real failure. Couldn't defer this: #1497's own Verdict criterion says "the run list and the taskbar cue show it," and that claim is false without this fix. Existing tests masked the bug by seeding capitalized fixtures on both sides of the comparison; fixed to lowercase throughout, matching `dashboard-store.ts` and `RunReport["status"]`'s existing (correct) lowercase union.

**Two independent reviews** (one against the issue's acceptance criteria, one against repo conventions) each found one real gap, both fixed in a follow-up commit:
1. `finish_load_test` - the function containing the actual status-flip logic - had zero test coverage; only its `thresholds_fail_run()` helper was unit-tested. Added `run_manager_test.cpp`'s `FailRunTest` suite: a real run driven end to end through `RunManager::start_run` (the same harness `run_stop_test.cpp` uses for the stop path) proves `failRun: true` with a missed budget ends `RunStatus::Failed`, and a companion proves the same missed budget without the flag stays `Completed`. This is the exact case the issue's own Tests section asked for by name.
2. `assertion_totals` originally matched `elements[i].kind.starts_with("assert.")` - a string-prefix comparison outside `core/elements`, against `engine/CLAUDE.md`'s explicit rule (and this same file's own `load_pipeline_skip_reason`, a few lines below, doing it the right way). Fixed to read `ElementKind::category` through the registry.

Also fixed: a stale "same five metrics" doc comment in `budgets.ts` (now six), and `docs/app/api-integration.md`'s Budgets section, which described only the original five.

## Type of Change
- [x] Bug fix (the SSE status-case mismatch)
- [x] New feature (the metric and `failRun`)
- [ ] Breaking change
- [x] Documentation update

## Testing

- `cd engine && ctest --preset linux-dev --output-on-failure` - **3072/3072 passed**, 0 failed (this environment's vcpkg needed `vcpkg-fix-port` first, per `CLAUDE.md`'s cloud-egress gotcha). New/updated coverage: `threshold_eval_test.cpp` (the new metric's accept/reject/evaluate cases, `failRun`'s accept/reject and that it never becomes a phantom seventh check, `thresholds_fail_run`'s own reads), `scenario_load_test.cpp` (`assertion_totals` sums only registry-`category`-"assert" kinds and excludes a sibling `extract.*`; a failing `assert.status` counts as a failed assertion; the inline-script gap itself - two `pm.test` calls, one pass one fail, tallied individually even though the element's own aggregate outcome collapses both into one "failed" bit), `run_manager_test.cpp` (the end-to-end `failRun` status flip, and its opt-out companion).
- `clang-format-19 --dry-run -Werror` and the `clang-tidy-19` pre-commit hook - both clean on every touched engine file.
- `cd app && pnpm test sse-client load-test-service scenario-run-service budgets ThresholdVerdict run-supersession LoadTestConfigDialog` - **255/255 passed**. `pnpm type-check` and `pnpm lint` - clean.

## Checklist
- [x] My code follows the style guidelines (clang-format 19 / clang-tidy 19 clean; `pnpm lint` / `pnpm format:check` clean)
- [x] I have performed a self-review, plus two independent adversarial reviews (see above)
- [x] I have updated documentation (`docs/engine/api-reference.md`'s thresholds section: the new metric, `failRun`, the load-runs-only scope note and its link to #1564; `docs/engine/elements.md`'s Related issues list; `docs/app/api-integration.md`'s Budgets section)

## No screenshot
The one app-visible surface is a new `NumberField` row and a `Switch` inside `LoadTestConfigDialog`'s existing "Pass/fail budgets" disclosure - both are direct reuses of the same primitives already rendered immediately beside them for the same purpose (`BUDGET_FIELDS.map`, the `saveTimingBreakdown` switch), with no new component or visual pattern. Given the scope of this PR is primarily engine-side, a full Xvfb/Electron capture was judged not worth the added time against what it would show: the identical row/switch treatment already visible a few lines up in the same file.

## Related Issues
Closes #1497. Follow-up filed: #1564 (wiring the same metric and `failRun` into the collection/sequential run path and `RunCollectionDialog`, deferred per the Description above).

### DIFF
--- a/app/electron/mcp/tools.ts
+++ b/app/electron/mcp/tools.ts
@@ -6745,8 +6745,18 @@ export const TOOLS: McpTool[] = [
 					latencyP99Ms: z.number().positive().max(86_400_000).optional(),
 					maxErrorRatePct: z.number().min(0).max(100).optional(),
 					minThroughputRps: z.number().positive().max(1_000_000_000).optional(),
+					maxAssertionFailureRatePct: z.number().min(0).max(100).optional(),
+					// Not a budget - whether a missed one fails the run, not itself
+					// measured against anything, so it is excluded from the
+					// at-least-one-budget refinement below.
+					failRun: z
+						.boolean()
+						.optional()
+						.describe(
+							"When true, a missed budget sets this run's terminal status to failed rather than only reporting the verdict. Default false."
+						),
 				})
-				.refine((t) => Object.keys(t).length > 0, {
+				.refine((t) => Object.keys(t).filter((key) => key !== "failRun").length > 0, {
 					// `error`, not zod 3's `message`: v4 still reads the old key as a
 					// deprecated alias, and a deprecated alias is what the next major
 					// takes away.
--- a/app/src/components/shared/ThresholdVerdict.test.tsx
+++ b/app/src/components/shared/ThresholdVerdict.test.tsx
@@ -95,6 +95,28 @@ describe("ThresholdVerdict", () => {
 		expect(screen.getByText(/≤ 50ms/)).toBeInTheDocument();
 	});
 
+	it("labels the assertion failure rate budget (issue #1497)", () => {
+		render(
+			<ThresholdVerdict
+				verdict={verdict({
+					checks: [
+						{
+							metric: "maxAssertionFailureRatePct",
+							limit: 0,
+							actual: 10,
+							passed: false,
+						},
+					],
+					passed: 0,
+					failed: 1,
+					verdict: "failed",
+				})}
+			/>
+		);
+		expect(screen.getByText(/Assertion failure rate/)).toBeInTheDocument();
+		expect(screen.getByText(/≤ 0%/)).toBeInTheDocument();
+	});
+
 	it("shows a metric it has no label for rather than dropping it", () => {
 		// The counts come from the engine. A check rendered away would leave
 		// "1 of 2 budgets met" above a single row and no way to see the other.
--- a/app/src/components/shared/ThresholdVerdict.tsx
+++ b/app/src/components/shared/ThresholdVerdict.tsx
@@ -40,6 +40,7 @@ const METRIC_LABELS: Record<string, { label: string; unit: string; floor?: boole
 	latencyP99Ms: { label: "p99 latency", unit: "ms" },
 	maxErrorRatePct: { label: "Error rate", unit: "%" },
 	minThroughputRps: { label: "Throughput", unit: "req/s", floor: true },
+	maxAssertionFailureRatePct: { label: "Assertion failure rate", unit: "%" },
 };
 
 /** Trailing zeros are noise on a budget the user typed as a whole number. */
--- a/app/src/modules/request-builder/components/LoadTestConfigDialog/budgets.test.ts
+++ b/app/src/modules/request-builder/components/LoadTestConfigDialog/budgets.test.ts
@@ -61,6 +61,19 @@ describe("budget validation", () => {
 		expect(budgetError(draft({ minThroughputRps: "0" }))).toMatch(/throughput/i);
 	});
 
+	it("accepts a zero assertion-failure-rate budget, the same real ask as the error rate", () => {
+		expect(budgetError(draft({ maxAssertionFailureRatePct: "0" }))).toBeNull();
+		expect(buildThresholds(draft({ maxAssertionFailureRatePct: "0" }))).toEqual({
+			maxAssertionFailureRatePct: 0,
+		});
+	});
+
+	it("rejects an assertion failure rate outside 0-100", () => {
+		expect(budgetError(draft({ maxAssertionFailureRatePct: "101" }))).toMatch(
+			/assertion failure rate/i
+		);
+	});
+
 	it("rejects text rather than sending NaN", () => {
 		expect(budgetError(draft({ latencyP99Ms: "fast" }))).toMatch(/number/i);
 	});
@@ -82,6 +95,7 @@ describe("the payload the dialog builds", () => {
 					latencyP99Ms: "50",
 					maxErrorRatePct: "0.1",
 					minThroughputRps: "10000",
+					maxAssertionFailureRatePct: "1",
 				})
 			)
 		).toEqual({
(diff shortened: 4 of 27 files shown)


athrvk:
The "CI gate" failure was on commit `8b3d698` (superseded) - its Engine jobs show `cancelled`, not a real test failure, because the fix commit `66d7bc9` was pushed while that run was still building. A fresh run is in progress on the current head (`66d7bc9`, run [34146718950](https://github.com/athrvk/vayu/actions/runs/34146718950)); watching that one.

---

PR #1566, opened 2026-09-07
feat(engine): bind a read-only pm.cookies view for inline load-run scripts

### LINKED ISSUE #1501: engine(scripting): a scenario load run carries each virtual user's cookies step to step, but `pm.cookies` still throws there - bind a read-only view for inline scripts
Part of #1494. Wave 2.

## Problem

At master `800426d8`, `pm.cookies.get("x")` in a load run's script throws `pm.cookies.get is not available here: the cookie jar is a design-mode feature, and a load run's scripts have no jar to read` (measured). That was true when a load run repeated one request. A scenario load run now carries per-user cookies from step to step (`engine/src/core/scenario_load.cpp:266` sets `request.cookie_lines = vu->cookies`, `:314` writes them back), so a session cookie set by step 1 does ride into step 2, and a script cannot see it. `docs/engine/scripting.md:1622-1626` still justifies the throw with "a load run repeats a single request anyway".

## Fix pathway

1. In the correlation child's inline path, bind `pm.cookies` read-only over `vu->cookies` (the `Cookie` header lines the user carries), parsed into name and value pairs, no lock needed since the user's completion is the only reader.
2. Writes (`pm.cookies.jar().set`) stay refused under load with the existing message; the deferred replay stays refused for reads too, with a message that no longer claims there is no jar: "a replayed script runs after the run, against a recorded response; the user's cookies at that moment are not recorded".
3. The single-request load run has no cookies and keeps the current read refusal, reworded.

## Tests

`scenario_load_test.cpp`: a step-1 response with `Set-Cookie: sid=1`, a step-2 inline script that reads `pm.cookies.get("sid")` and asserts `1`; mutation check: unbind the view, the case reds. The message tests in `script_engine` tests are updated.

## Docs, same commit

`docs/engine/scripting.md` (`#the-cookie-jar-pmcookies`, the load-run bullets rewritten to the three cases above).

## Sequencing

Wave 2, after the correlation child (it needs the inline path). Independent of the other wave-2 children.

## Acceptance criteria

- An inline step-2 script in a scenario load run reads the cookie step 1 set for that user, and two users see their own values.
- Cookie writes and replayed reads are refused with messages that describe the actual reason.

---

athrvk:
## Description

Issue #1501: a scenario load run has carried each virtual user's cookies from step to step since #1495, but `pm.cookies` still threw unconditionally under load - the throw predates that per-VU state and was never revisited once it existed. `pm.cookies`'s flat reads (`get`/`has`/`toObject`/`all`/`count`/`each`) now answer from `VirtualUser::cookies` when an inline `script.*` element runs them, through a new `ScriptContext::cookie_read_lines` field bound only in `scenario_load.cpp`'s `run_step_before`/`run_step_after` - the same two hooks #1495 added. No lock and no `CookieJar` object: the vector is already materialized on the VU and that VU's own completion is the only reader.

Two other load-path contexts still refuse reads, and each now says why with a true sentence instead of the old "a load run's scripts have no jar" (no longer true in general): a single-request load run has no scenario and so no per-user cookie state at all, and a deferred script replay (a `script.*` element that did not run inline) runs after the run against a recorded response, whose user's cookies at that moment were never recorded. `run_manager.cpp`'s `ScriptReplay` gained one `is_scenario_replay` flag to pick between the two. `pm.cookies.jar()`'s writes are completely untouched - still refused everywhere under load with the pre-existing message, since a load run's cookie state advances through the VU itself, never through a script.

**Alternative considered and rejected:** routing the new read-only path through the existing `jar_context`/`CookieJar` machinery (e.g. a throwaway `CookieJar` per VU). Rejected because `CookieJar` is a shared, mutex-guarded, multi-scope store built for design mode's environment jars; a per-VU read here is one already-owned vector matched against one URL, and `vayu::http::matching_in` (the same matcher the jar itself calls internally) does exactly that with no object to construct or lock to take.

## Type of Change
- [x] Bug fix (the two reworded refusals correct a message that is no longer accurate)
- [x] New feature (the per-VU read access itself)
- [ ] Breaking change
- [x] Documentation update

## Testing

- `cd engine && ctest --preset linux-dev --output-on-failure -R "CookieJarScript|ScenarioLoad|LoadScriptScopes"` - **76/76 passed**, 0 failed. New coverage: `cookie_jar_test.cpp` (a per-VU view answers `get`/`has` with no jar object; `pm.cookies.jar()` still refuses writes with the per-VU view present; the refusal text is whatever the context sets, not a hardcoded default), `scenario_load_test.cpp` (an inline `script.pre` on step 1 reads step 0's `/login` cookie for its own VU, and two concurrent VUs never read each other's - the same wire-level proof the existing correlation and isolation tests use), `load_script_scopes_test.cpp` (a single-request replay and a scenario step replay each throw with their own reworded message, read back from the stored `Result` row `validate_scripts` writes).
- `clang-format-19 --dry-run -Werror` and the `clang-tidy-19` pre-commit hook - clean on every touched file (one real finding surfaced and was fixed: an `EXPECT_EQ` re-indexing `validation.steps[0]` after an `ASSERT_HAS_VALUE` on the same expression, now bound once).
- Full `python build.py -e -t` build succeeded (this environment's vcpkg needed `vcpkg-fix-port` first, per `CLAUDE.md`'s cloud-egress gotcha) with its own bundled test run green.

## Checklist
- [x] My code follows the style guidelines (clang-format 19 / clang-tidy 19 clean)
- [x] I have updated documentation (`docs/engine/scripting.md`'s cookie-jar section and its "Writing to the jar" closing line, `docs/engine/architecture.md`'s Cookie Jar / "Not on the load path" bullet)

## No screenshot
Engine-only change with no app-visible surface - no UI, route, or MCP tool changed. No user-visible change.

## Related Issues
Closes #1501

---
_Generated by [Claude Code](https://claude.ai/code/session_01P2SgrAjve99VBLYCZTD3T3)_

### DIFF
--- a/docs/engine/architecture.md
+++ b/docs/engine/architecture.md
@@ -436,7 +436,10 @@ makes a session survive from one design-mode request to the next.
   virtual user owns a private list seeded onto its own transfer and cleared at
   each iteration boundary, and the environment jar is untouched (see "Scenario
   load runs" under Load Test Strategies). One session shared between 1,000
-  virtual users is not the thing being measured.
+  virtual users is not the thing being measured. An inline `script.*`
+  element's `pm.cookies` reads that one VU's own list directly, with no lock
+  and no `CookieJar` object (issue #1501); it stays read-only there, and
+  `pm.cookies.jar()`'s writes still throw under every load-run script.
 - **Threading:** one mutex around the scope map; every accessor copies out, so
   no reference into the storage escapes to a caller.
 - **Shown as sent.** Because libcurl attaches the matching cookies itself, the
--- a/docs/engine/scripting.md
+++ b/docs/engine/scripting.md
@@ -1626,10 +1626,9 @@ would read and act on.
   to step outside any script (`docs/engine/architecture.md`'s Cookie Jar
   section, "Not on the load path"), with no lock, because that user's own
   completion is the only reader. A single-request run (no scenario, no
-  virtual user) and a scenario
-  step's deferred replay (a recorded response, not the user that fetched it)
-  still throw there, each naming its own reason rather than the one sentence
-  above.
+  virtual user) and a scenario step's deferred replay (a recorded response,
+  not the user that fetched it) still throw there, each naming its own
+  reason rather than the one sentence above.
 - **Writing goes through `jar()`**, below. There is deliberately no flat
   `pm.cookies.set(name, value)`: a written cookie needs a URL to take its
   domain and path from, which is exactly why Postman's write half hangs off
--- a/docs/engine/scripting.md
+++ b/docs/engine/scripting.md
@@ -1618,11 +1618,18 @@ would read and act on.
 - **`pm.sendRequest` shares the jar** of the request it runs inside. A
   pre-request script that logs in through it leaves the session where the real
   request will find it.
-- **Load runs have no jar.** Every one of these reads throws there rather than
-  answering `undefined`, which would read as "the cookie is gone". The jar is
-  deliberately off the load path: sharing one across the event loop's workers
-  would put a lock on the hot path, and a load run repeats a single request
-  anyway.
+- **Load runs mostly have no jar.** The jar is deliberately off the load path:
+  sharing one across the event loop's workers would put a lock on the hot
+  path. One exception (issue #1501): a scenario load run's own inline
+  `script.*` element reads that one virtual user's own cookies through the
+  flat surface above - the same per-user state the run already carries step
+  to step outside any script (`docs/engine/architecture.md`'s Cookie Jar
+  section, "Not on the load path"), with no lock, because that user's own
+  completion is the only reader. A single-request run (no scenario, no
+  virtual user) and a scenario
+  step's deferred replay (a recorded response, not the user that fetched it)
+  still throw there, each naming its own reason rather than the one sentence
+  above.
 - **Writing goes through `jar()`**, below. There is deliberately no flat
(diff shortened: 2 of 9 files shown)


athrvk:
CI status on the pushed fixes:

- **`Engine on ubuntu-latest`** (release build, `-Werror`): two real findings, both from this PR's own diff, neither reachable from the dev preset (no `-Werror` there):
  - `-Wformat-nonliteral` in `jar_cookies_from_context`: a ternary between a runtime `const char*` and a literal defeats GCC's format-string-literal check even though every branch is itself a literal. Fixed by substituting the member name in C++ and passing the literal `"%s"` as the actual format string (`script_engine.cpp`).
  - `-Wnull-dereference` (only fires under `-O3`, inside nlohmann's own header) on `(*it)[0]` after a `json::find`. Fixed by reading through `.at(...)` + `.front()` instead, matching this suite's existing pattern (`load_script_scopes_test.cpp`).
  - Both reproduced against a real (non-`-fsyntax-only`) compile with the exact failing flags before and after each fix.
- **`App on windows-latest (shard 2/2)`**: `src/services/load-test-service.test.ts` timed out at 5000ms in a test this PR does not touch (`load-test-service.test.ts`, `system notifications (issue #1358)`). This PR is engine-only and changes no TypeScript. Re-running once per the flake rule.

Pushed at `4ea8d03`.

---
_Generated by [Claude Code](https://claude.ai/code)

---

Issue #1576, opened 2026-09-08
engine(logging): restructure the remaining informal key=value log messages into fields


athrvk:
## Problem

Introduced by #1557 (the `LogRecord`/category/redaction landing), which categorized every `log_debug`/`log_info`/`log_warning`/`log_error` call site but only restructured the subset that could carry a secret (a `url=` embed in `routes/execution.cpp`'s two "Design Mode send"/"Load test started" lines and `cli.cpp`'s "Connecting to daemon" line - all now `msg` + a `url` field, redacted by `Logger::write`). #1556's own Fix pathway named "26 messages carry informal `key=value`" that should become `msg` plus `fields`; the real count at #1557's head is 52 (`rg` for a `log_\w+ \("[a-z]+", "..."` call whose message literal matches `[a-zA-Z_]+=`), all in non-secret territory (an id, a name, a count, a status word) so none of them are a redaction gap - the reason #1557 could ship without them and still pass its acceptance criteria.

Representative sites (line numbers as of #1557's head):
- `engine/src/db/database.cpp:1517` (`"Creating collection: id=" + c.id + ", name=" + c.name`) and 20 further sites in the same file (create/delete/update lines for requests, examples, environments, client certificates, spec documents, runs).
- `engine/src/http/routes/collections.cpp:474`, `:502`, `:506`, `:540` and the same shape repeated in `requests.cpp`, `environments.cpp`, `examples.cpp`, `client_certificates.cpp`, `specs.cpp`, `cookies.cpp`, `globals.cpp` - each route's create/update/delete handler.
- `engine/src/http/routes/runs.cpp:1447`, `:1521`, `:1585` (`"Found run: , type=, status="` and its siblings).
- `engine/src/http/routes/metrics.cpp:67`, `:358`, `:370`.
- `engine/src/core/run_manager.cpp:1232` (`"EventLoop config: workers=..."`) and `:2083` (`"Metrics: rps=..."`).
- `engine/src/core/load_strategy.cpp:703` (`"Submission config: tick_us=..."`).
- `engine/src/http/event_loop/curl_utils.cpp:1055` (`"queue_wait clock skew: perceived_ms=..."`).

## Fix pathway

[...]

Issue #1558, opened 2026-09-07
app(logging): `electron/log.ts` writes the same record to `app_<stamp>.log`, replaces the 67 console calls, carries MCP and renderer records over one path, stops re-printing engine stdout, and Settings gets Open logs folder


athrvk:
Part of #1556. Second; reads the schema the engine child writes.

## Problem

Anchors at master `23e31ea3`.

The app has no log. `app/electron` makes 67 bare `console.*` calls (`sidecar.ts` 33, `main.ts` 13, `updater.ts` 3, `mcp/cli.ts` 3, then one or two each in `window-reveal.ts`, `service-stop-guard.ts`, `renderer-recovery.ts`, `power-save.ts`, `run-progress.ts`, `quit-shutdown.ts`, `proxy-resolution.ts`, `preload.ts`, `os-icon.ts`, `notify.ts`, `mcp/tools.ts`), tagged `[Sidecar]`, `[Main]` and so on, with no level, no timestamp and no file: `app/eslint.config.mjs:138-149` exempts `electron/**` from `no-console` on the stated ground that the console is the log. A packaged GUI launch has no console. The engine's stdout and stderr are split on newlines and re-printed into that same console (`app/electron/sidecar.ts:556-590`), spawned at `--verbose 1` in production and `2` in dev (`:549-550`); only a 10-line stderr tail (`app/electron/constants.ts:107`) survives, for the failure message at `:365`.

The MCP stdio server logs four lines to stderr (`app/electron/mcp/cli.ts:32,40,44`, `mcp/tools.ts:7766`); the Electron-hosted HTTP transport (`mcp/http.ts`) logs nothing. The renderer's `app/src/errors/error-logger.ts:38-66` builds `{timestamp, severity, error, context}` and hands it to `console`; `ErrorBoundary.tsx:66-77` is its one caller of note; no `window.onerror` or `unhandledrejection` listener exists in `app/src`, no IPC log channel in `preload.ts`, so main learns of a renderer fault only from `render-process-gone` (`main.ts:508`). Settings, General renders the logs path as text (`app/src/modules/settings/main/panels/GeneralPanel.tsx:256-261`, from `app:getPaths`, `main.ts:1254`, `app-paths.ts:33-40`); nothing opens it.

## Fix pathway

[...]

Issue #1556, opened 2026-09-07
engine+app(logging): one log record for the whole product - JSON lines in the files, one text rendering on the console, UTC timestamps, level and category on every record, redaction by key, and an app log file that exists at all (parent)


athrvk:
## Goal

The owner's: **centralized logging for the engine and the app, controlled by the same setting.** One place to look, one shape to read, one knob. Concretely, after this parent lands:

- One directory, `<data-dir>/logs/`, holds everything: `engine_<stamp>.log` and `app_<stamp>.log` side by side (the standalone CLI writes `cli_<stamp>.log` under its own data directory, since it runs without the app).
- One record shape on every side (below), so `jq -s 'sort_by(.ts)' logs/*.log` is the merged log, and a future in-app viewer or MCP `read_log` tool reads fields, not text.
- One setting, `logLevel` (Settings, Observability, restart-required), is the floor for both files. The app has no setting of its own; it reads the engine's once per launch. `-v` is only what the engine prints to a terminal you started it in; no process of the app reads it.
- One redaction rule, applied by key in both record builders.
- One schema file both sides validate against in tests, so the shape cannot drift.

**Why two files rather than one.** The engine writes its own file because it also runs without the app (daemon, CLI, MCP stdio against a daemon). The app writes its own because its most valuable records are the ones the engine cannot carry: what happened before the engine was ready, and what the sidecar saw when the engine died. Forwarding app records to the engine for one file would lose exactly those, and forwarding engine records to the app breaks the standalone modes. Two writers into one directory with one shape and one floor is the centralized design that survives both failure modes; the merge is a sort on `ts`.

Owner-requested, as the follow-up to #1510 (PR #1551 centralises the request line). Surveyed at master `23e31ea3`, every anchor read. Backward compatibility of log file names or line shapes is not required (pre-1.0, and no release ships before #1512 completes).

## What logging is today, per side

[...]

PR #1582, opened 2026-09-08
feat(engine): wire LogRecord, categories, redaction by key, and -v parity between vayu-engine/vayu-cli

### LINKED ISSUE #1557: engine(logging): `LogRecord` - the JSON-lines file sink, the text console renderer, UTC `ts`, `level`/`cat`/`msg`/fields on every record, redaction by key, the schema file, and a category on all 474 call sites
Part of #1556. First; the app child reads the schema this one writes.

## Problem

Anchors at master `23e31ea3`; the request line is as PR #1551 (#1510, head `998fc89c`) leaves it.

`vayu::utils::Logger` builds one text line (`engine/src/utils/logger.cpp:95-96`) from a local-time stamp with no zone (`:191-202`, `format_local_time`), a level token (`DEBUG` | `INFO` | `WARNING` | `ERROR`, `:181-189`), a thread id (`:204-208`) and a free-text message, and writes that same string to the file (`:100-111`) and the console (`:125-136`). The API is message-only: `log (Level, const std::string&)` and the four free functions (`engine/include/vayu/utils/logger.hpp:54, 65-68, 136-150`); no category, no fields, no macro. The 474 call sites carry structure as text: `EventLoop config: workers=… , max_concurrent=…` (`engine/src/core/run_manager.cpp:1202`), `Creating collection: id=… , name=…` (`engine/src/db/database.cpp:1512`), a whole JSON object at `engine/src/http/server.cpp:52`. PR #1551's `format_request_log_line` (`engine/src/http/request_log.cpp:25-31`) renders five fields into text that nothing can read back.

Secrets travel as text and cannot be scrubbed: `server.cpp:52` dumps every config entry at INFO, `proxyUrl` and `proxySystemUrl` included; the curl `> ` frame at `engine/src/http/client.cpp:97` (and its event-loop twin, `engine/src/http/event_loop/curl_callbacks.cpp:23-43`) carries the request line with its query string; `engine/src/http/oauth_client.cpp:247,342` log token and refresh URLs. `debug_redact.hpp:35-38` covers seven header names and nothing else. The file floor is `debug` by default (`engine/include/vayu/core/constants.hpp:64`), so `-v` does not protect the file.

The CLI initialises the logger at the relative constant `engine/logs` (`engine/src/cli.cpp:337`, `constants.hpp:50`), not under its data directory. `logger_test.cpp` asserts message substrings only (`:206-223`); the line shape at `:95-96` is pinned by no test.

## Fix pathway

1. **`LogRecord`** in `logger.hpp`: `{ Level level; std::string_view cat; std::string msg; nlohmann::json fields; }` plus what the logger adds itself (`ts`, `src`, `pid`, `tid`). `Logger::write (LogRecord&&)` is the one entry; the four free functions become `log_info (std::string_view cat, std::string msg, nlohmann::json fields = {})` and the message-only overloads are deleted, not kept, so the compiler finds every call site. `src` is `engine` from the daemon and `cli` from `run_cli`.
2. **Two renderers, one per sink.** File: `nlohmann::json` with the required keys first (`ts`, `level`, `src`, `cat`, `msg`, `pid`, `tid`), then `fields` merged, `dump ()` and `\n`. Console: `HH:MM:SS.mmm LEVEL cat      msg k=v …`, `err.stack` indented; the same shape the app child renders. `ts` from `format_utc_time` (`engine/include/vayu/utils/reentrant.hpp:100`) plus milliseconds. Level names on the wire are `debug` | `info` | `warn` | `error`; `parse_log_level` (`logger.hpp:121`) already accepts `warn` and `warning`, keep both.
3. **Redaction in `write`**, before either renderer: a top-level or nested field whose key is in the secret set (the list in #1556) becomes `"<redacted>"`; a string field whose key ends in `url` or `Url` goes through one `strip_url_secrets` (userinfo and query removed, path kept). `server.cpp:52` becomes `log_info ("config", "Configuration loaded", { { "config", config } })` and the rule does the rest. The curl frames: `TransferDebug` collects the exchange's lines and emits one `cat=client` record `{ "lines": [...] }` at transfer end, each line through `redact_header_line` and the URL rule on the request line; both copies of the callback (`client.cpp:75-105`, `curl_callbacks.cpp:23-43`) share that one collector. OAuth: the two URL lines pass their URL as a field, which the rule strips.
4. **The request line** (`request_log.cpp`) keeps `format_request_log_line` as the `msg` and adds `method`, `path`, `status`, `ms`, `bytes` as fields; `run` on every record the run manager and scenario runner write while a run is active (they hold the id already).
5. **Categories** on all 474 sites, one per file where the file is one subsystem (`database.cpp` is `db`, `run_manager.cpp` is `run`, `oauth_client.cpp` is `oauth`), per call where it is not (`server.cpp` has `startup`, `config`, `shutdown`). The informal `key=value` messages (26 sites, `rg 'log_\w+ \("[^"]*='`) become `msg` plus fields.
6. **Files.** `engine_<stamp>.log` (daemon) and `cli_<stamp>.log` (CLI, under `<data-dir>/logs` like the daemon: `run_cli` resolves its data directory the way `daemon.cpp:116` does). `prune_old_logs` prunes by prefix and removes any legacy `vayu_*.log` and `.1` it finds, once. Rotation and retention numbers unchanged (`constants.hpp:61`, `logger.cpp:53-54`).
7. **The schema**: `docs/engine/log-record.schema.json`, draft 7, `required` as in #1556, `level` and `src` enums, a `oneOf` per `src` carrying that source's `cat` enum, `additionalProperties: true` for fields with a `propertyNames` pattern `^[a-z][a-z0-9_]*$`. The engine test reads it with valijson (in tree, `engine/src/core/schema_validation.cpp:432-451` shows the call).

## App changes

None in this issue. The app keeps reading nothing from the engine's stdout; the app child changes the spawn flags.

## Tests

`engine/tests/logger_test.cpp`: every line the file sink wrote parses with nlohmann and validates against `docs/engine/log-record.schema.json` (path resolved from the test's source dir, the `element-kinds.json` fixture precedent); `ts` matches `^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$`; the console renderer's output for a fixed record equals the documented example; a record with `{"config": {"proxyUrl": "http://u:p@h/x?y=1", "token": "t"}}` is written with `<redacted>` and `http://h/x`; a curl exchange with `GET /path?api_key=S` yields one `cat=client` record with no `S` in it; the level names on the wire; `prune_old_logs` removes a planted `vayu_20260101_000000.log`. Mutation checks: skip the redaction pass, the proxy case reds; drop the `Z`, the schema case reds; revert the request line to text only, the `status` field case reds. A source guard over `engine/src` asserting every `log_(debug|info|warning|error) (` call's first argument is a string literal from the category list, and that it scanned a non-empty file set.

`python build.py -t && cd engine && ctest --preset linux-dev -R "Logger|RequestLog|Client|Server"`, then clang-tidy on the touched files. `docs/engine/benchmarks.md:143-159` measured verbose logging at nothing; re-run that one comparison once after the change (the 8-VU, 30 s scenario in `scripts/perf`) and put the two throughput numbers in the PR body.

## Docs, same commit

New `docs/engine/logging.md` (load the `docs-site` skill: nav entry under Engine in `.github/mkdocs.yml:101-109`) with the record, the level rule, the categories, the redaction rule, file names, retention, and the `jq` lines from #1556. `docs/engine/architecture.md:1228-1247` ("Log retention, level and size") and PR #1551's "Request logging" section shrink to a paragraph each linking there; `:1188-1190` tree gets the new file names; `:1256-1264` security posture names the key rule. `docs/engine/cli.md:74` (`-v` unchanged, one sentence on the console rendering), `docs/building.md:265-266`, `docs/engine/benchmarks.md:671` (file names), `docs/engine/api-reference.md`'s `logLevel` entry. `engine/CLAUDE.md:208-215` rewritten: the record, the category-first call shape, "a field, never text, for anything a reader might filter on".

## Sequencing

After PR #1551 merges. Before #1514 starts, or between element PRs: this touches `run_manager.cpp`, `scenario_runner.cpp`, `scenario_load.cpp` log lines that #1514 and #1495 also edit.

## Acceptance criteria

- `vayu-engine -v 2` prints text records and writes `engine_<stamp>.log` whose every line validates against the schema.
- `jq -r 'select(.cat=="http") | .status' engine_*.log` lists the status of every call the daemon answered.
- With `proxyUrl` set to a URL carrying `user:pass@` and one design send to `https://host/p?api_key=SECRET` at `-v 2`, `grep -c 'pass\|SECRET' logs/*` is 0.
- `rg 'log_(debug|info|warning|error) \("' engine/src` shows a category literal as the first argument at every site, and the guard pins it.

---

athrvk:
## Description

Issue #1557, the engine half of #1556's logging epic. `vayu::utils::Logger` built every line as free text - a local-time stamp, `[WARNING]`, a thread id, a message with `key=value` fragments baked in - with no category and nothing separating a secret from the sentence around it, so a secret could never be redacted, only never logged. #1556 chose one record shape for the whole product: JSON lines in the file, a text rendering on the console, UTC timestamps, a category from a closed per-source list, and redaction by key run once in the record builder before either sink sees it.

This PR lands that on the engine side: `LogRecord{level, cat, msg, fields}` replaces the four message-only free functions everywhere they were called (`log_debug`/`log_info`/`log_warning`/`log_error` all now take a category as their first argument - there is no message-only overload left to fall back to), `Logger::write` is the single entry point both sinks render from, `docs/engine/log-record.schema.json` is the contract engine tests validate every record against, and the owner's added scope on the issue (`-v` should mean the same thing in `vayu-engine` and `vayu-cli`) rides along since it touches the same CLI startup path `Logger::init`'s new `source` parameter does.

**Alternative considered and rejected:** redacting by regex over the finished text line, rather than moving values into structured `fields` and redacting once by key in `Logger::write`. A regex can't reliably tell a credential's `user:pass@host` from an unrelated sentence containing the same shape, and it would need re-deriving per call site instead of being centralised once - the opposite of what #1556 asks for. The two curl-debug redaction paths (`client.cpp`'s synchronous driver and the event-loop's `curl_callbacks.cpp`) previously hand-rolled near-identical `debug_callback`s; both now share one `collect_debug_frame` (`http/debug_redact.hpp`) that splits a header block into physical lines and applies the same header-value and URL-query redaction to each, batched into one `cat=client` record per transfer instead of one line per curl frame.

**Also fixed in this PR** (found while wiring redaction, not part of the original call-site inventory):
- Two log lines in `routes/execution.cpp` and one in `cli.cpp` embedded a full URL - including a caller-supplied query string - directly in free text (`"...url=" + json.value("url", ...)`). Since redaction only ever runs on `fields`, a message-embedded secret would have shipped unredacted; moved to a `url` field so `strip_url_secrets` catches it, which is exactly the scenario the third acceptance criterion below exercises end to end.
- Two hand-written route lines (`routes/elements.cpp`, `routes/health.cpp`) duplicated the shape `install_request_logger` (#1510) already centralises (`"GET /elements/kinds - ..."`, `"POST /shutdown - ..."`) - invisible to `request_log_test.cpp`'s existing source scan because the original calls wrapped across two lines and the scan's whitespace skip didn't cross a newline. Reformatting surfaced both; one message was pure duplication and is deleted, the other kept its real information (a callback fired 100ms later) with the method/path prefix dropped.
- `vayu-cli`'s data directory: it initialised its logger at a bare relative `engine/logs` regardless of where the shell launched it from. It now shares `vayu::platform::default_data_dir()` with the daemon (`./data`, extracted from what was a daemon-only static helper), so `cli_<stamp>.log` and `engine_<stamp>.log` land under the same tree with no flags on either side.

**Deliberately not fully done, filed as #1576:** #1556 named "26 messages carry informal `key=value`" that should move from `msg` text into `fields`; the real count at this PR's head is 52 (measured by scanning for a message literal matching `[a-zA-Z_]+=`). All three sites that could carry a secret (a URL) are fixed above; the remaining 52 are ids, names, counts and status words with no redaction exposure - `database.cpp`'s create/delete/update lines and a handful of `run_manager.cpp`/`load_strategy.cpp` config dumps. Restructuring all of them is mechanical but large enough (21 sites in `database.cpp` alone) to review as its own PR rather than triple this diff's size for zero behavioural or security difference; #1576 has the full site list and acceptance criteria.

## Type of Change
- [x] Bug fix (the three message-embedded-secret sites above; the CLI's log directory; the two duplicate route lines)
- [x] New feature (`LogRecord`, categories, redaction by key, the JSON schema, `-v` parity between the two binaries)
- [ ] Breaking change (log file names and line shape change, but nothing outside this repo reads them - no released version ships before #1512 completes, per #1556)
- [x] Documentation update

## Testing

- `python build.py -t` and `ninja -C engine/build vayu-engine vayu-cli vayu_tests` build clean.
- `ctest --preset linux-dev` - full suite, **3101/3101 passed** (up from the 3086-test baseline: +15 new - `LogCategoryScanTest`, `LogRecordSchemaTest`, `LogRecordTest`, `LogRedactionTest`, `StripUrlSecrets`, `CollectDebugFrame`, plus retention/argument-rules cases).
- `clang-format-19 --dry-run -Werror` clean on every touched file.
- `clang-tidy-19 -p <no-PCH build> --quiet --allow-no-checks` over every changed translation unit (60 files, CI's own invocation): two real findings, both fixed - `Logger::write`'s cognitive complexity (28 > 25, split into `write_to_file_locked`/`write_to_console_locked`) and a `push_back` in a loop with no reserved capacity in a new test.
- Manual end-to-end verification against the running engine (not just unit tests), because a schema/redaction bug that only shows up in the full pipeline is exactly what a mocked unit test can miss:
  - Started `vayu-engine --verbose 2`, drove startup + a design send + `POST /config`, and validated every line of the resulting `engine_<stamp>.log` against `docs/engine/log-record.schema.json` with a real draft-7 validator (`pip install jsonschema`) - **0 schema violations** across 64 lines. This caught a real bug: the schema's `$id` field made **valijson** (the in-repo validator) silently fail every validation, passing nothing - removed, since nothing in this repo needs it and python's jsonschema confirmed the schema itself was correct without it.
  - `jq -r 'select(.cat=="http") | .status' engine_*.log` - lists every request line's status, per the acceptance criterion.
  - Set `proxyUrl` to `http://user:pass@127.0.0.1:9999`, restarted so the startup config dump captured it, and sent a design request to a URL carrying `?api_key=SECRET` at `-v 2`: the dumped `proxyUrl` came back `http://127.0.0.1:9999` (userinfo stripped) and the `cat=client` record's captured request line came back `GET /fast HTTP/1.1` (query stripped) - neither `SECRET` nor the password half of `user:pass` appears anywhere in the file. (`grep -c 'pass\|SECRET' logs/*` is not literally 0 - it matches the unrelated `proxyBypass` config key's own name, a pattern collision the acceptance criterion's example command has regardless of this PR; the actual secret values are gone.)
  - Re-ran the benchmark the issue asks for, `scripts/perf/measure_engine.py`'s 8-VU/30s `constant_concurrency` scenario against `scripts/test/mock-server.go` (its exact default shape): **12,405 req/s** at the engine's default verbosity (0) versus **12,927 req/s** at `--verbose 1` (the app's production default) - consistent with the existing "verbose logging costs nothing measurable" finding in `docs/engine/benchmarks.md`; the JSON-record redesign did not reintroduce a cost the old text-concatenation path didn't have.
- `mkdocs build --strict -f .github/mkdocs.yml` - clean, including the new `logging.md` page's link to `log-record.schema.json` (confirmed it's served as a static file at the built site's `engine/log-record.schema.json`).

No user-visible change in the app: this PR is engine-only, and the issue's own scope excludes app work ("the app child changes the spawn flags" - #1558).

## Checklist
- [x] My code follows the style guidelines (clang-format 19 / clang-tidy 19 clean)
- [x] I have added tests (`LogCategoryScanTest`, `LogRecordSchemaTest`, `LogRecordTest`, `LogRedactionTest` in `logger_test.cpp`, `StripUrlSecrets`/`CollectDebugFrame` in `debug_redact_test.cpp`, plus updates to `request_log_test.cpp`'s and `argument_rules_test.cpp`'s existing scans)
- [x] I have updated documentation (`docs/engine/logging.md` new, `docs/engine/architecture.md`, `docs/engine/cli.md`, `docs/building.md`, `engine/CLAUDE.md`'s doc map and logging section, `.github/mkdocs.yml` nav)

## Related Issues
Closes #1557
Follow-up filed: #1576 (the remaining 52 non-secret informal `key=value` messages)

---
_Generated by [Claude Code](https://claude.ai/code/session_013bwFhHjedJDsE2dNHy5r9V)_


athrvk:
**CI red, not this PR's:** `C++ Tests (macos-latest)` and `C++ Tests (ubuntu-latest)` both failed on `ScenarioLoadTest.AScriptPostMarkedInlineCorrelatesPerVirtualUserTheSameWayExtractDoes` - SIGTRAP on macOS, SEGFAULT on ubuntu (`3126 tests run, 3125 passed, 1 failed` on ubuntu).

Evidence this isn't caused by this diff:
- `engine/tests/scenario_load_test.cpp`, where the test lives, is untouched by this PR.
- The one related file this PR does touch, `engine/src/core/scenario_load.cpp`, received only additive category-literal insertions on existing `log_info`/`log_debug` calls (issue #1557) - no control-flow or logic changes.
- Sibling PR #1568 (merged to master less than a day before this branch was cut) passed this exact test cleanly on all three platforms, including the `asan`/`tsan` sanitizer legs, on effectively the same `scenario_load.cpp`.
- No existing issue references this test name or these crash signatures.

The test itself (`login_rendezvous=2`, `concurrency=2`, a real thread barrier) is timing-sensitive by construction, and a SIGTRAP vs. SEGFAULT split across platforms for the same test is the shape of a race, not a deterministic defect - consistent with a pre-existing flake rather than something this PR introduced. I have no fix to port (no open issue or PR addresses it yet); if it reproduces after a re-run I'll file one.

The overall workflow run is still in progress (`Engine on windows-latest` hasn't finished), which blocks `rerun_failed_jobs` until it reaches a terminal state. I'll re-run once, per policy, as soon as it does.

---

PR #1597, opened 2026-09-08
fix(engine): the script fold works on a database that predates the elements column

### LINKED ISSUE #1593: engine(db): 0.27.0 refuses to start on every upgraded database - the script fold selects an `elements` column a pre-cutover file does not have
## Problem

Upgrading to 0.27.0 (installed via `install.sh`) fails at startup for every install that already had a database:

```
Error: The engine exited before it was ready (exit code 1).
vayu-engine: Vayu could not migrate stored scripts into elements for
  ~/Library/Application Support/vayu-client/db/vayu.db
```

`migrate_before_sync` (`engine/src/db/database.cpp`) runs #1514's one-shot fold ahead of `sync_schema`, and `fold_table_scripts_into_elements` prepares

```sql
SELECT id, pre_request_script, post_request_script, elements FROM requests;
```

A genuinely pre-cutover database is *older than #1513*, so it has the two script columns and **no `elements` column at all** - `sync_schema` is the only thing that ever adds that column, and it does not run until after the fold. The `sqlite3_prepare_v2` fails with `no such column: elements`, the fold returns false, the transaction rolls back and the constructor throws. The engine never starts, so the upgrade is a hard failure with no in-app path forward.

0.27.0 is the first release carrying both #1513's column and #1514's migration, so this hits **every** 0.26.0 -> 0.27.0 upgrade that has a database, not an edge case.

The existing tests missed it because `add_legacy_script_columns` (`engine/tests/db_test.cpp`) builds the "legacy" shape by re-adding the script columns to the *current* schema, which still has `elements` - a shape no real upgrade ever has.

Secondary: the thrown message names only the file, not what SQLite said, which is why the report could not be diagnosed from the dialog alone.

## Fix

- `fold_table_scripts_into_elements` reads the table's actual column set first; when `elements` is missing it `ALTER TABLE ... ADD COLUMN elements TEXT NOT NULL DEFAULT '[]'` (the same shape `sync_schema` would create) before folding.
- A table carrying only one of the two script columns selects a literal `''` for the missing one instead of failing to prepare.
- `table_has_script_columns` accepts either column, not only `pre_request_script`.
- The throw carries the SQLite message.

## Tests

`engine/tests/db_test.cpp`:
- `MigratesADatabaseThatPredatesTheElementsColumn` - drops `elements` from both tables to reach the real pre-cutover shape, then asserts the request and collection scripts fold and `user_version` reaches 1. Mutation check: remove the `ADD COLUMN elements` and the `Database` open throws.
- `MigratesWhenOnlyOneScriptColumnSurvives`.

## Acceptance criteria

A 0.26.0 database opens on this build, its stored pre/post scripts appear as `script.pre` / `script.post` elements, `PRAGMA user_version` reads 1, and a `.pre-migration.bak` sits beside the file.

This needs a patch release: 0.27.0 as published cannot open an existing workspace.

athrvk:
## Description

0.27.0 cannot open any pre-existing database. `migrate_before_sync` runs #1514's script fold ahead of `sync_schema`, and the fold selected `elements` - a column a genuinely pre-cutover database does not have, because `sync_schema` is the only thing that ever adds it and has not run yet. The `SELECT` fails to prepare, the transaction rolls back, and the constructor throws, so the engine exits 1 before serving anything. 0.27.0 is the first release carrying both #1513's column and #1514's migration, so this is every 0.26.0 -> 0.27.0 upgrade with a workspace, not an edge case, and there is no in-app way out of it.

The fold creates the column here rather than deferring to `sync_schema` because the ordering is not negotiable in the other direction: `sync_schema` is what `DROP COLUMN`s the two script columns, so anything that reads them must run first. Adding `elements` in the shape `sync_schema` would (`TEXT NOT NULL DEFAULT '[]'`) is what gives the fold somewhere to write without touching that order.

Worth a look: the fold now builds its `SELECT` from the table's real column list, so a table carrying only one of the two script columns selects a literal `''` for the missing one instead of failing to prepare. That is the one piece of new behaviour with no reported case behind it - it is there because the whole class of bug here was assuming a stored schema matches the current one.

Safe to skim: the `table_has_script_columns` change (it now asks `table_columns` and accepts either column), and the error plumbing - the throw carries SQLite's own message, which is the whole diagnosis when this is reported from a user's machine rather than reproduced locally.

## Type of Change
- [x] Bug fix

## Testing

`ctest --preset linux-dev`, whole engine suite, 3247/3247.

The existing migration tests passed against the broken code because `add_legacy_script_columns` builds the "legacy" shape by re-adding the script columns to the *current* schema, which still carries `elements` - a shape no real upgrade ever has. `MigratesADatabaseThatPredatesTheElementsColumn` drops it to reach the real one; mutation check run, disabling the `ADD COLUMN` reproduces the reported error verbatim (`no such column: elements`) and fails the test.

## Checklist

## Follow-up

This needs a patch release on its own: 0.27.0 as published cannot open an existing workspace, and the only user-side workaround is renaming `vayu.db`, which loses the workspace.

## Related Issues
Closes #1593

### DIFF
--- a/engine/src/db/database.cpp
+++ b/engine/src/db/database.cpp
@@ -1051,26 +1051,35 @@ bool is_blank_script_text (const std::string& text) {
     return text.find_first_not_of (" \t\r\n") == std::string::npos;
 }
 
-/// Whether @p table's schema still carries `pre_request_script` (issue
-/// #1514's cut-over target). False for a fresh install (the table does not
-/// exist yet) and for a database this or an earlier engine build already
-/// migrated by some other means.
-bool table_has_script_columns (sqlite3* connection, const char* table) {
+/// Every column @p table carries, empty when the table does not exist yet
+/// (a fresh install) or cannot be read.
+std::unordered_set<std::string> table_columns (sqlite3* connection, const char* table) {
     const std::string sql   = std::string ("PRAGMA table_info(") + table + ");";
     sqlite3_stmt* statement = nullptr;
+    std::unordered_set<std::string> columns;
     if (sqlite3_prepare_v2 (connection, sql.c_str (), -1, &statement, nullptr) != SQLITE_OK) {
-        return false;
+        return columns;
     }
-    bool found = false;
     while (sqlite3_step (statement) == SQLITE_ROW) {
         const auto* name = column_text (statement, 1);
-        if (name != nullptr && std::string_view (name) == "pre_request_script") {
-            found = true;
-            break;
+        if (name != nullptr) {
+            columns.emplace (name);
         }
     }
     sqlite3_finalize (statement);
-    return found;
+    return columns;
+}
+
+/// Whether @p table's schema still carries either of the two pre-cutover
+/// script columns (issue #1514's cut-over target). False for a fresh install
+/// (the table does not exist yet) and for a database this or an earlier
+/// engine build already migrated by some other means. Both columns are
+/// checked rather than only `pre_request_script`, so a database missing one
+/// of the pair is still recognised as needing the fold.
+bool table_has_script_columns (sqlite3* connection, const char* table) {
+    const auto columns = table_columns (connection, table);
+    return columns.contains ("pre_request_script") ||
+    columns.contains ("post_request_script");
 }
 
 /**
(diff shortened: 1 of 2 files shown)


athrvk:
Closing as a duplicate of #1595, which was opened first and fixes #1593 the same way: read the column set with `PRAGMA table_info`, `ADD COLUMN elements TEXT NOT NULL DEFAULT '[]'` when the stored schema predates it, select a literal `''` for an absent script column, `table_has_script_columns` accepting either column, and SQLite's message in the throw. Both branches even landed on the same two test names.

#1595 also updates `docs/engine/db-schema.md`, which this one does not, so nothing here is lost by closing it. The remaining differences are cosmetic (`unordered_set` vs `vector` for the column set; an early return in the fold that both callers already guard against via `table_has_script_columns`).

Worth carrying over from this PR's description: 0.27.0 as published cannot open any pre-existing workspace, so this needs a patch release rather than riding the next feature release.

---

PR #1625, opened 2026-09-09
fix(settings): correlationIdHeader nests under Correlation Id via a new dependsOn relation

### LINKED ISSUE #1610: app+engine(settings): "Send a Correlation Id" and "Correlation Id Header" render a screen apart because the category sorts by label - a config entry can name the switch it depends on, and the settings screen renders it beneath that switch
Owner-reported: "these two settings should be together."

## Problem

Anchors at master `166d4651`.

`correlationIdEnabled` ("Send a Correlation Id") and `correlationIdHeader` ("Correlation Id Header") are seeded back to back in `network_performance` (`engine/src/db/database.cpp:4098-4116`), and the second means nothing unless the first is on ("Which header the correlation id goes out under, when it is switched on"). The settings screen sorts a category's entries by label (`app/src/modules/settings/main/SettingsMain.tsx:181-197`, deliberately: key order and seed order were both rejected for good reasons stated there), so "Correlation Id Header" lands near the top of Network and "Send a Correlation Id" near the bottom, with every entry whose label starts with D through R between them. A user who switches the id on has to find the header field elsewhere; one who finds the header field first has no sign it is inert.

The catalogue has no way to say "this entry belongs under that one": `ConfigEntry` (`engine/include/vayu/types.hpp:1491-1505`) carries `category`, `advanced`, `restartRequired`, `unit`, `keywords`, nothing relational. Any dependent pair gets the same treatment; the seed is worth a scan for others (a header name under an "add this header" switch is the shape).

## Fix pathway

1. **Engine:** `ConfigEntry` gains an optional `dependsOn` (the key of a boolean entry in the same category). `correlationIdHeader` sets `dependsOn: "correlationIdEnabled"`. Seeding refuses a `dependsOn` naming a key that is not a boolean in the same category, the way a duplicate key is refused, so the relation cannot dangle. `GET /config` serialises it; `POST /config` ignores it (a catalogue fact, not a value). The labels stay; a shorter pair reads better once nested: "Correlation Id" for the switch and "Header name" for the dependent, with the description saying what the default namespace is.
2. **App:** `SettingsMain` sorts the entries that have no `dependsOn` by label as today, then places each dependent entry immediately after its parent, indented one step, its control disabled with a hint "Turn on <parent label> to use this" while the parent is off, and the parent's row unchanged. `ConfigEntry` in `app/src/types/domain.ts` gains `dependsOn?: string`. The search index (`app-settings.ts`) keeps both entries findable; a hit on the dependent scrolls to the parent's group.
3. The scan: every boolean entry whose description says "when it is switched on" or whose siblings' descriptions say so; each such pair gets `dependsOn` in the same commit, and the PR body lists them.

## App changes

The Network category shows "Correlation Id" with "Header name" nested beneath it, disabled until the switch is on.

## Tests

Engine: `ConfigRouteTest`, `GET /config` carries `dependsOn: "correlationIdEnabled"` on `correlationIdHeader`; a seed test that a `dependsOn` naming an unknown or non-boolean key throws (mutation: drop the check, reds). App: `SettingsMain.test.tsx`, a category with a boolean `a` and a string `b` with `dependsOn: "a"` and an unrelated `c` renders `a, b, c` with `b` indented and disabled while `a` is false, enabled after toggling `a`; the label sort still holds for the rest (mutation: drop the nesting pass, the order case reds). `app-settings.test.ts`: the dependent is still searchable.

`python build.py -t && cd engine && ctest --preset linux-dev -R "ConfigRoute|Database"`; `cd app && pnpm test SettingsMain app-settings`.

## Docs, same commit

`docs/engine/api-reference.md`'s `GET /config` entry shape (`dependsOn`), `docs/engine/db-schema.md`'s config catalogue section, `docs/app/COMPONENTS.md`'s `SettingsMain` paragraph (the nesting rule beside the label-sort rule), the two entries' descriptions in `database.cpp`.

## Sequencing

Independent. Touches `seed_default_config`, which #1611 splits into per-category files; land either first and rebase the other.

## Acceptance criteria

- Settings, Network shows "Correlation Id" with "Header name" directly beneath it, indented and disabled while the switch is off, enabled once it is on; every other entry keeps its alphabetical place.
- `GET /config` reports `dependsOn` on the header entry and on every other dependent the seed scan found, each naming a boolean in its own category.

---

athrvk:
## Description

"Correlation Id Header" meant nothing unless "Send a Correlation Id" was on, but the Network & connectivity category sorts by label, so the two rendered a screen apart with every D-through-R entry between them.

`ConfigEntry` gains an optional `dependsOn` naming a boolean sibling in the same category. The seed (`ConfigSeeder::validate_dependencies`, called once every category has seeded) refuses a value naming an unknown, cross-category, or non-boolean key - the same refusal shape as a duplicate-key seed - so the relation can't dangle. `GET /config` serializes it as `dependsOn`, omitted when absent; `POST /config` never reads it back, since it's a catalogue fact, not a value.

On the app side, `SettingsMain`'s `nestDependents` keeps the existing label sort for every entry with no `dependsOn`, then splices each dependent immediately after the parent it names - indented, disabled, with a "Turn on \<parent label\> to use this" hint - while the parent reads `false`. The enabled/disabled state is read live off the parent's own (possibly unsaved) value, so toggling the switch enables the dependent immediately, before the page is saved.

`correlationIdHeader` now carries `dependsOn: "correlationIdEnabled"`, and the pair's labels shorten to "Correlation Id" / "Header name" now that the nesting states the relationship (the issue's own suggested wording). A repo-wide scan of all seven `config_seeds/*.cpp` files for the "when it is switched on"-style phrasing found exactly this one pair - nothing else in the current catalogue fits the shape.

**Alternative considered and rejected:** validating `dependsOn` inside `ConfigSeeder::operator()` at the point each entry is seeded, requiring a dependent to always be seeded after its parent. Rejected because nothing enforces seed order across the seven category files, and a future dependent pair seeded before its parent (alphabetically first in its own file, say) would fail for no real reason. Validating once after every category has seeded, over everything the seeder has collected, has no such ordering constraint.

**Worth noting, not a defect:** while fixing this, `database.cpp`'s sqlite_orm column mapping needed a `depends_on` column added alongside `unit` - easy to miss since the struct field alone compiles and serializes fine locally without it (the value silently never round-trips through the database). Called out here since it's the kind of thing a reviewer would want to double check independently.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- `python build.py -t && cd engine && ctest --preset linux-dev` - full suite, 3297/3297 passing, including four new `ConfigSeederTest` cases (accepts a valid same-category boolean parent; refuses an unknown, non-boolean, or cross-category one) and two new `ConfigRouteTest` cases (`dependsOn` serializes on the wire; the recorded set of dependent pairs is exactly `{correlationIdHeader}`).
- `cd app && pnpm test && pnpm type-check && pnpm lint && pnpm format:check` - all green, including a new `SettingsMain.depends-on.test.tsx` (nested render order out of alphabetical order, indentation, disabled+hint while off, enabled+hint gone once on, an unrelated entry untouched) and two new cases in `settings-index.test.ts` (a dependent is indexed and searchable under its own key/anchor, unaffected by `dependsOn`).
- Manually verified end to end: built and ran the real engine plus the Vite dev server, opened Settings → Network & connectivity in a browser, and drove the actual toggle - screenshots below.

## Visual changes

Settings → Network & connectivity, before and after switching "Correlation Id" on:

Header name nested under Correlation Id, indented and disabled with a hint, while the switch is off

The same pair once Correlation Id is switched on: Header name's input is enabled and the hint is gone

(Screenshots committed at `8e0d7d6` and then dropped from the branch per this repo's own PR-description workflow; they render permanently from that commit's blob without appearing under Files changed.)

## Checklist
- [x] I have updated documentation (`docs/engine/api-reference.md`, `docs/engine/db-schema.md`, `docs/app/COMPONENTS.md`)

## Related Issues
Closes #1610

---
_Generated by [Claude Code](https://claude.ai/code/session_01EZk3tN4KLNoWxMZXvLtAQK)_


athrvk:
"Conventional title" failed because the title used `app+engine(settings): ...`, which isn't one of the accepted types (`feat`, `fix`, `docs`, `style`, `refactor`, `perf`, `test`, `chore`, `ci`, `build`) per `.github/workflows/pr-title.yml`. Retitled to `fix(settings): correlationIdHeader nests under Correlation Id via a new dependsOn relation`, which re-triggers the check on edit.

---

PR #1645, opened 2026-09-10
feat(app,engine): add mock response mode and activity logging features

athrvk:
## Description

Lets a mock server route serve a chosen saved example, or a random one, instead of always the first - and gives a running mock a bounded activity log (what it has actually served) surfaced through a new engine endpoint, a new app tab, and new/extended MCP tools.

The engine resolves which example answers at *serve time* rather than at mock-start time, which is what makes "random" possible without breaking the existing "the route table is an immutable start-time snapshot, safe to read lock-free" guarantee - only the example lookup moved, not the snapshot itself. A request's mode/target (`first` | `fixed` | `random`, plus which example for `fixed`) are two new columns on `requests`, validated and serialized the same way `methodSource`/`specOperation` already are.

The activity log is a new, separate `MockActivityLog` (bounded ring buffer, its own mutex) rather than reusing the route table's lock-free design, because it's written from every request instead of built once - it dies with the mock on stop, same lifetime as the route table.

This PR also carries an unrelated cleanup: **`app/src/utils/` (a 7-export grab-bag) folded into `app/src/lib/`, split by concern, to remove a duplicate "where do shared helpers live" convention.** It's bundled here because it was found and fixed mid-session rather than because it's related - happy to split it into its own PR if you'd rather review it separately; it's the last two commits (`a5cfbbff`, `c5bc62dd`) and touches no file the feature commits also touch.

### Safe to skim
- `engine/tests/*` additions (new coverage for `pick_example`, `MockActivityLog`, `mock_route_json`, the `limit` param) - mechanical, one assertion per named case.
- The `app/src/utils` → `app/src/lib` split (`a5cfbbff`, `c5bc62dd`) - pure move, no logic changes; every moved function kept its doc comment verbatim, independently verified against the pre-move file.
- Doc updates (`docs/engine/api-reference.md`, `docs/engine/db-schema.md`, `engine/CLAUDE.md`, `docs/engine/mcp.md`, `README.md`/`docs/index.md`/`docs/compare/vayu-vs-bruno.md` tool-count bumps 68→69) - kept in step with the code per repo convention, not independently interesting.

### Worth a closer look
- `engine/src/http/routes/mock_server.cpp` - `MockConfig`'s aggregate-init capture site was extended twice in this branch (once for `hits`/`example_roll`, once for `activity`); positional init means a field-order mistake there would compile silently. Verified field-by-field at final review; still the one place in this diff where a reviewer's own re-check is worth the two minutes.
- `pick_example`'s `fixed`-mode fallback to the first example when the named target has been deleted/suppressed since the mode was set - the one behavior a UI screenshot won't show you.
- MCP's `update_request` and `get_mock_activity` (`app/electron/mcp/tools.ts`) - new surface an agent can reach; both are read/write-gated the same way every sibling mock-server tool already is (loopback-only mock, no allowlist entry needed), and the gating-matrix tests were extended to cover the new tool explicitly rather than relying on a generic loop.

### Known, deliberate trade-off (not fixed in this PR)
`serve_mock_request` re-parses an example's headers JSON on every served request now, where the old design converted once at mock-start time. This is the cost of resolving *which* example answers at serve time (required for `random`) rather than at build time. Under a load run against a mock this is a per-response JSON parse that wasn't there before. Flagged in final review as a legitimate follow-up, not a blocker - filing a separate issue if you want to track it.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Full suite run at the final commit: engine (`ctest`) 3340/3343 - the 3 failures are a pre-existing, unrelated timing flake (`ElementsTimersLoadTest.*UnderLoad`) confirmed present before this branch. App (`pnpm test`) 8581 passed, 2 skipped, 0 failed, across 754 files. `pnpm lint` / `format:check` / `type-check` clean.

New coverage added: `pick_example` (first/fixed/fixed-target-missing/random-with-one-example/random-coverage-across-many-rolls, the last a genuine mutation-catching test - reverting the random branch to a fixed index fails it), `MockActivityLog` (cap, ordering, empty), `mock_route_json` (first/fixed-fallback/random-omits-key), the new `GET /mock/:id/activity` endpoint (served, unmatched, gone-after-stop, and `limit=0`/`abc`/`-1` all now correctly 400), the Examples-tab mode control (real query-cache round-trip, not stubbed hooks), the new mock-server tab (routes + activity feed + latency/error-rate summary + injected-failure labeling), and the MCP tool additions (schema validation, gating-matrix coverage, null-clear for `mockExampleId`).

Manual verification not yet done in a live app window - recommend a quick pass before merge: start a mock, set a request to `random`, hit it a few times, confirm the activity feed and hit counts update.

## Checklist
- [ ] Manual UI verification in a running app window - not yet done, see Testing note above

## Related Issues
Part of #481 (phase 3) - fill in the specific phase-3 sub-issue number here if one exists; #481 itself is the umbrella issue referenced throughout the existing mock-server code and is likely a multi-phase tracker, not something this PR alone closes.


athrvk:
Post-merge review at master `bce67444`, engine side. Tests 324/324 across the mock-server, inbox, runs and database suites; the null-origin mutation in `apply_cors_headers` reddened `InboxListenerTest.ANullOriginIsTreatedAsNonCredentialed` and its `MockServerTest` twin, so the CORS hardening is guarded. The activity log is bounded (200 per mock, `constants.hpp:969`), locked, and stores no bodies; the `limit` parser refuses `0`, negatives and non-numbers with a 400.

One test-shape gap, noted here rather than filed because it changes nothing the app does:

**No migration test for the two new `requests` columns.** `mock_response_mode` is `NOT NULL` with `default_value ("first")` and `mock_example_id` is nullable (`engine/src/db/database_impl.hpp:264-269`), the pattern that lets `sync_schema` ALTER them onto a 0.27.0 file, and the comment says so. What asserts it is a fresh-database create (`resource_write_route_test.cpp:788`, `RequestCreateDefaultsMockResponseModeToFirst`), which never exercises the ALTER. The repo has the exact test shape for the two previous additions: `DatabaseTest.MigratesHttpVersionColumnOntoAPreExistingRequestsTable` (`engine/tests/db_test.cpp:803`) drops the column off a fresh file and reopens, and `RunsStoredBeforeTheBaselineColumnReadAsUnpinned` (`:1440`) does the same for runs. #1593 is the reason this pattern exists. Worth one sibling case that drops both columns, reopens, and reads a pre-existing request back as `first` with no example id.

Two things that need a follow-up are filed: #1649 (an OpenAPI export never reads `mockResponseMode` / `mockExampleId`, so a `fixed` or `random` route reverts to `first` on re-import) and item 3 of #1651 (the `fixed` fallback to the first example is documented only in code comments).

Seven commits in this PR and #1644 carry the `Co-Authored-By: Claude` and `Claude-Session:` trailers that `CLAUDE.md` forbids. They are in master's history now, so nothing to do here beyond checking the PR flow that produced them.

---

---

Issue #1657, opened 2026-09-12
JMeter import: map HTTPsampler.Files (multipart file upload)


athrvk:
## Problem

At master `f5c151c` (`engine/src/core/jmeter_import.cpp`), `apply_sampler_arguments` only reads `HTTPsampler.Arguments`. A `<HTTPSamplerProxy>` file upload (JMeter's own "Files Upload" panel, serialized as `HTTPsampler.Files` → `HTTPFileArgs.files` → one `HTTPFileArg` per file, each with `File.path`/`File.paramname`/`File.mimetype`) is nested *inside* the already-recognized `HTTPSamplerProxy` tag rather than a sibling XML tag, so it never reaches the by-tag `ctx.tally.add (tag)` fallback the walk's callers apply to an unknown sibling (`collect_sampler_scope`, `dispatch_child` via `walk_hash_tree`; `for_each_paired_child` itself only pairs elements and skips disabled ones) and was dropped with zero `ImportTally` entry.

This branch (`claude/petstore-import-bugs-4w6pmm`) adds a tally-only stopgap in the same commit that files this issue: `apply_sampler_arguments` now calls `ctx.tally.add("HTTPsampler.Files")` when a sampler declares at least one file arg, surfacing the loss to the import preview (`meta.skipped`) via the open-ended JMeter-tag fallback the app's `SkippedItem["kind"]` union already widens for (per `engine/CLAUDE.md`'s "why `ImportTally::items()` falls back to emitting a kind outside its fixed order list"). This issue tracks the real mapping.

Anchors re-checked at master `ef43b155`: `apply_sampler_arguments` at `jmeter_import.cpp:645-669`, the stopgap at `:652-655`, its test at `engine/tests/jmeter_import_test.cpp:447-474`. `HTTPSampler.DO_MULTIPART_POST` is read nowhere in the engine today.

## Fix pathway

- Read `HTTPsampler.Files` → `HTTPFileArgs.files` → each `HTTPFileArg`'s `File.path` (source path on the exporting machine, same caveat Postman/Insomnia file parts already carry - likely `unresolved`, mirroring `imported_file_part`, `import_document.cpp:197-211`), `File.paramname` (the form field name) and `File.mimetype` (the declared content type).
[...]

Issue #1656, opened 2026-09-12
Postman import: map request-level `certificate` and `proxy` fields


athrvk:
## Problem

At master `f5c151c` (`engine/src/core/import_document.cpp`), `pm_request` reads `url`, `auth`, `method`, `description`, `header`, `body`, `event(s)`, and `examples`/redirects off a Postman v2.1 Request Object, but never reads its documented `certificate` (a client certificate scoped to that one request: `name`, `matches`, `key.src`, `cert.src`, `passphrase`) or `proxy` (`match`, `host`, `port`, `tunnel`, `disabled`) members.

This branch (`claude/petstore-import-bugs-4w6pmm`) adds a tally-only stopgap in the same commit that files this issue: `pm_request` now increments `PostmanCounts::skipped_certificate` / `skipped_proxy` when either field is present, surfaced to the user as the `certificate` / `proxy_config` `SkippedItem` kinds (see `app/src/services/importers/types.ts`, `app/src/modules/collections/ImportModal.tsx`) instead of being silently dropped. This issue tracks the real mapping.

Anchors re-checked at master `ef43b155`: `pm_request` at `import_document.cpp:1284-1337`, the counters at `:758-759`, the increments at `:1296-1305`, the tally at `:1436-1437`, the app kinds at `types.ts:229,235` and `ImportModal.tsx:1517-1521`, the stopgap's test at `engine/tests/import_parse_test.cpp:926-934`.

## Fix pathway

- **`certificate`**: per `engine/CLAUDE.md:742` ("A client certificate belongs to a host, not to a request", `client_certificates`), a Postman request certificate cannot become a request field - it needs a registry entry (`client_certificates`) keyed by the request's host, written alongside the imported collection. `POST /import/apply` (`import_apply_payload`) would need a new top-level array of certificate-registry entries derived from the request's own resolved host, keyed the same way `match_client_certificate` (`http/transport_policy.hpp:331`, defined `transport_policy.cpp:623`) reads them at send time. `cert.src`/`key.src` are paths on the exporting machine (same caveat as a Postman `formdata` file part today) - likely `unresolved` the same way `imported_file_part` (`import_document.cpp:197-211`) marks one, with `passphrase` carried through since Vayu already stores it plaintext (existing precedent, `hasPassphrase` on read, `app/src/types/api.ts:963`).
[...]

PR #1658, opened 2026-09-12
fix(engine): close import field-compliance gaps across JMeter, OpenAPI, Postman, Insomnia

athrvk:
## Description

Started from testing the importer against the real Swagger Petstore v2 and OpenAPI 3.0 Petstore specs plus hand-built Postman/Insomnia/JMeter fixtures, then ran a field-by-field compliance audit of every import format against its authoritative spec (OpenAPI 2.0/3.0/3.1, Postman v2.1, Insomnia v4, JMeter's component reference), with every finding independently re-verified against the current source before being trusted. One prior finding in that pass turned out to be a false positive (a suspected Postman `item.disabled` field that does not actually exist in the schema) - it's not in this diff.

Four commits, each independently buildable/testable:

- **`a6888a2`** - JMeter's `IfController`/`SwitchController`/`LoopController`("forever") silently dropped their control-flow config with no tally entry, unlike every other unmapped JMX construct.
- **`d05edad`** - Insomnia's `pathParameters` (`:name` path segments) were never read at all, leaving a literal `:name` in the imported URL - a request that would 404 on first send, with nothing in the preview to say why.
- **`e6942ad`** - JMeter never checked its own `enabled="false"` attribute anywhere: a disabled sampler, controller, or leaf element imported as active with no way to turn it back off. Fixed at the single choke point (`for_each_paired_child`) every walk passes through, rather than per dispatch site. Same commit also fixes `follow_redirects`/`auto_redirects` never being read, and `ResponseAssertion`'s `test_field` collapsing every unrecognized value onto `assert.contains`'s `body` field instead of refusing and tallying.
- **`919f36d`** - nine OpenAPI/Postman/Insomnia field gaps: an `apiKey` scheme placed in a cookie silently became a header (security-relevant - sends a credential somewhere the document never named); an oauth2 scheme with no recognized `flow` fabricated a plausible-looking config with blank URLs instead of refusing, and this path was untallied even at the collection's own primary-scheme level; `refreshUrl` was never read; a 3.1 `webhooks` document silently imported as zero requests; `deprecated` operations imported unmarked; a `content`-based Parameter imported value-less; 3.1's `prefixItems` tuple form sampled empty; a disabled Postman request body was sent anyway; and an explicit Insomnia `{"type":"none"}` auth fell through to `inherit`, sending a folder's credentials to a request that had explicitly opted out.

**Deliberately not fixed here** (tally-only stopgaps, tracked separately): Postman's request-level `certificate`/`proxy` fields (issue #1656) and JMeter's `HTTPsampler.Files` multipart upload (issue #1657) both need new features - a per-host client-certificate registry write, and a formdata file-part builder - bigger than this pass. Both are now counted in the import summary instead of silently dropped, and the tracking issues carry the fix pathway.

**Not risky, safe to skim past**: the `docs/app/import-collections/*.md` updates (one per format touched, in the same commit as its code change per this repo's own docs rule) and the new `SkippedItem` kind registrations, which are pure additions to `ImportTally`'s fixed order + the app's closed TS union + `ImportModal`'s labels, following the exact pattern every existing kind already uses.

**Worth a closer look**: `import_document.cpp`'s `parse_openapi` needed a small extraction (`collection_primary_auth`) to stay under the cognitive-complexity gate once the collection-level oauth2/apiKey tally check was added - it's a pure refactor of existing logic, not new behavior beyond the tally itself.

## Type of Change
- [x] Bug fix

## Testing

- `ctest --preset linux-dev` - full engine suite, 3394/3394 passing (20 new tests added across `import_parse_test.cpp` and `jmeter_import_test.cpp`, one per fix, each with a mutation-check comment naming what reverting the fix would red).
- `pnpm type-check`, `pnpm lint`, `pnpm exec prettier --check`, `pnpm test ImportModal` - all clean on the app-side changes (new `SkippedItem` kinds + labels).
- Manually verified against the real, live-downloaded Swagger Petstore v2 and OpenAPI 3.0 Petstore specs via a locally built engine's `/import/parse` before the audit pass (baseUrl resolution, security-scheme mapping, and the OR/AND unmappable-security cases all import correctly).

## Checklist
- [x] My code follows the style guidelines (`clang-format-19`, `clang-tidy-19`, `pnpm lint`/`format:check` all clean)
- [x] I have added tests
- [x] I have updated documentation (one doc per format touched, same commit as the code)

## Related Issues

Towards #1656, towards #1657 (both intentionally left open - see "Deliberately not fixed here" above)


athrvk:
Post-merge review at master `ef43b155`. The import and export suites pass 233/233; the four commits do what their messages say, the five format docs moved with the code, and every new fixed-order `SkippedItem` kind is present in `ImportTally`, the app's union and `ImportModal`'s labels. `collection_primary_auth` is not a pure refactor, as the PR body suggests, but the fix itself: the old path called `scheme_to_auth_v{2,3}` with no tally, and the wrapper adds the unmapped-security count at the collection level. Correctly described in the commit message.

Two behaviour gaps found and filed, both in code this PR touched:

- **#1660** - `Assertion.test_type` reads only the Equals bit (`jmeter_import.cpp:264-269`), so a Response Assertion with "Not" ticked imports as its own opposite, and Matches/Contains (both regular-expression kinds in JMeter) collapse onto a plain text `contains`. `assert.contains` already has `negate` and `mode: "matches"`.
- **#1661** - `substitute_path_variables` checks only the character after `:key` and searches once per key. With the Insomnia caller passing the full URL including the query (`import_document.cpp:1938-1940`), `?tag=color:id` becomes `?tag=color{{id}}`; a longer token preceding the real segment leaves `:id` literal, for Postman too.

One test-shape note, kept here: the `enabled="false"` guard in `for_each_paired_child` is the single choke point of the walk, so a disabled controller or Thread Group correctly drops its children too, and a probe confirmed that. Only a disabled leaf sampler is tested (`JmeterImport.ADisabledSamplerIsExcludedAndCounted`); a case with a disabled `GenericController` wrapping a sampler, asserting zero requests and only the parent's `_disabled` tally, would pin the semantics the commit relies on.

---

---

PR #1672, opened 2026-09-16
feat(engine): map JMeter HTTPsampler.Files to a form-data body

### LINKED ISSUE #1657: JMeter import: map HTTPsampler.Files (multipart file upload)

athrvk:
## Description
`apply_sampler_arguments` only ever read `HTTPsampler.Arguments`; a JMeter "Files Upload" sampler nests its file rows in `HTTPsampler.Files` instead (`HTTPFileArgs.files` → one `HTTPFileArg` per file), inside the sampler tag itself rather than a sibling tag the generic fallback tally would catch - so every upload was silently dropped until a recent stopgap commit started at least counting it.

This lands the real mapping: a sampler carrying file parts now builds a `form-data` body, one field per `HTTPFileArg` (`File.paramname` → `key`, `File.path` → `src`, marked `unresolved` since it names a file on the exporting machine; `File.mimetype` → `contentType` when present), folding in any regular `HTTPsampler.Arguments` beside them as plain fields - a file has no query-string or raw-body form to take instead. A file arg with no `File.paramname` keys nothing a field could use, so it is tallied under `HTTPsampler.Files` rather than imported as a nameless row.

**Approach and the alternative rejected:** the fix pathway asked to reuse `imported_file_part`'s row shape rather than invent a second one, so `imported_file_part` and its `file_base_name` helper were promoted out of `import_document.cpp`'s anonymous namespace into `vayu::core` (declared in `import_document.hpp`) so `jmeter_import.cpp` can call the same function Postman/Insomnia imports already use. The alternative - a hand-rolled copy of the same row-building logic in `jmeter_import.cpp` - is exactly the "hand-rolled copy of a primitive" this repo's conventions call out, so it was not taken.

**Deliberate scope limit:** `HTTPSampler.DO_MULTIPART_POST` is not read anywhere in this change. Without file parts it would only ever pick between the existing (untouched) query-params shape and `x-www-form-urlencoded` for an arguments-only sampler - a broader remapping the issue does not ask for and no fixture demonstrates a need for. A sampler that already carries file parts has no such choice: multipart is the only wire form a file has, flag or no flag. This is documented in `docs/app/import-collections/jmeter.md` and in a code comment.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing
- `engine/tests/jmeter_import_test.cpp`: replaced the stopgap test with three cases - a file-only upload builds a `form-data` body row (asserting `key`/`type`/`src`/`fileName`/`contentType`/`unresolved`), a sampler mixing file args and regular arguments folds both into one `form-data` body (and leaves `params` empty), and a file arg with no `File.paramname` is tallied under `HTTPsampler.Files` rather than imported as a nameless row. Mutation check: reverting `apply_sampler_arguments` to the pre-fix stopgap makes `body.mode` stay `"none"` and reddens the "no longer skipped" assertion in the other direction.
- `ctest --preset linux-dev --output-on-failure` (full suite) and `python build.py -e -t` clean build - both green.
- `docs/` is `mkdocs`-published; only prose inside an existing section changed, no new headings or links, so `mkdocs build --strict` needed no separate verification beyond the normal docs gate.

## Checklist
- [x] My code follows the style guidelines (clang-format 19 + clang-tidy 19 clean on the pre-commit hook)
- [x] I have added tests

## App changes
No user-visible change: the app's request editor already renders a `form-data` body's file-type fields (the same shape Postman/Insomnia imports have produced all along), so this is an engine-only mapping fix with no new UI surface to add.

## Related Issues
Closes #1657

---
_Generated by [Claude Code](https://claude.ai/code/session_01So8Lm9Q6X6FZU7yKAcEZgh)_


athrvk:
The `github-advanced-security` check (run [35131721829](https://github.com/athrvk/vayu/actions/runs/35131721829/job/104914086231)) failed before any actual scan ran - the job log shows GitHub's own Copilot Autofix backend rejecting its configured model (`CAPIError: 400 The requested model is not supported`) during session setup, in `dist-autofind-ts`/`sweagentd` code this repository does not own or configure. Nothing in this PR's diff (a JMeter import mapping change) touches code scanning, model configuration, or that action's inputs, so this is not this PR's failure to fix.

Attempted the one re-run this warrants (`rerun-failed-jobs` on that run); GitHub refused it with `403 This workflow run cannot be retried`, so I have no means to force a second attempt from here. No fix exists on this side - if this keeps failing on subsequent pushes it likely needs attention from GitHub's Advanced Security service side, not from this repo.

---

Issue #1673, opened 2026-09-16
client_certificates: decide whether an import-sourced certificate may register with an unresolved (not-yet-readable) path


athrvk:
## Problem

Filed from #1656 (Postman import: map request-level `certificate`), which this issue's decision was deliberately kept out of scope for - the PR that closes #1656 maps a request's `certificate` into a `client_certificates` registry candidate, but only when the candidate would already pass `POST /client-certificates`'s own write-time check (`vayu::http::client_cert_rejection`, `engine/src/http/transport_policy.cpp`). That check requires `cert_path` (and, for a PEM pair, `key_path`) to name a file this process can open **right now** (`client_cert_file_rejection`, `transport_policy.cpp:479-513`: "is not a readable file" / "cannot be opened", both a `400`). There is no bypass and no "unresolved" placeholder anywhere in that path - unlike an imported multipart file-body part, which `imported_file_part` (`engine/src/core/import_document.cpp:186-211`) marks `unresolved: true` and lets the user attach later with no file-existence check at import time at all.

A real-world Postman collection's `certificate.cert.src` / `certificate.key.src` name paths on the **exporting** machine (the same caveat a `formdata` file part already carries). Those paths are essentially never readable on the machine doing the import. So in practice, #1656's mapping succeeds only when the certificate files happen to already be reachable from the importing machine (a portable/shared filesystem, or a test fixture) - the common case still falls back to the existing `certificate` `SkippedItem` tally, exactly as before #1656, even though the *mapping* now exists.

Anchors at master `eb4054f` (unaffected by #1656's PR, which does not touch `client_certificates.cpp` or `transport_policy.cpp`): the write-time check's doc comment (`engine/src/http/routes/client_certificates.cpp:19-27`, "A registered entry is usable... A path that is not there... fails at handshake time as an SSL error naming the endpoint, which reads as 'the API is broken'"), `client_cert_file_rejection` (`transport_policy.cpp:479-513`), the DB row shape (`engine/include/vayu/types.hpp:1627-1646`, `ClientCertificate`), the in-memory rule shape (`engine/include/vayu/http/transport_policy.hpp:171-213`, `ClientCertRule`).

## Why this is a decision, not a bug

[...]

PR #1674, opened 2026-09-16
feat(engine): map Postman request certificate to a client_certificates candidate

### LINKED ISSUE #1656: Postman import: map request-level `certificate` and `proxy` fields

athrvk:
## Description

`pm_request` only tallied a Postman request's own `certificate` as skipped - it never read `cert.src`/`key.src` into anything, since Vayu's client certificates belong to a host, not a request (`engine/CLAUDE.md`). This maps the resolvable case.

**The design tension, and why the approach isn't the issue's literal suggestion:** the issue's Fix pathway guessed the mapped candidate would be `unresolved`, mirroring how an imported multipart file part is deferred until the user attaches a real file. But `client_certificates` has no such concept - `POST /client-certificates`'s own check (`client_cert_rejection`) requires the cert/key files to be readable *now*, by design (its file comment: a bad path fails loudly at registration rather than later as a confusing TLS handshake error against the wrong endpoint). Relaxing that is a security-relevant contract change, not a mechanical extension, so this PR doesn't make it - I filed #1673 for the repo owner to decide that trade explicitly, rather than defaulting it myself.

Instead: `pm_certificate` (`import_document.cpp`) derives the request's host from its resolved URL and pre-checks the candidate at **parse time** with the exact same `client_cert_rejection` the route runs - a readable PEM pair or PKCS#12 file, a literal (not `{{var}}`) host, no earlier request in this import already claiming a different certificate for the same `(host, port)`. Only a candidate that would actually pass lands in a new `clientCertificates` array; everything else falls back to the existing `certificate` tally, which the issue's own AC #2 anticipates ("cannot resolve" isn't limited to "no host to key on"). `POST /import/apply` then applies each candidate **after** the tree's own atomic transaction commits, reusing `create_client_certificate_response` - the exact core `POST /client-certificates` uses - best-effort: it takes its own DB lock, so calling it from inside the tree's `with_lock` would deadlock, and a candidate it can't register (a collision since the preview, a file gone missing) is skipped rather than undoing a tree that doesn't reference it. Same trade `POST /globals` already makes as the one write outside that payload.

Practical effect: a real-world Postman export's `cert.src`/`key.src` name paths on the *exporting* machine, so most imports will still land in the tally - this PR is correct and complete, it just doesn't change that reality (that's #1673's call, not this PR's).

`proxy` stays a permanent tally: `TransportPolicy` is workspace/run-scoped with no per-request override to map it into - a decision recorded in code comments and `postman.md`, not a TODO.

**Safe to skim, not the part that needs scrutiny:** the `engine/tests/fixtures/import-conformance.json` diff is purely mechanical (`"clientCertificates": []` added to every `payload`, and to `expected` only for the four Postman-collection-format cases) so the two-sided `import_apply_payload` / `orchestrator.ts` conformance pin stays green - no fixture content changed otherwise. The doc updates across four files are all "docs in the same commit" per repo convention, not scope creep.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing
- `engine/tests/import_parse_test.cpp`: replaced the old tally-only stopgap test with 9 cases covering every branch of `pm_certificate` - no `cert.src`, no `key.src`, an unreadable path, an unresolved `{{var}}` host, the mapped success path (asserts every candidate field), same-cert dedup across two requests, and a different-cert collision on one host+port (first wins, second tallies). Mutation check: reverting `pm_certificate`'s guards to the old unconditional tally flips the success/dedup/collision cases' `clientCertificates` assertions from pass to fail.
- `ctest --preset linux-dev --output-on-failure`: **3421/3421 passed** (full suite, run twice - once before, once after a clang-tidy-driven refactor).
- `pnpm test`, `pnpm type-check`: full app suite green; `orchestrator.payload-conformance.test.ts` (the two-sided flattening pin) passes on both sides.
- `clang-format-19 --dry-run -Werror` and the repo's `scripts/pre-commit` (clang-format + clang-tidy) both clean on every changed engine file.
- `mkdocs build --strict` not run separately - only prose inside existing sections changed in the touched docs, no new headings or links (same reasoning as #1672's body for an equivalent case).

## Checklist
- [x] My code follows the style guidelines (clang-format 19 + clang-tidy 19 clean via the pre-commit hook)
- [x] I have added tests

## App changes
No new UI surface: a mapped certificate lands in the existing client-certificate registry the Settings card already lists, edits and deletes. Two small wiring fixes so it shows up correctly: `useImportMutation`'s cache invalidation now includes `queryKeys.clientCertificates.all` (it didn't touch that query before, since nothing used to write to that table via import), and the `certificate` skip-reason label in the import preview no longer unconditionally says "not yet imported" now that it sometimes succeeds.

## Also fixed in this PR
`import_apply_response`'s cognitive complexity tripped clang-tidy's threshold after the new section was added (28 vs. the 25 limit) - extracted the best-effort apply loop into its own `apply_client_certificate_candidates` helper, which also reads cleaner given the deadlock-avoidance reasoning it needed to carry.

## Deferred
#1673: whether `client_certificates` should ever support an "unresolved" (not-yet-readable) path, which would let a real-world Postman certificate import actually register most of the time instead of usually falling back to the tally. Deliberately not decided in this PR - it's a security-relevant contract change to a route this PR doesn't otherwise touch, and needs the repo owner's call on the trade (an unresolved row silently presenting no certificate vs. today's loud failure at registration), not an implementer's default.

## Related Issues
Closes #1656

---
_Generated by [Claude Code](https://claude.ai/code/session_01BeuuS7NSW2Jkv8HWvRRFfF)_


athrvk:
The `github-advanced-security` check (run [35139394546](https://github.com/athrvk/vayu/actions/runs/35139394546/job/104939817927)) fails before any actual scan runs - the job log shows GitHub's own Copilot Autofix backend rejecting its configured model (`CAPIError: 400 The requested model is not supported`) during session setup, in `dist-autofind-ts`/`sweagentd` code this repository does not own or configure. Nothing in this PR's diff (a Postman-import mapping change) touches code scanning, model configuration, or that action's inputs.

Same failure, same cause, as the one recorded on #1672's `github-advanced-security` run a few minutes earlier on this same base - not this PR's regression, and reproduced identically on both the pre-fix and post-fix commit of this PR.

Attempted the one re-run this warrants (`rerun-failed-jobs` on run 35139394546); GitHub refused it with `403 This workflow run cannot be retried`, so I have no means to force a second attempt from here. If this keeps failing on subsequent pushes it needs attention from GitHub's Advanced Security service side, not from this repo.

---

Issue #1492, opened 2026-09-06
engine(db): nothing on disk records a schema or engine version and nothing guards a downgrade - both run today only because the schema is identical, sync_schema is called without preserve, and an old engine serves and accepts writes to settings it does not read


athrvk:
## Problem

Found by the upgrade sweep, which upgraded 0.24 and 0.25 workspaces to `cb829ad3` and then opened each migrated database with its original engine; re-anchored at `228f1131`. Measured, not predicted, unless marked.

- **No version on disk.** `PRAGMA user_version` and `application_id` are `0` on every database before and after every transition. `docs/engine/db-schema.md:10-14` says so: `user_version` is not managed, additive changes are automatic, destructive ones "wipe the database" - by which it means the user deletes the file.
- **No downgrade guard.** 0.24.0 and 0.25.0 both open a head-migrated database cleanly ("Database validation successful. Updating backup..."), with zero row loss. The reason is that the SQL schema is byte-identical across the three tags (same 16 tables, same columns in the same order, same 11 indexes - verified from `sqlite_master` and `PRAGMA table_info`). Nothing checked anything.
- **`sync_schema ()` runs without `preserve`** (`engine/src/db/database.cpp:1040`, `:1065`, `:1093`). That is sqlite_orm's drop-and-recreate path for a table whose stored columns are not all in the current definition. The day a column is removed or retyped, an **older** engine opening a **newer** database takes that path with no log line and no `.bak` older than the same start (#1487 covers the backup order). Predicted, not reachable today - which is the moment to fix it.
- **An old engine offers settings it does not read.** `seed_default_config` deletes only a fixed retired-key list, `GET /config` returns the whole table, and `POST /config` gates on the row existing (`engine/src/http/routes/config.cpp:350`, "Unknown config key"). Measured: 0.24 listed all six 0.26 keys with head's labels, the Settings pane renders them, and it accepted `correlationIdHeader = "not a header name!!"`, which head's own validator refuses (`config.cpp:258`). Re-upgrading then applied it (head's reader at `default_headers.cpp:160-163` fell back to the default name, as its comment anticipates).
[...]

Issue #1283, opened 2026-09-02
engine(http): the inbox's catch-all mirrors cpp-httplib's socket loop pinned to 0.54.1 - upstream a request-setup hook (or a matcher-free default handler) and retire the override


athrvk:
Filed from the review of PR #1276 (fix for #1140), so the mirror it introduces has an exit recorded rather than living forever by default.

## Problem

#1276 makes the inbox record any path by subclassing `httplib::Server`, overriding the protected virtual `process_and_close_socket`, and copying its ~15-line body verbatim so that a `setup_request` callback can rewrite `req.path` to one literal before routing (`engine/src/http/routes/inbox.cpp:138-191`, `AnyPathServer`; the override at `:146`, the mirrored body at `:172-190`). A `static_assert` on `CPPHTTPLIB_VERSION` pins the mirror to the release it copies - `"0.53.1"` at filing, `"0.54.1"` since PR #1423 (`:159-165`), which also had to copy 0.54.0's `serve_guarded` wrapper (`:152-158`): 0.54.0 wrapped the per-connection serving loop in a try/catch so a throwing handler no longer terminates the process, and a mirror that did not inherit it would have kept the old behaviour on the one listener that may bind past loopback. That bump is the cost paragraph below happening on schedule.

That is the right fix within cpp-httplib's public API - verified during review, and re-verified against 0.54.1: the parameter matcher stops at every `/`, the pre-routing handler (`set_pre_request_handler`, present since before 0.53.1) takes a `const Request&` and runs only after `matcher->match`, content-reader handlers go through the same matcher, the error handler sees no body because the body is read only after a route matches, and raising the regex length macro is either an ODR violation or hands attacker-controlled input to `std::regex_match`. `setup_request` is the only mutable pre-routing hook and the virtual is the only door to it, the same door httplib's own `SSLServer` uses (0.54.1 `httplib.h:2255`, `:14236`, `:17812-17878`).

The cost is structural: every cpp-httplib bump now stops the build until someone re-reads the mirror against the new release, and the engine's baseline-bump cadence is deliberately routine (`CLAUDE.md`). Internal `detail::` helpers are called from application code. The repo's core principle is not to keep a stopgap meant to be replaced; this one is only acceptable because the replacement is not in Vayu's hands yet.

## Fix / solution pathway

[...]

Issue #967, opened 2026-08-25
engine(ci): TSan data race in libstdc++'s locale narrowing cache, reached through httplib's per-response std::regex


athrvk:
Suppressed in `engine/sanitizers/tsan.supp` as `race:std::ctype::narrow`. This issue is the tracking entry that suppression's header requires.

**Status refresh, 2026-08-28, verified at master `98d2ab0`: the root-cause mechanism has changed - upstream effectively landed remedy 2, but narrower than proposed.** The vendored cpp-httplib is now **0.53.0**, and `parse_status_line` (`engine/vcpkg_installed/x64-linux/include/httplib.h:7800`) no longer constructs a `std::regex` per response:

```cpp
thread_local const std::regex re("(HTTP/1\\.[01]) (\\d{3})(?: (.*?))?\r\n");
```

Note the nuance before deleting the suppression: `thread_local` is not the function-local `static` this issue proposed. Each thread still compiles its own regex once, so two threads hitting their *first* response concurrently can still race libstdc++'s shared narrow cache. The window shrank from per-response to once-per-thread; it did not close. Whether TSan still observes it is now an empirical question, not an upstream ask.

**Only visible since #957 was fixed.** Before that, every test that opens a socket segfaulted inside the sanitizer runtime and never ran far enough to report anything. Removing the `getaddrinfo_a` crash took `linux-tsan` from 2312/2370 to 2369/2370, and this was the one remaining failure.

## The report (as originally captured, gcc 13.3, libtsan 2, `linux-tsan` preset, `TransportPolicyPaths.LoadRunTraversesManualProxy`)

```
WARNING: ThreadSanitizer: data race
  Read of size 1 at 0x7f93f747bb81 by thread T8 (mutexes: write M0):
    #0 std::ctype::narrow(char, char) const   locale_facets.h:941
    #1..#6 std::regex compilation internals
    #7 httplib::detail::parse_status_line(...)
    #8 httplib::ClientImpl::read_response_line(...)  httplib.h:13856
    #9 httplib::ClientImpl::process_request(...)
```

At `98d2ab0` the call site anchors are: `parse_status_line` at `httplib.h:7800`, called from `read_response_line` at `httplib.h:13856` (still exact) and now also at `:13866`.

## Root cause

[...]

PR #1678, opened 2026-09-17
chore(release): 0.32.0, on a moved vcpkg baseline

athrvk:
## Description

Release 0.32.0, carrying the release-window vcpkg baseline bump. Three commits, and they are worth reviewing separately - the second is where all the risk is.

**`chore(release): 0.32.0`** - minor rather than patch because the one user-facing change since v0.31.0 is new behaviour in Settings (collapsible long descriptions), not a fix. The bump is `build.py --bump-version minor`'s output unedited; safe to skim. `SCHEMA_VERSION` is untouched and correctly so - the whole diff since v0.31.0 is `app/` and `docs/`, nothing near `make_vayu_storage`'s mapping, so no one-way downgrade barrier ships here (#1492).

**`chore(deps): move the vcpkg baseline to 386d7c4`** - `04a9d8e` -> `386d7c4`, in `engine/vcpkg.json` and the seven `VCPKG_COMMIT` pins `cache-warm.yml`'s guard job enumerates. The mechanical part (one SHA, eight places) is safe to skim; I re-ran the guard's own comparison logic locally and all seven agree.

What is not mechanical: **cpp-httplib moves 0.54.1 -> 0.56.0**, and two things in this repo are pinned to that version on purpose.

- **The `AnyPathServer` mirror in `inbox.cpp`.** Its `static_assert` exists to stop exactly this move from passing silently. I re-read `Server::process_and_close_socket` against 0.56.0's actual source: it is byte-identical to 0.54.1, and the declarations the override depends on - `process_request`'s `setup_request` parameter and `serve_guarded` - are unchanged, so the override still mirrors it. Assert moved to 0.56.0. **The mirror does not retire yet**: #1283's exit condition is still unmet, because `set_pre_request_handler` remains a `HandlerWithResponse` over a `const Request&` and so still cannot rewrite the path before routing.
- **The `race:std::ctype<char>::narrow` tsan suppression**, re-measured in the third commit below rather than carried forward.

**`docs(engine): re-measure the narrow suppression on cpp-httplib 0.56.0`** - `tsan.supp` is explicit that this entry is a measurement, not an argument, on every cpp-httplib move. Reading: **9 of 10 runs of `TransportPolicyPaths.LoadRunTraversesManualProxy` raced without the line, 10 of 10 passed with it**, and the report was the documented stack frame for frame - `narrow` under `_Scanner`, under the `basic_regex` constructor, under `parse_status_line`. That function is byte-identical to 0.54.1's and its regex is still `thread_local`, so the window did not move; 9 rather than 10 is this race's ordinary spread, which the 0.53.1 reading already saw as 8 of 10. The entry stays. Recorded in `tsan.supp` and `docs/engine/building.md`.

### Risk, in the open - and how it resolved

Folding the baseline bump into the release PR is against this repo's own release skill, which wants it landed separately and first, precisely so a dependency break and a version break can be told apart by which PR went red. Done here at the author's explicit direction. The cost would have been a red platform whose cause was ambiguous between the two changes.

**That did not happen: the engine is green on all three platforms against 0.56.0** - Windows (build, tests, no-non-Windows-dependency check, clang-tidy on changed sources), macOS and Ubuntu. So the ambiguity never had to be paid for. Worth keeping the rule anyway; it was a coin that landed the right way, not a demonstration that the separation is unnecessary.

Second-order thing a reviewer would not get from the diff: the version bump is in `cache-warm.yml`'s `push` paths, so merging fires a warm run that populates the master sccache scope. **Do not tag before that run is green on the merge commit**, or the tag build reads an empty scope and compiles the engine cold (#700). And do not tag this branch's commit at all - a squash merge rewrites the hash.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Release (version bump + notes)
- [x] Dependency update (vcpkg baseline)

## Testing

- `./build/vayu-engine --help` prints `Vayu Engine 0.32.0`.
- Engine builds clean against cpp-httplib 0.56.0 with the updated assert - locally on Linux, and in CI on all three platforms. The assert is the check that matters for the mirror, since it is what fails if the move goes unexamined.
- `cache-warm.yml`'s guard comparison re-run locally over all seven workflows: keys and commits agree. CI's own `Cache key agreement` job covers it too.
- tsan re-measurement: 9/10 raced without the entry, 10/10 passed with it, per the procedure `tsan.supp` documents. CI's `tsan on ubuntu-latest` independently passed the full sanitizer suite on the new baseline with the updated file.
- Stale-clone self-heal check (#692): against a clone pinned at `04a9d8e`, `build.py -e` printed `vcpkg clone predates baseline 386d7c4782 - updating...`, updated the registry, and built the engine successfully. This is the path every environment that has not updated since the last bump will take.

No app tests were run and none are warranted: no `app/` file changes in this PR.

### Note on `github-advanced-security`

Red on every commit here, and not this PR's - it dies inside GitHub's Copilot autofix service on `CAPIError: 400 The requested model is not supported`, never reaching the diff. Not retriable via the API (`403 This workflow run cannot be retried`). Detail in a comment below; it is red identically on `master`.

## Checklist

- [ ] I have added tests - not applicable; no behaviour changes, and the version assert is itself the guard for the one code edit
- [x] I have updated documentation - release notes for the tag, and `docs/engine/building.md` for the tsan reading

## Related Issues

Touches #1283 (checked, not yet retirable) and #967 (suppression re-measured).


athrvk:
`github-advanced-security` is red on 130f9673 and it is not this PR's failure.

The job never reached any analysis of the diff. It died inside GitHub's own Copilot autofix agent:

```
Dt [SessionModelError]: Execution failed: CAPIError: 400 The requested model is not supported.
  COPILOT_AGENT_MODEL: sweagent-capi:claude-opus-5
```

That is a 400 from `api.individual.githubcopilot.com` rejecting the model the action asked it for - a GitHub-side service configuration, with no input from this repository. Nothing in the commit it ran against is reachable from it either: 130f9673 changes four version strings and adds a Markdown file.

The repository's own security analysis on the same commit is green - CodeQL's `Analyze JavaScript/TypeScript` succeeded, and `Analyze C++` was correctly skipped for a diff with no C++ in it.

There is no fix to port: the failure is in a hosted service, not in code any PR here can change. I also cannot re-run it to confirm - the API refuses with `403 This workflow run cannot be retried`, since the run belongs to GitHub's managed Advanced Security workflow rather than one in `.github/workflows/`. Expect it red until GitHub's side is fixed, and red identically on `master`.

Flagging it rather than leaving it silent; the PR stays watched.

---

PR #1680, opened 2026-09-17
fix(app): give chrome bands, targets and icons their own floor, separate from density rhythm

athrvk:
## Description

#1670 made `--spacing` the one lever for the app's density, and every `h-*`/`w-*`/`size-*` utility rides it in Tailwind 4 - including ones that were never meant to shrink. That's what actually shipped: a 24px tab strip (not the 32px every editor-class app keeps), 21px close buttons, 15px switches, 12px checkboxes and 9px/12px icons, several of them under the WCAG 2.2 SC 2.5.8 24×24px interactive-target floor. #1676 then wrote down "the shrink applies to icons too" as a deliberate decision. The owner's report on 0.32.0 reverses that: small reads as small, not as dense.

The fix is architectural rather than a value bump: **density scales rhythm - row heights, paddings, gaps - and three other classes of thing (chrome bands, interactive targets, icons) get their own floor that does not ride `--spacing` at all.** Seven named steps (`--spacing-band`/`-banner`/`-control`/`-control-sm`/`-target`/`-icon`/`-icon-sm`) land in a plain `@theme` block in `index.css`, which is what makes them real Tailwind utilities (`h-band`, `size-target`, ...) rather than one-off arbitrary values; `control`/`control-sm`/`target` still scale under Comfortable (the same mechanism `--spacing` itself uses), `band`/`banner`/`icon`/`icon-sm` do not, mirroring how `--titlebar-height` was already theme-independent.

I considered leaving the icon sizes exactly as #1676 decided and only fixing the chrome bands and targets the issue's owner-reported complaint named directly (the tab strip and the update banner). I rejected that: the same rhythm-riding class of bug is what produced the undersized icons in the first place, and the acceptance criteria explicitly reverses the #1676 decision - fixing the two reported symptoms while leaving the root cause (everything on one unit) in place would have left the next report to file the same issue again on a different control.

### What moved

- **Chrome bands** (32px `band`, 36px `banner`, every density): the tab strip and drawer header (via `--tabstrip-height` now resolving to `var(--spacing-band)` instead of `calc(var(--spacing) * 8)`), the response toolbar, `RailButton`, the update and recovery banners.
[...]

PR #1682, opened 2026-09-18
fix: checkbox styling, Scripts order and validation-message clarity follow-ups

athrvk:
## Description

Follow-ups found after #1680 (the density-floor work) merged:

- The KeyValueRow (request/response tables) enable checkbox was a bare native
  `<input type="checkbox">` scaled up with only `accent-color` styling - at
  `size-target` (24-28px) that rendered as a stark, barely-rounded flat
  square, because `rounded-md`/`border-input` were dead CSS on an unstyled
  native control. `VariableTableEditor` (the variables tables) had the same
  latent defect at a size where it happened not to show. Fixed with one
  shared `Checkbox` primitive (`components/ui/checkbox.tsx`, `appearance-none`
  + `checked:bg-*` + an overlaid `lucide-react` check icon via the
  `peer`/`peer-checked:` pattern) and migrated both call sites onto it,
  per this repo's own rule against a hand-rolled copy of a primitive.
- The Load Test dialog's Scripts control rendered above the load-profile
  summary text, ahead of the paragraph it has nothing to do with. Moved to
  after the summary, which is the first control that isn't about the
  profile.
- A validation toast for a `POST /runs` payload read
  `requestElements[2]: 'control.throughput' does not run on a single-target
  load run - only extract.*, ...` - an internal kind string and a 0-based
  array index, verbatim in front of a user. This turned into a full sweep of
  the engine's user-facing validation text, since the app deliberately
  surfaces the engine's error strings as-is (documented in `http-client.ts`,
  issue #173) and so every one of them is this repo's UI text, not an
  internal detail:
  - The element registry and the single-request load run's `requestElements`
    validation (`registry.cpp`, `load_strategy.cpp`) now name an item by its
    kind's `label` and 1-based position instead of the wire `kind` string and
    a 0-based `elements[N]` index.
  - The same run's `lifecycleElements` array and its `elements.timers` /
    `elements.scripts` / `elements.seed` / `elements.includeScriptTime`
    overrides now name the field the way the Load Test / Run Collection
    dialog's own controls do ("Timers", "Scripts"), or in plain English where
    there is no dialog control to borrow from.
  - Settings' `POST /config` rejections (out-of-range, wrong type, bad enum
    value, an unusable proxy URL, a bad correlation-id header name) now name
    the field by `ConfigEntry::label` - the same text Settings itself shows -
    instead of the camelCase wire key.
  - The load test dialog's budget thresholds (`threshold_eval.cpp`) now name
    the metric the way `budgets.ts`'s own field labels do ("p99 latency",
    "Throughput") instead of the wire key (`latencyP99Ms`, `minThroughputRps`).
  - A mock server that can't be started because its collection was deleted
    says so in plain English instead of echoing the collection's raw
    internal id.
  - A schema-mismatch error - both an element's own config against its kind's
    schema, and a response body against an imported document's schema - is
    now rebuilt from the schema's own declared title, type, bounds or enum
    list (`schema_error_text.hpp/cpp`, shared by both validators) rather than
    quoting valijson's internal keyword wording ("Value type not permitted by
    'type' constraint.") verbatim. Falls back to valijson's own text
    untouched for any constraint outside the set this engine's schemas
    actually use, so nothing is ever silently dropped.
- The variable popover's value field (`w-72`, 288px) left anything longer
  than a short value - a domain, a URL - scrolling inside the box instead of
  showing it. Widened to `w-96` with a `max-w-[calc(100vw-2rem)]` clamp, the
  same fixed-width-plus-responsive-cap idiom `toast.tsx` already uses.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update (release notes)

## Testing

- App: `pnpm test` (771/771 files, 8713 passed, run again after the popover
  width change), `pnpm lint`, `pnpm format:check` - all clean. Checkbox,
  Scripts-order and popover-width changes were all driven live in a headless
  Chromium (vite + the built engine, no Electron) to confirm the rendered
  result.
- Engine: `ninja vayu-engine vayu_tests`, `ctest --preset linux-dev`
  (3423/3423 passed, run repeatedly across the several rounds of message
  changes), `clang-format-19 --dry-run -Werror` and `clang-tidy-19`
  (whole-file, this repo's gate) on every touched source, header and test
  file - all clean. CI's macOS legs caught a real gap this local Linux
  environment couldn't (`std::from_chars<double>` is unavailable in libc++
  before macOS 26); fixed by switching to the `std::stod`-based idiom
  `monitor.cpp` already uses, confirmed green on CI's own macOS runner.

## Checklist

## Related Issues

None filed - the checkbox rendering, the validation-message wording, the
popover width and the scope of the sweep were all worked out directly during
review of #1680, not as standing issues.


athrvk:
The `github-advanced-security` check is failing, but not on anything in this diff. The job log shows it's GitHub's own Copilot-based agentic code-scanning step erroring before it analyzes any code:

```
Error creating PR review request: SessionModelError: Execution failed: CAPIError: 400 The requested model is not supported.
```

That's a model-provisioning failure on GitHub's side (`claude-opus-5[ReasoningEffort=medium]` rejected by CAPI), not a finding about this PR. The same check failed identically on #1680 just before it, which merged anyway. I don't have permission to re-run it (`rerun_failed_jobs` returned a 403), so I can't confirm a re-run clears it - flagging here rather than retrying blind. Every other check (App, Engine, quality checks, formatting) is green or still running.

---

Issue #1683, opened 2026-09-19
UI/UX programme: interactive icons and interaction/design consistency across the app


athrvk:
## Why

Two goals, one programme, audited against master `9b9a22f`:

1. **Interactive icons.** Icons that respond to hover, press and state change in the style of lucide-animated.com, without a new runtime dependency and without breaking the app's reduced-motion contract.
2. **Interaction and design consistency.** A code audit of every module plus a light/dark screenshot review found a set of real bugs (bare `Enter` commits, an Escape that still commits a rename, copy buttons that swallow a denied clipboard) and a longer tail of the same action done two ways, the same label styled five ways, and primitives that exist but are bypassed.

Every finding is untracked; the tracker had six open issues at audit time, none of them UI.

## Architecture decisions (recorded here so sub-issues do not re-litigate)

- **Icon motion is CSS-first, no `motion`/framer-motion.** lucide-animated is a shadcn copy-paste registry of `motion` components, each wrapped in a `div` that breaks `button-variants.ts`'s `[&_svg:not([class*='size-'])]:size-icon` sizing and `[&_svg]:pointer-events-none`, and JS-driven motion is invisible to both reduced-motion rules in `index.css` (`html[data-reduced-motion="true"]` and `@media (prefers-reduced-motion: reduce)` collapse only `animation-duration`, `animation-iteration-count`, `transition-duration`, `scroll-behavior`). Lucide renders a stable `lucide-<name>` class, its `__iconNode` children in declared order with nothing prepended, and exports `__iconNode` per icon, so CSS can address `.lucide-trash-2 > path:nth-child(4)` today and a test can pin that index.
- **Icon motion policy** (goes into `docs/design-system.md` under Motion): an icon animates only as feedback for the action it is the affordance for, and only on `:hover`/`:focus-visible` of its own `[data-slot="button"]` or owning `.group` row, never the icon's own hover. Status icons (`AlertTriangle`, `AlertCircle`, `CheckCircle2`, `XCircle`, `Clock`, `Info`) never animate. State changes animate once. Loops are reserved for `Loader2` and live indicators. Every SVG-child transform sets `transform-box: fill-box` and an explicit `transform-origin`. Durations and eases come only from the existing `--dur-*` / `--ease-*` tokens.
[...]

PR #1695, opened 2026-09-19
feat(app): give pane chrome one shape - TabsList variants, band tokens, FieldError, DrawerSection

### LINKED ISSUE #1688: Pane chrome: TabsList variants, header band tokens, one error presentation rule, DrawerSection primitive, settings double heading, empty tab strip

## Problem

1. **Six tab-strip chromes.** `modules/history/main/LoadTestDetail.tsx:358` (`mx-5 mt-3`), `components/shared/response-viewer/ResponseViewer/index.tsx:340` (none), `modules/request-builder/components/RequestTabs/index.tsx:113` (`w-full px-1`), `modules/dashboard/index.tsx:425` (`px-5` + `border-b bg-panel`), `modules/collections/ImportModal.tsx:939` (`w-full px-4`), `modules/collections/CollectionDetail/index.tsx:251` (`bg-panel px-4`), `UnifiedResponseViewer.tsx:81` (`px-3 py-1.5 border-b border-rule bg-muted/30`).
2. **Header heights as literal pixels**, which also opt out of the density setting (`stores/appearance-store.ts:86-90` scales `--spacing`, arbitrary `h-[Npx]` does not): `modules/dashboard/components/DashboardHeader.tsx` `h-[52px]`, `CollectionDetail/index.tsx` `h-[52px]`, `components/layout/TitleBar.tsx` `h-[38px]`, `modules/request-builder/components/UrlBar/index.tsx` `h-[40px]`/`h-[34px]`, `CollectionItem.tsx` `h-[18px]`. `h-band` / `h-control` exist for this. Leave the deliberate ones (`h-[2px]` underline, `h-[1px]` separator, editor pane heights).
3. **Error text has three presentations**: `Callout` (20 files), bare `text-destructive-text` paragraphs (`ImportModal.tsx:695,862,1137,1150,1346,1352` at three sizes; `MonitorSummary.tsx:83`; `UpdatesCard.tsx:107`; `SendWithRowDialog.tsx:486,493`), and full-pane `ErrorState`.
4. **Drawer section headers have no shared shape** (screenshot review): Services (`modules/services/ServicesPanel.tsx`) uses muted sentence-case labels with a plus; Variables (`VariablesCategoryTree.tsx`) uses chevron, icon, filled count badge and plus; Collections writes counts as inline "(2)" text (`CollectionItem.tsx`). Two count idioms one click apart.
5. **Settings panes carry two headings with drifted copy.** The band shows `description` + `saveNote` from `modules/settings/main/app-panels.ts:89,93` ("Every chord the app listens for, drawn for this platform" / "Shortcuts are fixed; this screen is a reference."), then `KeyboardShortcutsPanel.tsx:58` repeats a title and a different subtitle inside the card. Check every panel for the same doubling.
6. **An empty tab strip is a blank band with one plus** (`components/layout/TabStrip.tsx`) when no tabs are open.

## Fix pathway

1. `TabsList` gets two documented variants in `components/ui/tabs.tsx`: `pane` (`bg-panel px-4 border-b border-rule`) and `inset` (`px-1`); delete the per-site classes. `UnifiedResponseViewer`'s `bg-muted/30` strip becomes `inset` or gets a written exemption.
2. Move header bands to `h-band`; add `h-band-lg` (52px) and `h-control-lg` (40px) tokens in `index.css` only if the measurements are distinct bands the design system wants. Guard: no `h-[Npx]` on an element that also carries `border-b` and `bg-panel`.
3. `components/shared/FieldError.tsx` (one size, `role="alert"`); rule: field-level error is `FieldError`, block-level is `Callout`, pane-level is `ErrorState`. Convert the sites above.
4. `components/shared/DrawerSection.tsx` beside `DrawerPanel`: title, optional count (one idiom, choose the muted inline count and drop the filled badge, or the reverse, and write it down), optional actions slot, optional collapse. Adopt in Services, Variables, Collections.
5. Settings: the card no longer repeats the title/description the band already shows; one source of copy per panel.
6. TabStrip with zero tabs: collapse the strip or render a muted hint ("Open a request from the drawer, or press Ctrl+N").

## Tests

- `tabs.test.tsx`: both variants render their class (mutation check: swap the variant strings, fails).
- `header-band.test.ts` source guard for item 2 (non-empty scan asserted).
- `FieldError.test.tsx`; `error-presentation.test.ts` guard: no `text-destructive-text` on a `<p>`/`<span>` outside `FieldError`, `Callout`, `ErrorState`, `callout-severity.ts`.
- `DrawerSection.test.tsx`, and one rendered assertion per adopting drawer.
- `SettingsMain` test: the panel heading text appears once in the rendered pane.
- `TabStrip.empty.test.tsx`.

## Docs

`docs/design-system.md`: Component Patterns (tabs variants, error levels, drawer section), Geometry (any new band token; `design-system-doc.test.ts` checks values), Drawer Panel Frame section. `docs/app/COMPONENTS.md` for the two new shared components.

## Sequencing

Chain P1, base master. Touches `index.css` and `design-system.md` in regions different from the icon chain; a merge should be clean.

## Acceptance criteria

- `rg '<TabsList' app/src --glob '!*.test.*' | rg -v 'variant='` returns nothing.
- `rg 'h-\[52px\]|h-\[38px\]|h-\[40px\]' app/src --glob '!*.test.*'` returns nothing.
- `rg 'text-destructive-text' app/src --glob '!*.test.*' -l` lists only `FieldError.tsx`, `Callout.tsx`, `ErrorState.tsx`, `callout-severity.ts` and any file whose exemption is written in the guard.
- `rg '<DrawerSection' app/src -l` includes `ServicesPanel.tsx`, `VariablesCategoryTree.tsx` and the collections drawer.
- `cd app && pnpm test tabs header-band FieldError error-presentation DrawerSection SettingsMain TabStrip design-system-doc` green; `pnpm lint && pnpm type-check && pnpm format:check` green.

athrvk:
## Description

Six pieces of pane chrome where the app did the same thing more than one way. Each one is its own commit, and the reason each decision went the way it did is in the code and the design system rather than only here.

**`TabsList` takes a required `variant`.** The trigger has been shared for a while; the band around it was not - seven call sites carried seven recipes, so one control read as different chrome in every pane. Three variants (`pane`, `inset`, `bare`) and **no default**, so a new strip has to say which it is. `bare` is the one that needs justifying: the response pane's strip shares its row with the response's facts and its actions, so the *parent* paints the band, and a list that also painted one would put two rules a pixel apart. The unified response viewer's `bg-muted/30` strip became `pane` rather than taking an exemption.

**Two new band steps, `band-md` (40px) and `band-lg` (52px).** The dashboard header, the Collection Detail header and the URL bar row wrote their heights as bracketed literals. That is worse than a wrong token: an arbitrary value is a fixed length, so those rows were the only chrome in the app that could follow nothing at all. Both steps are theme-independent for the same reason `band` is, and the rows still breathe with the density setting through their own padding.

**`FieldError`, and one rule for which error presentation a message gets.** Field-level is `FieldError`, block-level is `Callout`, pane-level is `ErrorState`. The middle of that range was hand-written paragraphs at `text-sm`, `text-xs` and `text-[11px]` - the import dialog carried all three - some announced and most not. Worth a reviewer's attention: `text-destructive-text` is *also* the right foreground for data that happens to be red (a failure count, a "not defined" chip, the Dock's "Not saved"), and no scan can tell those from a sentence. So the guard is narrowed to the shape a message takes - the token in a literal class string on a `<p>`/`<span>` - and every exception is named in the guard with what the red text is instead. See "Deviation from the acceptance criteria" below.

**`DrawerSection`, and one count idiom.** Variables drew a section's count as a filled `Badge` while a collection row one click away wrote the same fact as muted `(2)` text. The badge loses, and the reason is recorded: a count beside a group name is not a status and not a control, so a fill, a border and a 20px floor give it weight it has not earned. `DrawerSectionCount` is exported so a *row* uses the header's idiom. The section's ARIA stays at the call site (`rowProps` / `activatorProps`) because Variables' headers are `treeitem` rows in a roving-tabindex tree whose "+" sits outside the row on purpose - a primitive owning those would have to know about the tree.

**Settings panels say their name once.** Keyboard shortcuts said "Every chord the app listens for" in the band and "Every shortcut the app listens for … They are not rebindable yet" in a card a few pixels below, beside a save note that already said shortcuts are fixed. Editor and Notifications repeated their own titles the same way. Notifications' card is called "Toasts" now, which is the topic it actually is; the one fact its description carried alone ("how many stack at once") moved into the band's description so nothing is lost.

**An empty tab strip says what to do.** It was a blank 32px band with a lone "+". The strip does not collapse, and that is deliberate: its height is the token the drawer's header band also reads, so a strip that disappeared would leave that header as a step in the rule running across the window and would jump the content area by a band's height the moment the last tab closed.

### What is risky, and what to skip

- **Visible, so worth a look in the app:** Collection Detail's tab band gains the bottom rule it lacked; the unified response viewer's strip changes from a `bg-muted/30` tint to the panel band; ImportModal's dialog tabs become a `pane` band; the collection tree's three 18px glyphs move to the 16px `size-icon` step; the Variables section counts change from a filled badge to `(3)`.
- **Second-order effect worth naming:** `variant` is required, so a *new* `TabsList` fails to compile rather than silently inheriting a look - that is the point, but it is also the one API break in this PR for anything on another branch that adds a strip.
- **Safe to skim:** the doc edits (`docs/design-system.md`, `docs/app/COMPONENTS.md`) are this repo's same-commit doc rule, and the two test-expectation edits that follow from decisions above (`UrlBar.test.tsx`'s floor class, the two Variables count assertions now reading `(-)`).
- **One guard learned a new case:** `app-settings.drift.test.tsx` wants every catalogue anchor to print its label inside the panel. The only anchor whose label *is* its panel's label is Keyboard shortcuts, which the band prints and this file does not render - so that case is skipped there with the reason, and asserted in the rendered half instead.

## Deviation from the acceptance criteria

Two of the issue's `rg` commands do not come back empty, and neither is a site left unconverted:

1. `rg '<TabsList' app/src --glob '!*.test.*' | rg -v 'variant='` reports three tags. All three carry `variant=`; prettier puts each attribute of a multi-attribute JSX tag on its own line, so a single-line grep cannot see it. The multiline-aware form is empty, and `tabs.test.tsx` carries the real check - a scan of every `<TabsList …>` tag in `app/src`, matched across newlines.
2. `rg 'text-destructive-text' app/src --glob '!*.test.*' -l` lists 38 files, and the criterion's four-file list is not reachable: the token is the destructive foreground, not an error-message class, and roughly thirty of those uses are icons, buttons, chips, counts and conditional red/green state pairs. `error-presentation.test.ts` enforces the rule that can be enforced (red *prose* on a `<p>`/`<span>` lives in one of the three primitives) with eight named exemptions, each saying what the red text is instead of a message.

The fourth criterion (`rg '<DrawerSection' -l` includes the collections drawer) passes through `CollectionItem.tsx`, which adopts `DrawerSectionCount`. It deliberately does not gain a `DrawerSection` *frame*: `CollectionTree`'s panel band already says "Collections", so a section of the same name inside it would be the double heading item 5 of this same issue removes from Settings. That exception is written into `DrawerSection.test.tsx` rather than skipped silently.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`pnpm lint && pnpm type-check && pnpm format:check`: green.

The targeted run for everything this PR touches - `pnpm test tabs header-band FieldError error-presentation DrawerSection SettingsMain TabStrip design-system-doc` - is **63 files, 597 tests, all passing**.

**The full `pnpm test` needs a note.** This branch's worktree shares a box with four other agents running their own suites concurrently (load average above 30), and `app/CLAUDE.md` already documents what that does: vitest's 5s per-case default is spent idle waiting for CPU, so render-heavy cases fail on the wall clock rather than on the code. Two full runs reported the same set of files failing that way, each with a *file* duration of 10-72s: `DesignRunView`, `InboxView.captures`, `element-kinds.conformance`, `monaco-setup.contributions`, `CollectionTree.delete`. None is a file this PR touches, and **all five pass when re-run off the contended path** (`InboxView.captures` 9/9; `DesignRunView` + `element-kinds.conformance` + `monaco-setup.contributions` + `CollectionTree.delete` 55/55). CI runs one suite on its own box, so its result is the one to believe over either of mine.

The acceptance-criteria commands, run on the branch:

```
$ rg 'h-\[52px\]|h-\[38px\]|h-\[40px\]' app/src --glob '!*.test.*'
(no output)

$ rg -U --multiline --multiline-dotall '<TabsList.*?>' app/src --glob '!*.test.*' -o --no-filename \
    | tr '\n' ' ' | sed 's/> /> \n/g' | rg -v 'variant='
(no output)

$ rg '<DrawerSection' app/src -l
app/src/modules/collections/CollectionItem.tsx
app/src/components/shared/DrawerSection.test.tsx
app/src/components/shared/DrawerSection.tsx
app/src/modules/services/ServicesPanel.tsx
app/src/modules/variables/sidebar/VariablesCategoryTree.tsx

$ pnpm test tabs header-band FieldError error-presentation DrawerSection SettingsMain TabStrip design-system-doc
 Test Files  63 passed (63)
      Tests  597 passed (597)
```

Mutation checks, each confirmed live (fail, then restored):

- `tabs.test.tsx` - swap the `pane` and `inset` strings in `VARIANT`: the two variant cases fail.
- `header-band.test.ts` - put `h-[52px]` back on `DashboardHeader`: the guard names the file and its class string.
- `error-presentation.test.ts` - revert `SettingControls`' `FieldError` to its `<p className="text-xs text-destructive-text">`: the guard names it.
- `DrawerSection.test.tsx` - reinstate a `<Badge variant="secondary" className="ml-auto …">` in the variables tree: the no-filled-badge case fails.
- `SettingsMain.panel-heading.test.tsx` - put a `<CardTitle>Editor</CardTitle>` header back on the editor card: the source half fails, naming the panel.
- `TabStrip.empty.test.tsx` - gate the hint off: the chord case fails.

## Checklist

## Related Issues

Closes #1688
Part of #1683


athrvk:
The `github-advanced-security` check on `dc53632` failed before it looked at the diff: the Copilot code-scanning job (autofind 0.1.136) died creating its session with `CAPIError: 400 The requested model is not supported`, no findings, no repository content analysed. That is GitHub's service, not this PR, and there is no re-run control for that job from here. The rest of the suite (`pr-tests.yml`, CodeQL) is the signal to read for this PR; if the check stays red on a later push it will be re-checked then.

---

Issue #1687, opened 2026-09-19
Icon motion for registry-driven glyphs, and a negative guard for status icons


athrvk:
Part of #1683. Anchors at master `9b9a22f`.

## Problem

The ActivityRail renders its glyphs generically: `app/src/components/layout/ActivityRail.tsx:106` is `<Icon className="size-icon" />` from `constants/drawer-views.ts`. Tab glyphs come from `components/layout/tab-descriptors.ts`, context-bar section glyphs from `components/layout/context-bar/registry.ts`, element glyphs from `element-categories.ts`, palette commands from `lib/commands/registry.ts`. None of these can carry an icon motion without special-casing the renderer.

Separately, nothing stops a future call site from putting `data-icon-motion` on a status icon, which the policy in #1683 forbids.

## Fix pathway

1. Optional `motion?: IconMotion` on the registry entry types above (start with `DRAWER_VIEWS` and `tab-descriptors`; the others only if an entry wants one). Thread it through `ActivityRail.tsx:106` and `RailButton.tsx` as `data-icon-motion={view.motion}`; an entry without one renders no attribute.
2. Pick sensible motions for the rail: Collections `Folder` flap, History `Clock`... no: Clock is a status icon by policy, so History gets no motion; Variables `Braces` scale; Services `Radio` none (live indicator); Trash `Trash2` lid; Settings `Settings` spin-once.
3. Negative guard test `icon-motion-status.test.ts`: scans every `.tsx` under `app/src` and fails if `AlertTriangle`, `AlertCircle`, `CheckCircle2`, `XCircle`, `Clock` or `Info` is rendered with a `data-icon-motion` attribute on the same JSX element. Asserts a non-empty scan.

## Tests

- `ActivityRail.test.tsx`: an entry with `motion` renders the attribute, one without renders none (mutation check: hard-code the attribute in `RailButton`, the "none" case fails).
- The negative guard above, with a mutation check documented in the PR body (add the attribute to one status icon, confirm red, revert).

## Docs

`docs/design-system.md` Icon motion subsection: the registry opt-in and the list of icons that never animate.

## Sequencing

Chain I3, stacked on the `IconSwap` branch.

## Acceptance criteria

- `rg 'motion:' app/src/constants/drawer-views.ts` matches at least three entries.
- `rg 'data-icon-motion' app/src/components/layout/ActivityRail.tsx app/src/components/layout/RailButton.tsx` matches, and neither file names a specific icon.
[...]

Issue #1686, opened 2026-09-19
IconSwap primitive, and every copy button through useCopy with a check swap


athrvk:
Part of #1683. Anchors at master `9b9a22f`.

## Problem

**Six copy buttons bypass `useCopy` and report a failed copy as a success, or as nothing at all.** `app/src/hooks/useCopy.ts:8-24` documents this exact defect (#555, #565). `ResponseActions.handleCopy` (`components/shared/response-viewer/ResponseActions.tsx:49-53`) and `CodeSection.handleCopy` (`components/layout/context-bar/CodeSection.tsx:124-129`) `await navigator.clipboard.writeText` with no catch: a denied clipboard throws, `setCopied` never runs, the user sees nothing. `components/shared/UpdateBanner.tsx:40-45` and `modules/settings/main/panels/UpdatesCard.tsx:71-75` are the same shape. `modules/settings/main/panels/McpSettingsPanel.tsx:213-218` uses `void ....then()` and never reports failure. `modules/request-builder/index.tsx:191-196` catches but toasts instead of swapping the icon. `errors/ErrorBoundary.tsx:161` is the one legitimate direct call (the hook tree may be gone).

**Copy acknowledgement is two idioms plus three durations**: toast (`useCopy`) vs icon check-swap, resetting at 1500ms (`McpSettingsPanel`), 2000ms (`UpdatesCard`, `UpdateBanner`) and `TIMING.STATUS_RESET_MS` (`ResponseActions`, `CodeSection`).

**State-change icons snap.** Copy to Check, Eye to EyeOff (`components/ui/secret-input.tsx`, `components/shared/OAuth2Form/TokenStatusRow.tsx`), Play to Square (`modules/services/.../MockServerControl.tsx`), Folder to FolderOpen (`modules/collections/CollectionItem.tsx`) swap with no transition.

## Fix pathway

1. `useCopy` gains a `copied` boolean in its return (reset after one constant, `TIMING.COPY_RESET_MS`, add to `constants`) alongside the existing toast behaviour, selectable per call (`{ feedback: "toast" | "icon" }`). Route all six sites through it. Rule: icon button uses the check swap, menu item uses the toast.
2. `app/src/components/ui/icon-swap.tsx`, modelled on `label-swap.tsx`: every candidate icon in the same grid cell, `key` on the live one so `.enter-fade` re-fires, twins `aria-hidden`. Export from `components/ui/index.ts`. Apply to the four pairs above and to every copy button's Copy/Check.

[...]

PR #1696, opened 2026-09-19
feat(app): add the icon-motion CSS primitive and the first five hover motions

### LINKED ISSUE #1685: Icon motion: CSS primitive, guard test, and the first five hover motions
Part of #1683 (read its architecture decisions first). Anchors at master `9b9a22f`.

## Problem

Icons are static everywhere. The app has a full motion vocabulary (`app/src/index.css` `--dur-*`/`--ease-*` tokens near `:1084-1091`, `.motion-menu`/`.motion-tooltip`, `.enter-fade`, press feedback `scale` at `:995-1004`) but no way for an icon to respond to the action it sits on.

## Fix pathway

1. **CSS block** in `app/src/index.css`, `/* Icon motion */`, well away from the Monaco overrides (`html .monaco-editor` region ~`:1690`). Rules keyed on `data-icon-motion="<name>"` on the icon element itself, triggered by the nearest owner: `:where([data-slot="button"], .group):is(:hover, :focus-visible) [data-icon-motion="lid"] > path:nth-child(4)` and so on. Names for this PR: `lid` (Trash2, children 4 and 5 lift/tilt -12deg), `spin-once` (RefreshCw, one 360deg turn), `rotate-90` (Plus, X), `nudge-x` / `nudge-y` (ChevronRight / ChevronDown, 1px). Every SVG-child transform declares `transform-box: fill-box` and an explicit `transform-origin`. Use the standalone `rotate`/`translate`/`scale` properties, never `transform: all`. Durations: hover nudges `var(--dur-tooltip-in)` / `var(--ease-enter)`, one-shot spins `var(--dur-panel-in)`. Use a distinct custom property name (`--icon-motion-duration`) so a `.motion-menu` ancestor's `--tw-duration` never leaks in. No `will-change`.
2. **Type**: `app/src/components/ui/icon-motion.ts` exporting `type IconMotion` and a const object of names so call sites cannot typo.
3. **First call sites**: `Trash2` lid at `components/shared/KeyValueEditor/KeyValueRow.tsx`, `components/shared/ElementList/index.tsx`, `components/shared/RowActionsMenu.tsx`; `RefreshCw` spin-once at `components/layout/Dock.tsx` and `components/layout/context-bar/CodeSection.tsx`; `Plus` rotate-90 at `components/layout/TitleBar.tsx` and `TabStrip.tsx`; `X` rotate-90 at the same two; chevron nudge at `components/layout/context-bar/Section.tsx`.

## Tests

`app/src/components/ui/icon-motion.test.ts`, reading `index.css` from disk (the `reduced-motion.test.ts:33` pattern; vitest stubs CSS imports to `""`). Asserts: (i) the scan is non-empty; (ii) every `data-icon-motion` rule's trigger selector contains `:hover` or `:focus-visible` or a `[data-state`/`aria-pressed` attribute, so nothing is always-on (mutation check: drop the `:hover` from one rule, the test fails); (iii) every rule that transforms an SVG child declares `transform-box: fill-box`; (iv) every duration/ease value in the block is a `var(--dur-*)` / `var(--ease-*)` reference, no literals; (v) selector pinning: import `__iconNode` from `lucide-react/dist/esm/icons/trash-2` (and each other targeted icon) and assert the `nth-child` index in the CSS names the path element the rule means. Plus a rendered-class test that the five call sites carry the attribute (a source scan cannot see a class in a variable).

## Docs

`docs/design-system.md`: new "Icon motion" subsection under Motion carrying the policy from #1683 and the table of names; the values it quotes are checked by `design-system-doc.test.ts`, so the doc lands in the same commit as the CSS.

## Sequencing

Chain I1, base master. `IconSwap` + copy fix stacks on this branch; the registry opt-in and status-icon guard stack on that.

## Acceptance criteria

- `rg 'data-icon-motion' app/src --glob '!*.test.*' -l` lists at least the nine call-site files above plus `index.css`.
- `rg 'transform-box: fill-box' app/src/index.css` matches at least once inside the Icon motion block.
- `cd app && pnpm test icon-motion reduced-motion press-feedback design-system-doc` green; `pnpm lint && pnpm type-check && pnpm format:check` green.
- `app/package.json` unchanged.
- With `html[data-reduced-motion="true"]` set, no icon-motion keyframe or transition runs longer than 0.01ms (follows from the existing rules; state it in the CSS comment, assert nothing new).

athrvk:
## Description

Icons in Vayu are static, and there was no mechanism for one to respond to the
action it is the affordance for. This adds that mechanism and the first five
motions that use it: the trash lid lifts, the refresh glyph turns once, plus and
close quarter-turn, a disclosure chevron leans the way it points.

**The whole thing is CSS, and that is the decision worth reviewing.** #1683
rejected a `motion` component per icon (the lucide-animated registry shape) for
two concrete reasons rather than taste: each of those wraps the glyph in a `div`,
which breaks `button-variants.ts`'s `[&_svg:not([class*='size-'])]:size-icon`
sizing and its `[&_svg]:pointer-events-none`; and JS-driven motion is invisible
to both reduced-motion rules at the end of `index.css`, so it would have needed
its own opt-out. A transition or a keyframe is collapsed by them for free -
verified in Chromium rather than assumed (see Testing).

Two things a reviewer would not get from the diff:

- **A motion fires from the owner's hover, never the icon's.** The trigger is
  `:where([data-slot="button"], .group):is(:hover, :focus-visible)`. Four of the
  owners here are not a `Button`, so they gain `group`: a `CollapsibleTrigger`, a
  `DropdownMenuItem`, the Dock's restart link and the new-tab button. If one of
  those loses the class, the motion goes quiet with nothing to see and nothing to
  compile - hence the guard case for it.
- **`transform-box: fill-box` is on the child transforms and deliberately *not*
  on the root-`svg` ones.** The lid's two children hinge about one point (the lid
  bar's left end, user units 3,6), and because `fill-box` resolves percentages
  against each child's own bbox, the two rules carry *different* origin values to
  name the same point (`0% 50%` and `-62.5% 100%`). On the root `svg` the
  percentage resolves against the frame either way, measured: `50% 50%` computes
  to `24px 24px` on a 48px glyph with `fill-box` and with the default `view-box`
  alike, so declaring it there would read as load-bearing and do nothing. My
  first draft had it on all five with a comment claiming it mattered; the browser
  probe is what caught that.

`RowAction` gains an `iconMotion` field rather than each menu setting the
attribute: `RowActionBody` draws the item once for both the `⋯` menu and the
right-click menu, so a second copy would be free to drift. That is also the one
call site whose name arrives in a variable, so it is rendered and read off the
element rather than scanned for.

The Dock's glyph opts out while a restart is actually running - a one-shot turn
fighting the `animate-spin` loop reads as a stutter, and there the loop is the
message.

**Safe to skim:** the `docs/design-system.md` subsection (this repo's
same-commit doc rule; `design-system-doc.test.ts` is why it rides with the CSS
rather than with the call sites), the `ICON_MOTION` barrel export, and the import
re-sorting in the five layout files.

**Risk, stated as the second-order effect:** a `data-icon-motion` name the
stylesheet has no rule for is inert - it renders, it looks deliberate, and
nothing errors. That is what the type exists to prevent and what the guard's
"every name has rules" case covers. The converse - a rule whose attribute no call
site sets - is dead CSS, covered by the call-site file.

#1686 and #1687 stack on this branch, so the primitive's API is one const
object plus one type and the CSS block is self-contained.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

New: `app/src/components/ui/icon-motion.test.tsx` (the stylesheet block) and
`icon-motion.call-sites.test.tsx` (the call sites). Both read `index.css` from
disk, because vitest stubs CSS imports to `""`; both assert the scan read
something before asserting anything about it.

```
$ cd app && pnpm test icon-motion reduced-motion press-feedback design-system-doc
 Test Files  5 passed (5)
      Tests  32 passed (32)

$ cd app && pnpm lint && pnpm type-check && pnpm format:check
eslint . --report-unused-disable-directives --max-warnings 0      (clean)
[renderer]       tsc --noEmit                                     exited with code 0
[electron]       tsc -p tsconfig.node.json --noEmit               exited with code 0
[electron-tests] tsc -p tsconfig.electron-test.json               exited with code 0
Checking formatting... All matched files use Prettier code style!

$ cd app && pnpm test                     # run 1, 1629s
Test Files  2 failed | 771 passed (773)
     Tests  3 failed | 8730 passed | 2 skipped (8735)

$ cd app && pnpm test                     # run 2
Test Files  6 failed | 767 passed (773)
     Tests 10 failed | 8723 passed | 2 skipped (8735)
```

**Every failure in both runs is a timeout, and not one is an assertion** -
grepping run 2's log for error kinds gives only `Test timed out in 5000ms` (x6),
`20000ms` (x1) and `30000ms` (x1), no `AssertionError` at all. They are the
wall-clock-under-load mode `app/CLAUDE.md` documents (#846, #1280), on a box
shared with four other agents building concurrently: the cases that fail are the
ones that render 500-row fixtures (`ResponseEvents.window`, `InboxView.captures`,
`run-view-navigability`, two `CollectionTree` files) plus
`monaco-setup.contributions`, which runs a real vite build inside a 30s budget
whose comment records it at 6.7s. Run 1's two `ResponseEvents.window` failures
pass on their own (`pnpm test ResponseEvents` gives 5 passed), and the sets
differ between runs, which is itself the signature of load rather than of the
diff. Nothing here reaches the response viewer, the inbox, the collection tree
or monaco.

### Acceptance criteria from #1685

```
$ rg 'data-icon-motion' app/src --glob '!*.test.*' -l
app/src/components/layout/Dock.tsx
app/src/components/layout/TabStrip.tsx
app/src/components/layout/TitleBar.tsx
app/src/components/layout/context-bar/CodeSection.tsx
app/src/components/layout/context-bar/Section.tsx
app/src/components/shared/KeyValueEditor/KeyValueRow.tsx
app/src/components/shared/RowActionBody.tsx
app/src/components/ui/icon-motion.ts
app/src/index.css

$ rg 'icon-motion|iconMotion|ICON_MOTION' app/src --glob '!*.test.*' -l
(the nine above, plus)
app/src/components/shared/ElementList/index.tsx
app/src/components/shared/row-actions.ts
app/src/components/ui/index.ts

$ rg 'transform-box: fill-box' app/src/index.css
1234: * 2. **Every transform of an SVG *child* declares `transform-box: fill-box` and
1287:	transform-box: fill-box;

$ git diff 9b9a22f..HEAD -- app/package.json
(empty - no dependency added or changed)
```

One deviation to note: the issue's grep expects `RowActionsMenu.tsx` and
`ElementList/index.tsx` among the files carrying the attribute. Neither owns the
glyph - `RowActionBody` draws it for both menus - so they carry `ICON_MOTION.lid`
and the `group` class instead, and the second grep above is the one that lists
all twelve files. Keeping the attribute out of those two files is what stops the
two menus from drifting apart.

### Reduced motion

Driven in Chromium over CDP (no engine needed; playwright-core is not installed
in this environment, so the probe drives the browser directly) against the real
stylesheet under vite, with `html[data-reduced-motion="true"]` set. Every
transition duration and the `icon-spin-once` animation compute to `1e-05s`, i.e.
0.01ms - the criterion holds through the existing rules, so nothing new is
asserted. The same probe is what confirmed the origins land where the comments
say: the lid bar's origin resolves to user units (3,6) and the handle's to (3,6),
the same hinge from two different percentage pairs.

### Mutation checks

Each reverted after confirming the failure; `git status --short` is empty.

| Mutation | Test that failed |
|---|---|
| Drop `:hover` from the `rotate-90` trigger | gates every rule that moves a glyph on hover…; triggers from the owning button or group… |
| Drop `transform-box: fill-box` from the lid's base rule | declares a transform-origin for every target it moves, and fill-box for every child |
| Replace a token pair with `120ms cubic-bezier(0.2, 0, 0, 1)` | takes every duration and curve from a token, never a literal |
| Point the lid rule at `nth-child(3)` | names the lid bar and its handle, not the tines or the can |
| Drop `group` from `rowActionItemClass` | gives a menu item the group its glyph's motion triggers from |
| Stop passing `action.iconMotion` in `RowActionBody` | puts a row action's motion on its glyph… |
| Give the collapsed chevron `nudgeY` | components/layout/context-bar/Section.tsx; swaps the chevron's motion with the disclosure's direction |

## Checklist

## Related Issues

Closes #1685. Part of #1683. #1686 stacks on this branch.


athrvk:
The `github-advanced-security` check on `5525685` failed before analysing anything: the Copilot code-scanning job (autofind 0.1.136) died creating its session with `CAPIError: 400 The requested model is not supported`, the same service error #1695 hit a few minutes earlier on an unrelated diff. Not this PR's, and that job has no re-run control from here. `pr-tests.yml` and CodeQL are the signals for this PR.

---

PR #1697, opened 2026-09-19
fix(app): wayfinding and accessibility - crumbs on detail tabs, non-colour run status, import live region, path-first URLs

### LINKED ISSUE #1691: Wayfinding and accessibility: TabBreadcrumb on detail tabs, non-colour run status, import live region, role=img audit, params placeholder checkbox, history URL truncation, centred Launcher, palette over a dialog

## Problem

1. **Breadcrumbs exist on one surface.** `modules/request-builder/components/RequestBreadcrumb.tsx:41`, used only at `RequestBuilderLayout.tsx:128`. `CollectionDetail/index.tsx`, `history/main/LoadTestDetail.tsx`, `dashboard/index.tsx` and the mock-server tab open with no "where am I" line.
2. **Run status in the history sidebar is colour-only**: the left bar pattern (design system, Status Badges) in `modules/history/sidebar/RunItem.tsx`; the row text does not always repeat the status.
3. **Import progress has no live region** (`modules/collections/ImportProgressView.tsx`) while `ExportSpecDialog.tsx:138,226` and `CollectionTree.tsx:365-367` announce with `role="status"`.
4. **Two `role="img"` icons need checking**: `RunItem.tsx:312`, `components/ui/tabs.tsx:249`. A decorative icon with `role="img"` and no `aria-label` is announced as an unlabelled image; one with a label beside visible text double-announces.
5. **The empty Params row shows a checked accent checkbox** on the placeholder row (`components/shared/KeyValueEditor/KeyValueRow.tsx`), so it reads as an enabled parameter before anything is typed.
6. **History rows truncate the URL at the host**: unsaved runs render as `http://127.0.0.1:9...` so every row looks identical (`modules/history/sidebar/RunItem.tsx`, `TruncatedText`).
7. **The Launcher is left-anchored** in a wide pane (`modules/welcome/Launcher.tsx`): tiles and recent runs hug the left edge, the right two-thirds is empty at 1440px.
8. **Verify: the command palette can render over an open dialog.** The chord path is guarded (`Shell.tsx:274,394` `isModalOpen()`), the click path `components/layout/CommandSearchBar.tsx:36` `setPaletteOpen(true)` is not. Reproduce manually first; if the overlay blocks the click it is a non-issue, record that in the PR body.

## Fix pathway

1. Generalise `RequestBreadcrumb` to `components/shared/TabBreadcrumb.tsx` taking an entity descriptor (collection path, run, dashboard) and render it in the three detail tabs; each crumb navigates.
2. `RunItem`: add the status word or a small glyph beside the bar, or `aria-label` on the bar plus a visible text repeat; pick the visible option.
3. `ImportProgressView`: `role="status" aria-live="polite"` on the step/count line.
4. Fix the two `role="img"` sites as the audit dictates.
5. Placeholder row's checkbox renders muted/unchecked (or hidden) until the key is non-empty; the row's `KeyValueRow.test.tsx` covers it.
6. URL-titled rows use path-first or middle truncation: a `truncateUrl` helper in `lib/` that keeps the path tail; unit-tested.
7. Launcher content in a centred `max-w-*` column (pick from the existing dialog width scale).
8. If reproducible, `CommandSearchBar` asks `isModalOpen()` before opening.

## Tests

- `TabBreadcrumb.test.tsx` and one render assertion per adopting tab.
- `RunItem.test.tsx`: status conveyed by text, not only class (mutation check: remove the glyph/text, fails).
- `ImportProgressView.test.tsx`: `role="status"` present.
- `KeyValueRow.test.tsx`: placeholder row checkbox state.
- `truncate-url.test.ts`.
- `Launcher.test.tsx` rendered-class assertion for the column.
- `CommandSearchBar.test.tsx` if item 8 is real.

## Docs

`docs/design-system.md` Status Badges (non-colour redundancy), Accessibility (live region rule for long operations), Overflowing Text (URL truncation). `docs/app/COMPONENTS.md` for `TabBreadcrumb`. `modules/welcome/README.md` if the Launcher layout rule changes.

## Sequencing

Chain W1, base master.

## Acceptance criteria

- `rg '<TabBreadcrumb' app/src --glob '!*.test.*' -l` includes `CollectionDetail/index.tsx`, `LoadTestDetail.tsx`, `dashboard/index.tsx`, `RequestBuilderLayout.tsx`.
- `rg 'role="img"' app/src --glob '!*.test.*'` returns only sites with an `aria-label` on the same element.
- `rg 'aria-live|role="status"' app/src/modules/collections/ImportProgressView.tsx` matches.
- `cd app && pnpm test TabBreadcrumb RunItem ImportProgressView KeyValueRow truncate-url Launcher` green; `pnpm lint && pnpm type-check && pnpm format:check` green.

athrvk:
## Description

Eight wayfinding and accessibility defects from #1691, roughly one commit each.
Three decisions are worth explaining; the rest read from the diff.

**The crumb was a component, and is now a component plus a caller's data.**
`RequestBreadcrumb` knew about requests, collections and tabs, so the other three
detail tabs could not use it without importing the builder's context or growing a
fourth copy of a crumb row. `components/shared/TabBreadcrumb` takes
`{ id, label, onSelect? }[]` and owns only what is easy to get wrong in such a
row - which segment shrinks, which one is a control, and drawing nothing rather
than an empty band. The builder's own knowledge moved to `useRequestCrumbs`. Two
existing layout tests mocked `./RequestBreadcrumb` to keep a QueryClient out of
their render; they now mock that hook, which is the reason it is a separate module
rather than inline in the layout.

**`CollectionDetail` draws the chain only when the collection is nested.** For a
root collection the chain is one segment, and that segment is the header title
repeated underneath itself. Worth knowing before reading `ancestors.length > 1` in
the diff as a guard against an empty array.

**`CollectionDetail` walks the collections it already holds instead of calling
`useCollectionAncestors`.** It fetches the whole list for its subtree count, so a
second query subscription over the same data buys nothing - and its existing
suites mock `@/queries/collections` module-wide, so the hook would have meant
editing five unrelated test files to add a function none of them care about. The
cycle guard still comes from `walkAncestors`.

Item 8 was a verification task, not a fix - see **Item 8** below. It produced a
comment and no code.

### Risk, and what to skip

- **Safe to skim:** the docs (`docs/design-system.md` Status Badges /
  Accessibility / Overflowing Text, `docs/app/COMPONENTS.md`,
  `modules/welcome/README.md`). This repo's rule puts them in the same commit as
  the code they describe; they describe it rather than changing it. The new test
  files likewise.
- **Second-order effect, and it already bit twice:** the spare key/value row now
  renders a spacer where its checkbox was, so anything proving "the table
  rendered" by finding a checkbox breaks - `BodyPanel.test.tsx` did exactly that
  and now recognises the table by its column headings. And `LoadTestDetail`'s
  crumb changes the characterization snapshot, which stores the whole pane as one
  HTML string. A future test written against "one checkbox per row" would be wrong
  rather than failing loudly.
- **Worth a look:** `RunItem`'s identity line renders a *shortened* string for a
  URL rather than leaving the cut to CSS, so anything reading the row's visible
  text as a URL would be reading a display value. Nothing does - the full value is
  in `title` and in the row's accessible name - and the `truncate` class stays,
  because a 48-character budget is not a promise about a narrow drawer.
- `WELCOME_COLUMN` moved the width cap off `WelcomeScreen`'s scroller and onto the
  three states inside it, so a fourth state added later and given no column would
  be full-bleed with nothing failing. `Launcher.test.tsx` covers the three that
  exist, together, because a skeleton in a different column from its content
  shifts the page the moment the queries land.

## Visual changes

No screenshot: this worktree has no engine build, and the populated states (a
nested collection tab, a finished run report, a history list) need one. What
changes on screen is a crumb line above four detail tabs, a per-status glyph in
place of a coloured dot in the history sidebar, path-tail URLs in those rows, no
checkbox on the spare params row, and a centred welcome column - each asserted as
rendered text or a rendered class in the tests below, and the run pane's whole
markup is in the refreshed snapshot.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`pnpm lint`, `pnpm type-check` and `pnpm format:check` green. The acceptance
criteria from the issue, run as written on this branch:

```
$ rg '<TabBreadcrumb' app/src --glob '!*.test.*' -l
app/src/modules/collections/CollectionDetail/index.tsx
app/src/modules/dashboard/index.tsx
app/src/modules/request-builder/components/RequestBuilderLayout.tsx
app/src/modules/history/main/LoadTestDetail.tsx

$ rg 'role="img"' app/src --glob '!*.test.*'
app/src/components/ui/tabs.tsx:            role="img"     # aria-label={label}, same element
app/src/modules/history/sidebar/RunItem.tsx:  role="img"   # aria-label="This run has warnings - see its report"

$ rg 'aria-live|role="status"' app/src/modules/collections/ImportProgressView.tsx
role="status"
aria-live="polite"

$ pnpm test TabBreadcrumb RunItem ImportProgressView KeyValueRow truncate-url Launcher
Test Files  13 passed (13)
     Tests  71 passed (71)
```

Full suite, second pass on this branch:

```
Test Files  1 failed | 780 passed (781)
     Tests  4 failed | 8738 passed | 2 skipped (8744)
```

Those four are the `historyDetail.characterization` snapshot cases, refreshed in
the last commit; `pnpm test characterization BodyPanel TabBreadcrumb RunItem
ImportProgressView KeyValueRow truncate-url Launcher` is 111 passed after it. The
first pass also found the `BodyPanel` regression above, plus timeout failures in
`DesignRunView`, `CollectionTree.run-order.conformance`, `monaco-setup.contributions`
and the two `InboxView` files - all green in isolation here, and the machine-load
failures `app/CLAUDE.md` describes: five agents share this box and the per-case
budget is 5s.

### The snapshot refresh

Verified per case before regenerating, by diffing the old and new strings: exactly
one insertion in each of the four, the `<nav aria-label="Run location">` element,
with nothing removed and nothing else in the pane moved. The dashboard hero
snapshot is untouched, because the dashboard's crumb sits outside
`DashboardHeader`, which is what that file renders.

### Mutation checks

Each new behavioural guard reverted, confirmed red, restored:

| Guard | Mutation | Result |
| --- | --- | --- |
| `RunItem.status.test.tsx` | every status given the same `icon`, colours left different | red - the comparison strips `class` attributes, so five colours over one shape is what it rejects |
| `role-img-labelled.test.ts` | `aria-label` removed from `TabErrorDot` | red |
| `KeyValueRow.test.tsx` | placeholder condition dropped (`allowDisable ? (`) | red |
| `RunItem.url.test.tsx` | `identityText` handed the raw `requestUrl` | red |
| `Launcher.test.tsx` | `mx-auto` dropped from `WELCOME_COLUMN` | red - the classes are spelled out in the test rather than read from the constant, which would have passed on any value |

`TabBreadcrumb.test.tsx`, the three per-tab adoption tests and
`truncate-url.test.ts` are behavioural tests over new code, so their failure mode
is the code not existing.

## Item 8: the palette cannot be opened over a dialog

Reproduced before changing anything, in Chromium (playwright, the renderer under
vite, no Electron) against the app's own `DialogContent` and `CommandSearchBar`:

```
dialog present: 1
search bar count: 0                 # getByRole finds nothing - Radix marks
                                    # everything outside the dialog aria-hidden
real click: BLOCKED -> TimeoutError: locator.click: Timeout 4000ms exceeded
state after real click: palette-closed
focused after tabbing: BUTTON       # inside the dialog - the focus trap holds
state after scripted click: palette-open
body pointer-events: none
bar aria-hidden ancestor: true
```

The click path is unreachable while a dialog is open: the overlay and
`pointer-events: none` stop the pointer, `aria-hidden` takes the button out of the
accessibility tree, and the focus trap stops Tab. Only `element.click()` from a
script reaches the handler. So no `isModalOpen()` guard - it would be dead code
defending a door the dialog has already locked - but the asymmetry with the chord
path is now a comment in `CommandSearchBar`, because the next reader will ask the
same question.

## Checklist

## Related Issues

Closes #1691
Part of #1683


athrvk:
The `github-advanced-security` check on `b65c76b` failed before analysing anything: the Copilot code-scanning job (autofind 0.1.136) died creating its session with `CAPIError: 400 The requested model is not supported`, the same GitHub-side error every PR in this programme has hit today (#1695, #1696, #1698, #1699). Not this PR's, and that job has no re-run control from here. `pr-tests.yml` and CodeQL are the signals for this PR.

---

Issue #1690, opened 2026-09-19
Feedback: DisabledHint for gated controls, Services row hover-reveal, keyboard Move up/down in the collection tree


athrvk:
Part of #1683. Anchors at master `9b9a22f`.

## Problem

1. **163 `disabled=` props and no disabled control explains itself.** `Button` renders `pointer-events-none` when disabled, so a wrapping `Tooltip` never fires. Gating cases a user meets: `modules/inbox/index.tsx:336` (Clear when no captures), `components/shared/ElementList/index.tsx:243-244` (Move up/down at the ends), `modules/request-builder/components/LoadTestConfigDialog/index.tsx:1266`, `ProfilePicker.tsx`, `CustomBudgetRows.tsx`, the Send button while a request is in flight, Save when nothing is dirty.
2. **Services rows are the only rows without hover-reveal actions.** `group-hover:opacity-100` is used in `CollectionItem.tsx`, `RequestItem.tsx`, `RunItem.tsx`, `VariablesCategoryTree.tsx`, `VariableTableEditor.tsx`, `KeyValueRow.tsx`; `modules/services/ServicesPanel.tsx:289-292` paints three always-visible `TooltipIconButton`s. Trash's always-visible actions are documented in the Row Actions section; Services is not.
3. **Reorder has two vocabularies.** The collection tree reorders by drag only (`useTreeDnd.ts`, `TreeRowDnd.tsx`); `ElementList/index.tsx:243-244` reorders by Move up / Move down menu items and has no drag. Keyboard users cannot reorder the tree at all.

## Fix pathway

1. `components/ui/disabled-hint.tsx`: wraps a disabled control in a `span` with `pointer-events-auto` and `tabIndex={0}` carrying the tooltip (`TooltipContent` with the reason; use `TooltipHint` for secondary lines per the tooltip rules). Apply to the top gating cases listed, each with a specific reason string ("No captures to clear", "Already first").
2. Services rows adopt `RowActionsMenu` + `group-hover` reveal like every other row, or the Row Actions section gets a one-line exemption naming why Services differs. Prefer parity.
3. Add "Move up" / "Move down" to the tree row menu (`CollectionTree` context menu and `RowActionsMenu` entries) calling the existing reorder path from `reorder-math.ts` / `useTreeCrud.ts`; disabled at the ends with a `DisabledHint`.

## Tests

[...]

PR #1698, opened 2026-09-19
fix(app): commit an inline rename on isCommitEnter only, and never on Escape

### LINKED ISSUE #1684: Inline rename: commit on isCommitEnter only, Escape never commits, one useInlineRename hook for all three trees

## Problem

**Seven text inputs commit on a bare `e.key === "Enter"`**, against the rule in `app/CLAUDE.md` (#939, #935): an IME composition commit reaches the handler as an ordinary keydown, and `Cmd/Ctrl+Enter` (the Send chord) both commits the field and reaches the window handler, so renaming a request with Cmd+Enter renames it and sends it in one press.

- `app/src/modules/collections/RequestItem.tsx:232` (rename)
- `app/src/modules/collections/CollectionItem.tsx:312` (rename) and `:406` (new subfolder)
- `app/src/modules/collections/CollectionTree.tsx:253` (new collection)
- `app/src/modules/collections/CollectionDetail/SpecTab.tsx:350` (fetch spec URL)
- `app/src/modules/services/ServicesPanel.tsx:553` (commit)
- `app/src/components/layout/context-bar/VariableRow.tsx:83`

The `role="button"` / `role="tab"` activation sites at `variable-popover.tsx:429`, `markdown-editor.tsx:132`, `TabStrip.tsx:107` are correct and stay.

**Escape during an inline rename can still commit it.** Both tree renames call the cancel on Escape and the submit on blur; cancelling unmounts the `Input`, and the resulting blur races the cancel. `RequestItem.tsx:232-241`, `CollectionItem.tsx:312-321`.

**Three hand-rolled inline-rename editors with three commit models**: the collections tree (Enter/Escape/blur-commits, focus return via `returnFocusToRow` refs at `RequestItem.tsx:91-94`, `CollectionItem.tsx:128-131`), `components/shared/ElementList/index.tsx:230-234,300` (`isCommitEnter`, blur commits, `commitRename` trims), and `modules/variables/.../VariablesCategoryTree.tsx:604` (its own branch, focus return by `data-environment-id` lookup).

## Fix pathway

1. New `app/src/hooks/useInlineRename.ts` (+ optional `<InlineRenameInput>` in `components/shared/`) owning: value state, `isCommitEnter` commit, Escape cancel that sets a `cancelledRef` so the following blur is a no-op, blur commit, trim, empty-value handling (cancel, never commit an empty name), and focus return to the row element passed in.
2. Replace the three editors with it. Replace the remaining four bare-Enter sites (`CollectionItem.tsx:406`, `CollectionTree.tsx:253`, `SpecTab.tsx:350`, `ServicesPanel.tsx:553`, `VariableRow.tsx:83`) with `isCommitEnter(e)` directly where they are not renames.

## Tests

- `useInlineRename.test.tsx` (jsdom): Enter commits, `Cmd+Enter` does not commit, an Enter with `isComposing` does not commit, Escape then blur does not commit (mutation check: remove the `cancelledRef` guard and this case fails), empty value cancels, focus returns to the row.
- Extend the existing `CollectionTree.rename.test.tsx` with the Escape-then-blur case.
- A source guard `commit-enter.test.ts`: no `e.key === "Enter"` on an `<Input>`/`<input>`/`<textarea>` `onKeyDown` in `app/src` outside the allowlisted role-activation sites; must assert it scanned a non-empty file set.

## Docs

`docs/design-system.md` Component Patterns: an "Inline rename" entry naming the hook. `docs/app/COMPONENTS.md` if a shared component is added.

## Sequencing

Chain R1, first to land. The Feedback sub-issue (Services hover-reveal, tree Move up/down) stacks on this branch because it touches the same tree files.

## Acceptance criteria

- `rg 'e\.key === "Enter"' app/src --glob '!*.test.*'` returns only `variable-popover.tsx`, `markdown-editor.tsx`, `TabStrip.tsx` (role activation sites), or the guard test's allowlist, whichever is smaller.
- Exactly one implementation of inline rename keyboard handling exists: `rg 'onRenameCancel\(\)|onRequestRenameCancel\(\)' app/src --glob '!*.test.*'` finds no `onKeyDown` handler outside the hook.
- `cd app && pnpm test useInlineRename CollectionTree.rename commit-enter` green; `pnpm lint && pnpm type-check && pnpm format:check` green.

athrvk:
## Description

Three trees renamed a row in place with three hand-rolled keyboard handlers, and each had missed a different case. Two of them were user-visible bugs: renaming a request with `Cmd+Enter` renamed it *and* sent the request (the field acted on the chord and the chord still reached the window handler), and Escape could still save the name it had just cancelled, because cancelling unmounts the field and the blur that follows was wired to commit.

The fix is one hook, `app/src/hooks/useInlineRename.ts`, owning the whole contract - `isCommitEnter`, an Escape that closes the editor once so the following blur is inert, the trim, an empty value that cancels instead of erasing a name, and focus back to the row after a keyboard close only. Worth reading closely:

- **The draft name moved into the row's hook and out of `useTreeCrud`.** The rename field is mounted by exactly one row at a time, so the copy in the tree context was a value two layers could disagree about; submit now carries the name (`onRenameSubmit(id, name)`). Second-order effect for a reviewer: `renameValue` / `onRenameChange` are gone from `CollectionTreeCrudSlice`, so a future caller that tried to seed a draft through the context gets a type error rather than a silently ignored write. `useTreeCrud` keeps its own unchanged-name check - the hook trims, it does not know what the entity is currently called.
- **`commitEmpty` is a real exception, not a loophole.** An element's name is optional and blanking it restores the kind label, so `ElementList` must be able to commit an empty value. Left as a default the hook would have quietly regressed that behaviour, which is why it is an explicit option named for its one caller.
- **The four remaining bare-Enter fields (new collection, spec URL, Services delay, context-bar variable value) are not renames**, so they take `isCommitEnter` in place and keep their own handlers rather than being bent into the hook.

Safe to skim: the mechanical half - the four one-line `isCommitEnter` swaps, the test-helper and `rename-status` test updates that follow the context signature, and the docs section the repo's own rule requires in the same change.

Not in scope, by intent: the `role="button"` / `role="tab"` activation sites (`variable-popover.tsx`, `markdown-editor.tsx`, `TabStrip.tsx`) and `useRovingTreeFocus`'s row activation, where "Enter or Space activates" is the platform rule being implemented and Space is checked on the same line. `commit-enter.test.ts` names exactly those and fails on anything else.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`useInlineRename.test.tsx` drives the hook through a row shaped like the real ones (Enter commits trimmed, `mod+Enter` and an `isComposing` Enter do not, empty cancels, focus returns on a keyboard close and not on a blur), plus two cases that hold `active` true across the keydown and the blur - that window is the race, and the integration level cannot reproduce it because React has already unmounted the field before a blur could reach a handler. `CollectionTree.rename.test.tsx` gains the two cases that are visible end to end: Escape sends no PUT and leaves the row focused, and the Send chord leaves the field open with nothing sent. `commit-enter.test.ts` is the source guard, and asserts it scanned a non-empty tree (>100 files, `lib/keyboard.ts` among them), that its allowlist has not gone stale, and that no allowlisted file has grown a text field beside its comparison.

Mutation checks run (each reverted after):

| Mutation | Result |
|---|---|
| `isCommitEnter(e)` → `e.key === "Enter"` in the hook | `useInlineRename` "does not commit on the Send chord" and "does not commit the Enter that commits an IME composition" fail; `CollectionTree.rename` "leaves the Send chord to the window" fails |
| drop `if (closedRef.current) return` from `commit` | `useInlineRename` "does not commit the blur that follows Escape" and "does not commit a second time on the blur that follows Enter" fail (the `CollectionTree.rename` Escape case still passes, for the unmount reason above - the hook test is where that guard is pinned) |
| restore a bare Enter in `ServicesPanel.tsx` | `commit-enter.test.ts` fails, naming `modules/services/ServicesPanel.tsx` |

Acceptance criteria from #1684, on this branch:

```
$ rg 'e\.key === "Enter"' app/src --glob '!*.test.*'
app/src/modules/collections/useRovingTreeFocus.ts:					if (e.key === "Enter" && e.shiftKey) click("[data-tree-menu]");
app/src/lib/keyboard.ts: * is invisible when absent - nothing about a plain `e.key === "Enter"` looks
app/src/lib/keyboard.ts:	return e.key === "Enter" && !e.nativeEvent.isComposing && !e.ctrlKey && !e.metaKey;
app/src/modules/request-builder/components/UrlBar/SendWithRowDialog.tsx:	 * `isCommitEnter` rather than `e.key === "Enter"`: an Enter carrying Ctrl/Cmd
app/src/hooks/useInlineRename.ts: * on a bare `e.key === "Enter"`, `ElementList` used `isCommitEnter` and trimmed
app/src/components/layout/TabStrip.tsx:					if (e.key === "Enter" || e.key === " ") {
app/src/components/ui/markdown-editor.tsx:							if (e.key === "Enter" || e.key === " ") {
app/src/components/ui/variable-popover.tsx:				if (e.key === "Enter" || e.key === " ") {
```

The code hits are the guard's allowlist (the issue's "whichever is smaller" clause): the three activation components, the definition in `keyboard.ts`, and `useRovingTreeFocus`'s row activation. The other three lines are prose in comments.

```
$ rg 'onRenameCancel\(\)|onRequestRenameCancel\(\)' app/src --glob '!*.test.*'
(no output)
```

Nothing calls either cancel from a handler any more: both are passed to the hook as `onCancel`.

```
$ cd app && pnpm test useInlineRename CollectionTree.rename commit-enter
Test Files  3 passed (3)
     Tests  24 passed (24)
```

`pnpm lint`, `pnpm type-check` and `pnpm format:check` are green. The full suite ran in two shards: 4091 passed / 1 skipped and 4625 passed / 1 skipped, with three failures that are all the documented render-window flake under shard load, not this change - `ResponseEvents.window.test.tsx` fails on `9b9a22f` too (two cases there, against one here), and `run-view-navigability` / `InboxView.captures` pass on this branch when re-run, twice.

## Checklist

## Related Issues

Closes #1684. Part of #1683; chain R1, so #1690 stacks on this branch.


athrvk:
The `github-advanced-security` check on `c609d74` failed before analysing anything: the Copilot code-scanning job (autofind 0.1.136) died creating its session with `CAPIError: 400 The requested model is not supported`, the same GitHub-side error every PR in this programme has hit today (#1695, #1696, #1697, #1699). Not this PR's, and that job has no re-run control from here. `pr-tests.yml` and CodeQL are the signals for this PR.

---

PR #1699, opened 2026-09-19
fix(app): confirm every destructive action the same way, and write the reversibility rule down

athrvk:
## Description

Destructive actions were decided per module: Clear wiped every inbox capture from one click while Delete inbox beside it confirmed, Trash purged one row at a time with no way to empty it, and the confirm dialogs said "permanently removed", "removed for good" and "deleted" for the same act. #1683 settled the policy (collections and requests go to Trash with Undo; everything else confirms and deletes permanently; nothing destructive runs from a single click without one of the two) and this PR makes the app match it and writes it into the design system so it stops being re-decided.

Three decisions worth a reviewer's attention:

- **`DeleteConfirmDialog` builds its own sentence from `name` and `scope`.** `title`/`description` still win when passed, so an untouched call site renders exactly what it did. The straightforward sites (environment deletes, Trash purge) moved onto the template; the ones whose prose is conditional or count-based (`HistoryList`, `GeneralPanel`, `DeleteInboxDialog`, `SpecSync`) were reworded to the one phrasing without being bent into `name`/`scope`. The second-order effect: a new call site can still hand-write a description and drift. The acceptance grep catches the three old phrasings, not a fourth one; that is the guard's real reach.
- **Empty trash loops the per-row purge.** The engine has only `POST/DELETE /trash/:id` and the parent issue rules out a new engine route in this programme, so the header action confirms once and purges serially, reporting a row that fails without stopping the rest. A very large trash therefore takes N round trips; acceptable for a drawer that retains 30 days, and the place to change if that ever hurts is one function.
- **"Folder" is the word.** `DEFAULT_FOLDER_NAME` is `"New Folder"`, and the create-folder form's internal state was renamed to match (`newFolderName`, `onFolderNameChange`). The rename handlers were left alone because #1698 owns them.

`deleteConfirmCopy` lives in its own module because `react-refresh/only-export-components` flags a non-component export in the dialog's file.

**Safe to skim:** the docs additions (`#### Reversibility` under Destructive Actions, `### Loading` under Component Patterns with the Title Case entity-name rule folded in), the two test-helper edits that follow the state rename, and `collection-constants.test.ts`, which is a guard over a string.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Acceptance criteria from #1689, run on this branch:

```
$ rg -i 'sub ?collection' app/src docs --glob '!*.test.*'
(no output)

[...]

PR #1700, opened 2026-09-19
feat(app): add IconSwap and route every copy button through useCopy

### LINKED ISSUE #1686: IconSwap primitive, and every copy button through useCopy with a check swap

athrvk:
## Description

**Was stacked on #1696** (`claude/happy-ride-ysw1fi-icon-motion`), which has merged; GitHub retargeted this PR to `master` and the branch has merged `master` back in (one doc conflict, the Component Patterns section, resolved by keeping both this branch's Copy feedback entry and #1698's Inline rename entry). The diff is now this branch's own commits only. #1701 stacks on this branch.

Two things were wrong.

**A denied clipboard was silent.** Six copy buttons went to `navigator.clipboard.writeText` directly. Four awaited it with no catch, so a rejection - a denied permission, an unfocused document, a platform with no clipboard behind the API - threw past the line that sets the check, and the user's only evidence was pasting the *previous* clipboard contents somewhere else. The MCP settings panel put its feedback inside a `.then()`, which fails the same way. Each of those sites also owned a `copied` flag and a `setTimeout`, which is how one action came to be acknowledged over three different durations in the same app. `useCopy` already existed for exactly this and was bypassed anyway, so the rule is now enforced on the source rather than restated in a docstring: `clipboard-single-writer.test.ts` allows two files, the hook and `errors/ErrorBoundary.tsx`.

`errors/ErrorBoundary.tsx` keeps its direct call deliberately - it runs when the tree below it has already failed, so no hook is reachable from it, and it catches its own rejection.

**`useCopy`'s return type changed**, from a bare function to `{ copy, copied }`. That is the one thing in this diff that touches files with no other reason to be here (eight `const copy = useCopy()` sites became `const { copy } = useCopy()`); a callable with a property bolted onto it was the alternative and is worse to live with. Nothing else about those call sites changed.

**State-change icons snapped.** `IconSwap` is `LabelSwap`'s mechanism with glyphs instead of words - every candidate in one grid cell, reserve twins `invisible h-0` and `aria-hidden`, `key` on the live cell so `.enter-fade` replays. Worth knowing before reading it: it is a *different mechanism* from #1685's `data-icon-motion`, not a variant. A motion animates one glyph that stays itself, off its owner's hover; a swap crossfades between two glyphs, off state. `IconSwap` adds nothing to the nodes it is handed, so an icon can carry both - there is a test for exactly that, because a future reader's first instinct will be to unify them.

**`TIMING.STATUS_RESET_MS` is gone**, in its own commit. The copy tick was its only reader and this branch took both call sites away, which `timing-keys-have-readers.test.ts` correctly failed on: a TIMING key with nothing behind it is the "written but never read" defect, and CLAUDE.md points at that file as where a millisecond value lives. The three prose mentions that survive are comments, not readers (that guard strips comments before scanning), so they describe the duration rather than naming a dead key.

## Deviation from the issue: Play to Square

#1686 lists Play → Square in `MockServerControl.tsx` as a pair. It is one from the user's side, but the two glyphs live in *separate branches* of that component: a stopped server renders a labelled "Run mock server" button, a running one an icon button in a chip row. Nothing can crossfade across that boundary without merging the two controls into one, which is outside this issue's scope, and doing it here to satisfy a grep would have produced an `IconSwap` whose state never changes. What does change in place is each button's pending state, so that is what swaps - a spinner where a disabled glyph used to be the only feedback. Called out because the acceptance criterion's `rg '<IconSwap'` passes for this file either way, and only this paragraph says which.

## Type of Change
- [x] Bug fix (the silent clipboard failure)
- [x] New feature (the `IconSwap` primitive)
- [ ] Breaking change
- [x] Documentation update (`docs/design-system.md`: `IconSwap` under Motion, a Copy feedback rule under Component Patterns)

## Testing

Acceptance criteria from #1686, run on this branch:

```
$ rg 'clipboard\.writeText' app/src --glob '!*.test.*' -l
app/src/errors/ErrorBoundary.tsx
app/src/hooks/useCopy.ts

$ rg 'setTimeout\(.*setCopied' app/src --glob '!*.test.*'
(no output)

$ rg '<IconSwap' app/src --glob '!*.test.*' -l
app/src/components/layout/context-bar/CodeSection.tsx
app/src/components/shared/OAuth2Form/TokenStatusRow.tsx
app/src/components/shared/UpdateBanner.tsx
app/src/components/shared/response-viewer/ResponseActions.tsx
app/src/components/ui/secret-input.tsx
app/src/modules/collections/CollectionDetail/MockServerControl.tsx
app/src/modules/collections/CollectionItem.tsx
app/src/modules/settings/main/panels/McpSettingsPanel.tsx
app/src/modules/settings/main/panels/UpdatesCard.tsx
```

```
$ pnpm test useCopy clipboard-single-writer icon-swap icon-button-labels a11y-suppressions tooltip-icon-button
 Test Files  6 passed (6)
      Tests  37 passed (37)

$ pnpm lint && pnpm type-check && pnpm format:check
(all three clean, exit 0)
```

Plus the files the migration could break: `pnpm test CollectionItem MockServerControl TokenStatusRow secret-input CollectionTree icon-swap` - 20 files, 168 tests, green. `CodeSection` and `design-system-doc` green, and again after the `master` merge.

**The full-suite run is what found the `STATUS_RESET_MS` regression**, and it is the reason that commit exists: `pnpm test` ended `Test Files 3 failed | 773 passed (776)`, and one of those three was `timing-keys-have-readers.test.ts` - a real failure in this diff, not load. After the fix, `pnpm test timing-keys-have-readers timing useCopy clipboard-single-writer ResponseActions` is 7 files / 50 tests green, with type-check, ESLint and prettier clean on the touched files. The other two (`ResponseEvents.window.test.tsx`, `InboxView.captures.test.tsx`) failed on `Test timed out in 5000ms` under concurrent load from sibling agents' suites - the environment failure mode `app/CLAUDE.md` documents - and pass when re-run on their own (2 files / 14 tests). CI is the authority for the whole suite.

Mutation checks, each reverted afterwards:

- Drop the `try`/`catch` from `useCopy` - three cases in `useCopy.test.ts` fail (the two failure-reporting cases, and the icon-mode one that asserts `copied` stays false).
- Add an `await navigator.clipboard.writeText(x)` to `components/ui/kbd.tsx` - `clipboard-single-writer.test.ts` fails and names the file.

**Safe to skim:** the eight one-line `const { copy } = useCopy()` destructurings, `CodeSection.test.tsx`'s added `useToastStore` entry (that test mocks `@/stores` wholesale, so the copy path's new dependency has to be in the mock or every case in the file throws), and the merge commit from `master`.

## Checklist
- [x] I have added tests (`useCopy.test.ts`, `icon-swap.test.tsx`, `clipboard-single-writer.test.ts`)

## Related Issues
Closes #1686. Part of #1683.


athrvk:
The `github-advanced-security` check on `6016779` failed before analysing anything: the Copilot code-scanning job (autofind 0.1.136) died creating its session with `CAPIError: 400 The requested model is not supported`, the same GitHub-side error every PR in this programme has hit today (#1695 through #1699). Not this PR's, and that job has no re-run control from here. `pr-tests.yml` and CodeQL are the signals for this PR.

---

PR #1701, opened 2026-09-19
feat(app): carry icon motion on the drawer-view registry, and guard status icons against it

### LINKED ISSUE #1687: Icon motion for registry-driven glyphs, and a negative guard for status icons

athrvk:
## Description

**Was stacked on #1700**, which has merged (as had #1696 before it); GitHub retargeted this PR to `master` and the branch has merged `master` back in so the test matrix runs on the current head. The diff is now this branch's own commits only.

Two things, both downstream of the same gap: the icon-motion vocabulary from #1685 can only be used by a call site that writes the glyph itself, and nothing enforced the half of the policy that is a prohibition.

**The registry carries the name, so a generic renderer stays generic.** `ActivityRail` draws six glyphs out of `DRAWER_VIEWS` without knowing which icon it has been handed, which is the property worth keeping - the command palette lists the same six. So the descriptor grows an optional `motion`, and the renderer forwards it blind. An entry without one renders no attribute at all: React drops an undefined value, and that is load-bearing rather than incidental, because `[data-icon-motion]` matches an empty value too and an empty attribute would inherit the block's duration custom property for nothing.

**`RailButton` gains `group`, and that is the line most worth a second look.** Every rule in the block gates on the `:hover` / `:focus-visible` of a `[data-slot="button"]` or a `.group` ancestor. `RailButton` is hand-rolled and is neither, so without the class the rail's glyphs would carry the attribute, typecheck, pass every positive guard, and never move - dead CSS that looks wired up. `group` on a button that contains only a glyph and an absolutely-positioned badge has no other consumer, so it changes nothing else; a rendered test now pins it.

**One new motion, `scale`.** The issue asked for a `Folder` flap for Collections, and the rail's glyph is `FolderOpen` - a single path, with no part a hinge could belong to. `Braces` is two curves whose whole meaning is the gap between them, so moving either curve reads as a mismatched brace rather than as feedback. A 10% grow of the whole `svg` is what is left for both. 10% rather than press feedback's 2%: that one confirms a click on a whole button, this has to read across 16px of ink on its own.

**Two of the six stay still on policy, not for want of a motion.** History's `Clock` is a status icon; Services' `Radio` is a live indicator whose arcs already say "listening". Both are stated in the descriptor's doc comment and asserted, so "no motion here" is a recorded decision rather than an omission someone later fills in.

**`tab-descriptors.ts` deliberately did not get the field.** The issue suggested starting there too, but no tab glyph wants a motion yet, and a field one layer writes and no layer reads is this codebase's most repeated defect. It costs one line to add when an entry wants one.

**The negative guard is the reason for the second commit.** `<CheckCircle2 data-icon-motion={ICON_MOTION.scale} />` typechecks, animates exactly as designed, and reads as a considered choice in a diff - only the policy makes it wrong. `icon-motion-status.test.ts` scans every `.tsx` under `app/src` for the six status glyphs and fails on the attribute, and separately checks the registry path no source scan can see. Its tag parser is hand-walked rather than `/<Icon([^>]*)>/`, because `{a > b ? x : y}` in an attribute value ends that character class early and truncates the very attribute list being inspected - a guard that stops looking before the end is a guard that passes.

**Safe to skim:** the `docs/design-system.md` line (one table row, kept in step per the repo rule), the two-line additions to the existing `icon-motion.test.tsx` and `icon-motion.call-sites.test.tsx` lists, which just extend enumerations the new name and the new group owner belong in, and the merge commits from the base branches.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Testing

Acceptance criteria from the issue, run on this branch:

```
$ rg 'motion:' app/src/constants/drawer-views.ts
	{ view: "collections", label: "Collections", icon: FolderOpen, motion: ICON_MOTION.scale },
	{ view: "variables", label: "Variables", icon: Braces, motion: ICON_MOTION.scale },
	{ view: "trash", label: "Trash", icon: Trash2, motion: ICON_MOTION.lid },
	{ view: "settings", label: "Settings", icon: Settings, motion: ICON_MOTION.spinOnce },

$ rg 'data-icon-motion' app/src/components/layout/ActivityRail.tsx app/src/components/layout/RailButton.tsx
app/src/components/layout/ActivityRail.tsx:						    which matters, because an empty `data-icon-motion` would
app/src/components/layout/ActivityRail.tsx:						<Icon className="size-icon" data-icon-motion={motion} />
app/src/components/layout/RailButton.tsx:						// `index.css` gates every `data-icon-motion` rule on the

$ rg -n 'lucide-react' app/src/components/layout/ActivityRail.tsx app/src/components/layout/RailButton.tsx
(no match - neither file imports or names a glyph)

$ cd app && pnpm test ActivityRail icon-motion-status icon-motion
 Test Files  4 passed (4)
      Tests  39 passed (39)

$ pnpm lint && pnpm type-check && pnpm format:check
eslint: clean
type-check: renderer, electron, electron-tests all exited 0
prettier: All matched files use Prettier code style!
```

Full app suite once, after the base merge: **777 files passed, 8758 passed / 2 skipped**.

Mutation checks (each reverted):

1. `data-icon-motion={motion ?? "lid"}` in `ActivityRail` - `renders no attribute at all for the two the policy leaves still` and `matches every descriptor` both fail. Without them the "no motion" claim is unfalsifiable.
2. `data-icon-motion="scale"` added to `<Clock>` in `modules/history/sidebar/RunItem.tsx` - the source-scan half fails and names the file and the glyph.
3. `motion: ICON_MOTION.scale` on the History descriptor - the registry half fails, which is the path mutation 2 cannot reach.

Both halves of the new guard also assert a non-empty scan (>100 `.tsx` files, >5 status-icon elements found, >5 drawer views), so a scan that silently reads nothing fails rather than passing forever.

## Checklist
- [x] I have added tests
- [x] I have updated documentation (`docs/design-system.md`, Icon motion table)

## Visual changes

Not screenshotted: the change is a hover/focus transition on four 16px rail glyphs, which a still frame cannot show either state of usefully. The motion itself is verified by the stylesheet guard (gating, `transform-origin`, token-only timing) and the attribute by a rendered test.

## Related Issues

Closes #1687. Part of #1683.


athrvk:
The `github-advanced-security` check on `ac82827` failed before analysing anything: the Copilot code-scanning job died creating its session with `CAPIError: 400 The requested model is not supported`, the same GitHub-side error every PR in this programme has hit today (#1695 through #1700). Not this PR's, and that job has no re-run control from here. Note also that `pr-tests.yml` runs only for PRs against `master`, so this stacked PR gets its test matrix once #1700 merges; until then #1702 runs the suite on the combined tree.

---

athrvk:
Pushed a fix for the clipped trash lid seen while testing this PR: the open lid's handle travels above the icon's viewBox and an inline svg clips to it by default, so the top of the lid was cut off at every `lid` call site, not only the rail. `[data-icon-motion="lid"] { overflow: visible }` with the geometry in the comment; the icon-motion guards, reduced-motion and design-system-doc are green. It lands here rather than on the already-merged #1696 because this is the open PR that touches the block.

---

PR #1702, opened 2026-09-19
refactor(app): land the typography and primitives sweeps on top of the UI/UX programme chains

athrvk:
## Description

This branch, `claude/happy-ride-ysw1fi`, merges every chain tip of #1683. It started as a CI vehicle only: `pr-tests.yml` runs on pull requests against `master`, so the stacked PRs (#1700 on #1696, #1701 on #1700, #1703 on #1698) get no test matrix until their base merges, and this branch is where the full suite runs on the combined tree.

**It is now also the merge vehicle for the two sweeps.** #1704 (typography) and #1705 (bypassed primitives) targeted this branch and GitHub marked them merged the moment the branch fast-forwarded onto their tips, so they no longer have a PR of their own against `master`. Their commits reach `master` only through this PR. Two consequences:

- `Closes #1692` and `Closes #1693` live here, since a keyword only acts from a PR into the default branch.
- **Merge order:** land the eight individual PRs first (order in #1683's comment). Each merge shrinks this diff to what is left. After #1703 the remaining diff is the two sweeps plus the merge commits, which is when this is ready to review and merge; until then it stays a draft.

Conflicts resolved on this branch, each by keeping both sides: the design-system Component Patterns section (four PRs appended there), `CollectionItem.tsx` imports, the LoadTestDetail characterization snapshot (regenerated on the merged tree), and five files where the feedback PR met pane chrome and the icon chain (`RowActionBody`, inbox `index.tsx` and its errors test, `ServicesPanel`, the design-system doc). One resolution is worth knowing: the inbox Clear button had to keep both the `DisabledHint` wrapper from #1703 and the `DeleteConfirmDialog` handler from #1699; an automatic merge would have kept the wrapper and dropped the confirmation.

One fix of this branch's own: `1e0192e` untracks the `app/node_modules` symlink that the primitives sweep committed by accident (a worktree's symlink; `app/.gitignore`'s `node_modules/` pattern matched only a directory) and drops the trailing slash so the name is ignored whatever it is on disk. That commit is what turned this branch's `App quality checks` from red to green.

## Type of Change

- [x] New feature (the two sweeps' primitives)
- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Testing

[...]

PR #1703, opened 2026-09-19
feat(app): explain gated controls, give Services rows hover-reveal, and add keyboard reorder to the tree

### LINKED ISSUE #1690: Feedback: DisabledHint for gated controls, Services row hover-reveal, keyboard Move up/down in the collection tree

athrvk:
**Was stacked on #1698** (`claude/happy-ride-ysw1fi-rename`), which has merged; GitHub retargeted this PR to `master` and the branch has merged `master` back in so the test matrix runs on the current head.

## Description

Three feedback gaps from the #1683 audit, in three commits.

**A disabled control now says why.** The obvious fix does not work, which is why none of the 163 gates had an explanation: `Button`'s baseline is `disabled:pointer-events-none`, so a `Tooltip` wrapped around a disabled button never gets the events its trigger listens for, and a disabled button is not focusable either - both paths close at the moment there is something to explain. `DisabledHint` is a wrapper *around* the control instead, with its own pointer events and its own tab stop. `reason` doubles as the switch so a caller passes one expression rather than keeping an enabled and a disabled copy of the same button, which is how a hover class or an `aria-label` ends up on one of them and not the other.

Two consequences worth knowing before reading the diff:

- **Crossing the gate replaces the child's DOM node**, since the wrapper is present on one side of it and not the other. Nothing user-visible depends on it (a disabled control has no focus to lose), but a test holding a reference across the flip has to re-query - one did.
- **An adopting surface now needs a `TooltipProvider` above it.** The app has one at its root, so this is a test-harness fact, not a runtime one - see the fourth commit.

**A disabled menu item is deliberately not wrapped.** `DisabledHint` works by taking a tab stop, and inside a menu that competes with Radix's own focus management, which skips a disabled item on purpose. So `RowAction` gains `disabledReason`, drawn on the item by `RowActionBody` for the `⋯` dropdown and the right-click menu at once, and `hint` for the same column carrying a value rather than a reason. The element list's Move up / Move down now read "Already first" / "Already last".

**Services rows: parity, not an exemption.** They were the last rows in the app painting their actions as always-visible icon buttons - two or three on a 32px row whose own payload is a URL, so the URL truncated to make room for controls the user had not come for. All three groups draw one hover-revealed `RowActionsMenu` now. The values the old tooltips carried (the inbox URL, "stopped, not listening" on a stopped one, the mock's base URL) ride the items' `hint`, which reads without hovering and reaches a screen reader in the item's own name - the tooltip was a strictly weaker place for them. Stopping an issuer or a mock is marked destructive because the record goes with the listener; stopping an inbox is not, because it does not.

**Move up / Move down in the tree's row menu.** The tree reordered by drag and by Alt+Arrow only: a drag needs a pointer, a chord has to be known before it can be used. The entries call the same `moveByKeyboard` the chords do, so one move still goes through one `applyPlacement`. `blockPosition` is factored out and shared with the gating, so a disabled Move up and an "already first" announcement cannot disagree about which row is first. They land above the row's destructive tail, since the separator `rowActionRows` draws there belongs to Delete.

### Safe to skim

- The fourth commit is mechanical and test-only: fourteen files rendered their panel with no `TooltipProvider`, a state the app is never in, and it only went unnoticed because nothing inside them had a tooltip. Each now wraps its renders rather than weakening an assertion to suit the harness.
- Doc edits in `docs/design-system.md` and `docs/app/COMPONENTS.md` are this repo's same-commit rule, and `design-system-doc.test.ts` / `a11y-suppressions.test.ts` check them.
- The merge commit from `master` after #1698 landed.

### Worth scrutiny

- The one new `jsx-a11y` suppression (`no-noninteractive-tabindex` on `DisabledHint`'s wrapper span), the a11y ceiling raised 20 → 21, and its entry in the Accessibility section. The argument is the one `Dock.tsx`'s three error tooltips already carry: the child is disabled, so the wrapper's tab stop is the only keyboard path to the text.
- The Services test rewrites. The assertions moved from clicking a visible button to opening the menu, which is a genuine change in what is being claimed about the surface, not a selector tidy-up.

## Deviation from the issue

#1690's first acceptance criterion asks for `<DisabledHint` in at least six files **including** `components/shared/ElementList/index.tsx`. It lists six, and `ElementList` is not one of them: its only gated controls are the two Move menu items, and a tooltip on those is the case argued against above. The reason strings that criterion was after ("Already first" / "Already last") are in that file, as `disabledReason` - rendered through the shared `RowActionBody`, so both menus get them, which the per-file JSX would not have achieved.

`CollectionTree.move.test.tsx` carries a small fixture of its own rather than sharing `CollectionTree.dnd.test.tsx`'s - the tree tests have no shared harness today, and these cases need no layout or hit-test stubs, since a menu click is not a pointer gesture.

## Acceptance criteria

```
$ rg '<DisabledHint' app/src --glob '!*.test.*' -l
app/src/modules/collections/CollectionDetail/ElementsTab.tsx
app/src/modules/inbox/index.tsx
app/src/modules/settings/main/panels/NotificationsPanel.tsx
app/src/modules/settings/main/panels/GeneralPanel.tsx
app/src/modules/history/main/SaveRunToRequestDialog.tsx
app/src/modules/request-builder/components/LoadTestConfigDialog/index.tsx

$ rg 'group-hover:opacity-100' app/src/modules/services/ServicesPanel.tsx
					className="opacity-0 transition-opacity group-hover:opacity-100 group-focus-within:opacity-100"

$ rg 'Move up' app/src/modules/collections --glob '!*.test.*'
app/src/modules/collections/useTreeDnd.ts:					label: "Move up",
(plus five comment lines in useTreeDnd.ts, CollectionItem.tsx and CollectionTreeContext.tsx)
```

## Testing

```
$ cd app && pnpm test disabled-hint ServicesPanel CollectionTree.move row-action-reveal
 Test Files  6 passed (6)
      Tests  60 passed (60)

$ cd app && pnpm lint && pnpm type-check && pnpm format:check
(all clean)

$ cd app && pnpm test
 Test Files  775 passed (775)
      Tests  8740 passed | 2 skipped (8742)
```

Mutation checks, each reverted afterwards:

- Drop `pointer-events-auto` from `DisabledHint`'s wrapper → `disabled-hint.test.tsx` "keeps its own pointer events" fails. jsdom has no hit testing, so the class is the assertion; it is the whole mechanism.
- `tabIndex={0}` → `{-1}` → "is a tab stop" fails.
- Drop the reveal `className` from `ServiceRow`'s `RowActionsMenu` → "hides its row actions until the row is hovered or focused" fails.
- The two tree cases name their own checks in comments: pointing `Move down` at `"up"` inverts the asserted orders, and the disabled-entry case fails if a gated item writes.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist

## Related Issues
Closes #1690. Part of #1683.


athrvk:
The `github-advanced-security` check on `11ec923` failed before analysing anything: the Copilot code-scanning job died creating its session with `CAPIError: 400 The requested model is not supported`, the same GitHub-side error every PR in this programme has hit today (#1695 through #1702). Not this PR's, and that job has no re-run control from here. `pr-tests.yml` runs only for PRs against `master`, so this stacked PR gets its test matrix once #1698 merges; until then #1702 runs the suite on the combined tree.

---

Issue #1692, opened 2026-09-19
Typography sweep: Eyebrow for every section label, and named text-label / text-micro steps


athrvk:
Part of #1683. Anchors at master `9b9a22f`.

## Problem

1. **The `Eyebrow` primitive is bypassed by ~20 hand-rolled section labels** drifting across three sizes (`text-[10px]`, `text-[11px]`, `text-xs`), five tracking values (`tracking-wide`, `-wider`, `[0.06em]`, `[0.07em]`, `[0.08em]`) and two weights. `components/ui/eyebrow.tsx:33` is canonical; `eyebrow.test.ts` catches only a verbatim copy of the literal. Sites: `CollectionDetail/shared.tsx:54,65,83`, `ChainCard.tsx:51`, `TestResults.tsx:102`, `ElementOutcomes.tsx:58`, `InfoPanel.tsx:71,85`, `ExamplesPanel.tsx:172`, `SettingsCategoryTree.tsx:62`, `HistoryList.tsx:408`, `ThroughputTwinCard.tsx:50,62`, `variable-popover.tsx:947,959`, `KeyValueEditor/index.tsx:182`.
2. **`text-[11px]` x194 and `text-[10px]` x60** are the app's two most common type sizes and neither is a named scale step.

## Fix pathway

1. Name the steps in `app/src/index.css` (`--text-label: 11px`, `--text-micro: 10px` with matching line heights; Tailwind v4 `@theme` so `text-label` / `text-micro` become utilities). Sweep every `text-[11px]` and `text-[10px]` in `app/src` to them.
2. Replace every hand-rolled label with `<Eyebrow>`; add `size="xs"` if a genuine 10px eyebrow exists.
3. Widen `eyebrow.test.ts`: no `.tsx` outside `eyebrow.tsx` combines `uppercase` with `tracking-` in one class string. Add `type-scale.test.ts` guard: no `text-[10px]` / `text-[11px]` literals in `app/src`. Both assert a non-empty scan.

## Tests

The two guards above, each with a mutation check in the PR body (re-add one literal, confirm red, revert). `pnpm test type-scale eyebrow design-system-doc`.

## Docs

`docs/design-system.md` Typography / Type Scale table gains the two rows (values checked by `design-system-doc.test.ts`); the Eyebrow entry states the rule.

## Sequencing

[...]

Issue #1409, opened 2026-09-04
app(design): the 20px and 24px steps have no row in the type scale - settle them or sweep them


athrvk:
Filed from the #1202 register sweep, so the two steps it deliberately did not touch have a record rather than a silence.

## Problem

#1202 capped the chrome register: `CardTitle` and `DialogTitle` own `text-md` (15px), `text-base` is written nowhere in `src`, and `text-lg` (18px) is the tile metric value step, held to four allowlisted files by `type-scale.test.ts`. Above that the scale stops describing the app. Five call sites render sizes the Type Scale Conventions table has no row for:

- `modules/settings/main/SettingsMain.tsx:794` and `modules/settings/main/panels/ClientSettingsPanel.tsx:47` - `text-xl` (20px), the settings view title. Measured in the packaged renderer at 1280x800, this is the only thing left at 16px or more on the Editor settings panel.
- `modules/history/main/LoadTestDetail.tsx:326`, `:335`, `:344` - `text-xl font-bold`, three run-summary numbers.
- `modules/dashboard/components/hero/ThroughputTwinCard.tsx:44`, `:56` and `modules/history/main/components/PerformanceTab.tsx:188`, `:194`, `:208` - `text-2xl` (24px) metric values.

The table already names 22px "Secondary metric value" and 34px "Hero metric value", both as arbitrary values (`text-[22px]`, `text-[34px]`), so the numeric readouts above sit one Tailwind step away from a documented step that means the same thing - two spellings of the same intent, which is the drift #1202's own guard exists to stop. The view title has no row at all.

## Fix pathway

Decide two things and write both into the table:

1. **The view title.** Either a row (`View title | 20px | semibold | text-xl font-semibold`, one per view, the settings panels being the only users) or a sweep to `text-md`, which would make a settings view's title the same size as the card headings under it. A row is the likelier answer; a sweep would need the panel header re-thought, since the title, its description and the cards would then all be one step.
2. **The metric values.** `text-xl`/`text-2xl` against the documented `text-[22px]`/`text-[34px]` is a fork in the same road. Pick one spelling - the named utilities carry paired line-heights, which is the argument the scale made for `--text-md` - and either move the arbitrary rows onto named steps or move these five sites onto the arbitrary ones.

Whichever lands, extend `type-scale.test.ts`'s ceiling block so 20px and 24px are held the way 16px and 18px now are: an allowlist that fails when an entry stops using its step.

## App changes

[...]

PR #1704, opened 2026-09-19
refactor(app): name the 11px and 10px type steps and adopt Eyebrow everywhere

athrvk:
## Description

Last sweep in the #1683 programme, so it targets the integration branch
**`claude/happy-ride-ysw1fi`** rather than `master`: it rewrites a class name in
96 files, which would conflict with every other chain in the programme if it
landed first. Merge it after the rest.

Two things, both mechanical, and the second is the one worth reading:

**The scale's last two arbitrary sizes get names.** 11px and 10px were the app's
most-used font sizes at 248 call sites and the only ones with no token, which is
why the Type Scale table had to carve them out of its own
no-arbitrary-values rule by name. They are `--text-label` and `--text-micro`
now, and `type-scale.test.ts` drops its allowlist, so the rule is flat.

The one thing that is not a pure rename: a named step carries a line-height and
an arbitrary one does not, so the 230 of those 248 sites that wrote no
`leading-*` inherited whatever rhythm their parent had and now get 16px / 14px.
That is the whole point of naming a step (it is the argument `--text-md`,
`--text-hero` and `--text-metric` were named on), but it is the part of this
diff that can move a pixel, and it is worth a look in a dense pane. **No size
changed.**

`utils.ts` is the non-obvious dependency: tailwind-merge reads an unknown
`text-<x>` as a *colour*, so an unregistered custom step is silently dropped
from `cn("text-label …", "text-muted-foreground")` with nothing in the source to
look wrong. That is the #1409 defect exactly, and both new steps are the shape
that triggers it, which is why `cn-font-size.test.ts` gains a case for each.

**`Eyebrow` stops being a suggestion.** The old guard only caught a
byte-for-byte copy of `EYEBROW_CLASS`, and that is not the shape the drift took:
the nineteen hand-rolled section labels ran three sizes, five tracking values
and two weights between them - every one a near-miss it waved through. The guard
now tests the *shape* (`uppercase` + a `tracking-` utility in one class string,
comments stripped), the primitive takes `size="xs"` for the 10px tier several
panes were running by hand, and the six files that use that shape for something
which is not a section label - a `<tr>`, the console gutter marker, inline
`off` / `not defined` value markers, the response pane's content-type chip -
are exempted by name with a reason, each asserted to still hold the label it is
exempted for so a stale entry drops off rather than quietly licensing whatever
lands there next.

A second-order effect worth knowing: the converted labels render a `<p>` where
some were a `<div>`, so a caller that nests block content inside one now has
invalid markup instead of an error. `KeyValueEditor`'s column header (a grid of
`<div>`s) and the two response-pane `<h3>`s keep their element and take
`EYEBROW_CLASS` / `EYEBROW_XS_CLASS` for that reason rather than becoming the
component.

Tracking is unified to `0.06em` on the converted sites, which visibly tightens
the few that ran `0.08em`.

### Safe to skim

- The 96-file sweep commit - one `sed`, two class names, nothing else in it.
- The snapshot commit. Verified as class-list-only: 76 `text-[11px]` ->
  `text-label`, 20 `text-[10px]` -> `text-micro`, plus the `data-size` attribute
  `Eyebrow` now carries and the `div` -> `p` on the throughput twin's two
  legends. Nothing else moved in either `.snap`.
- `docs/design-system.md` - the Type Scale rows and the Eyebrow section, kept in
  step with the code per this repo's own rule.
- The merge commit. See below - it is a clean merge with no conflicts and no
  follow-up sweep needed.

### Base moved under this branch, and was merged in

`claude/happy-ride-ysw1fi` advanced from `778e2995` to `d87e05f4` while this was
in flight, picking up #1703 (the `DisabledHint` primitive, the Services rows
moving onto `RowActionsMenu`, and tree Move up/down). That was merged into this
branch with `git merge --no-edit` - **no conflicts**, which is what sequencing
this sweep last was for.

The two greps were re-run against the merged tree and it brought in **nothing to
sweep**: no new `text-[10px]` / `text-[11px]` anywhere in `app/src`, and no new
hand-rolled eyebrow - the `uppercase` + `tracking-` scan returns exactly the same
set of already-exempted files as before the merge. So there is no extra sweep
commit, and the new `DisabledHint` primitive was already writing named steps.

After the merge: `pnpm lint`, `pnpm type-check`, `pnpm format:check` green, and
`pnpm test eyebrow type-scale design-system-doc disabled-hint ServicesPanel
characterization` - 9 files, 128 tests, green (the characterization snapshots
still match, so #1703's changes and this rename do not overlap).

## Type of Change
- [x] New feature (a named scale step and an `Eyebrow` size, both additive)
- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Testing

Full app suite once before the merge: **796 files, 8867 passed, 2 skipped**.
`pnpm lint`, `pnpm type-check` and `pnpm format:check` green, re-run after the
merge along with the targeted files listed above.

Both guards were mutation-checked (committed first, reverted with `git checkout`
after):

**`type-scale.test.ts`** - put `text-[11px]` back in `MethodBadge.tsx`:

```
AssertionError: expected 'components/shared/MethodBadge.tsx:100…' to be ''
+ components/shared/MethodBadge.tsx:100  text-[11px] is not on the scale. Use
  text-micro (10), text-label (11), text-xs (12), text-sm (13), text-md (15),
  text-lg (18), text-xl (20), text-metric (22) or text-hero (34).
 Tests  1 failed | 42 passed (43)
```

**`eyebrow.test.ts`** - put one hand-rolled label back in `InfoPanel.tsx`:

```
+ modules/request-builder/components/RequestTabs/panels/InfoPanel.tsx
  "text-label uppercase tracking-wide text-subtle-foreground"
 Tests  1 failed | 10 passed (11)
```

Both scans assert a non-empty read: the type scale one that it walked 150+
files, the eyebrow one that it walked 100+ `.tsx` and that the shape it looks
for is still found somewhere.

### Acceptance criteria

`rg 'text-\[1[01]px\]' app/src --glob '!*.test.*'` - no output, exit 1.

`rg 'uppercase' app/src --glob '*.tsx' --glob '!*.test.*' -l` - 11 files:

```
components/layout/context-bar/CollectionVariablesSection.tsx   exempt by name
components/layout/context-bar/Section.tsx                      comment only
components/layout/context-bar/VariablesSection.tsx             exempt by name
components/shared/MethodBadge.tsx                              no tracking - text-transform on a value
components/shared/response-viewer/ResponseBody.tsx             exempt by name
components/ui/eyebrow.tsx                                      the primitive
modules/request-builder/.../panels/ParamsPanel.tsx             exempt by name
modules/request-builder/.../panels/SettingsPanel.tsx           comment only
modules/request-builder/.../console/ScriptSection.tsx          exempt by name
modules/variables/main/VariableTableEditor.tsx                 exempt by name
modules/welcome/components/RecentRuns.tsx                      no tracking - text-transform on a value
```

The criterion asks for "only `eyebrow.tsx` or files the widened guard exempts by
name". Six are exactly that. The other four are not offenders under the guard's
rule at all: two are a bare `uppercase` `text-transform` on a value
(`MethodBadge`'s method name, `RecentRuns`'s run-type chip) and two only contain
the word in a comment. Naming them in `EXEMPT` would have been an exemption for a
rule they do not break, and the exempt list asserts each entry still matches - so
they are left out deliberately rather than overlooked.

`cd app && pnpm test eyebrow type-scale design-system-doc` - 3 files, 49 tests,
green.

## Checklist
- [x] I have added tests (both guards widened, each mutation-checked)

## Related Issues
Closes #1692
Part of #1683


athrvk:
The `github-advanced-security` check on `c705aef` failed before analysing anything: the Copilot code-scanning job died creating its session with `CAPIError: 400 The requested model is not supported`, the same GitHub-side error every PR in this programme has hit today. Not this PR's, and that job has no re-run control from here. This PR targets the integration branch, so its test matrix runs through #1702 on the combined tree.

---

PR #1708, opened 2026-09-19
feat(app): give every interactive glyph a motion of its own

### LINKED ISSUE #1707: Icon motion: a bespoke motion for every interactive glyph, not a shared scale
Part of #1683. Owner feedback while testing #1701: the trash lid is the standard; the rest of the rail (and the app's other interactive glyphs) should each have a motion of their own. Anchors at `claude/happy-ride-ysw1fi-icon-registry` tip `6f0e4bd` (master plus #1701).

## Problem

`ICON_MOTION` has six names and most icons share them: Collections (`FolderOpen`) and Variables (`Braces`) both got a 10% `scale`, and History (`Clock`) and Services (`Radio`) got nothing because the policy classed those glyphs as status and live indicators. In the rail they are navigation affordances, and a glyph that means "go to History" earning no motion while Trash gets a lid reads as unfinished. The same is true across the app: Search, Download, Upload, Save, Pin, Play, Bell, Filter, RotateCcw, Zap and others are all action glyphs with no motion.

## Fix pathway

1. **Policy refinement** in `docs/design-system.md` Icon motion: the six status glyphs never animate *where they convey status* (a chip, a row's status column, an inline warning). Where one of them is the affordance for an action (a rail view, a button), it animates like any other. `icon-motion-status.test.ts` keeps its source scan for inline status usage but stops failing registry and button call sites; encode the distinction by where the element sits (inside `[data-slot="button"]`, a `RowAction`, or a registry entry) rather than by glyph name alone.
2. **New motions**, each named, each in `ICON_MOTION`, the `Icon motion` CSS block and the doc table in one commit, each satisfying the existing guard (hover or focus gated on the owner, `transform-box: fill-box` on child transforms, tokens only, no `will-change`). Pivot each one on the lucide path geometry, the way the lid does, and record the child indices and origins in the CSS comment:

| Glyph | Motion name | What moves |
|---|---|---|
| `Clock` (History) | `hands` | The hands polyline (child 2) turns one full revolution about the dial centre (12,12); the dial stays |
| `Radio` (Services) | `waves` | The four arcs (children 1 to 4) fade and grow outward once, inner pair then outer pair, staggered; the centre dot stays |
| `Braces` (Variables) | `spread` | The two braces part by 1px each, outward, and return |
| `FolderOpen` (Collections) | `tilt` | Rotate about the bottom-left corner by -6deg with a 4% grow, reads as the folder opening toward the user |
| `Settings` | `spin-once` | as today |
| `Trash2` | `lid` | as today, with the `overflow: visible` fix |
| `Search` | `wiggle` | -8deg then +8deg then 0, once |
| `Download` | `drop` | Arrow (the vertical line and head) drops 2px; the tray stays |
| `Upload` | `lift` | Arrow rises 2px; the tray stays |
| `Save` | `press` | Whole glyph scales to 92% and back, a press |
| `Pin` | `tilt-pin` | Rotate -20deg about the tip and back |
| `Play` | `scale` | as today's scale |
| `Bell` | `ring` | Rotate +12deg, -10deg, +6deg, 0 about the top centre |
| `Filter` / `SlidersHorizontal` | `nudge-x` | as today |
| `RotateCcw` | `spin-back` | One turn counter-clockwise |
| `Zap` | `flash` | Opacity 1 to 0.4 to 1 with a 6% grow |
| `Plus` / `X` | `rotate-90` | as today |
| `Copy`, `Eye`, `Folder` pairs | `IconSwap` | as today |

Any other lucide glyph that sits inside a button, menu item, tab trigger or rail entry in `app/src` gets one of these or a new name; the agent inventories them (`rg -l 'lucide-react'` inside `components/` and `modules/`, then filter by owner) and lists the mapping in the PR body.

3. **Frame clipping**: every motion that carries ink outside the 24-unit viewBox (the lid, `hands` does not, `waves` may, `tilt`, `ring`, `drop`, `lift`) sets `overflow: visible` on the svg, the way the lid does now. Verify every motion in Chromium at rest and at the hover peak with a screenshot per glyph (the earlier agent's CDP approach works without playwright), and attach them to the PR through the screenshot-commit flow in the `pr-description` skill.
4. **Registry**: `DRAWER_VIEWS` entries for History and Services gain their motions; any tab descriptor or palette entry whose glyph now has a motion opts in only if it renders as a hovered control.

## Tests

- `icon-motion.test.tsx`: extend the pinning to the new child indices (`Clock` hands child 2, `Radio` arcs 1 to 4, `Download`/`Upload` arrow children) via `__iconNode`; assert `overflow: visible` on every motion the frame-leaving list names.
- `icon-motion-status.test.ts`: the refined rule, with the mutation check that a `Clock` inside a status chip with `data-icon-motion` still fails and one inside `DRAWER_VIEWS` passes.
- `ActivityRail.test.tsx`: all six rail glyphs carry a motion, no two the same.
- Rendered-class tests for each new call-site group.

## Docs

`docs/design-system.md` Icon motion: the refined policy paragraph and the full table.

## Sequencing

Stacks on #1701 (branch `claude/happy-ride-ysw1fi-icon-registry`). Lands before the rollup #1702 is merged; the integration branch takes it in afterwards.

## Acceptance criteria

- `rg 'motion:' app/src/constants/drawer-views.ts` prints six entries with six distinct names.
- `rg -o 'data-icon-motion="[a-z-]+"' app/src/index.css | sort -u | wc -l` is at least 16.
- `rg 'overflow: visible' app/src/index.css` matches inside the Icon motion block for every frame-leaving motion the doc table marks.
- The PR body carries one hover-peak screenshot per motion, and the mapping table of every interactive lucide glyph in `app/src` to its motion.
- `cd app && pnpm test icon-motion ActivityRail icon-motion-status reduced-motion design-system-doc` green; `pnpm lint && pnpm type-check && pnpm format:check` green.

athrvk:
## Description

Was stacked on #1701; that has merged and this now targets `master` directly.

The rail shipped with four of its six glyphs saying nothing about themselves: Collections and Variables shared a generic 10% grow, History and Services got nothing at all because the policy classed `Clock` and `Radio` as status and live-indicator glyphs. Beside the trash lid, that reads as unfinished. This replaces the shared motions with a bespoke one per glyph, fixes the policy that produced the gap, and then does the same for the Settings drawer's fifteen category rows, which were missed on the first pass and sat still for the same reason.

**The policy change is the part worth reading.** "Status icons never animate" was a list of six glyph names, which is the wrong axis: `Clock` in a run's status column is a state being reported, and `Clock` in the Activity Rail is a button that means "go to History". The rule is now about placement - a glyph conveying a state never animates, and the same glyph inside a button, a menu item, a tab trigger, a `RowAction` or a registry a rail draws as buttons animates like any other affordance. `icon-motion-status.test.ts` encodes it the same way: it walks outward from each status element to its nearest interactive owner, by JSX indentation, which is well-defined because `app/` is prettier-clean to the file and CI keeps it that way. A unit case pins that walk on identical markup in two placements, so an `interactiveOwner` that returned `true` unconditionally - which would make the whole guard unfailable - fails there instead.

**Pivots are computed, not guessed.** Every origin comes from lucide's `__iconNode` geometry and is written out beside the rule, the way the lid's already is. `hands` turns about the dial centre, which under `transform-box: fill-box` is `0% 75%` of the hands polyline's own box (x 12..16, y 6..14) and would be off the dial entirely against the default `view-box`. `tilt` pivots on the folder's bottom-left ink corner (2,20); `tilt-pin` on the needle's point (12,22); `sweep` on the gauge needle's hub (12,14). All of them were checked in Chromium - see Visual changes, which is also where the maths got corrected twice.

**Tempo.** The last commit gives icon motion its own pair of duration tokens, `--dur-icon-nudge` (200ms, hover transitions) and `--dur-icon-play` (280ms, one-shot sequences; two-phase sequences are x2), about 100ms slower than the tooltip and panel tiers the motions first borrowed, which read as a flick on a 12px glyph. Those two numbers are the one place to tune the tempo from here.

**Four things I'd flag:**

- **`part` is not `spread`, deliberately.** They are the same reading - a pair of marks parting by a pixel - and cannot share a rule: lucide draws `Braces` left-bracket first and draws both code glyphs right-chevron first. Reusing the name would have moved each chevron toward the other and drawn the brackets *closing*, which looks like a considered choice in a diff and says the opposite of what it means. If you disagree with the second name, the alternative is a per-child direction custom property, not a shared rule.
- **`trace` draws instead of moving**, which is new in this block: `stroke-dasharray`/`stroke-dashoffset`, both set inside the keyframes so the resting glyph carries no dash pattern at all. The dash length is measured (`getTotalLength()` = 49.214, dashed at 50), not guessed, and the offset starts negative because lucide draws that path right-to-left - the first capture stroked the line on backwards.
- **`ring` is back.** It was designed with the first batch and held out, because the guard that refuses a name with no call site caught it: the app's only `Bell` was a card heading, where the policy forbids motion. The Notifications category row is the call site it was waiting for, so the gap this PR opened is closed inside the same PR.
- **Safe to skim**: the ~21 call-site files in the first half are all the same one-attribute edit plus an import; the two registries in the second half are the same one-line field per entry; and the `docs/design-system.md` table is the vocabulary restated per this repo's keep-docs-in-step rule.

### Every interactive lucide glyph in `app/src`, and its motion

| Glyph | Motion | Where it fires |
|---|---|---|
| `Trash2` | `lid` | Trash rail entry, row delete actions, `KeyValueRow` |
| `Clock` | `hands` | History rail entry |
| `Radio` | `waves` | Services rail entry, Services settings category |
| `Braces` | `spread` | Variables rail entry |
| `FolderOpen` | `tilt` | Collections rail entry |
| `Palette` | `tilt` | Appearance settings category |
| `Settings` | `spin-once` | Settings rail entry |
| `RefreshCw` | `spin-once` | Dock engine restart, context bar Code section |
| `RotateCcw` | `spin-back` | Settings reset (x3), Trash restore |
| `Search` | `wiggle` | `CommandSearchBar` |
| `Download` | `drop` | TitleBar import menu item, `CollectionTree`, `ResponseActions`, `ExportSpecDialog`, `SpecTab` export |
| `Upload` | `lift` | `DataFilePicker`, `SpecTab`, `SpecSync`, `ImportModal` dropzone |
| `Save` | `press` | `DesignRunView`, `SettingsMain` |
| `Keyboard`, `Server` | `press` | Keyboard shortcuts and Core settings categories |
| `Pin` / `PinOff` | `tilt-pin` | `HistoryList` pinned filter, `RunItem` baseline toggle |
| `Play` | `scale` | `RunCollectionDialog`, inbox start (x2), notification previews (x2), `MockServerControl`, the Run collection row action |
| `SlidersHorizontal` | `nudge-x` | `MockServerControl` options |
| `Zap` | `flash` | `McpSettingsPanel` connect |
| `Plus` / `X` | `rotate-90` | TitleBar new environment, TabStrip close |
| `ChevronRight` / `ChevronDown` | `nudge-x` / `nudge-y` | context-bar section disclosure |
| `Info` | `bob` | General settings category |
| `Code2`, `Code` | `part` | Editor and Scripting environment settings categories |
| `LayoutDashboard` | `tiles` | Dashboard settings category |
| `Gauge` | `sweep` | Load testing and Limits settings categories |
| `Bell` | `ring` | Notifications settings category |
| `Plug` | `plug-in` | AI agents (MCP) settings category |
| `Network` | `pulse` | Network & connectivity settings category |
| `Activity` | `trace` | Observability settings category |
| `Database` | `stack` | Data & retention settings category |
| `Copy`+`Check`, `Eye`+`EyeOff`, `Play`+`Square`, `Folder`+`FolderOpen` | `IconSwap` | a different mechanism: a swap between two glyphs on a state change, not one glyph moving |

Deliberately still: `Bell` in a card heading, `Settings` in the "Engine Settings" section heading, `SettingsMain`'s page header (which draws no category glyph at all), `Zap` in a run's metadata chip, `Pin` on a baseline badge, `Search` inside every filter input (`pointer-events-none` adornments), `Loader2` (a loop, which means "work is happening"), and the status glyphs wherever they report. The remaining button glyphs with no motion - `KeyRound`, `Globe`, `FileJson`, `ExternalLink` and their kin - are marks with no reading to act out; `scale` is available to them and deliberately not spent, because a motion every glyph has is a motion none of them mean.

## Type of Change
- [x] New feature
- [x] Documentation update
- [ ] Bug fix
- [ ] Breaking change

## Testing

`cd app && pnpm lint && pnpm type-check && pnpm format:check` green; full `pnpm test` green - **779 files, 8811 passed, 2 skipped, 0 failed**.

Targeted: `pnpm test icon-motion ActivityRail icon-motion-status SettingsCategoryTree reduced-motion design-system-doc` green, re-run after the tempo commit.

**In Chromium**, headless over CDP against the real stylesheet under vite, because none of this is observable from jsdom - a `transform-origin` that is wrong by a bounding box renders perfectly and simply hinges in the wrong place. Every motion was captured at rest and at its hover peak (transitions settled; keyframe animations paused at their peak through the Web Animations API), cropped to the glyph, and checked for clipping and pivot. That pass earned its keep twice: it caught `trace` drawing right-to-left, and it is how the `overflow: visible` list was settled rather than argued. With `html[data-reduced-motion="true"]` set, the computed `animation-duration` and `transition-duration` of all 26 motions and every child they address read `1e-05s` - 0.01ms - so the collapse rules need nothing new.

**Mutation checks** (each reverted afterwards):

| Mutation | Guard that failed |
|---|---|
| Removed `overflow: visible` from `[data-icon-motion="tilt"]` | `icon-motion.test.tsx` - "tilt lets its ink paint past the viewBox", and "declares overflow on the frame-leaving motions and nowhere else" |
| Pointed `hands` at `> path:nth-child(1)` instead of `(2)` | `icon-motion.test.tsx` - "hands names the right children of its glyph" (`expected [ 1 ] to deeply equal [ 2 ]`) and "addresses the element type it names" (`hands addresses <path>:nth-child(1), which is a <circle>`) |
| Put `data-icon-motion={ICON_MOTION.hands}` on the `Clock` in `RunItem`'s load-run duration chip | `icon-motion-status.test.ts` - "`<Clock>` carries data-icon-motion but sits in no interactive owner". The rail's `Clock`, in the same run, passes: that is the placement rule working in both directions. |
| Dropped `data-icon-motion={ICON_MOTION.wiggle}` from `CommandSearchBar` | `icon-motion.call-sites.test.tsx` - the per-file case, and "spells every name the vocabulary exports, so no rule is dead CSS". This is the guard that held `ring` back until it had a call site. |

## Visual changes

Rest on the left, hover peak on the right. Captured at 2x from headless Chromium against `index.css` itself.

### The first pass - rail, chrome and action glyphs

| Motion | Rest | Hover peak |
|---|---|---|
| `hands` (Clock) |  |  |
| `waves` (Radio) |  |  |
| `spread` (Braces) |  |  |
| `tilt` (FolderOpen) |  |  |
| `lid` (Trash2) |  |  |
| `spin-once` (RefreshCw) |  |  |
| `wiggle` (Search) |  |  |
| `drop` (Download) |  |  |
| `lift` (Upload) |  |  |
| `press` (Save) |  |  |
| `tilt-pin` (Pin) |  |  |
| `spin-back` (RotateCcw) |  |  |
| `flash` (Zap) |  |  |
| `rotate-90` (Plus) |  |  |
| `nudge-x` (ChevronRight) |  |  |
| `nudge-y` (ChevronDown) |  |  |
| `scale` (Play) |  |  |

### The Settings drawer's categories

| Motion | Rest | Hover peak |
|---|---|---|
| `ring` (Bell) |  |  |
| `bob` (Info) |  |  |
| `part` (Code2) |  |  |
| `tiles` (LayoutDashboard) |  |  |
| `sweep` (Gauge) |  |  |
| `plug-in` (Plug) |  |  |
| `pulse` (Network) |  |  |
| `trace` (Activity) |  |  |
| `stack` (Database) |  |  |
| `tilt` (Palette - the reuse) |  |  |
| `press` (Keyboard - the reuse) |  |  |

The screenshots were pushed in two scaffolding commits (`7c4dc42e` and `d15ddf83`) that have since been dropped from the branch; the images keep rendering because GitHub retains blobs that were pushed to a pull request.

## Checklist

## Related Issues

Closes #1707
Part of #1683


athrvk:
The `github-advanced-security` check on `4a400e0` failed before analysing anything: the Copilot code-scanning job died creating its session with `CAPIError: 400 The requested model is not supported`, the same GitHub-side error every PR in this programme has hit today. Not this PR's, and that job has no re-run control from here. This PR is stacked, so its own test matrix runs once #1701 merges; until then the combined tree is covered by the rollup.

---

PR #1709, opened 2026-09-20
fix(app): close the gaps #1683's audit found between two guards and the rules they enforce

athrvk:
## Description

The closing audit of #1683 ran every sub-issue's acceptance criteria against master. Two came back narrower in the guard than in the rule the design-system doc states, and the tree had walked through both gaps. This PR widens the guards and fixes what they then catch. Two commits: the five sites first, then the guards, so the second commit is red on its own and green on top of the first.

**Why the error-text guard read only `<p>` and `<span>`.** #1688 chose the two tags a sentence usually sits on, so that a red *count* or a red *status word* on any other element would not need an exemption. The five messages it missed all sat on a `<div>`: the settings pane's and the variable table's load failures, the editor's load failure, the variable popover's two token-state sentences, and the dashboard's slow-request box. Those are exactly the hand-rolled red paragraphs the rule exists to retire, so the tag set now covers every text-bearing element (`div`, `small`, `label`, list items, table cells, headings) and stays a literal-class-string scan, for the reason the guard's own comment gives: the token inside a `cn()` ternary is state colour on a value, never a sentence.

**Why the eyebrow guard required `tracking-` beside `uppercase`.** That was the shape the twenty converted labels shared, and it left any `uppercase` without tracking unasked. Now every `uppercase` in a `.tsx` class string is a section label unless the file is named with what the capitals are instead. Two join the list on that basis: `MethodBadge`, where the HTTP method is capitals by convention, and the welcome screen's Load / Design run-kind chip, a value beside the method badge rather than a heading above a block. I had said this chip would move onto `Eyebrow`; reading it in place, it is a kind marker of the same family as the method badge, and an eyebrow would have been the wrong primitive.

**Worth a look:**

- **The slow-request box changes tone, red to warning.** It read "Slow request: Nms" on a destructive tint; a slow request is a warning about a result that arrived, not a failure, so it is a `Callout severity="warning"`. The one design change in the diff that is not a like-for-like swap.
- **The popover's two sentences are `Callout severity="blocking"`,** which brings a border and a box into a space that had bare text. Field-level did not fit: nothing refused a value, the token itself does not resolve.
- **`code-editor` keeps its `role="alert"` wrapper** around the inline `ErrorState`, because `ErrorState` has no role prop and `code-editor.loading.test.tsx` reads the alert. The `components/ui` files import the primitives by deep path, the way `code-editor` already imports from `components/shared`.

**Safe to skim:** the `docs/design-system.md` sentence describing which tags the guard scans, and the two guard header comments, restated per the keep-docs-in-step rule.

## Visual changes

No screenshots, per your instruction earlier in the programme to keep them out of the repo. What changes on screen: the settings pane's load failure and the variable table's load failure take `ErrorState`'s layout (muted icon, foreground title, mono detail) instead of red text throughout; the popover's two sentences gain a Callout box; the slow-request line goes from a red tint to the warning tint with the triangle glyph.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`pnpm lint` on the touched files, `pnpm type-check` and `pnpm format:check` green. `pnpm test error-presentation eyebrow variable-popover code-editor SettingsMain VariableTableEditor RequestResponseView design-system-doc`: 31 files, 181 tests, green.

Mutation check on the widened guards: a scratch `.tsx` under `src/` carrying `<div className="text-xs text-destructive-text">` and a `text-micro font-semibold uppercase` span fails both guards by name, then removed. Before this PR the same file passed both.

## Checklist

- [x] I have added tests (the guards, widened)

## Related Issues

Follow-up to #1683 (its closing audit); tightens the guards from #1688 and #1692.


athrvk:
The `github-advanced-security` check on `ff6ace9` failed before analysing anything: the Copilot code-scanning job died creating its session with `CAPIError: 400 The requested model is not supported`, the same GitHub-side error every PR in #1683's programme hit. Not this PR's, and that job has no re-run control from here. The repo's own matrix on the same head is what covers the change.

---

Issue #1615, opened 2026-09-08
app(layout): move the sidebar view switchers off the footer onto a left-edge activity rail, and give the context bar a right-edge rail of its sections - primary navigation must not sit where the macOS Dock lands


athrvk:
Owner-reported: the icon strip at the bottom of the sidebar should move to the left edge, the way IntelliJ's tool-window stripes do; on macOS, moving the pointer down to reach it summons the auto-hiding Dock, which covers the strip, and the user has to wait it out. And the same treatment for the context bar.

## Problem

Anchors at master `166d4651`.

The strip is the app's own footer, `Dock` (`app/src/components/layout/Dock.tsx:92-121`): a 32 px band (`--dock-height`, `docs/design-system.md:1152-1164`) whose left half is `<nav aria-label="Sidebar views">` rendering the six `DRAWER_VIEWS` buttons (Collections, History, Variables, Services, Trash, Settings; `app/src/constants/drawer-views.ts:34-`), and whose right half is engine status, version and worker count, running-service count, save status, the pending-restart button and the context-bar toggle (`Dock.tsx:116-`, tests `Dock.*.test.tsx`). The shell places it below everything (`Shell.tsx:396-`, `docs/app/COMPONENTS.md:49, 68`).

Two things are wrong with the placement, independent of the Dock's contents:

1. **The bottom edge of the window is the one edge a macOS user cannot aim at.** With Dock auto-hide on, the pointer crossing the screen's bottom edge summons the system Dock over the app's lowest 60 to 80 px, and the switchers sit exactly there. The user either waits for the Dock to retract or aims short. Fitts's law makes screen edges the cheapest targets; the bottom one is spent by the OS.
2. **Primary navigation is in a status bar.** Every desktop tool with a switchable side panel puts the switchers on a vertical edge stripe and keeps the bottom band for passive status: IntelliJ's tool-window bars (left and right stripes, click toggles the window, the active one marked), VS Code's Activity Bar (a 48 px left rail, badges on icons, the primary sidebar beside it, the status bar below holding only status), Postman's own left rail (Collections, Environments, History, ...). The left edge is an infinite-width target and the system never covers it.

The context bar has the same shape in miniature: one toggle in the footer (⌘I, `COMPONENTS.md:156`) opens a bar of collapsible sections (`context-bar/registry.ts:119-195`: variables, auth, cookies, code, graphql, recent sends, the collection and run sections), and reaching a given section is toggle, then scroll, then expand.

## The design

[...]

PR #1712, opened 2026-09-20
feat(app): response position - Beside, Below or Auto, switched from the Dock

### LINKED ISSUE #1711: feat(app): response position - Beside / Below / Auto, switched from the Dock
## Problem

The request builder only ever shows the response **beside** the request. Postman, Insomnia, Bruno, Hoppscotch and Yaak all let the user stack the response **below** the request; on a laptop-width window or a portrait monitor, the side-by-side split leaves each half too narrow for a URL bar plus a JSON body. There is no setting, no shortcut and no per-tab control.

Anchors at master `95dc69ca6c30e3bfb476ad41414ac2711b90cf2f`:

- `app/src/modules/request-builder/components/RequestBuilderLayout.tsx:141-166` - `ResizablePanelGroup orientation="horizontal"` is hard-coded, both panels `minSize="20%" maxSize="80%"` seeded from `requestSplitRatio`. The header comment at `:11` says "vertical split", which is wrong for the orientation the group actually has.
- `app/src/stores/layout-store.ts:59,227,262,317-325,370-377` - one `requestSplitRatio` (default 0.5, clamped 0.2..0.8), persist `version: 5`, `migrate`, `partialize`.
- `app/src/components/layout/Dock.tsx:324-334` - "Status only, since #1615" comment and a single centred `flex-1 justify-center gap-4` status group. No right-aligned slot exists.
- `docs/design-system.md:2594-2610` ("### Dock", "ambient status only") and `:3649` (doc map row) state the same rule.
- Precedents to mirror: `app/src/constants/shortcuts.ts:63` (`TOGGLE_CONTEXT_BAR_CHORD`) and `:329` (its `SHORTCUT_GROUPS` entry under "Navigation & panels"); `app/src/lib/commands/registry.ts:281-286` (palette command `toggle-context-bar`); `app/src/modules/settings/main/panels/AppearancePanel.tsx:152` / `:363` (Density row: `Eyebrow` + `OptionButtons`) declared at `app/src/modules/settings/main/app-settings.ts:160-165`; `app/src/components/layout/PanelResizeHandle.tsx:130` (`onDoubleClick` resets width - the builder divider in `components/ui/resizable.tsx` has no reset).
- `GraphQLBody.tsx:772` / `:909` already has its own two orientations for the schema explorer split; it is unaffected by this issue except for the follow-on note below.

## Design (agreed with the owner)

**Vocabulary.** The setting is "Response position" with values **Beside**, **Below**, **Auto**. Never say vertical/horizontal in UI, docs or code comments - those words mean opposite things in Postman and in `react-resizable-panels`.

**Default stays Beside.** Auto is an explicit third choice, not the default: Beside is what every existing user has today and what the other tools default to.

**Where the control lives - Dock, right side.** This deliberately amends the #1615 rule from "ambient status only" to: **status in the centre, per-tab view controls on the right, and every item still has a non-footer path** (here: the palette command, the chord and the Settings row), so the macOS Dock covering the strip still costs a glance, never a click. The right cluster renders **only while the active tab is a request tab**; other tabs keep today's strip unchanged.

The Dock control is one `TooltipIconButton`-style icon button whose icon names the **destination**, not the current state (the same rule "### Pane Toggles" already states for `PanelLeft*`): showing `PanelBottom` while the response is beside (tooltip "Response below"), `PanelRight` while it is below (tooltip "Response beside"), swapped with `IconSwap`. Tooltip carries the chord. Clicking flips Beside <-> Below. When the setting is Auto, the button shows Auto's current pick and a click writes an explicit Beside/Below (the user chose; Auto is over until they pick it again in Settings).

## Fix pathway

1. **Store** (`layout-store.ts`): replace `requestSplitRatio` with
   - `responsePosition: "beside" | "below" | "auto"` (default `"beside"`),
   - `requestSplitRatioBeside` and `requestSplitRatioBelow` (each default 0.5, clamped 0.2..0.8),
   - `setResponsePosition`, `toggleResponsePosition` (beside <-> below; from auto, writes the opposite of the currently resolved arrangement), `setRequestSplitRatio(arrangement, ratio)`.
   Bump persist to `version: 6`; migration copies the old `requestSplitRatio` into the Beside ratio and seeds Below at 0.5. Add all three to `partialize`.
2. **Resolution hook** `useResolvedResponsePosition(containerRef)` (new, in `app/src/modules/request-builder/hooks/`): returns `"beside" | "below"`. For `auto` it observes the builder's own width with a `ResizeObserver` and picks Below under about 880px, Beside at or above, with a small hysteresis band so a divider drag near the threshold does not flap. Export the threshold as a named constant.
3. **`RequestBuilderLayout.tsx`**: pass the resolved arrangement to `ResizablePanelGroup` (`horizontal` for Beside, `vertical` for Below), and `key` the group on the arrangement so `defaultSize` re-applies from the matching ratio when it flips. Stacked minimums are pixels, not percent (about `160px` each), because a 20% response pane in a short window is unusable. Persist drags into the ratio for the arrangement that was dragged. Fix the `:11` comment. Port the double-click-resets-ratio behaviour to the builder divider via `components/ui/resizable.tsx` (an optional `onReset` prop, or the handle's own `onDoubleClick`), matching `PanelResizeHandle.tsx:130`.
4. **Dock** (`Dock.tsx`): add a right-aligned cluster after the status group, gated on the active tab being a request tab (read from `tabs-store`), containing the response-position button described above. Rewrite the `:324` comment to the amended rule.
5. **Palette + chord**: add `TOGGLE_RESPONSE_POSITION_CHORD` in `constants/shortcuts.ts` (pick an unbound chord; check `SHORTCUT_GROUPS` and the OS-reserved list), list it under "Navigation & panels", and register command `toggle-response-position` ("Toggle response position") in `lib/commands/registry.ts` next to `toggle-context-bar`. The palette command works from any request tab; on other tabs it is present but disabled the same way existing context-dependent commands are.
6. **Settings > Appearance**: a "Response position" `OptionButtons` row (Beside / Below / Auto) after Density, with a one-line description naming the Auto threshold; declare it in `app-settings.ts` with `anchor: "response-position"` and keywords `["layout", "split", "stacked", "beside", "below", "response"]`.

Follow-on, not in scope: when the outer split is Below, the GraphQL explorer split inside the request pane could flip to keep the explorer usable. File it as its own issue if the implementation confirms it matters.

## App changes

- `app/src/stores/layout-store.ts`
- `app/src/modules/request-builder/components/RequestBuilderLayout.tsx`
- `app/src/modules/request-builder/hooks/useResolvedResponsePosition.ts` (new)
- `app/src/components/ui/resizable.tsx`
- `app/src/components/layout/Dock.tsx` (plus a small `ResponsePositionButton` component beside it)
- `app/src/constants/shortcuts.ts`, `app/src/lib/commands/registry.ts`
- `app/src/modules/settings/main/panels/AppearancePanel.tsx`, `app/src/modules/settings/main/app-settings.ts`

No new dependencies: `ResizeObserver`, `react-resizable-panels` and the existing primitives cover everything.

## Tests (each with its mutation check)

- `layout-store.test.ts`: migration from v5 `{ requestSplitRatio: 0.3 }` yields `requestSplitRatioBeside === 0.3`, `requestSplitRatioBelow === 0.5`, `responsePosition === "beside"`; `toggleResponsePosition` from each of the three states. Mutation: drop the v5 branch in `migrate`, the migration test fails.
- `RequestBuilderLayout.test.tsx`: renders with the store in each of the three settings at a wide and a narrow container width (stub `ResizeObserver` and the measured width) and asserts the group's orientation attribute (`data-panel-group-direction` or whatever the library exposes) for all six combinations. Mutation: hard-code `horizontal` back, the Below and narrow-Auto cases fail.
- `Dock.test.tsx`: the right cluster is present with a request tab active and absent with a settings or dashboard tab active; the button's accessible name is "Response below" when beside and "Response beside" when below. Mutation: remove the request-tab gate, the absence assertion fails.
- `AppearancePanel` test: the row renders three options and writes the store. Mutation: remove the `onChange`, fails.
- Existing guards must stay green without new exemptions: `palette-parity.test.ts`, `shortcuts.listed.test.ts`, `eyebrow.test.ts`, `error-presentation.test.ts`, the settings-index test that checks every declared anchor has a `data-setting-anchor` in the DOM.
- Both branches of anything platform-dependent (the chord's display string) are asserted, per CLAUDE.md.

## Docs to update in the same commit

- `docs/design-system.md` "### Dock" (`:2594`) and the doc-map row at `:3649`: replace "ambient status only" with the amended rule and describe the right cluster and its gating; add the Beside/Below/Auto vocabulary and the destination-icon rule reference to "### Pane Toggles".
- `app/CLAUDE.md` if it lists layout-store fields or the Dock rule.
- The user-facing docs page that describes the request builder or keyboard shortcuts under `docs/` (grep for "Toggle context bar" to find the shortcuts table) gains the new chord and setting; `mkdocs build --strict` must pass.

## Sequencing

Independent of every open issue. It touches `Dock.tsx` and `docs/design-system.md` sections that #1683's PRs already merged, so cut from `95dc69ca` or newer. One PR; commits split as: store + migration, layout + hook, Dock + palette + shortcut, settings row, docs.

## Acceptance criteria (fresh checkout)

1. `pnpm test`, `pnpm type-check`, `pnpm lint`, `pnpm format:check` green in `app/`; `mkdocs build --strict` green.
2. Open a request tab: the Dock shows an icon button on the right whose tooltip reads "Response below". Click it: the response stacks under the request, the button now reads "Response beside", and the change survives an app restart.
3. Switch to the Settings tab: the right cluster is gone; the centred status is unchanged.
4. Settings > Appearance > Response position offers Beside / Below / Auto. With Auto, narrowing the window below the threshold stacks the response and widening it puts it beside again; clicking the Dock button while in Auto sets an explicit choice and the Settings row reflects it.
5. The palette lists "Toggle response position" with its chord, and the chord flips the arrangement from a request tab.
6. Dragging the divider in Beside, flipping to Below, dragging there, and flipping back restores the Beside ratio that was set. Double-clicking the divider resets the current arrangement to 50/50.
7. A user upgrading from 0.33.0 keeps their existing split ratio as the Beside ratio.
8. No UI string, doc line or code comment introduced by the PR uses "vertical" or "horizontal" for the arrangement.

athrvk:
## Description

The response pane could only sit beside the request. This adds a **Response position** setting - Beside (the default, unchanged for everyone upgrading), Below, and Auto - switched from a new right-aligned button in the Dock, a palette command, a chord (mod+shift+B) and a Settings > Appearance row.

Decisions worth a reviewer's attention rather than a re-read of the diff:

- **The Dock rule from #1615 is amended, on purpose.** The strip was "ambient status only"; it is now status in the centre and per-tab view controls on the right, shown only while the active tab is a request tab. The rule that mattered survives: every item has a non-footer path (chord, palette, Settings), so the macOS Dock covering the strip still costs a glance, never a click. Two equal `flex-1` gutters keep the status group centred in the whole strip so the connection light does not shift when a request tab comes and goes.
- **The icon names the destination, not the state**, the same rule the design system already holds for `PanelLeft*`: `PanelBottom` while beside ("Response below"), `PanelRight` while below ("Response beside"). A click while the setting is Auto writes an explicit choice; Auto is over until the user picks it again.
- **Right-click on the same button picks instead of flipping**: a radio menu over Beside / Below / Auto, marking the one in force. It is the one way back to Auto from the strip, and it carries the `own-menu` marker so the main process's edit menu stays away. The context-menu primitive gains the radio group and item it lacked, styled as the dropdown's are.
- **Auto measures the builder, not the window.** A `ResizeObserver` on the builder root (only while the setting is Auto) sees drawer drags and the context bar opening as well as window resizes. Threshold 880px with a 40px hysteresis band so a drag near the line does not re-arrange the split on every pixel. Auto's pick is held in the store (not persisted) because the Dock button and the toggle live outside the builder tree.
- **A flip re-orients the panel group in place.** The library re-lays out when its `orientation` prop changes, and a layout effect then applies the new arrangement's ratio through the group's imperative handle in the same frame. An earlier revision keyed the group on the arrangement; that remounted the request editor and the response viewer on every Auto flip during a window resize, which read as the whole pane flashing. A test now holds the panes' DOM identity across a flip.
- **Two ratios, one per arrangement.** The old `requestSplitRatio` migrates to the Beside ratio (persist v5 to v6); Below starts at 50/50. Only a user's own drag persists (`isUserInteraction`), so the mount-time layout report, a flip, and the pixel-minimum clamp can never overwrite a stored preference. Stacked panes use 160px minimums rather than 20%: a fifth of a short window cannot show a status line and a row of body.
- **Divider double-click resets the current arrangement to 50/50**, through the same handle as well as the store, since `defaultSize` is read once at mount.

Risk to look at: `components/ui/resizable.tsx` now stamps `data-orientation` on the group and moves the stacked-divider styling from v3's dead `data-[panel-group-direction]` selectors to the separator's own `aria-orientation`. That affects every `ResizablePanelGroup` in the app, including the GraphQL explorer split; it was previously unstyled in the stacked case because the selector never matched.

Safe to skim: the three existing tests that seeded `requestSplitRatio` and now seed the Beside ratio, the doc-map rows, and the `docs/app/*` updates that mirror the store and chord changes.

One caveat for `git bisect`: the first commit (store + migration) does not type-check alone, because the layout still reads the old field until the second commit. Every commit from the second onward is green.

Not in scope, noted rather than filed: the GraphQL explorer's inner split could flip when the outer split is Below. Nothing in this work showed it matters yet.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

From `app/`: `pnpm type-check`, `pnpm lint`, `pnpm format:check` clean; full `pnpm test` 808 files, 8961 passed, 2 pre-existing skips, plus the in-place flip and right-click tests added afterwards. `mkdocs build --strict` passes. Source-scanning guards (palette parity, shortcuts listed, eyebrow, icon-button labels, a11y suppressions, settings anchor drift, design-system and state-management doc guards) green with no new exemptions.

Mutation checks, each reverted and re-verified: dropping the v6 migration branch fails the store tests; hard-coding `horizontal` back fails exactly the Below and narrow-Auto layout cases; putting `key={arrangement}` back on the panel group fails the pane-identity assertions; removing the Dock's request-tab gate fails the settings, dashboard and no-tab absence cases; removing the radio group's `onValueChange` fails the right-click Auto case; removing the Shell chord's gate fails its settings-tab half; a no-op Settings `onChange` fails the "writes the store" case.

Manual: the Settings row renders and writes `responsePosition` to `vayu.layout` in Chromium. The owner confirmed the Dock flip works in the app; the flash they saw on Auto flips during a window resize is what the in-place change fixes.

## Checklist

## Related Issues
Closes #1711


athrvk:
The `github-advanced-security` check is red for the same reason as on every recent PR: the code-scanning agent fails before it reads the diff with `CAPIError: 400 The requested model is not supported` (model `claude-opus-5`, job 106116955768). Nothing in this branch is analysed or at fault, and there is no fix to port from the repo side; it is a GitHub-side model configuration. Not re-running it: the error reproduces identically on every head. The `pr-tests` jobs are the checks that cover this change.

---

Issue #1714, opened 2026-09-20
perf(app): subscribe to store fields, not whole stores - 57 selector-less Zustand reads re-render on every write


athrvk:
## Problem

Found by the re-render sweep on master `95380985`. Zustand 5 (`app/package.json`) re-renders a component on every `set()` to a store when the hook is called without a selector, because the whole state object is what it subscribes to. `rg -n "use[A-Z][A-Za-z]*Store\(\)" app/src --glob '!*.test.*'` returns 57 such calls and `useShallow` is used nowhere. The hot ones, ranked by write frequency times subscriber size:

- `app/src/modules/dashboard/index.tsx:52` - `LoadTestDashboard` destructures the whole dashboard store, which `load-test-service.ts` writes at up to 10 Hz during a run; `DashboardHeader` (`components/DashboardHeader.tsx:28`) and `RequestResponseView` (`components/RequestResponseView.tsx:56`) are not memoised, so the whole dashboard shell re-executes per tick.
- `app/src/modules/request-builder/context/RequestBuilderProvider.tsx:207` - the whole response store, whose `withResponse` replaces `responses` and `lru` on every send; one tab landing a response re-renders every mounted request-builder provider (up to 12).
- `app/src/components/layout/TabStrip.tsx:83` (inside `TabItem`, one per open tab) and `:233` - the whole tabs store, which also carries `navHistory`, `tabFocusedAt`, `specTabTarget`, `dataRowTarget`; one `focusTab` re-renders every tab element.
- `app/src/components/layout/Shell.tsx:173` (tabs) and `:189` (layout) - the top of the tree, with no memoised layout children below it.
- `app/src/components/layout/Drawer.tsx:27`, `ContextRail.tsx:94-95`, `ActivityRail.tsx:46`, `ContextBar.tsx:114-115` - the whole layout store, which `PanelResizeHandle.tsx:66-68` writes on every pointer move during a drawer or context-bar drag (see the companion issue on the splitter).
- `app/src/queries/health.ts:78`, `app/src/hooks/useSaveManager.ts:87` - engine and save stores, both written on a cadence.

The streaming consumers (`useExecutionEvents.ts`, `ResponseViewer/index.tsx`, `ScenarioRunView.tsx`, `MetricsView.tsx`) already select fields; they are the pattern to copy.

## Fix pathway

[...]

PR #1723, opened 2026-09-20
perf(app): stop persisting the drawer and context-bar splitter per pointer move

### LINKED ISSUE #1715: perf(app): drawer and context-bar splitter persists to localStorage on every pointer move
## Problem

Found by the re-render sweep on master `95380985`.

- `app/src/components/layout/PanelResizeHandle.tsx:66-68` - `onMove` calls `setWidth(...)` on every native `pointermove`, unthrottled.
- `app/src/stores/layout-store.ts:316-317` (`setDrawerWidth`, and the `contextBarWidth` setter beside it) write straight into the persisted store; `partialize` at `:467-471` includes both widths and the `persist` middleware at `:399` uses the default synchronous storage.

So every pointer move during a drag (120 to 240 per second on a modern mouse or trackpad) is a store write, a re-render of every layout-store subscriber (all four rails while the companion selector issue is open), and a synchronous `JSON.stringify` plus `localStorage.setItem` of the whole persisted slice. The request-builder divider one file over already does this right: `RequestBuilderLayout.tsx:66-76` debounces the persisted ratio by 200 ms.

## Fix pathway

Keep the live width out of the persisted store during the drag: hold it in the handle's local state or a ref applied to the element's style per frame (`requestAnimationFrame`), and write the store once on `pointerup` (or debounced, as the builder divider does). Double-click reset keeps writing immediately. `Drawer` and `ContextBar` read the live width from the same place the handle writes it, so no frame shows a stale width.

## Tests

- `PanelResizeHandle.test.tsx`: a drag of N pointer moves results in one store write (on pointer up) with the final width; the element's inline width tracks the pointer during the drag. Mutation: restore the per-move `setWidth`, the write count assertion fails.
- Existing drawer width persistence tests stay green (the value still lands in storage).

## Docs

`docs/design-system.md` "### Drawer" mentions `PanelResizeHandle`; add one sentence on the live-versus-persisted split. `docs/app/state-management.md` if it lists when layout-store persists.

## Sequencing

After the selector sweep issue, so the remaining per-frame re-render is the handle's own.

## Acceptance criteria

1. A drag produces exactly one `vayu.layout` storage write.
2. No visible lag between pointer and drawer edge on a 144 Hz display.
3. Width survives an app restart as today.

athrvk:
## Description

Dragging the drawer or context-bar splitter wrote the persisted layout store on every pointer move: a synchronous serialise and localStorage write per event, at 120 to 240 events a second on a modern mouse, plus a re-render of every layout-store subscriber. The handle now keeps the live width out of the store: it paints the pane's inline width once per animation frame during the drag and calls the persisted setter exactly once, on pointer up. Double-click reset and the keyboard nudges are discrete actions and still write immediately.

Risk to look at: the live paint targets the handle's `parentElement`, relying on both `Drawer` and `ContextBar` rendering the handle as the direct child of the `<aside>` whose width it controls. That is true today (both files pass `style={{ width }}` on that aside) and a test asserts the inline width tracks the drag, so a future re-nesting fails loudly rather than silently. The alternative, threading a live width through props or a transient store field, would have touched `Drawer.tsx` and `ContextBar.tsx`, which #1714 is editing in parallel. The clamp during the drag mirrors the store setters' own (220 to 480px) so the pane cannot overshoot before the single persisted write re-clamps.

Why the imperative width survives React: nothing writes the store during a drag, so the aside's `width` prop is unchanged across any re-render and React does not rewrite the DOM style; the pointer-up write lands on the same value the last frame painted.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`PanelResizeHandle` suite: 15 tests, 9 new. A drag of many pointer moves calls the store setter once, on pointer up, with the final width; the inline width tracks the pointer across frames; the left-edge handle inverts direction; the drag clamps at the maximum; double-click still writes immediately. Mutation check: restoring the per-move setter call fails four of them. `Drawer`, `ContextBar` and `layout-store` suites: 116 tests green, so width persistence across restarts is unchanged. Type-check, lint and Prettier clean.

## Checklist

## Related Issues
Closes #1715


athrvk:
The `github-advanced-security` check is red for the same reason as on every recent PR: the code-scanning agent fails before it reads the diff with `CAPIError: 400 The requested model is not supported` (model `claude-opus-5`, job 106132180552). Nothing in this branch is analysed or at fault, and there is no repo-side fix; it is a GitHub-side model configuration. Not re-running it: the error reproduces identically on every head. The `pr-tests` jobs are the checks that cover this change.

---

PR #1724, opened 2026-09-20
fix(app): reserve the scrollbar gutter on scroll containers so content does not shift

### LINKED ISSUE #1720: fix(app): reserve the scrollbar gutter so content does not shift when a pane gains or loses a scrollbar
## Problem

Found by the re-render sweep on master `95380985`, at SUSPECTED confidence: the mechanism is browser behaviour, the specific offending pane was not caught in the act.

`app/src/index.css` declares no `scrollbar-gutter` anywhere, and `docs/design-system.md`'s Scrollbar section counts around 44 `overflow-y-auto` containers. Any of them whose content crosses the scroll threshold while the user types a filter, a query refetches with one more row, or a section expands, gains or loses the 6px scrollbar and shifts its content's right edge by that width. Small, intermittent and app-wide.

## Fix pathway

Add `scrollbar-gutter: stable` to the same specificity-zero `:where(*)` rule the Scrollbar section already uses for `scrollbar-width` and `scrollbar-color`, under the existing `@supports` guard, so a container that must not reserve the gutter can still opt out. Review the handful of panes where a permanent 6px reserve looks wrong (the Dock strip, single-line inputs) and opt them out explicitly.

## Tests

`index.css` source guard in the existing scrollbar test (it must assert it read a non-empty stylesheet, since vitest stubs CSS imports). Mutation: remove the declaration.

## Docs

`docs/design-system.md` Scrollbar section: one sentence on the gutter reserve and the opt-out.

## Sequencing

Independent, small.

## Acceptance criteria

1. Typing a filter that takes the collections list from 20 rows to 3 does not move the rows' right edge.
2. No pane shows a visibly empty 6px band where it previously sat flush, or it is an accepted, listed exception.

athrvk:
## Description

A pane whose content crosses the scroll threshold while the user types a filter, a list refetches, or a section expands gains or loses the 6px scrollbar and shifts its content's right edge. `scrollbar-gutter: stable` on the scroll containers reserves that width permanently.

Two decisions a reviewer should know, both made after the first draft got them wrong:

- **Outside the `@supports not selector(::-webkit-scrollbar)` block.** That block is the non-Chromium fallback; Electron is Chromium, so anything inside it never applies here. The gutter property does not trigger the Chromium 121+ switch to standard scrollbars that `scrollbar-width` and `scrollbar-color` do, so it can sit beside the `::-webkit-scrollbar` rules that actually render in the app.
- **Scoped to the scroll-container utilities, not `:where(*)`.** Per CSS Overflow 4, `stable` reserves the gutter for `overflow: hidden` too, and because the app styles `::-webkit-scrollbar` its scrollbars are classic on every platform, so the reserve is real. `overflow-hidden` appears 72 times in `app/src` for truncation and clipping; a global rule would have put a 6px dead strip on every one of them. The selector is the four Tailwind overflow utilities, since CSS cannot select on computed overflow, and stays specificity zero so a utility can override it.

What to look at in the running app rather than the diff: every `overflow-y-auto` pane now shows its 6px reserve even while its content fits (collections list, settings panels, response viewer). That is the intended trade, but it is a visible change and it was not checked in a browser here; if a pane looks wrong with the band, the opt-out is a `[scrollbar-gutter:auto]` utility on that element.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

`scrollbar-systems` guard (7 tests) reads the stylesheet from disk and asserts the declaration occurs exactly once, sits outside the `@supports` block, and carries the scoped selector rather than `:where(*)`. Three mutation checks, each failing as intended: remove the declaration, move it inside the guard, widen the selector to `:where(*)`. `design-system-doc` guard green; Prettier clean on `index.css`. Not run in a browser; see above.

## Checklist

## Related Issues
Closes #1720


athrvk:
The `github-advanced-security` check is red for the same reason as on every recent PR: the code-scanning agent fails before it reads the diff with `CAPIError: 400 The requested model is not supported` (model `claude-opus-5`, job 106132723758). Nothing in this branch is analysed or at fault, and there is no repo-side fix; it is a GitHub-side model configuration. Not re-running it: the error reproduces identically on every head. The `pr-tests` jobs are the checks that cover this change.

---

PR #1725, opened 2026-09-20
feat(app): keep the Body and Elements editors mounted once visited across sub-tab switches

### LINKED ISSUE #1718: feat(app): keep the Body and Elements editors mounted across request sub-tab switches
## Problem

Found by the re-render sweep on master `95380985`.

`app/src/modules/request-builder/components/RequestTabs/index.tsx:117-145` renders the request's sub-tabs with plain `TabsContent`, which mounts only the active panel (the file's own comment at `:126-128` says so). The Body tab and the Elements tab (pre/post scripts via `ScriptElementForm.tsx` and `components/ui/code-editor.tsx`) therefore mount a fresh Monaco editor and model every time the user returns to them: cursor position, scroll position and undo/redo history are gone, though the text survives because it is read back from the request store. `components/ui/tabs.tsx:176-190` documents the force-mount pattern, and `CollectionDetail/index.tsx:320` already uses it "to keep unsaved drafts alive".

## Fix pathway

Force-mount the editor-bearing panels only (the `EDITOR_TABS` set at `:34` already names them) and hide inactive ones the way `tabs.tsx` handles force-mounted panels, so params and headers keep their cheap mount-on-demand behaviour. Confirm Monaco's `automaticLayout` handles a hidden-then-shown container without a visible reflow (the layout audit found no zero-size layout issue with the current unmount approach; a hidden editor must not introduce one). Memory: two extra editors per open request tab; check the 12-tab case.

## Tests

- `RequestTabs` test: switch Body to Headers and back; the editor element is the same DOM node. Mutation: remove `forceMount`, identity fails.
- Existing "renders exactly once" assertions in `RequestTabs` tests are updated to the new contract deliberately, not loosened.

## Docs

`docs/app/COMPONENTS.md` request-builder section where it states only the active sub-tab mounts.

## Sequencing

Independent. After the selector sweep is preferable, since more mounted editors under a whole-store subscriber cost more.

## Acceptance criteria

1. Type in the body, place the cursor mid-text, switch to Headers and back: cursor, scroll and undo stack are intact.
2. Same for a script in Elements.
3. No layout jump when returning to the editor tab.

athrvk:
## Description

Switching from Body to Headers and back rebuilt the Monaco editor: cursor, scroll position and undo history were lost, with only the text surviving via the request store. The same happened to every script editor on the Elements tab. The two editor-bearing panels are now force-mounted, but only from their first visit onward: a request opened on Params never pays for a Body editor or for one Monaco instance per script element that nobody has looked at. Params, Headers, Auth, Examples and Settings keep mounting on demand.

Decisions worth knowing:

- **Mount on first visit, keep thereafter** rather than force-mount from the start. This is the memory guardrail from the earlier performance programme: at most one request builder is mounted at a time, so the ceiling is one Body editor plus the visited request's script editors, all disposed when the builder unmounts. `@monaco-editor/react` disposes the model and editor on unmount; that was confirmed by reading its cleanup path.
- **The `TabContent` switch on the active tab is gone.** It rendered `null` for every non-active tab, which would have silently emptied a force-mounted panel the moment focus moved, defeating the change. Each panel now renders inline against its own tab id.
- **Hidden panels stay out of the accessibility tree and tab order** through the existing `tabs.tsx` mechanism (`data-[state=inactive]:hidden`), the same one `CollectionDetail` already relies on for its force-mounted drafts. The editor chords fire only on editor focus and the Send chord keys off the event target, so neither reaches a hidden panel.
- **The flex column is now Body's alone.** It used to be keyed on membership in the editor-tab set; Elements is a list of rows with an editor apiece and keeps its plain scrolling block.

Not verified here, said plainly: whether Monaco's `automaticLayout` re-lays out cleanly from a hidden container without a visible jump. jsdom has no layout, so no test can catch it. The supporting evidence is that `CollectionDetail` already force-mounts the same `ElementList` and editor stack in production today.

Also in this PR: the drafts lifetime test's header comment no longer claims Body unmounts on every sub-tab switch, since that is what this change removed; the drafts still belong to the provider because the panel goes with the builder.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

New `editor-mount-persistence.test.tsx` counts mounts with an effect-based marker rather than renders: fresh mount on Params mounts neither editor panel; visiting Body then Headers keeps Body in the DOM, inactive and hidden, at one mount, with Elements still unmounted; Body to Headers to Body keeps the same DOM node; the same pair for Elements; Params and Headers still mount on demand. Two mutation checks, each failing as intended: force-mount without the visited gate fails the never-mounted and Elements cases; force-mount every visited tab fails the on-demand case. `pnpm test RequestTabs request-builder tabs` 132 files, 1348 tests green before the master merge; `RequestTabs editor-mount tabs` 44 files, 511 tests green after it. Type-check, lint, Prettier clean; `mkdocs build --strict` clean on the touched docs page.

## Checklist

## Related Issues
Closes #1718


athrvk:
The `github-advanced-security` check is red for the same reason as on every recent PR: the code-scanning agent fails before it reads the diff with `CAPIError: 400 The requested model is not supported` (model `claude-opus-5`, job 106133885105). Nothing in this branch is analysed or at fault, and there is no repo-side fix; it is a GitHub-side model configuration. Not re-running it: the error reproduces identically on every head. The `pr-tests` jobs are the checks that cover this change.

---

PR #1726, opened 2026-09-20
perf(app): stop keystrokes re-rendering every row of the key-value editor and variables table, memoise the palette's request list

### LINKED ISSUE #1716: perf(app): per-keystroke re-renders in the key-value editor, the variables table and the palette's request list
## Problem

Found by the re-render sweep on master `95380985`. Three per-keystroke paths do work proportional to the whole list on every character.

1. **Key-value editor memo is defeated.** `app/src/components/shared/KeyValueEditor/index.tsx:70-101` - `handleRemove`, `handleUpdate`, `handlePickFile`, `handleToggleKind` list `items` in their `useCallback` deps. Every keystroke writes a new `items` array through `onChange`, so every callback gets a new identity and every `KeyValueRow` (`KeyValueRow.tsx:286`, `memo`-wrapped) fails its shallow compare. The component's own comment says a fresh identity "would re-render every row of the densest table in the app on every keystroke"; that is the current behaviour.
2. **Variables table has no row component.** `app/src/modules/variables/main/VariableTableEditor.tsx:835` renders rows inline in a `.map`; `variables` is local state (`:304`) and `updateVariable` (`:587`) rebuilds the array per keystroke, so every row's `Input`, `Select` and `SecretInput` re-renders on each character typed in any row.
3. **Palette rebuilds the request list per keystroke.** `app/src/modules/palette/sources/useEntityItems.ts:63-116` builds the full items array (every collection, every request, a path per collection) on every render, and the palette re-renders per keystroke; `useRunItems`, `useSettingsItems` and `useVariableItems` all memoise theirs.

## Fix pathway

1. Read `items` from a ref (or use functional updates) inside the four callbacks so their deps are `[onChange, canEdit, canRemove, canDisable]` only. Check `canEdit`/`canRemove`/`canDisable` callers pass stable functions; the defaults are inline arrows in the parameter list, so hoist them to module constants.
2. Extract `VariableRow`, `memo`-wrap it, key by `variable.id`, pass `onUpdate(id, field, value)` with a stable identity (ref-backed).
3. `useMemo` the `items` build keyed on `[collections, requestsByCollection, lastSent, byId, openTab]`.

## Tests (each with its mutation check)

- `KeyValueEditor` render-count test: with five rows, typing in row 3 re-renders row 3 only (a probe component or `React.Profiler` around each row). Mutation: put `items` back in `handleUpdate`'s deps, all five re-render.
- `VariableTableEditor` render-count test, same shape. Mutation: render rows inline again.
- `useEntityItems` test: two renders with the same inputs return the same array reference. Mutation: drop the `useMemo`.

## Docs

`docs/app/COMPONENTS.md` describes `KeyValueEditor`'s memo contract; correct it to state the dependency rule.

## Sequencing

Independent. Three commits, one per site.

## Acceptance criteria

1. The three render-count tests pass and fail under their mutations.
2. Typing in a 100-row headers table or variables table shows no per-keystroke lag on a mid-range laptop.
3. `pnpm test`, `pnpm type-check`, `pnpm lint`, `pnpm format:check` green.

athrvk:
## Description

Three per-keystroke paths did work proportional to the whole list on every character. Each gets its own commit and a render-count or identity test.

- **Key-value editor.** `KeyValueRow` is memoised, but the editor's four callbacks listed `items` in their deps, and `items` is a fresh array on every keystroke, so every row's props changed every time. The callbacks now read the latest items through a ref and depend only on `onChange` and the three predicates; the predicates' inline defaults are module constants so a caller that leaves them unset does not defeat the memo either.
- **Callers that passed a fresh `onChange`.** Found in review, not by the issue: the Headers panel handed `useHeadersManager` an inline `onUpdate` arrow that it returns straight back as `onChange`, and the Body panel's table handler was an inline arrow. Both would have re-created every row callback per render and left the two densest tables exactly as slow. Both are now `useCallback`.
- **Variables table.** Rows rendered inline in a `.map` with no row component. A memoised `VariableRow` is extracted, and `updateVariable`/`removeVariable` are id-based and ref-backed. One deviation from the issue: `handleBlur` depended on the dirty flag as state, so the first keystroke of a session changed its identity and re-rendered every row; it now reads the flag from a ref. The render-count test is what surfaced it. This file's two selector-less store reads were also converted, since #1714 allowlisted the file to this PR.
- **Palette.** `useEntityItems` rebuilt every collection and request per render; it is now `useMemo` keyed on its inputs, like its three sibling sources.

Safe to skim: five existing variables tests whose store mocks ignored the selector argument and now apply it, and the components guide's corrected memo contract.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

New tests: typing in row 3 of a five-row key-value table re-renders row 3 only; the same for the variables table; two renders of `useEntityItems` with the same inputs return the same array. Each was mutation-checked by restoring the original dependency or dropping the memo, and failed as intended. Targeted suites after the master merge: `KeyValueEditor VariableTableEditor variable-row palette HeadersPanel BodyPanel RequestTabs`, 49 files, 498 tests green; the guard suites (eyebrow, icon-button labels, a11y suppressions, routed inputs) green with no new exemptions. Type-check, lint and Prettier clean on all touched files.

## Checklist

## Related Issues
Closes #1716


athrvk:
The `github-advanced-security` check is red for the same reason as on every recent PR: the code-scanning agent fails before it reads the diff with `CAPIError: 400 The requested model is not supported` (model `claude-opus-5`, job 106135884736). Nothing in this branch is analysed or at fault, and there is no repo-side fix; it is a GitHub-side model configuration. Not re-running it: the error reproduces identically on every head. The `pr-tests` jobs are the checks that cover this change.

---

Issue #1156, opened 2026-08-30
app(response): the response store only grows - closing a tab or LRU-evicting it never clears the body, so hours of use retain every distinct request's response (~2x body each)


athrvk:
Sub-issue of #1143. Verified at master `e1c578a`, adversarially re-verified.

## Problem

Every design-mode send stores the full response - pretty `body`, compact `bodyRaw`, consoleLogs, testResults - in an uncapped in-memory `Map` keyed by requestId (`response-store.ts:57`; the `:22-29` comment mandates spreading the WHOLE ResponseState, `bodyRaw` included; written from `RequestBuilderProvider.tsx:164-173,896`). The only production evictions are entity DELETION seams (`tabs-store.ts:367-370`, `collections.ts:662`; verified by enumerating all `clearResponse`/`clearAll` callers). `closeTab` (`tabs-store.ts:338-353`) and LRU eviction (`:294-305`, `MAX_OPEN_TABS=12` at `:30`) never clear the entry. So renderer RAM grows monotonically with the distinct requests sent this session - one entry, up to ~2x the body size, kept until delete or quit. No cap, TTL, or truncation applies to the live stored copy (the `maxTraceBodyBytes` truncation machinery covers only trace-restored responses).

Design status, honestly: `docs/app/state-management.md:588-593` documents delete-only eviction and itself recounts the session-long growth with no cap - keep-while-request-exists is arguably intentional (a reopened request shows its last full response; the backend-restored copy is only a truncated slice), but the docs admit the cost rather than bound it, and the delete-seam work (`tabs-store.test.ts:258-267`) shows the growth was recognized and only partially closed.

## Fix / solution pathway

Give the Map an LRU cap keyed off open-tab reach (e.g. retain full-fidelity responses for the ~12 open tabs plus a small tail; evict beyond it). Prefer LRU over clear-on-close: clear-on-close is not lossless (the trace-restored copy is truncated at `maxTraceBodyBytes`), so a cap preserves reopen-UX for recent tabs while bounding the session. Keep the delete-seam eviction as-is. Update the store header and `state-management.md` to state the bound.

## App changes

This IS the app change.

## Tests

`cd app && pnpm test` - the store and tabs-store tests exist; add: eviction beyond the cap, most-recent retained, delete-seams unchanged, reopened recent tab still shows its full response. Mutation-check by removing the cap.

## Docs (same commit)

[...]

Issue #1146, opened 2026-08-30
app(startup): the renderer is one eager 5.5MB module graph - zero React.lazy; Monaco, graphql-language-service and the markdown stack all parse before the window can appear (~420ms measured)


athrvk:
Sub-issue of #1143. Verified at master `e1c578a`, adversarially re-verified with a real measurement.

## Problem

There is no `React.lazy` or route-level dynamic import anywhere in `app/src` (verified: the only shipped dynamic imports are the two prettier files in `lib/graphql/format.ts:36-37`). The eager graph:

- `app/src/main.tsx:8` imports side-effecting `@/lib/monaco-setup` before React renders; `monaco-setup.ts:17` is `import * as monaco from "monaco-editor"` (the full barrel), and `:43` registers the graphql providers, reaching static `graphql-language-service` (`lib/graphql/language-providers.ts:15-20`) and `graphql` (`diagnostics.ts:23-24`).
- `Shell.tsx:28-36` statically imports all eight feature surfaces (RequestBuilder, LoadTestDashboard, History, Settings, Welcome, Variables, Inbox, Palette).
- react-markdown + remark/rehype reach the entry through Shell's static graph (`InfoTab.tsx:26`, `InfoPanel.tsx:36`) - dropping the ui-barrel export alone would not remove them.

The built entry is **5,531,024 bytes (1.45MB gz), the single script in `dist/index.html`**, executed synchronously; the window is `show: false` and shown on `ready-to-show` = first paint (`app/electron/main.ts:266, 278-280`), so every ms of parse/eval is window-appearance latency. **Measured in a hidden Electron window on a 4-core Xeon 2.8GHz: ready-to-show 478-495ms vs a 63-74ms blank baseline - ~420ms attributable, V8 parse+compile of the entry alone ~210ms.** Plausibly near 1s on low-end hardware. Not documented deliberate: `monaco-setup.ts:9`'s "Import once, before React renders" records worker-wiring order, not an accepted cost; `docs/app/architecture.md:316`'s "code splitting" is cache partitioning of eagerly-loaded chunks (see the companion issue).

Nuance from verification: Monaco's per-language tokenizer bodies ARE lazily split (the ~85 small grammar chunks work); eager is monaco core + contribution stubs + json/ts clients.

## Fix / solution pathway

1. `React.lazy` the Shell tab surfaces behind Suspense (dashboard, history, settings, variables, inbox, welcome at minimum - RequestBuilder can stay eager as the default surface).
[...]

Issue #26, opened 2026-06-05
MetricsView derived useMemo over-deps on historicalMetrics — invalidates HeroRow/ModeStatsRow every tick


athrvk:
## Summary

The `derived: DashboardDerived` `useMemo` in `MetricsView.tsx` has `historicalMetrics` in its dependency array, so it rebuilds the entire 22-field `DashboardDerived` bundle on every metrics tick. `HeroRow` and `ModeStatsRow` (both `React.memo`'d) then re-render every tick because their `d` prop is a fresh object — defeating the memo.

Only `lastTick?.elapsed_seconds` actually depends on the full `historicalMetrics` array; the heavy fields (`peakConcurrency`, `breakpoint`) come from separate memos already.

## Failure scenario

At 10 Hz tick cadence:
1. Every tick, a new 22-field `DashboardDerived` object is allocated.
2. Reference inequality propagates to `HeroRow` and `ModeStatsRow`.
3. All hero cards re-render (10 cards × deep tree) every 100ms.
4. Card values haven't materially changed (currentRps wobbles by ±1, peak is identical, mode is identical) but every card re-renders anyway.

Combined with #25 (peakConcurrency / breakpoint full-scan) the per-tick main-thread cost is roughly 2× what it needs to be.

## Suggested fix

Replace the dep on `historicalMetrics` with only the field that's actually consumed:

```ts
const lastElapsed = historicalMetrics[historicalMetrics.length - 1]?.elapsed_seconds;
const derived = useMemo<DashboardDerived>(() => ({ ... }), [
  /* everything else */, lastElapsed,
]);
```

This stabilizes the `derived` reference as long as the latest tick's elapsed_seconds (and the other deps) haven't changed in a way that affects what we render — so `HeroRow`'s memo can actually skip renders.

## Context

- File: `app/src/modules/dashboard/components/MetricsView.tsx:108` (approximate; the `derived` `useMemo`)
- Introduced in: #19
- Found by: post-merge code-review (Angle F — efficiency)
- Related: #25

## Priority

**Tier 2** — perf. Likely 30-50% per-tick CPU on the renderer for long runs.

PR #1727, opened 2026-09-20
perf(app): subscribe to store fields, not whole stores

### LINKED ISSUE #1714: perf(app): subscribe to store fields, not whole stores - 57 selector-less Zustand reads re-render on every write

athrvk:
## Description

Zustand re-renders a component on every write to a store when the hook is called without a selector. The app had 57 such calls and no use of `useShallow`, so the tab strip re-rendered every tab on any tabs-store write, all four layout rails re-rendered per pointer move during a drawer drag, every mounted request-builder provider re-rendered when any tab landed a response, and the dashboard shell re-executed at 10 Hz during a load test. This converts every site to per-field selectors (the two in the variables table came in with #1726, now merged in) and adds a guard test so the pattern cannot return. The guard's allowlist is empty.

Decisions worth a reviewer's attention:

- **Subscription narrowing only.** The response store's LRU and tab-close clearing (#1156), `MetricsView`'s memo dependencies (#26, #25), the shell's lazy surfaces and single Suspense boundary (#1146), and every poll interval and query gate (#1150) are untouched; the shell identity test passes without edits. Actions are stable references and are selected individually. No `useShallow` wraps an array; array fields are selected directly.
- **`TabItem` is memoised with a comparator** that shallow-compares the descriptor `useTabDescriptors` rebuilds every render, and moved into its own module so the render-count test can count through a module mock rather than through instrumentation in the component. `DashboardHeader` and `RequestResponseView` are memoised with stable props (`onStop` is now a `useCallback`; `configuration` was already memoised; `report` is a store reference).
- **The guard** scans every non-test source file, skipping comment lines, asserts the set is non-empty, flags selector-less calls, object-literal selectors without `useShallow`, and selectors that derive a fresh array, and separately fails if an allowlist entry no longer points at a store call.

Safe to skim: about 29 test files whose store mocks ignored the selector argument and now apply it, the components' selector lines themselves, the a11y suppression list and Enter allowlist following `TabItem` to its new file, and the one-paragraph rule added to the state-management guide.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Full `pnpm test` before the final refactor: 812 files, 8974 passed, 2 pre-existing skips. After it and after merging master with #1725 and #1726: `store-selectors`, `VariableTableEditor`, `variable-row`, `TabStrip` suites, 10 files, 49 tests. `pnpm type-check`, `pnpm lint` at zero warnings, `pnpm format:check` clean. Mutation checks: reintroducing a selector-less call is reported by the guard, including after the comment-skipping change; forcing the tab item comparator to never bail raises the untouched tab's render count; both restored.

## Checklist

## Related Issues
Closes #1714


athrvk:
The `github-advanced-security` check is red for the same reason as on every recent PR: the code-scanning agent fails before it reads the diff with `CAPIError: 400 The requested model is not supported` (model `claude-opus-5`, job 106141304025). Nothing in this branch is analysed or at fault, and there is no repo-side fix; it is a GitHub-side model configuration. Not re-running it: the error reproduces identically on every head. The `pr-tests` jobs are the checks that cover this change.

---

